PR #2, opened 2012-08-28
Add the --duration and --split flags to the python wrapper for rosbag.

mjcarroll:
Changing so that duration and size flags are recognized through the python command line.

Modification taken from https://code.ros.org/trac/ros/ticket/3810

Quote:
[moved from  https://code.ros.org/trac/ros-pkg/ticket/5335]

Currently rosbag record doesn't support --duration argument (it ignores it).

It would be useful to have such feature, in particular to record multiple topics for a given time, since -l (number of messages) is not useful in such case (it records the same number of messages for each topic).

This affects to ros-electric, as well as previous version (at least, diamondback).

### DIFF
--- a/tools/rosbag/src/rosbag/rosbag_main.py
+++ b/tools/rosbag/src/rosbag/rosbag_main.py
@@ -108,10 +108,8 @@ def record_cmd(argv):
         if not options.duration and not options.size:
             parser.error("Split specified without giving a maximum duration or size")
         cmd.extend(["--split"])
-        if options.duration:
-            cmd.extend(["--duration", options.duration])
-        if options.size:
-            cmd.extend(["--size", str(options.size)])
+    if options.duration:    cmd.extend(["--duration", options.duration])
+    if options.size:        cmd.extend(["--size", str(options.size)])
     if options.node:
         cmd.extend(["--node", options.node])
 


wjwwood:
What is this change for?  This diff doesn't seem to do much....

mjcarroll:
So the size and duration params can be passed without the need for the
split flag in the actual cpp binaries of rosbag.

The extra indent in the python wrapper makes it so that you have to have
the split flag in order for duration or size to take effect.

My use case was:

rosbag record --duration 30 topic1 topic2 topic3

Which should record a 30 second long bag, but instead records indefinitely.

~mc mobile

On Aug 28, 2012, at 19:57, William Woodall notifications@github.com wrote:

What is this change for? This diff doesn't seem to do much....

—
Reply to this email directly or view it on
GitHubhttps://github.com/ros/ros_comm/pull/2#issuecomment-8111292.

wjwwood:
As Dirk just pointed out to me, the indention is different.

There was some confusion though, because I thought in this instance the patch was incorrect, because `--size` and `--duration` are used by `--split` to indicate how to split bag files.  But there seems to be the ability for `--size` and `--duration` to work independently of `--split` in the c++ implementation.  Therefore the patch is valid.

I will apply this patch and that will make the behavior as follows:

`rosbag record --duration 30 topic1 topic2 topic3` -> Records a 30 second bag file.
`rosbag record --duration 30 --split topic1 topic2 topic3` -> Records indefinitely, splitting bag files at 30 second intervals

I will, however, also open an issue stating that `--size` and `--duration` so always specify overall recording and `--split` should be moved into `--split-duration` and `--split-size` which might be mutually exclusive options, but could be used with `--duration` and `--size`.

mjcarroll:
I can probably add that, as it's something that I'm interested in.

And yes, I should have told you it was just a few indents.  It was noted in
the original issue, but I didn't copy it over.

~mc mobile

On Aug 28, 2012, at 20:19, William Woodall notifications@github.com wrote:

As Dirk just pointed out to me, the indention is different.

There was some confusion though, because I thought in this instance the
patch was incorrect, because --size and --duration are used by --split to
indicate how to split bag files. But there seems to be the ability for
--size and --duration to work independently of --split in the c++
implementation. Therefore the patch is valid.

I will apply this patch and that will make the behavior as follows:

rosbag record --duration 30 topic1 topic2 topic3 -> Records a 30 second bag
file.
rosbag record --duration 30 --split topic1 topic2 topic3 -> Records
indefinitely, splitting bag files at 30 second intervals

I will, however, also open an issue stating that --size and --duration so
always specify overall recording and --split should be moved into
--split-duration and --split-size which might be mutually exclusive
options, but could be used with --duration and --size.

—
Reply to this email directly or view it on
GitHubhttps://github.com/ros/ros_comm/pull/2#issuecomment-8111641.

wjwwood:
The issue is here: https://github.com/ros/ros_comm/issues/4

PR #6, opened 2012-09-05
Logging

ahendrix:
I've update the way rospy does logging so that it uses the python logging module throughout, and I've attempted to keep the original behavior.


ahendrix:
I've added the ROS services to allow log-level adjustment, built on top of the previous patches.

PR #7, opened 2012-09-21
Initial IPv6 support

schneider42:
Adds initial IPv6 support for ros_comm.

It is enabled by setting ROS_IPV6=on. See http://www.ros.org/wiki/ros_comm6 for more details.


roalter:
This pull request is also available as fork at https://github.com/roalter/ros_comm. Either this pull request can be integrated or the fork goes upstream again.

schneider42:
Merged the with the latest upstream version of ros_comm

PR #47, opened 2013-01-05
Roswtf

mirzashah:


utilities/roswtf/src/roswtf/py_pip_deb_checks.py
```diff
@@ -0,0 +1,228 @@
...
+    for key in py_to_deb_packages:
+        check_and_print(key)
+
+
+def is_command_runnable(cmd_name):
+    """Indicates if a command is runnable on Unix/Linux"""
+    try:
+        rc = subprocess.call('command -v ' + cmd_name,
+            shell=True,
+            stdout=open(os.devnull, 'w'),
```
wjwwood: You should just use `subprocess.PIPE`, that will block the I/O

utilities/roswtf/src/roswtf/py_pip_deb_checks.py
```diff
@@ -0,0 +1,228 @@
...
+
+def wtf_check(ctx):
+    """Check implementation function for roswtf"""
+    from roswtf.rules import warning_rule, error_rule
+    for r in warnings:
+        warning_rule(r, r[0](ctx), ctx)
+    for r in errors:
+        error_rule(r, r[0](ctx), ctx)
+
+#Not used currently
```
wjwwood: Does this mean the code below this is unused?

If so, then remove it and put that code in another branch if you want to save it for later?

wjwwood:
Also I will show you how to update the pull request and then squash the commits into a single commit which looks nicer on the groovy-devel commit history.

Especially in something like ros_comm which has like 22 packages in it, the commit messages should indicate what is being affected, for example:

`Integrated in core python script checks and rosdep db check`

Should probably be:

`roswtf: Integrated in core python script checks and rosdep db check`

PR #50, opened 2013-01-08
Use roslaunch arg resolution in rosparam text

stonier:
Would be very useful to do things like

```
rosparam param="app_lists">$(arg app_lists)</rosparam>
```

or even use `$(find ...)` embedded in the raw yaml. This works for `<param.../>` where the value is a tag attribute.  Obvious advantage with rosparam is to be able to parse lists and dictionaries. Over the last three years I keep running into this and seem to be finding more situations where it's desirable now I'm using args alot more.

The fix here is a simple one-liner. It works, but is there intentionally a reason this wasn't implemented?

### DIFF
--- a/tools/roslaunch/src/roslaunch/xmlloader.py
+++ b/tools/roslaunch/src/roslaunch/xmlloader.py
@@ -207,7 +207,7 @@ class XmlLoader(loader.Loader):
             
             # load is the default command            
             cmd = cmd or 'load'
-            self.load_rosparam(context, ros_config, cmd, param, file, _get_text(tag), verbose=verbose)
+            self.load_rosparam(context, ros_config, cmd, param, file, self.resolve_args(_get_text(tag), context), verbose=verbose)
 
         except ValueError as e:
             raise loader.LoadException("error loading <rosparam> tag: \n\t"+str(e)+"\nXML is %s"%tag.toxml())


dirk-thomas:
I think there might be a problem with that approach. The tag may contain i..e backticket stuff like `$(ROS_ROOT)` which would than be expanded.

Another way to approach your use case would be to add a value attribute to the rosparam tag which can be used instead of the content of the tag. That way only desired stuff goes into it and gets expanded as any other attribute.

stonier:
Good point. I'll try out the alternative idea and if it works, send a pull request with that.

dirk-thomas:
Since adding a value attribute will be significantly different from this description marking this pull request as closed. Looking forward for a pull request achieving the goal with a value attribute.

stonier:
ok

jbohren:
@dirk-thomas 
1. Why would non-portable environment variable usage like "$(ROS_ROOT)" go into rosparam text? 
2. Why should this be made into a value attribute, especially when the text is going to be full of line-breaks and formatting?

If we're worried about backwards-compatibility, why not just add an additional attribute that _enables_ subarg substitution in a given `<rosparam>` tag?

jbohren:
And that's not even an issue if someone puts an environment variable like "$ROS_ROOT" into text that gets run through argparse. It will only be a problem if someone is trying to do bash command subtitution with "$()" which sounds like they're doing something wrong already. See the additional test case in my PR: https://github.com/ros/ros_comm/pull/206

dirk-thomas:
Can we be sure that no code is using "$(..)" inside rosparam text? I would say no and therefore prefer not to break BC if there is no extremely strong need/desire for it.
- The "value" attribute was a suggested way to achieve the behavior as it is consistent with how substitution already happens.
- A boolean flag like "enables" would also be a way to implement. But it is obviously less intuitive for the user.
- For changing the behavior in an backward incompatible way I would need either strong prove that it is nowhere known to be used and/or votes from several people to intentionally break BC.

I am agnostic to how the feature gets realized.

jbohren:
What about adding an attribute called "argparse" which is true by default for all tags (like it is now) and allow users to disable argparse for any given tag, including `<rosparam>`?

dirk-thomas:
What is the difference to the third bullet if the attribute is true by default (except with the option to disable it)? It changes current behavior which means it might break working code for users and they need to make modification to keep it working.

jbohren:
Well, I think most people would agree that the fact that argument substitution doesn't happen in rosparam tags is a bug. If we add an attribute to enable argparse and the attribute is true by default then:
1. the attribute can be used in all tags
2. rosparam tags behave in an expected way

Additionally, I think the chance that there are any `$(...)` strings in rosparam tags is very low. The chance that someone put a valid argparse command into a rosparam tag without expecting it to be parsed by argparse is even lower. On the _very_ rare chance that someone wants to have `$(...)` strings in their rosparam yaml, then they can enable this flag.

An alternative option is to turn it on by default, but instead of throwing an exception when it reads a `$(...)` pattern which isn't a valid argparse command, the parser could just leave the string as-is.

stonier:
I'm fairly agnostic as well, but I'm inclined to think that two almost identical ways to feed input into the rosparam tag is a bit strange. Some digging:
- backticks are already not allowed in rosparam, e.g. `<rosparam param="dude">`echo $(ROS_ROOT)`</rosparam>` results in lots of profuse bleeding:

```
error loading <rosparam> tag: 
        Error within YAML block:
        while scanning for the next token
found character '`' that cannot start any token
  in "<unicode string>", line 1, column 1:
    `echo $(ROS_ROOT)`
    ^

YAML is:
`echo $(ROS_ROOT)`
XML is <rosparam param="dude">`echo $(ROS_ROOT)`</rosparam>
```
- You can feed dollar signs into the current rosparam and it will push the string onto the rosparam server, e.g. `<rosparam param="dude">$(ROS_ROOT)</rosparam>`:

```
> rosparam get /dude
$(ROS_ROOT)
```
- Running it through `resolve_args` as per this pull request, of course, results in an error:

```
Unknown substitution command [ROS_ROOT]. Valid commands are ['find', 'env', 'optenv', 'anon', 'arg']
```

An alternative option may be in this case to have a flag passed to `resolve_args` telling it to be quiet - i.e. don't produce an error for an unrecognised command, just pass the command/arg string through. That way you _shouldn't_ break anything and you get the extra functionality for free. Only downside is that it might take a wee bit more debugging if you accidentally spell 'ifnd' wrong. And even then, you could print a warning to help users.

In any case, this should be a hydro feature update rather than a groovy change. I'm all for avoiding breaking things in a release, but definitely also in favour of new, useful features with a new release.

dirk-thomas:
After discussion the topic again the expected breakage of existing code when the parameter value is substituted is considered to problematic to change behavior. Therefore we should go with an optional boolean attribute for <rosparam> called `subst_value` which defaults to `false`. If it is set to `true` the value of the rosparam will be processed with the substitution function.

@jbohren if you could create a pull request for that it can be applied (even to groovy-devel) since it only adds new functionality without altering existing behavior.

Issue #37, opened 2012-12-21
Check licensing on ifaddrs.py in rosgraph


ablasdel:
https://github.com/ros/ros_comm/blob/master/tools/rosgraph/src/rosgraph/ifaddrs.py

apparent source:
http://carnivore.it/2010/07/22/python_-_getifaddrs

schneider42:
I'll ask the author if he want to license it under BSD. If necessary I'll reimplement the file.

dirk-thomas:
Any news on this? Else we should reimplement it...

schneider42:
I've asked the author directly, but had no response. I'll try again on his blog, and reimplement it over the weekend if nothing happens.

schneider42:
I've added a pull request for this issue: https://github.com/ros/ros_comm/pull/130

dirk-thomas:
That file has been removed and replaced with an existing Python module.

PR #130, opened 2013-01-20
Issue #37: Use netifaces package instead of non-BSD ifaddrs file

schneider42:
I've removed the ifaddrs.py file and the code is now using the python-netifaces package.

The change is straight forward, but i currently do not have a ROS environment to test it properly.

It has a slightly different behavior with link local addresses. It returns them as "<address>:<interface>", while the other implementation returns them as "<address>. I am not sure if this has an impact and it would be great if someone could quickly test it.


dirk-thomas:
Please comment on the issue as soon as you have tested the modified version.
Does it also work with older versions of netifaces, i.e. 0.5 on Oneiric?

PR #178, opened 2013-02-25
fixed error message, was displayed the wrong way

kalectro:


dirk-thomas:
Thanks for the patch. I will apply a modified version because line 114/115 has the same issue.

PR #187, opened 2013-03-06
Fixed #186, roslaunch --files works even when anon ids are not unique.

### LINKED ISSUE #186: roslaunch --files prints error if multiple anon nodes have same id
When running:

```
roslaunch --files package_name launch_file.launch
```

I get the following error:

```
roslaunch file contains multiple nodes named [/$(anon foo)].
Please check all <node> 'name' attributes to make sure they are unique.
Also check that $(anon id) use different ids.
```

Note that when actually launching the launch file, this will not be a problem because the anonymous node names will be expanded with unique suffixes.  Also, the list of files does not relate to the expansion of anonymous node names, so the file list should be printable regardless.

This is similar in spirit to bugs #94 and #65, both of which suffer because roslaunch simply prints an error and quits when it finds non-unique node names instead of doing as much as possible to help you find your error.

hershwg:
I'm pretty sure it is safe to expand anonymous node names when printing the list of files loaded.

This just changes a single False to True.

### DIFF
--- a/tools/roslaunch/src/roslaunch/rlutil.py
+++ b/tools/roslaunch/src/roslaunch/rlutil.py
@@ -235,7 +235,7 @@ def print_file_list(roslaunch_files):
     from roslaunch.config import load_config_default, get_roscore_filename
     import roslaunch.xmlloader
     try:
-        loader = roslaunch.xmlloader.XmlLoader(resolve_anon=False)
+        loader = roslaunch.xmlloader.XmlLoader(resolve_anon=True)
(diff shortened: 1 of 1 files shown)


hershwg:
Here is an example launch file which shows the problem:

```
<launch>
  <node name="$(anon foo)" pkg="rostopic" type="rostopic" args="echo bar"/>
  <node name="$(anon foo)" pkg="rostopic" type="rostopic" args="echo bar"/>
</launch>
```

put that in foo.launch.  Then

```
roslaunch foo.launch
```

will succeed, but

```
roslaunch --files foo.launch
```

will fail.  This patch makes it so they both work.

Issue #101, opened 2013-01-09
roslaunch: Allow substitution args in rosparam list


130s:
Migrated from https://code.ros.org/trac/ros/ticket/4015
Reported by:    weiin   Owned by:   kwc
Priority:   minor 
Last modified ago: Jul 2012

---

I am trying to achieve this:  http://answers.ros.org/question/39085/what-is-the-correct-way-to-set-args-in-rosparam/ As roslaunch parses text inside <rosparam> using yaml parser, substitution args cannot be passed into it when there is a parameter list.

malban:
+1  This would make managing launch configurations much easier.

dirk-thomas:
Duplicate of #50.
As discussed there this should be achieved differently.

PR #203, opened 2013-03-29
Adding substitution_args support for rosparam tags (yaml-based rosparam)

jbohren:
Also added a simple unit test.

The substitution happens to the raw text, before it goes into the YAML parser.

This is the feature requested here: https://github.com/ros/ros_comm/issues/101


dirk-thomas:
Duplicate of #50.

PR #204, opened 2013-03-29
Replacing '%' in ipv6 address by '/' to avoid syntax conflict

yichoe:
In Windows, roswtf throws the error while checking local network configuration. It comes from  the expression of IPv6. The IPv6 uses a zone index to provide routing information. The zone index is appended to the address, separated by '%'. This notation causes syntax conflict with Python when roswtf deals with it.

This syntax conflict happens when roswtf appends IP address to the context to just show it to user after checking all. The new is to avoid this conflict through replacing '%' in IP addresses by '/'.

PS. For URI, ISOC is finding solution to avoid conflict like here:

http://tools.ietf.org/html/draft-fenner-literal-zone-02

### DIFF
--- a/utilities/roswtf/src/roswtf/network.py
+++ b/utilities/roswtf/src/roswtf/network.py
@@ -52,8 +52,14 @@ def ip_check(ctx):
     global_ips = [ ip for ip in resolved_ips if not ip.startswith('127.') and not ip == '::1']
 
     remote_ips = list(set(global_ips) - set(local_addrs))
-    if remote_ips:
-        return "Local hostname [%s] resolves to [%s], which does not appear to be a local IP address %s."%(socket.gethostname(), ','.join(remote_ips), str(local_addrs))
+
+    # replace '%' symbol in ipv6 address by '/' to avoid syntax conflict
+    refined_ips = list()
+    for val in remote_ips:
+        refined_ips.append(val.replace('%', '/'))
+        
+    if refined_ips:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The patch should not alter the IPv6 address with the zone index itself but use a different way to construct the string.

dirk-thomas:
After reviewing the current implementation I do not see a way this could break. Three `%s` are substituted with three values. Even if one of the strings contain `%` characters it would still work.

I will close the pull request as I think it is not valid.

If you can provide more information (i.e. the values of `socket.gethostname()`, `remote_ips` and `local_addrs` as well as the expected and actual output) I will look into it again.

todayman:
I also have this issue.  Here are the values you requested:

```
socket.gethostname() =  MyComputer
global_ips =  ['fe80::226:4aff:fe0c:ec46%eth0', 'fe80::226:8ff:fee9:21c5%wlan0', '128.220.224.124', '10.188.166.120']
local_addrs =  ['127.0.0.1', '128.220.224.124', '10.188.166.120']
remote_ips =  ['fe80::226:4aff:fe0c:ec46%eth0', 'fe80::226:8ff:fee9:21c5%wlan0']
```

At  utilities / roswtf / src / roswtf / rules.py:62, I get the error: `TypeError: float argument required, not dict`.

Maybe this isn't the best fix, but it is a real issue.  I'm runing Linux.

PR #211, opened 2013-04-08
Patch for missing exception handling when using netifaces in rosgraph/network.py

francoisferland:
We encountered a small bug when using the current ros_comm version in groovy. When netifaces.ifaddresses(...) is used on one of our CAN bus network interfaces, it throws an unhandled exception. This might actually be the driver's fault, but we cannot do anything about it right now. 

This occurred when roscore was launched, and it meant we couldn't launch anything on our robot. I tracked the source of the problem to a recent addition in rosgraph/network.py, and I simply added some exception handling/ignoring. Similar exception handling was done with netifaces in a different condition a few lines ahead of that addition. Unfortunately, I cannot test the change with many different configurations.

### DIFF
--- a/tools/rosgraph/src/rosgraph/network.py
+++ b/tools/rosgraph/src/rosgraph/network.py
@@ -215,7 +215,7 @@ def get_local_addresses():
             try:
                 v4addrs.extend([addr['addr'] for iface in ifaces if socket.AF_INET in netifaces.ifaddresses(iface) for addr in netifaces.ifaddresses(iface)[socket.AF_INET]])
                 v6addrs.extend([addr['addr'] for iface in ifaces if socket.AF_INET6 in netifaces.ifaddresses(iface) for addr in netifaces.ifaddresses(iface)[socket.AF_INET6]])
-            except: pass
+            except ValueError: pass
 
         if use_ipv6():
             return v6addrs + v4addrs
--- a/tools/rosgraph/src/rosgraph/network.py
+++ b/tools/rosgraph/src/rosgraph/network.py
@@ -209,8 +209,14 @@ def get_local_addresses():
         # unix-only branch
         import netifaces
         ifaces = netifaces.interfaces()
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The patch looks bogus as it iterates twice over "ifaces":
- once in the outer for loop
- once in the list comprehension

francoisferland:
You're right ... I admit I wrote it pretty fast, and I'm not that well versed in Python list comprehensions. I will look into it further.

francoisferland:
This should be better.

dirk-thomas:
You mentioned that the same exception is also handled "above". Do you mean the this line: https://github.com/ros/ros_comm/blob/groovy-devel/tools/rosgraph/src/rosgraph/network.py#L207
It is catching a KeyError while your patch catches a ValueError. Can you please describe which function raised what exception in your case?

francoisferland:
Yes, that's where I saw it. This condition doesn't seem to get run however, since _use_netifaces is forced to False (see line 73 and 84). But the _is_unix_like_platform() case ignores it and use netifaces anyway.

The problematic call is on netifaces.ifaddresses(). In our case, it throws a ValueError exception with message "You must specify a valid interface name". Our interface name is correctly enumerated in netifaces.interfaces() and it's a valid network interface, just not from ifaddresses() point of view. I guess the KeyError catch at line 207 is there to avoid cases where the 'addr' field isn't present in ifaddresses() output, but it wouldn't catch the same error.

I've looked for similar (not related to ROS) cases, and I found this old, still active bug report: https://bugs.launchpad.net/ubuntu/+source/netifaces/+bug/753009. This is pretty much the same problem we have, but it looks like it might take a while before it gets patched.

dirk-thomas:
Thanks for providing the patch.
I have created an updated pull request (#213) based on your feedback which cleans up some legacy code and makes the exception handling a bit more explicit. I will mark this pull request as closed in favor of #213.

jspricke:
I've had the same problem on Ubuntu 12.04 and fixed it by installing python-netifaces from Ubuntu 12.10.

Issue #210, opened 2013-04-04
rospy: logging levels in ros-fuerte-ros-comm 1.8.16-0precise-20130313-0010-+0000


jarvisschultz:
It seems this issue is likely related to the changes mentioned in issues #14 and #5 and pull request #6.  It seems that in this version the `set_logger_level` and `get_loggers` services are not implemented correctly.  The following snippet adapted from [this tutorial](http://www.ros.org/wiki/rospy_tutorials/Tutorials/WritingPublisherSubscriber) can illustrate the point:

```
#!/usr/bin/env python
import rospy
from std_msgs.msg import String

def talker():
    pub = rospy.Publisher('chatter', String)
    rospy.init_node('talker', log_level=rospy.INFO)
    while not rospy.is_shutdown():
        str = "hello world %s" % rospy.get_time()
        rospy.loginfo(str)
        pub.publish(String(str))
        rospy.sleep(1.0)

if __name__ == '__main__':
    try:
        talker()
    except rospy.ROSInterruptException:
        pass
```

All I have changed is explicitly set the `log_level` to be `rospy.INFO`.  What actually happens is that the logging actually gets set to DEBUG level, and tons of messages are printed to the terminal.  Attempting to use _rxconsole_ to adjust this level results in a warning that says

```
Failed to call service [/talker/get_loggers]. Did the node go away?
```

Note that in the previous repository version (v1.8.15) the logging does not work either.  I can set the value explicitly in the `init_node` method, but it always ends up behaving like INFO in the terminal.

dirk-thomas:
It looks like that #6 changed the semantic of the log_level argument to only work correctly with logging log levels (i.e. `logging.INFO`). I will check that with @ahendrix who provided the pull request.

dirk-thomas:
Needs backport of #212.

dirk-thomas:
Backported in a2c724551340a7a47516577a4aa7a3b70cb6129e

PR #212, opened 2013-04-08
fix log level of RosOutHandler by converting rospy level to expected logging level (#210)

dirk-thomas:


dirk-thomas:
@tfoote @wjwwood Please review and merge.

tfoote:
Is it worth keeping the backwards compatibility of a private variable? If so maybe add a comment on why that's there. 

otherwise +1 for merge

dirk-thomas:
I removed the private variable from the client module.

wjwwood:
+1

PR #213, opened 2013-04-09
catch ValueError on netifaces.ifaddresses(), cleanup disabled code

dirk-thomas:
Based on patch provided in #210.

@tfoote @wjwwood Please review and merge.


wjwwood:
+1

Issue #128, opened 2013-01-20
rospy topic connection fails if publisher restarts on same xml-rpc port


adamantivm:
When using a fixed xml-rpc port number for a ROS node which publishes a topic, if the node is restarted then topic listeners will not re-connect to the publisher to receive new messages, even when the publisher is following the appropriate topic registration procedure on start-up (and unregistration on shutdown).

This problem probably went unnoticed so far because until now each new ros process was randomly choosing their xml-rpc port.

The procedure I'm using for reproducing this is unfortunately not completely straight-forward. I'm using a lightly patched version of rospy which allows specifying an xml-rpc port from parameters: https://github.com/ros/ros_comm/pull/127. I have cloned and made this fork of ros_comm and configured a ROS workspace to use this version of rospy instead of the installed one from repos.

I built the following small test program which uses rospy to start a node on a fixed port number and publish a test message:

```
#!/usr/bin/python
import roslib; roslib.load_manifest('rospy')
import rospy
from std_msgs.msg import String
import sys

if len(sys.argv) > 1:
    port = int(sys.argv[1])
else:
    port = 33330

rospy.init_node('testport',port=port,log_level=1,tcpros_port=port+1)

publisher = rospy.Publisher('/testpub', String)
rospy.sleep(1)

r = rospy.Rate(2)
i = 0

while not rospy.is_shutdown():
    publisher.publish("Out goes msg %d" % i)
    r.sleep()
    i += 1

print "Shutting down, please wait"
publisher.unregister()
rospy.sleep(1)
```

In order to reproduce the issue, open a roscore and a rostopic echo, then start test.py, let it run for a little while, kill it and start it again.
In other words:

```
roscore
rostopic echo /testpub
./test.py
# wait 5 seconds and hit ^C
./test.py # for the second time
```

You will notice that the first time it runs, rostopic can echo the published messages but the second time (and subsequent times) it cannot, unless rostopic echo is restarted.

This may be related with issue #111 if unregistration is indeed failing

adamantivm:
Looking at a network trace I can see that both times the publisher is making the appropriate registerPublisher xmlrpc call (and also unregisterPublisher when shutting down), but only the first time it gets a requestTopic xmlrpc call from the subscriber - the second time it is ignored

adamantivm:
After some digging in the rospy code everything seems to indicate that Subscriber disconnection from unregistered Publishers is not implemented, is this possible?

[...]

PR #215, opened 2013-04-10
fix closing and deregistering connection when connect fails (#128)

dirk-thomas:
addressed #128


dirk-thomas:

wjwwood:
I assume this line:

https://github.com/ros/ros_comm/blob/8231b4b1559a10d436424868fe45ca232500a692/clients/rospy/src/rospy/impl/tcpros_base.py#L784

does `self.done = True` at some point?

dirk-thomas:
Yes, the parent class sets the boolean flag (https://github.com/ros/ros_comm/blob/8231b4b1559a10d436424868fe45ca232500a692/clients/rospy/src/rospy/impl/transport.py#L98) and calls the callback which performs the `remove_connection()` which was missing before.

wjwwood:
+1

PR #226, opened 2013-05-14
roscpp: fix issues with the packet reordering in the UDP transport

xqms:
This patch improved the UDP transport from aborting immediately on the first missing packet (!) to working in a stable manner for me.

When an out-of-order packet arrived, only the requested read size of the
packet was saved in the reordering buffer. The rest was discarded, which
led to protocol synchronization problems.

Also, out-of-order blocks should be discarded entirely.

### DIFF
--- a/clients/roscpp/src/libros/transport/transport_udp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_udp.cpp
@@ -340,6 +340,9 @@ int32_t TransportUDP::read(uint8_t* buffer, uint32_t size)
   {
     TransportUDPHeader header;
 
+    // Get the data either from the reorder buffer or the socket
+    // copy_bytes will contain the read size.
+    // from_previous is true if the data belongs to the previous UDP datagram.
     uint32_t copy_bytes = 0;
     bool from_previous = false;
     if (reorder_bytes_)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you please provide some example programs you used to validate the changes? At best in the form of unit tests?

xqms:
Here is a small package that demonstrates the problem: http://x-quadraht.de/roscpp_udp_test.tar.bz2

It contains a shell script `do_test.sh` that setups a virtual low-quality network connection and runs a launch file with a sample publisher and subscriber. As the script requests root privileges via sudo, it's not really suited for a real unit test.

If I run the `do_test.sh` on my machine with the stock roscpp, I get the following error message within a few seconds:

```
[ERROR] [1369931930.706535776] /subscriber: a message of over a gigabyte was predicted in tcpros. that seems highly unlikely, so I'll assume protocol synchronization is lost.
```

The reception of messages then silently (!) stops.

The patched roscpp keeps running and receiving messages.

dirk-thomas:
Thank you for the patch - I have also committed the (slightly modified) test script.

Regarding the "silently (!) stops" receiving message: the error message should be pretty clear about closing the connection (hence not receiving any further data). I think that should be enough to inform the user about the problem.

PR #227, opened 2013-05-16
adding function for checking whether all nodes have launched

YoheiKakiuchi:
Adding functions for checking whether all nodes have launched.
One is function which can be called through python.
One is method for xmlrpc.

### DIFF
--- a/tools/roslaunch/src/roslaunch/netapi.py
+++ b/tools/roslaunch/src/roslaunch/netapi.py
@@ -93,3 +93,29 @@ def list_processes(roslaunch_uris=None):
             pass 
     return procs
 
+def list_complete(roslaunch_uris=None):
+    """
+    @param roslaunch_uris: (optional) list of XML-RPCS. If none
+        are provided, will look up URIs dynamically
+    @type  roslaunch_uris: [str]
+    @return: list of roslaunch complete
+    @rtype: [bool]
+    """
+    if not roslaunch_uris:
+        roslaunch_uris = get_roslaunch_uris()
+        if not roslaunch_uris:
+            return []
+
+    comps = []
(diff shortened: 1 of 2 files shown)


tools/roslaunch/src/roslaunch/netapi.py
```diff
@@ -93,3 +93,29 @@ def list_processes(roslaunch_uris=None):
...
+        try:
+            r = xmlrpclib.ServerProxy(uri)
+            code, msg, val = r.get_registrations_complete()
+            if code == 1:
+                comps.append(val)
+        except:
+            #import traceback
+            #traceback.print_exc()
+            # don't have a mechanism for reporting these errors upwards yet
+            pass
```
dirk-thomas, line 120: This looks a bit weird. By just passing the calling code can not even figure out which uri has actually been completed and which not.

dirk-thomas:
Please also provide some more context for the patch. How should the new API be used in the future? Will you extend the patch to expose the functionality somehow?

YoheiKakiuchi:
One big problem in roslaunch is that user can not decide the order to execute process.
We would like to use programs which were not written with ROS manner,
We would like to control timing to execute process.
So, it is useful to know all nodes launched in a launch file.

dirk-thomas:
The "problem" in roslaunch about the unspecified startup order is an intentional design decision. Please see http://ros.org/wiki/roslaunch/Architecture#line-13 for more information.

dirk-thomas:
Due to lack of response, the fact that the proposed API should not be exposed anywhere in the CLI and the inconsistent/incomplete API/return value (as commented above) this issue is getting closed.

If you have any further feedback please feel free to comment and/or update the pull request.

PR #230, opened 2013-05-23
roslaunch completion for args

dornhege:
Extends the default completion in rosbash by completing roslaunch args, using the --ros-args functionality in roslaunch.

The completion in rosbash could be removed once this is released.


dirk-thomas:
The patch does not work for me. The roslaunch-complete script always raises an RLException that specific args are not defined. I am not sure how it is supposed to work - could you please mention an example roslaunch file and commands which you tried that show the new completion for arguments?

Can you also base you patch against the existing auto completion in rosbash? That would keep the diff minimal and easier to test. It could be moved over to roslaunch separately (if necessary).

dornhege:
That's odd. I'm sure this is related to #231 as currently the --ros-args command will fail for required args. So this should work for any launch file with args that have default values and will work for any once #231 is fixed. An example would be `roslaunch turtlebot_dashboard turtlebot_dashboard.launch` (completes the base arg).

However, you should never see an RLException. The default behavior should be to do nothing in that case and just use normal completion.

dornhege:
Regarding the completion: Do you only want to see the diff or should I move the completion to rosbash in the ros/ repository?

dirk-thomas:
I think a patch against rosbash would be the easiest to apply. I have looked a bit into #231 and I don't see any straight forward way to realize it.

Regarding the auto completion: I couldn't get it to work. It also does conceptually not work when passing a launch file as a single argument (instead of package name plus launch file).

dornhege:
When you say "couldn't get it to work": Are you referring to the full completion or just running roslaunch-complete? Obviously both should work, but that would help me to determine what might be wrong.

Passing launch files as single arguments should also work. I can do the following:

```
$ roscd turtlebot_dashboard/launch/
$ roslaunch-complete turtlebot_dashboard turtlebot_dashboard.launch 
base
$ roslaunch-complete turtlebot_dashboard.launch 
base
$ roslaunch turtlebot_dashboard.launch ba<TAB>
$ roslaunch turtlebot_dashboard.launch base 
$ roslaunch turtlebot_dashboard turtlebot_dashboard.launch ba<TAB>
$ roslaunch turtlebot_dashboard turtlebot_dashboard.launch base 
```

dirk-thomas:
Only the new roslaunch-complete part seems to not work for me. But I have only tried it with launch files with (required) arguments. Can you update this pull request to match the one in ros/ros#19 - so remove the env hook?

dirk-thomas:
Waiting for ros/ros#19 to work completely.

dirk-thomas:
I gave it another try. None of you examples above auto complete for me. Only `--<TAB>` auto completes the available options.

roslaunch-complete actually returns the valid arguments but it also outputs warning from pacpkages in my workspace. I will post a comment on the ros pull request how to update it that it works.

PR #233, opened 2013-05-24
modified roslaunch (find PKG) to consider path behind it for resolve strategy

dirk-thomas:
This is a reduced patch of #228 which only fixes the behavior of `$(find)` by considering the path following the command for the resolving.


tfoote:
+1

dirk-thomas:
Since there have been no further comments on this pull request I will merge it and update #228 to be potentially mergable again.

Issue #234, opened 2013-05-29
rosnode_ping) Not handled error "expected string or buffer"


130s:
When a node "node_a" is shut down while looping `rosnode_ping('node_a', 1)`, following error occurs and loop exits.

```
expected string or buffer
  File "/usr/lib/python2.7/unittest/case.py", line 332, in run
    testMethod()
  File "/home/n130s/catkin_ws/src/ros_comm/tools/rosnode/test/test_rosnode.py", line 252, in test_rosnode_kill
    is_pingged = rosnode.rosnode_ping(n, max_count=1)
  File "<string>", line 349, in rosnode_ping
  File "/usr/lib/python2.7/xmlrpclib.py", line 1547, in __init__
    type, uri = urllib.splittype(uri)
  File "/usr/lib/python2.7/urllib.py", line 1070, in splittype
    match = _typeprog.match(url)
```

ros_comm SHA bef19f434f9363daa907082a7ce099265da6711a

dirk-thomas:
The code you reference (test_rosnode.py", line 252) seems to be different from the repository. Can you please describe if you have modifications to your code base and if yes reference them somehow?

130s:
Looks like I referred to [the commit](https://github.com/130s/ros_comm/commit/658817cafbff2461605c3cbcddea839278ffeca4#diff-45f4389bb904a1558d132cf73991554bR252) that only exists in [my pull request](https://github.com/ros/ros_comm/pull/244) that is closed.

% I'm sorry, I just found that I didn't respond to your comment in the PR just mentioned above..

dirk-thomas:
Since both related PRs ended up being closed I will also mark this ticket as closed. If the issue is still relevant please comment with more information and it can be reopened.

PR #235, opened 2013-05-29
rosnode_ping) Fix to https://github.com/ros/ros_comm/issues/234

130s:
@dirk-thomas Please review. I'm not sure if this patch doesn't break anything else.


tools/rosnode/src/rosnode/__init__.py
```diff
@@ -346,7 +346,8 @@ def rosnode_ping(node_name, max_count=None, verbose=False):
                             if verbose:
                                 print("node url has changed from [%s] to [%s], retrying to ping"%(node_api, new_node_api))
                             node_api = new_node_api
-                            node = xmlrpclib.ServerProxy(node_api)
+                            if node_api:
+                                node = xmlrpclib.ServerProxy(node_api)
```
dirk-thomas, line 350: In case that `node_api` is `None` it seems to be a bad idea to continue. Shouldn't it output an error message and return False?

PR #239, opened 2013-06-10
Fix crash for bag migration

cmansley:
This appears to be a copy and paste error. You should check the
rename rule order and not the last element of the list.

### DIFF
--- a/tools/rosbag/src/rosbag/migration.py
+++ b/tools/rosbag/src/rosbag/migration.py
@@ -763,7 +763,7 @@ class MessageMigrator(object):
                     rulechain.chain[otherind].location, r.location, r.old_type, r.location)
                 return
             else:
-                if rulechain.rename and (r.order >= rulechain.chain[-1]):
+                if rulechain.rename and (r.order >= rulechain.rename.order):
                     print >> sys.stderr, "WARNING: Update rule [%s] has order number larger than rename rule [%s]. Ignoring"%(
                         r.location, rulechain.rename.location)
                     return


dirk-thomas:
Can you please provide your migration rule file which exhibits the issue?

cmansley:
So, I have a migration rule that induces this bug. However, I cannot provide that information publicly. 

I can describe in detail what happens. It appears as if around line 549, if the `iteritems` function provides the rules out of order, it is possible for a rename rule to be in the rulechain before an update rule is in the rule chain. Even though the ordering of the two rules is reversed. 

Then, there is a check:

``` python
if rulechain.rename and (r.order >= rulechain.chain[-1]):
```

which crashes because there are no rules in the rule chain, only a single rename. If you look right before the diff in the file, you can see all of the other checks, ensure `rulechain.chain` exists before checking it.

All I did was make the if-check match the debug print statement it encloses and the offending file works just fine. I believe this is a copy and paste error with line 752.

dirk-thomas:
By reading the source I agree with the changes of the pull request. But still I wanna try it with a rule file first before merging it. If you can't provide a simple version of yours I will create one for verifying it.

cmansley:
Yeah, I apologize about that. I tried to create a "fake" migration rule that would trigger the failure, but I was unsuccessful. I think it has something to do with the ordering coming out of the dict. So, maybe the hashes need to be specific.

dirk-thomas:
I was able to reproduce the issue with two bag migration rules: one which renames message `Foo` into `FooRenamed` and one which just modifies the fields of `Foo`.

This currently raises `IndexError: list index out of range` since `rulechain.chain` in line 766 is an empty array. With the patch applied loading of the migration rules works. Depending on their order number the warning from line 767 is printed.

PR #243, opened 2013-06-19
fix 'rostopic echo' for submessages of type uint8[] (fix #242)

### LINKED ISSUE #242: rostopic echo of uint8[] seems to be displayed as a string, rather than array
Trying to rostopic echo a message with a uint8[] array (see the k_effort field in [atlas_msgs/AtlasCommand](https://bitbucket.org/osrf/drcsim/src/7ff63e8aab47973b10539420d5a8685aa2bb9384/ros/atlas_msgs/msg/AtlasCommand.msg?at=default) in drcsim as an example) causes the array not to be displayed properly; it seems to be interpreted as a string.

dirk-thomas:


dirk-thomas:

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -407,7 +407,7 @@ def _sub_str_plot_fields(val, f, field_filter):
     return None
 
 
-def _str_plot(val, time_offset=None, current_time=None, field_filter=None):
+def _str_plot(val, time_offset=None, current_time=None, field_filter=None, type_information=None):
```
tfoote, line 410: This argument does not appear to be used.

dirk-thomas, line 410: Correct, but still necessary to have the same signature as `custom_strify_message`.

tfoote:
+1

wjwwood:
+1

PR #245, opened 2013-06-28
CRITICAL BUG: proper escaping for ROS_PACKAGE_NAME definition

jbohren:
Without this, sometimes this ends up as (for a packange named `foo_pkg`):
`-DROS_PACKAGE_NAME="foo_pkg"`

Which breaks everything, since on the command line this needs to be the following so that the quotes end up in the variable:
`-DROS_PACKAGE_NAME=\"foo_pkg\"`

See this cmake thread here: http://www.cmake.org/pipermail/cmake/2007-June/014611.html

It looks like the [old rosbuild method](http://www.ros.org/wiki/rosbuild#ROS_.23defines) used to use single quotes:
`ROS_PACKAGE_NAME='"foo"'`

PR #246, opened 2013-06-29
Adding explicit cmake policy to rosconsole-extras.cmake

jbohren:
I don't know if this jives with the philosophy with the catkin/ros_comm maintainers but adding this to the `rosconsole-extras.cmake` prevents the user error (incorrect camke_minimum_required version) reported in https://github.com/ros/ros_comm/pull/245 from preventing packages to build. This `cmake_policty()` command should be local to this file only. I chose `cmake_policy()` instead of `cmake_minimum_required()` due to it's clearer meaning.

See the follwoing for reference: http://www.cmake.org/Wiki/CMake/Policies#Fixing_an_Interface_Breaks_Work-Arounds

### DIFF
--- a/tools/rosconsole/cmake/rosconsole-extras.cmake
+++ b/tools/rosconsole/cmake/rosconsole-extras.cmake
@@ -1,4 +1,5 @@
 # ros_comm/tools/rosconsole/cmake/rosconsole-extras.cmake
+cmake_policy(VERSION 2.8.3)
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I would rather not change the policies since this will affect other code parts which have no way of knowing that this is happening internally. Could you check if the alternate patch from #247 also fixes the issue for you?

jbohren:
Ahh yeah, #247 works too. Thanks!

PR #247, opened 2013-06-29
force policy before setting pp definition to ensure correct escaping (#245)

dirk-thomas:


jbohren:
:+1:

dirk-thomas:
Thanks for taking the time to investigate it.

jbohren:
> Thanks for taking the time to investigate it.

Right back at you.

PR #248, opened 2013-07-01
Separate rosbag_storage from rosbag

tfoote:
rosbag_storage contains the file manipulation tools but does not depend on the communication libraries. Allowing rosbag to be used at the API level w/o compiling against the communication libraries.  

rosbag_storage elements switched to console_bridge from rosconsole

This needs to be paired with: https://github.com/ros/roscpp_core/pull/12


dirk-thomas:
Please update the pull request so that it can be reviews and merged.

dirk-thomas:
Closing in favor of #299

PR #260, opened 2013-07-30
add convenience function for rostest with gtests

dirk-thomas:


dirk-thomas:
@wjwwood Please review.

hershwg:
Yay!  Thanks for working on this @dirk-thomas .  The example usage here in message_filters has 4 lines per test:
- test
- if test built
  - link libraries for test
- endif

Looks to me like the conditional could be removed if the containing conditional that says `if( CATKIN_ENABLE_TESTING)` were changed to `if( CATKIN_ENABLE_TESTING AND GTEST_FOUND )`.  Does that seem right?  Obviously that's only applicable to a given situation if all the tests in the conditional need gtest, but that is a common situation for me at least.

wjwwood:
+1

dirk-thomas:
Yes, the user land code might be able to check the conditions before to make sure `add_rostest_gtest()` always generates a target. Else it needs to check target existence before adding dependencies.

PR #264, opened 2013-08-02
make rospy nodes killable while waiting for master (fix #262)

### LINKED ISSUE #262: rospy nodes are hard to kill if master is not running
If you run a node before starting the master, you get the expected message:

```
Unable to register with master node [http://localhost:11311]: master may not be running yet. Will keep trying.
```

However, it doesn't respond to Ctrl-C or kill commands. The only way to stop it is to find the pid and use kill -9.

dirk-thomas:

### DIFF
--- a/clients/rospy/src/rospy/impl/registration.py
+++ b/clients/rospy/src/rospy/impl/registration.py
@@ -51,7 +51,7 @@ try:
 except ImportError:
     import xmlrpclib as xmlrpcclient
 
-from rospy.core import is_shutdown, xmlrpcapi, \
+from rospy.core import is_shutdown, is_shutdown_requested, xmlrpcapi, \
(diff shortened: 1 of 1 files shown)


wjwwood:
+1

clients/rospy/src/rospy/impl/registration.py
```diff
@@ -162,8 +162,17 @@ def clear(self):
         """
         Remove all registration listeners
         """
-        with self.lock:
+        if not is_shutdown_requested():
+            with self.lock:
+                del self.listeners[:]
+        else:
+            # when being in shutdown phase the lock might not be lockable
```
tfoote, line 169: For reference the expected holder of this lock is higher in this call stack.

tfoote:
+1

PR #269, opened 2013-08-05
change roslaunch resolve order (fix #256)

### LINKED ISSUE #256: roslaunch does not accept a package name if a file with the same name exists in the current directory
Reproduce with:

```
roscreate-pkg roslaunch_test
cd roslaunch_test
mkdir bin
mkdir launch
touch bin/roslaunch_test
echo "<launch/>" > launch/example.launch
cd bin
roslaunch roslaunch_test example.launch
```

Error output:

```

roslaunch: error: The following input files do not exist: example.launch
```

Without the file in `bin/` or with another working directory roslaunch executes without errors (and exits immediately as there are no nodes).

I am using roslaunch 1.9.47 installed from the binary package repository in ROS hydro.

dirk-thomas:
@tfoote Please review.

### DIFF
--- a/tools/roslaunch/src/roslaunch/rlutil.py
+++ b/tools/roslaunch/src/roslaunch/rlutil.py
@@ -85,24 +85,26 @@ def resolve_launch_arguments(args):
     if not args:
         return args
     resolved_args = None
-    top = args[0]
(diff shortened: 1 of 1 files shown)


tfoote:
+1

PR #272, opened 2013-08-08
Rolling buffer for ROSBAG

baxelrod:
I needed to have a rolling buffer of messages that I could dump after I saw an error occur.
It seems that having this functionality in vanilla rosbag would be quite useful, so I'm submitting a pull request to discuss it. 

Here's something I hacked together to do this. You can start it by running rosbag record --rolling-buffer 30s. 
edit: If you're willing to merge this, I'm willing to clean the code up, and make other small changes, write documentation etc. 

I listen to a topic, and when I get a message on that topic I do a dump. 

@jonbinney @KaijenHsiao


tfoote:
There's definitely been discussion about doing something like this.  I think overall it's a great idea. 

One other use case is to dump the data when something crashes and as such an inverse logic on the trigger might be useful, aka dump data if I stop hearing a watchdog. 

We should review the API and merge this.  The other option might be to keep it as a standalone application like a black box recorder.  But if we can support it at the library level that's great too.

baxelrod:
I'm not sure that I would put the watchdog functionality inside of rosbag
itself.

Since the dump is triggered by a message, any node can easily trigger it.
One could easily enough write a watchdog node that would print an error and
trigger a dump.

Edit: removed some junk that was added when I replied by email. 
—
Reply to this email directly or view it on
GitHubhttps://github.com/ros/ros_comm/pull/272#issuecomment-22367541
.

baxelrod:
Bump, what do you feel about that?

tools/rosbag/src/record.cpp
```diff
@@ -62,8 +62,8 @@
       ("topic", po::value< std::vector<std::string> >(), "topic to record")
       ("size", po::value<int>(), "The maximum size of the bag to record in MB.")
       ("duration", po::value<std::string>(), "Record a bag of maximum duration in seconds, unless 'm', or 'h' is appended.")
-      ("node", po::value<std::string>(), "Record all topics subscribed to by a specific node.");
-
+      ("node", po::value<std::string>(), "Record all topics subscribed to by a specific node.")
+      ("rolling-buffer", po::value<std::string>(), "Keeps a rolling buffer of messages of the specified number of seconds and dumps it upon recieving a message to the buffer_dump topic");
```
dirk-thomas, line 66: Spelling: recieving

tools/rosbag/src/record.cpp
```diff
@@ -189,6 +189,41 @@
...
+        multiplier = 3600.0;
+      else
+        throw ros::Exception("Duration unit must be s, m, or h");
+
+      opts.rolling_buffer_duration = ros::Duration(duration * multiplier);
+      if (opts.rolling_buffer_duration <= ros::Duration(0))
+        throw ros::Exception("Duration must be positive.");
+    }
+    else
+	ROS_ERROR("no rolling buffer");
```
dirk-thomas, line 225: Indendation and please enclose single line commands (especially macros) in curly braces.

tools/rosbag/src/recorder.cpp
```diff
@@ -206,7 +216,7 @@ int Recorder::run() {
     ros::NodeHandle nh;
     shared_ptr<int> count(new int(options_.limit));
     shared_ptr<ros::Subscriber> sub(new ros::Subscriber);
-    *sub = nh.subscribe<topic_tools::ShapeShifter>(topic, 100, boost::bind(&Recorder::doQueue, this, _1, topic, sub, count));
+    *sub = nh.subscribe<topic_tools::ShapeShifter>(topic, 10000, boost::bind(&Recorder::doQueue, this, _1, topic, sub, count));
```
dirk-thomas, line 219: Why are you increasing the queue so drastically?

tools/rosbag/src/recorder.cpp
```diff
@@ -491,15 +524,17 @@ void Recorder::doRecordSnapshotter() {
                 return;
             queue_condition_.wait(lock);
         }
-        
+
```
dirk-thomas, line 527: Please avoid whitespace changes in pull requests.

tools/rosbag/src/rosbag/rosbag_main.py
```diff
@@ -115,9 +116,11 @@ def record_cmd(argv):
     if options.size:        cmd.extend(["--size", str(options.size)])
     if options.node:
         cmd.extend(["--node", options.node])
+    if options.rolling_buffer:
+        cmd.extend(["--rolling-buffer", options.rolling_buffer])
 
     cmd.extend(args)
-
+
```
dirk-thomas, line 123: Please avoid whitespace changes in pull requests.

tools/rosbag/src/recorder.cpp
```diff
@@ -507,13 +542,15 @@ void Recorder::doRecordSnapshotter() {
             ROS_ERROR("Error writing: %s", ex.what());
             return;
         }
+        ROS_ERROR("started writing");
```
dirk-thomas, line 545: Why are these ROS_ERROR?

tools/rosbag/src/recorder.cpp
```diff
@@ -507,13 +542,15 @@ void Recorder::doRecordSnapshotter() {
...
 
         while (!out_queue.queue->empty()) {
             OutgoingMessage out = out_queue.queue->front();
-            out_queue.queue->pop();
-
+		
             bag_.write(out.topic, out.time, *out.msg);
+            out_queue.queue->pop();
```
dirk-thomas, line 551: Why is the pop moved below the write?

tools/rosbag/src/recorder.cpp
```diff
@@ -491,15 +524,17 @@ void Recorder::doRecordSnapshotter() {
...
             queue_condition_.wait(lock);
         }
-        
+       
         OutgoingQueue out_queue = queue_queue_.front();
+        std::queue<OutgoingMessage> hackyCopy = *out_queue.queue;//should do a deep copy (this is of shared pointers and strings)
         queue_queue_.pop();
-        
         lock.release()->unlock();
+        out_queue.queue = &hackyCopy;
```
dirk-thomas, line 532: Even without really understanding what this is good for the block looks pretty suspicious to me. Please explain what you are trying to achieve with that.

dirk-thomas:
Besides the comments on individual code blocks I can't see how this patch is supposed to work. I would imagine that the patch must do something like this:
- add the command line option
- if the option is set it should
  - not write the data to the file anymore
  - but move the data into a queue (which can grow arbitrary)
- listen to the trigger and dump the content of the new queue

I was not able to find where especially the steps of the second bullet are performed. Please try to describe where and how this is done so that I can understand the changes.

baxelrod:
Hello Dirk,

This code isn't ready to be merged. What I was really hoping to learn was
whether such a feature, similar to what I had hacked together was desired.
Just to be clear, please don't merge this as is, that's not what I'm trying
to do here.

My initial question was: would you like me to make a clean, tested version
of the feature I showed here, that meets the requirements you memtioned in
your list? If so, I have a couple more questions to ask before implementing
something intended to actually ship with rosbag.

I'm sorry if a pull request was not the right way of asking this question.

dirk-thomas:
Sorry, I did not get that in the first place. I only took a short view through the changes.

Yes, such a feature would be considered valuable. If you can continue working on this that would be great. If you have any further questions please feel free to ask. And I agree with you previous comment that a watchdog can easily implemented outside which than uses the generic trigger message.

In the future I think it would be good to fill a ticket describing the feature, potentially pointing to your forked repo containing a prototype. That way it is clearer that it is not intended to merged. And when the implementation is complete you can easily create the pull request.

dirk-thomas:
Since the pull request is not yet in a state where it can be merged I will close it for now. If you have an updated version please comment so that it can be reopened or open a new pull request.

PR #273, opened 2013-08-13
add rosbag_migration_rule package to export rules without depending on rosbag

dirk-thomas:


dirk-thomas:
Based on the feedback I will move the new package into its own repo before merging the pull request.

dirk-thomas:
@tfoote Please review.

tfoote:
+1

PR #275, opened 2013-08-20
Catch exceptions when computing roslaunch dependencies.

ahendrix:
It's possible for the dependency calculation to throw exceptions if arguments don't exist, and possibly in other cases as well. Catch them and attempt to proceed.

### DIFF
--- a/tools/roslaunch/src/roslaunch/rlutil.py
+++ b/tools/roslaunch/src/roslaunch/rlutil.py
@@ -195,7 +195,12 @@ def check_roslaunch(f):
     
     errors = []
     # check for missing deps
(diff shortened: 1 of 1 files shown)


ahendrix:
This is deeper than I thought; parse_launch in depends.py isn't building up a context before evaluating args, so any <include>, <node> or <test> tags that reference an arg will fail.

This causes roslaunch-check to crash when parsing complex launch files such as https://github.com/ros-interactive-manipulation/pr2_object_manipulation/blob/groovy-devel/manipulation/pr2_object_manipulation_launch/launch/laser%2Bstereo-perception.launch

PR #276, opened 2013-08-20
Fix arg parsing when computing dependencies in roslaunch-check

ahendrix:
parse_launch in depends.py isn't building up a context before evaluating args, so any `include`, `node` or `test` tags that reference an arg will fail.

This causes roslaunch-check to crash when parsing complex launch files such as https://github.com/ros-interactive-manipulation/pr2_object_manipulation/blob/groovy-devel/manipulation/pr2_object_manipulation_launch/launch/laser%2Bstereo-perception.launch

### DIFF
--- a/tools/roslaunch/src/roslaunch/depends.py
+++ b/tools/roslaunch/src/roslaunch/depends.py
@@ -83,7 +83,43 @@ class RoslaunchDeps(object):
     def __str__(self):
         return "nodes: %s\nincludes: %s\npkgs: %s"%(str(self.nodes), str(self.includes), str(self.pkgs))
 
-def _parse_launch(tags, launch_file, file_deps, verbose):
+def _get_arg_value(tag, context):
+    name = tag.attributes['name'].value
+    if tag.attributes.has_key('value'):
+        return resolve_args(tag.attributes['value'].value, context)
+    elif name in context['arg']:
+        return context['arg'][name]
+    elif tag.attributes.has_key('default'):
+        return resolve_args(tag.attributes['default'].value, context)
+    else:
+        raise RoslaunchDepsException("No value for arg %s"%(name))
+
+def _parse_arg(tag, context):
+    name = tag.attributes['name'].value
+    if tag.attributes.has_key('if'):
+        val = resolve_args(tag.attributes['if'].value, context)
+        if val == '1' or val == 'true':
+            return (name, _get_arg_value(tag, context))
+    elif tag.attributes.has_key('unless'):
+        val = resolve_args(tag.attributes['unless'].value, context)
+        if val == '0' or val == 'false':
+            return (name, _get_arg_value(tag, context))
+    else:
+        return (name, _get_arg_value(tag, context))
+
+def _parse_subcontext(tags, context):
+    subcontext = {'arg': {}}
+
+    if tags == None:
+       return subcontext
+    
+    for tag in [t for t in tags if t.nodeType == DomNode.ELEMENT_NODE]:
+        if tag.tagName == 'arg':
+            (name, val) = _parse_arg(tag, context)
+            subcontext['arg'][name] = val
+    return subcontext
+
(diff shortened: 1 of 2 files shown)


dirk-thomas:
`roslaunch pr2_object_manipulation_launch laser+stereo-perception.launch` already fails with the following error for me:

```
Invalid <arg> tag: openni_launch
ROS path [0]=/opt/ros/groovy/share/ros
ROS path [1]=/opt/ros/groovy/share
ROS path [2]=/opt/ros/groovy/stacks. 

Arg xml is <arg name="kinect_launchfile" unless="$(arg use_freenect)" value="$(find openni_launch)/launch/openni.launch"/>
```

This is the same error message `roslaunch-check` generates. So at least for me the result of `roslaunch` and `roslaunch-check` is identical as it should be.

Is that the same case for you?
- if yes, can you please clarify what the issue is supposed to address?
- if no, can you please provide more information what you do in order to reproduce your different behavior?

trainman419:
I assume that launch file works; most of the interactive manipulation examples and/or other launch files use it.

Unfortunately, I don't have a good example or a way to reproduce this on my local machine, and I no longer have access to a PR2 to debug on.

I suspect you could reproduce this with one or two launch files:

a.launch has a parameter, which specifies another launch file to include
b.launch is a simple launch file that can be included. it probably doesn't need to do anything

dirk-thomas:
After playing with it a bit more I got the following one working:

```
ROBOT=sim roslaunch pr2_object_manipulation_launch laser+stereo-perception.launch
```

but the `roslaunch-check` equivalent failing:

```
ROBOT=sim /opt/ros/groovy/share/roslaunch/scripts/roslaunch-check /opt/ros/groovy/stacks/pr2_object_manipulation/manipulation/pr2_object_manipulation_launch/launch/laser+stereo-perception.launch
```

I will look into it.

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -195,7 +195,16 @@ def check_roslaunch(f):
     
     errors = []
     # check for missing deps
-    base_pkg, file_deps, missing = roslaunch.depends.roslaunch_deps([f])
+    try:
+        base_pkg, file_deps, missing = roslaunch.depends.roslaunch_deps([f])
+    except rospkg.common.ResourceNotFound as r:
+        errors.append("Could not find package: %s included from %s"%(str(r), f))
```
dirk-thomas, line 201: Can you please update the line to include `[]` around the variables to follow the same style in these files:

```
errors.append("Could not find package [%s] included from [%s]"%(str(r), f))
```

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -195,7 +195,16 @@ def check_roslaunch(f):
...
+        missing = {}
+        file_deps = {}
+    except roslaunch.substitution_args.ArgException as e:
+        errors.append("Could not resolve arg %s in %s"%(str(e), f))
```
dirk-thomas, line 205: errors.append("Could not resolve arg [%s] in [%s]"%(str(e), f))

tools/roslaunch/src/roslaunch/depends.py
```diff
@@ -83,7 +83,43 @@ def __repr__(self):
...
+def _get_arg_value(tag, context):
+    name = tag.attributes['name'].value
+    if tag.attributes.has_key('value'):
+        return resolve_args(tag.attributes['value'].value, context)
+    elif name in context['arg']:
+        return context['arg'][name]
+    elif tag.attributes.has_key('default'):
+        return resolve_args(tag.attributes['default'].value, context)
+    else:
+        raise RoslaunchDepsException("No value for arg %s"%(name))
```
dirk-thomas, line 95: raise RoslaunchDepsException("No value for arg [%s]"%(name))

dirk-thomas:
Superseded by #290.

Issue #263, opened 2013-08-02
roscpp missing getParam() for lists and dictionaries


b-adkins:
"roscpp's parameter API supports all of [ROS Parameter Server's data types], though **it is only easy to use strings, integers, floats and booleans.**" ([roscpp Parameter Server](http://www.ros.org/wiki/roscpp/Overview/Parameter%20Server))

_Why?_ I don't see why an application developer needs to learn XmlRpc and re-implement the wheel. I believe it's quite doable to reconstruct lists and dictionaries (at least of primitives) entered in YAML to the parameter server. 

I propose a patch with templated varieties of NodeHandle::getParam() and param::get() that:

Extracts lists as

``` C
// C++ STL style:
std::vector<type> the_list_cpp;

// Possibly C style:
type* the_list_c;
```

Extracts dictionaries as

``` C
// C++ STL style:
std::map<std::string, type> the_dictionary;

// C style:
char* keys[];
type* values;
```

Proposed interface:
(n_max is the maximum elements the passed buffer(s) can hold, to prevent buffer overrun)

``` C
// C++ STL style:
template<typename T> bool getParam(const std:string& key, std::vector<T>& l) const;
template<typename T> bool getParam(const std:string& key, std::map<std::string, T>& d) const;

// C style:
template<typename T> bool getParam(const std:string& key, T* l, int n_max) const;
template<typename T> bool getParam(const std:string& key, char* s[], T* v, int n_max) const;
```

The C++-style versions would be very easy to use. The C-style types may be useful for performance.

I've essentially already written these for another project. I would like to offer my code as part of this proposed patch.

EDIT: For easy reference, here is <ros/param.h>:
https://github.com/ros/ros_comm/blob/groovy-devel/clients/roscpp/include/ros/param.h

Also note that the methods would need to be wrapped NodeHandle:
https://github.com/ros/ros_comm/blob/groovy-devel/clients/roscpp/src/libros/node_handle.cpp

jbohren:
I think many of us have implemented similar things. It would be great to see some of them added to the core libraries.
See my vector param implementation here: https://github.com/jhu-lcsr/terse_roscpp/blob/master/include/terse_roscpp/param.h

Do you have an implementation of your own floating around somewhere?

b-adkins:
Sorry for the delayed response. I have getters for vectors and maps 
which use the same conventions as the ros code. I'll post them when I 
figure out how to use GitHub.

Bea

b-adkins:
[...]

PR #279, opened 2013-09-03
roscpp: Adding std::vector and std::map rosparam sets/gets

jbohren:
This patch adds the ability to call `ros::param::get()` and ros::param::set() with `std::vector<T>` and `std::map<std::string,T>` where T is one of (`std::string`, `double`, `float`, `int`, `bool`). 

Also some trivial doc fixes.

### DIFF
--- a/clients/roscpp/include/ros/node_handle.h
+++ b/clients/roscpp/include/ros/node_handle.h
@@ -1384,20 +1384,92 @@ if (handle)
    * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
    */
   void setParam(const std::string& key, double d) const;
-  /** \brief Set a integer value on the parameter server.
+  /** \brief Set an integer value on the parameter server.
    *
    * \param key The key to be used in the parameter server's dictionary
    * \param i The value to be inserted.
    * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
    */
   void setParam(const std::string& key, int i) const;
-  /** \brief Set a integer value on the parameter server.
+  /** \brief Set a boolean value on the parameter server.
    *
    * \param key The key to be used in the parameter server's dictionary
    * \param b The value to be inserted.
    * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
    */
   void setParam(const std::string& key, bool b) const;
+
+  /** \brief Set a string vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param vec The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string& key, const std::vector<std::string>& vec) const;
+  /** \brief Set a double vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param vec The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string& key, const std::vector<double>& vec) const;
+  /** \brief Set a float vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param vec The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string& key, const std::vector<float>& vec) const;
+  /** \brief Set a int vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param vec The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string& key, const std::vector<int>& vec) const;
+  /** \brief Set a bool vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param vec The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string& key, const std::vector<bool>& vec) const;
+
+  /** \brief Set a string vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param map The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string& key, const std::map<std::string, std::string>& map) const;
+  /** \brief Set a double vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param map The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string& key, const std::map<std::string, double>& map) const;
+  /** \brief Set a float vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param map The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string& key, const std::map<std::string, float>& map) const;
+  /** \brief Set a int vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param map The value to be inserted.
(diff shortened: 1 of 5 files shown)


jbohren:
Fix for #263

isucan:
I have needed this in the past. 
+1

jbohren:
@wjwwood Where should unit tests for this crap go?

wjwwood:
There is an external package, `test_roscpp`, which holds the tests (as they are integration tests and have dependencies that roscpp wouldn't otherwise have):

https://github.com/ros/ros_comm/tree/groovy-devel/test/test_roscpp

jbohren:
> There is an external package, `test_roscpp`, which holds the tests (as they are integration tests and have dependencies that roscpp wouldn't otherwise have):
> 
> https://github.com/ros/ros_comm/tree/groovy-devel/test/test_roscpp

Ah cool. I'll add some unit tests there, and then ping back.

jbohren:
Ok. Unit tests are in place (and passing): https://github.com/ros/ros_comm/pull/279/files#diff-4

Currently, it will automatically convert between float/double/int/bool but not string, but that's easy to change if we want.

Repro:

``` sh
unset CMAKE_PREFIX_PATH
source /opt/ros/hydro/setup.sh
mkdir -p ws/src
git clone git@github.com:jhu-lcsr-forks/ros_comm.git -b feature-roscpp-composite-rosparam  ws/src/ros_comm
cd ws
catkin_make
catkin_make tests
catkin_make run_tests_test_roscpp_rostest_test_launch_params.xml
```

b-adkins:
Nice! Thanks for slogging through the unit tests.

Is there a reason to define a function for every single vector<primitive> and map<primitive>? A single function template would be much shorter. And acceptable types could be limited with boost, though other types wouldn't compile anyway, so I wouldn't bother. Is it to hide the implementation in the cpp file?  Is that worth doing?

Is there a reason to mess with type safety? Int to float is pretty obvious ([0, .1] should not fail), but float to int? If users mean an int, that's what they should enter! I think, since YAML is already parsed to numbers, and because C++ is strongly typed, we should do strict type safety.

jbohren:
> Is there a reason to define a function for every single vector and map? A single function template would be much shorter. And acceptable types could be limited with boost, though other types wouldn't compile anyway, so I wouldn't bother. Is it to hide the implementation in the cpp file? Is that worth doing?

A globally-exposed function template would be much shorter to write, but the end cost would then be put onto the user and the user's compiler. When templates are meant to take arbitrary or a broad spectrum of types, then they need to be exposed to the user, but when they're only used to simplify implementation, there's no real benefit to exposing them, it just increases the time it would take to compile every binary which includes "ros/ros.h". Also, adding all the wrapper functions was the easiest part of the patch.

Additionally, by not exposing the templated code in what should arguably be one of the simplest parts of ROS, then you make the error messages much clearer when someone tries to pull the wrong type off of the parameter server. 

> Is there a reason to mess with type safety? Int to float is pretty obvious ([0, .1] should not fail), but float to int? If users mean an int, that's what they should enter! I think, since YAML is already parsed to numbers, and because C++ is strongly typed, we should do strict type safety.

Well, float to int, or double to int, or int to bool are all reasonable casts, they just come with the cost of loss of precision. Even in C++, this doesn't even generate a warning when you compile it, let alone an error. I think in this case we should be as consistent as we can be with C++ standards. In this case, if it's possible to `static_cast()` the parameter type to the output type, we should let the user do so.

b-adkins:
Ah, I see.

" In this case, if it's possible to static_cast() the parameter type to the output type, we should let the user do so."
Sounds good to me.

b-adkins:
I thought of another, slight enhancement: It seems to me like filling an xmlRpcValue with a std::vector/std::map should have its own function in that class. But I'm not sure what the best way to call that templated code would be. Can constructors be templated? Member functions? Perhaps name it after an appropriate stl function? 

Or would it have to be a non-member function? Of the form 

``` cpp
template<class T> void createFromVector(const std::vector<T>& vec, xmlRpcValue& val);
```

or even an operator

``` cpp
template<class T> xmlRpcValue& operator<<(xmlRpcValue& val, const std::vector<T>& vec);
```

jbohren:
@b-adkins Maybe more advanced APIs can be discussed later on after we've gotten someone to review / merge this simple one.

Anyone?

wjwwood:
@dirk-thomas just got back from Germany, maybe he can review it.

dirk-thomas:
I just took a first look and the patch looks very good. I will add some comments but they should be all be pretty minor. It would be great if you could address them (even if some are doc and code style only) before the merge - probably squash your commits afterwards.

clients/roscpp/include/ros/node_handle.h
```diff
@@ -1483,7 +1667,21 @@ bool callback(ros::ServiceEvent<std_srvs::Empty, std_srvs::Empty>& event)
...
+   * no subsequent getCached() calls with the same key will query the master --
+   * they will instead look up in the local cache.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param[out] i Storage for the retrieved value.
+   *
+   * \return true if the parameter value was retrieved, false otherwise
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  bool getParamCached(const std::string& key, float& d) const;
```
dirk-thomas: Can you change the parameter name to `f` and also update the doc string which currently refers to the parameter as `i`?

clients/roscpp/include/ros/node_handle.h
```diff
@@ -1399,6 +1399,78 @@ bool callback(ros::ServiceEvent<std_srvs::Empty, std_srvs::Empty>& event)
...
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string &key, std::map<std::string,std::string> map) const;
+  /** \brief Set a double vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param map The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string &key, std::map<std::string,double> map) const;
```
dirk-thomas: According to the code style a `,` should be followed by a space. Can you update that for all signatures with vector and map? (Sorry, but that would make it more consistent with the existing code base)

clients/roscpp/include/ros/node_handle.h
```diff
@@ -1399,6 +1399,78 @@ bool callback(ros::ServiceEvent<std_srvs::Empty, std_srvs::Empty>& event)
    */
   void setParam(const std::string& key, bool b) const;
 
+  /** \brief Set a string vector value on the parameter server.
+   *
+   * \param key The key to be used in the parameter server's dictionary
+   * \param vec The value to be inserted.
+   * \throws InvalidNameException If the parameter key begins with a tilde, or is an otherwise invalid graph resource name
+   */
+  void setParam(const std::string &key, std::vector<std::string> vec) const;
```
dirk-thomas: Wouldn't it be better is the value is passed as a `const reference`? (Some for all similar vector and map signatures)

dirk-thomas: Can you please make the position of the `&` consistent in the code? Most of the existing code in the class has it "on the left" side: `const std::string& key, ...`.

clients/roscpp/include/ros/param.h
```diff
@@ -84,6 +87,80 @@
...
 ROSCPP_DECL void set(const std::string& key, bool b);
 
+
+/** \brief Set a string vector value on the parameter server.
+ *
+ * \param key The key to be used in the parameter server's dictionary
+ * \param vec The vector value to be inserted.
+ * \throws InvalidNameException if the key is not a valid graph resource name
+ */
+ROSCPP_DECL void set(const std::string& key, const std::vector<std::string> &vec);
```
dirk-thomas: Same comment as above - please make the position of the `&` consistent (as for `key`).

jbohren:
@dirk-thomas, all requested fixes made in latest commits.

dirk-thomas:
The `setParam(..., std::map ...)` methods are still passing the map without const? Can you add the const there too or is there a reason why that does not work?

Can you please also squash the commits at the end - that keeps the log more compact.

jbohren:
@dirk-thomas The const for `setParam(...)` were just missing. Consider it squashed.

dirk-thomas:
Great, thank you very much for implementing the feature and taking the time to iterate (on my pedantic comments :smile: ).

dirk-thomas:
#302 might be a regression of this pull request.

jbohren:
> #302 might be a regression of this pull request.

Yeah I think this could happen if someone adds `using namespace std;` before the ROS param.h header gets included. Ruben's fix looks good, though.

catskul:
These utility functions are a huge win. Thanks!

Any reason we shouldn't include std::list as well?

PR #284, opened 2013-09-20
rospy: use threading module rather than thread

wjwwood:
This pull request modifies rospy to eliminate the use of the Python `thread` module and to instead use the Python `threading` module universally.  Reasons for doing this:
- Remove usage of deprecated module (must be imported as `_thread` in Python3)
- Allows the coverage tool in python to cover callbacks for services in rospy

For more information about the coverage bug with the `thread` modules, see:

https://bitbucket.org/ned/coveragepy/issue/244/coverage-doesnt-work-with-thread-not

I mostly replace lines like:

``` python
import thread
...
thread.start_new_thread(func, (arg1, arg2))
```

With sections like this:

``` python
import threading
...
threading.Thread(target=func, args=(arg1, arg2)).start()
```

Not keeping a reference to the Thread object after creation appears to be OK, according to this to this website:

http://bytes.com/topic/python/answers/160905-my-thread-safe-premature-garbage-collection

### DIFF
--- a/clients/rospy/src/rospy/impl/registration.py
+++ b/clients/rospy/src/rospy/impl/registration.py
@@ -39,10 +39,6 @@
 import socket
 import sys
 import logging
-try:
-	import _thread
-except ImportError:
-    import thread as _thread
 import threading
 import time
 import traceback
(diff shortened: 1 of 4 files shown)


wjwwood:
So... I don't get 100% test's passing with this change... Though I am dubious of the stability of these tests...

wjwwood:
This is absolutely not passing the tests on my Mac. Why, I have no idea, but I am going to revert the change and reopen until we can figure it out.

wjwwood:
I reverted it here: aacd6beb4ca2f8d09279b2e8f07a580bce134ac3

wjwwood:
And reopened it here: #285

PR #285, opened 2013-09-20
rospy: use threading module rather than thread

wjwwood:
This is a reopening of #284.

### DIFF
--- a/clients/rospy/src/rospy/impl/registration.py
+++ b/clients/rospy/src/rospy/impl/registration.py
@@ -39,10 +39,6 @@
 import socket
 import sys
 import logging
-try:
-	import _thread
-except ImportError:
-    import thread as _thread
 import threading
 import time
 import traceback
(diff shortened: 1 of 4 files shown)


wjwwood:
I have a theory:

The TCPServer class spawns a thread in which it calls an "inbound_handler" which might be spawning threads as well. This might be the source of the behavior change, i.e. calling `thread.start_new_thread` in a thread may behave differently from `threading.Thread.start`.

wjwwood:
This is the issue, `thread.start_new_thread` is equivalent to `threading.Thread` in daemon mode. This is pretty bad, it basically means that when the script ends these threads go on. There should probably be a new pull request for making rospy threads cleanly shutdown.

wjwwood:
http://stackoverflow.com/questions/4003783/translate-thread-start-new-thread-to-the-new-threading-api

wjwwood:
Ok, there is a new commit, which converts this:

``` python
import thread
...
thread.start_new_thread(func, args=(arg1, arg2))
```

Into:

``` python
import threading
...
t = threading.Thread(target=func, args=(arg1, arg2))
t.setDaemon(True)
t.start()
```

wjwwood:
I am running tests now on my mac and on Linux.

wjwwood:
```
% catkin_test_results build/test_results
Summary: 312 tests, 0 errors, 0 failures
```

wjwwood:
@dirk-thomas @tfoote for review.

tfoote:
+1  I'm amazed how few lines this patch is.

dirk-thomas:
+1 Please squash your commits since the in-between state is pretty broken.

wjwwood:
Squashed.

PR #287, opened 2013-10-01
Allow rosbag "chunk threshold" to be adjusted when recording

kwatts:
For large files (above a few GB), opening bags can take a significant amount of time. The open step is slow because it seeks through the bag file, loading chunk data. By increasing the chunk size for a bag, we can reduce the opening time significantly.

I would like to allow users to be able to configure the chunk size of a bag file when recording. This will allow me to record bag files with larger chunk sizes.


dirk-thomas:
+1 thanks for the patch.

kwatts:
Thanks, Dirk!

PR #288, opened 2013-10-01
Add option to disable terminal title setting

cmansley:
In some larger projects, the terminal windows are instantiated with correct titles, which roslaunch promptly overwrites. It would be nice to have an option to disable roslaunch from messing with the terminal title.


dirk-thomas:
+1 thank you for the enhancement.

PR #289, opened 2013-10-02
fix result file naming for wet rostests when being built in-source (fix ros/catkin#512)

dirk-thomas:
@wjwwood Please review


wjwwood:
+1 works for me on my failure case:

ros/catkin#512

PR #291, opened 2013-10-04
Fixed compilation of base64.h on clang + libc++

severin-lemaignan:
If `<ios>` is not included, the compilation fails with the following
error:

```
In file included from
ros_comm-1.9.47/utilities/xmlrpcpp/src/XmlRpcValue.cpp:5:
ros_comm-1.9.47/utilities/xmlrpcpp/include/base64.h:231:13:
error: incomplete type 'std::__1::ios_base' named in nested name
specifier
_St |= _IOS_FAILBIT|_IOS_EOFBIT; return _First; // unexpected EOF
^~~~~~~~~~~~
```

### DIFF
--- a/utilities/xmlrpcpp/include/base64.h
+++ b/utilities/xmlrpcpp/include/base64.h
@@ -13,6 +13,12 @@
 # include <iterator>
 #endif
 
+#if defined(__clang__)
+    #include <ios>
+#else
+    #include <iosfwd>
+#endif
+
 static
 int _base64Chars[]= {'A','B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z',
 				     'a','b','c','d','e','f','g','h','i','j','k','l','m','n','o','p','q','r','s','t','u','v','w','x','y','z',


wjwwood:
What version of OS X, Xcode, Command Line tools, and clang are you using? I just rebuilt yesterday, but I don't have this problem on my mac.

wjwwood:
I guess I assumed OS X, are you on Linux?

severin-lemaignan:
Yes, I'm on linux. I've reproduced the issue with both ubuntu 13.04 and OpenNao 1.14 (gentoo-based).

dirk-thomas:
Why does the patch contain an else block? Currently it seems to work fine on several supported platforms.

```
#include <iosfwd>
```

Is there any reason why not to use `ios.h` generally? Are there platform which do not support that?

severin-lemaignan:
To be frank, it's a blind copy-paste from this suggestion: http://lists.cs.uiuc.edu/pipermail/cfe-dev/2011-September/017017.html

Since `<iosfwd>` contains only forward declarations, it may be faster to compile on the configurations that support it. But `<ios>` is definitely also ok.

PR #296, opened 2013-10-11
added param, has_param and launch_file substitution arg

xaedes:
added `param`, `has_param` and `launch_file` substitution arg

`param` can substitute ros params
`has_param` can be substituted with bool value whether a ros param exists (e.g. for use in `if` or `unless`)
`launch_file` can be substituted with the name of (one of) the launch file(s) that is parsed until this point, without a given index of which launch file is wanted, the last one is used

also fixed an issue in `_launch_tag` where a ros_config instance where given to the ifunless decorator instead of a context

additionally it is now possible to access `ros_config` for more sophisticated substitution args

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -264,11 +264,10 @@ def _param(resolved, a, args, context, ros_config):
     process $(param param_name)
     
     :returns: updated resolved argument, ``str``
-    :raises: :exc:`SubstitutionException` If param_name invalidly specified
-    :raises: :exc:`SubstitutionException`: if param_name does not exist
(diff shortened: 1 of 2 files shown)


progtologist:
This is a very good functionality for roslaunch.
If I understand correctly, you have implemented loading of arguments using the parameter server, correct?
Will this pull request be merged eventually to groovy and hydro?

tfoote:
roslaunch's API is a core interface for ROS and needs to go through a community review via the REP process. http://www.ros.org/reps/rep-0000.html

k-okada:
+1, param substitution may very useful

garaemon:
+1.
It would be great feature to write re-usable launch file

dirk-thomas:

PR #298, opened 2013-10-11
this fixes #297

### LINKED ISSUE #297: $(anon) does not behave as described documentation
$(anon) is documented as following:
http://wiki.ros.org/roslaunch/XML#substitution_args

> `$(anon name)`
> e.g. `$(anon rviz-1)`. Generates an anonymous id based on `name`. `name` itself is a > unique identifier: multiple uses of `$(anon foo)` will create the same "anonymized" name.  This is used for &lt;node&gt; `name` attributes in order to create nodes with anonymous names, as ROS requires nodes to have unique names. For example:
> 
> ``` xml
> <node name="$(anon foo)" pkg="rospy_tutorials" type="talker.py" />
> <node name="$(anon foo)" pkg="rospy_tutorials" type="talker.py" />
> ```
> 
> Will generate an error that there are two &lt;node&gt;s with the same name.

In fact this example would just run fine because $(anon foo) gets resolved differently.
I tracked the problem to the point where the handling of the &lt;node&gt; tag is called. The problem is that a copy of the context is created and thrown away afterwards. The $(anon) resolvings are stored in these. Thus the second substitution `$(anon foo)` won't resolve to the same as the first substitution.

``` python
# clone the context so that nodes' env does not pollute global env
n = self._node_tag(tag, context.child(''), ros_config, default_machine, verbose=verbose)
```

> roslaunch/src/roslaunch/xmlloader.py (line number 604, commit #65ea73aa40c9431cbe2cf11798907ce64d139b42)

xaedes:
this fixes #297

### DIFF
--- a/tools/roslaunch/src/roslaunch/xmlloader.py
+++ b/tools/roslaunch/src/roslaunch/xmlloader.py
@@ -601,13 +601,17 @@ class XmlLoader(loader.Loader):
                         self._recurse_load(ros_config, tag.childNodes, child_ns, \
                                                default_machine, is_core, verbose)
             elif name == 'node':
-                # clone the context so that nodes' env does not pollute global env
-                n = self._node_tag(tag, context.child(''), ros_config, default_machine, verbose=verbose)
+                # backup the contexts env_vars and revert them afterwards so that nodes' env does not pollute global env
+                env_vars_backup = context.env_args[:]
+                n = self._node_tag(tag, context, ros_config, default_machine, verbose=verbose)
+                context.env_args = env_vars_backup
                 if n is not None:
                     ros_config.add_node(n, core=is_core, verbose=verbose)
             elif name == 'test':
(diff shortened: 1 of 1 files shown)


tfoote:
I think it might make more sense to simply pass in a copy of the context or change the method to not mutate the context but copy internally instead of mutating and then reverting.

xaedes:
As I stated in #297 it was actually implemented that way and this is what causes $(anon) to not work the way it is meant to be.
`_node_tag` ~~and `_test_tag`~~ in xmlloader.py needs the actual context. But I will examine other points in the code to create a copy of the context that can be polluted by env.

xaedes:
Copying of the context just for the env tag inside of nodes works:
https://github.com/xaedes/ros_comm/commit/0b6d6ec4153ae8716e3e20556801b8026329e10e

I  did another pull request for it (#300) and close this.

PR #300, opened 2013-10-12
fixes issue #297, another approach

### LINKED ISSUE #297: $(anon) does not behave as described documentation

xaedes:
As stated in #298 here is another approach to fix #297.

### DIFF
--- a/tools/roslaunch/src/roslaunch/xmlloader.py
+++ b/tools/roslaunch/src/roslaunch/xmlloader.py
@@ -361,6 +361,10 @@ class XmlLoader(loader.Loader):
             # it inherits from its parent
             remap_context = context.child('')
 
+            # each node gets its own copy of <env> arguments, which
+            # it inherits from its parent
+            env_context = context.child('')
+
             # nodes can have individual env args set in addition to
             # the ROS-specific ones.  
             for t in [c for c in tag.childNodes if c.nodeType == DomNode.ELEMENT_NODE]:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Merged to _indigo-devel_ with a unit test covering preventing regressions in the future: 28259f1380704d7054ff0b8ab6a5509f70d0f799

Thank you for providing the PR.

PR #301, opened 2013-10-13
add roscpp dependency

vrabaud:
client.py imports stuff from roscpp


dirk-thomas:
S***\* but is necessary until those logger messages/service are moved out of roscpp.

PR #302, opened 2013-10-15
roscpp: Fix clashing namespaces for ros::param::set

smits:
Compiling on RHEL with gcc 4.4.6 resulted in the following error:

```
/usr/lib/gcc/x86_64-redhat-linux/4.4.6/../../../../include/c++/4.4.6/bits/stl_set.h: In function 'void ros::param::setImpl(const std::string&, const std::vector<T, std::allocator<_CharT> >&)':
/usr/lib/gcc/x86_64-redhat-linux/4.4.6/../../../../include/c++/4.4.6/bits/stl_set.h:87: error: 'template<class _Key, class _Compare, class _Alloc> class std::set' is not a function,
/home/codac-dev/rpmbuild/BUILD/ros-hydro-core-1.9.50/src/roscpp/include/ros/param.h:161: error:   conflict with 'void ros::param::set(const std::string&, const std::map<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool, std::less<std::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<const std::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool> > >&)'
```

Apparently the set was clashing with std::set for some reason.

Signed-off-by: Ruben Smits ruben.smits@intermodalics.eu

### DIFF
--- a/clients/roscpp/src/libros/param.cpp
+++ b/clients/roscpp/src/libros/param.cpp
@@ -124,32 +124,32 @@ template <class T>
     xml_vec[i] = vec.at(i);
   }
 
-  set(key, xml_vec);
+  ros::param::set(key, xml_vec);
 }
 
 void set(const std::string& key, const std::vector<std::string>& vec)
 {
-  setImpl(key, vec);
+  ros::param::setImpl(key, vec);
 }
 
 void set(const std::string& key, const std::vector<double>& vec)
 {
-  setImpl(key, vec);
+  ros::param::setImpl(key, vec);
 }
 
 void set(const std::string& key, const std::vector<float>& vec)
 {
-  setImpl(key, vec);
+  ros::param::setImpl(key, vec);
 }
 
 void set(const std::string& key, const std::vector<int>& vec)
 {
-  setImpl(key, vec);
+  ros::param::setImpl(key, vec);
 }
 
 void set(const std::string& key, const std::vector<bool>& vec)
 {
-  setImpl(key, vec);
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you please clarify what exactly you are compiling and how? Is it just _ros_comm_ or also any custom code?

It seems that the invocation of `set()` (https://github.com/smits/ros_comm/blob/9c133c6a761f73accbe674af9df1f64c0db49051/clients/roscpp/src/libros/param.cpp#L127) is ambiguous when compiling `param.cpp`. This would make sense if any code before would define `using namespace std` but since it is happening in a cpp file rather than a header (and ROS does not use `using namespace` in headers) I don't know where that should be coming from.

smits:
I was compiling the ROS Hydro Bare flavour from source in a catkin workspace on ...

```
lsb_release -a
LSB Version:    :core-4.0-amd64:core-4.0-noarch:graphics-4.0-amd64:graphics-4.0-noarch:printing-4.0-amd64:printing-4.0-noarch
Distributor ID: RedHatEnterpriseWorkstation
Description:    Red Hat Enterprise Linux Workstation release 6.3 (Santiago)
Release:    6.3
Codename:   Santiago
```

using

```
gcc --version
gcc (GCC) 4.4.6 20120305 (Red Hat 4.4.6-4)
```

dirk-thomas:
I can reproduce the issue under Ubuntu Quantal when using gcc-4.4.

PR #306, opened 2013-10-24
refactoring to make log4cxx an optional dependency

dirk-thomas:
@gerkey @tfoote @wjwwood Please review.

Especially any feedback on the interface for custom backends would be appreciated (see https://github.com/ros/ros_comm/compare/log4cxx?expand=1#diff-1e993bcf45a700893f4d675ec053f1aaR57).


wjwwood:
Reviewing this is going to take a while, I'll try to get to it by Friday.

dirk-thomas:
Please take some time to review this in the near future so that we can pass it on for external testing/review.

tfoote:
In the diff there are 20+ instances of "new" but no deletes.

tools/rosconsole/src/rosconsole/rosconsole.cpp
```diff
@@ -292,197 +283,112 @@ TokenPtr createTokenFromType(const std::string& type)
...
-    else if (event->getLevel() == log4cxx::Level::getWarn())
-    {
-      color = COLOR_YELLOW;
-    }
-    else if (event->getLevel() == log4cxx::Level::getInfo())
-    {
-      color = COLOR_NORMAL;
-    }
-    else if (event->getLevel() == log4cxx::Level::getDebug())
+#if 0
```
wjwwood, line 299: Should this block be left, or something else done with it.

tools/rosconsole/src/rosconsole/rosconsole.cpp
```diff
@@ -292,197 +283,112 @@ TokenPtr createTokenFromType(const std::string& type)
...
-    std::stringstream ss;
-    ss << color;
-    V_Token::iterator it = tokens_.begin();
-    V_Token::iterator end = tokens_.end();
-    for (; it != end; ++it)
-    {
-      ss << (*it)->getString(event);
-    }
-    ss << COLOR_NORMAL;
+    std::string token = results[1];
```
wjwwood, line 306: Is there guaranteed to be a value at `[1]`? Should the length be checked? I'm not familiar with the `match_results` type.

tools/rosconsole/src/rosconsole/rosconsole.cpp
dirk-thomas, line 299: This code was just moved around. I would just keep it - I guess it was only used for debugging purposes.

tools/rosconsole/src/rosconsole/rosconsole.cpp
dirk-thomas, line 306: `[0]` contains the full string, `[1]` the first matched block within parenthesis. Since the match was successful (while loop) and the regex contains a parenthesis there must be one result.

dirk-thomas:
@tfoote (https://github.com/ros/ros_comm/pull/306#issuecomment-27522589) The allocated objects are either globals which we currently have no way of cleaning up correctly as well as log appenders which are managed within log4cxx.

dirk-thomas:
I have integrated all changes. Any further feedback? Else I will hand it over for external review/testing.

wjwwood:
Nope looks good to me.

tfoote:
+1 overall

Issue #169, opened 2013-02-07
rospy publisher hanging when subscriber drops connection (e.g. because of wifi)


advaitjain:
I'm submitting an issue for this question on ROS answers:
http://answers.ros.org/question/29387/rospy-publisher-hanging-when-subscriber-drops-connection-because-of-wifi-is-this-normal/

Summary:
If a rospy publisher is connected to a subscriber on a different computer and you disconnect the ethernet cable, the publisher hangs after about a minute.

Issue can be replicated with the rospy talker node running on a different computer than the listener node.

The cause might be here:
https://github.com/ros/ros_comm/blob/groovy-devel/clients/rospy/src/rospy/impl/tcpros_base.py#L621

forrestv:
I ran into the same problem. It seems that rospy's use of blocking writes to sockets is indeed the cause.

mikeferguson:
@tfoote I'm going to bring this back up -- as hydro tf/tf2 will use pure python publishers and this is going to cause a major change in how tf publishers operate over wifi/unreliable links compared to Groovy and earlier.

As can be seen in https://github.com/ros/ros_comm/blob/groovy-devel/clients/rospy/src/rospy/impl/tcpros_base.py#L620, Ken never finished implementing async or retry logic, which means that if you have say, a remote RVIZ connected to a python node publishing tf, you will now have the python node hang if the wifi link drops out. You might recall the problems we had with a certain robot when python nodes started randomly hanging (in that case, it was an /odom topic).

flixr:
This is indeed a big problem... is anyone working on this?
I will probably need to wait until this is fixed before I can switch to hydro.

dirk-thomas:
The actually issue has been in ROS since forever. _rospy_ exposes a synchronous blocking publisher API. The problem is that addressing this issue requires asynchronous IO operations. In order to not block the publisher when one subscriber is not able to receive the information (drop-out of wireless, reading the data too slow) each message must be queued per connections and being delivered to the socket from a separate thread (as it is done in _roscpp_)

This can not be done easily since there is no good way to keep the current API for publishing. E.g. signaling error conditions can not be done via exceptions since the error will happen asynchronously.

May be the change @mikeferguson mentioned which he states to be new in "hydro tf/tf2" can be addressed within tf itself. @tfoote any ideas on that?

tfoote:
[...]

PR #308, opened 2013-10-29
implement optional queueing for rospy publications (#169)

dirk-thomas:
Specifying the maximum queue size when creating a publisher (https://github.com/ros/ros_comm/blob/69fb651de8f547c7d48d57ea9cc7c78a492d77d9/clients/rospy/src/rospy/topics.py#L742) will make all `publish() calls being handled asynchronously. The functionality is implemented as an non-default option in order to not change the default behavior.

There are several side effects to this change:
- changing the internal behavior to asynchronous IO will expose error cases only after a repeated `publish()` call to the user code (since they don't happen immediately but asynchronously)
- this implementation will actually start dropping messages on the publisher side if the queue size is exceeded which is not the case for the current/default blocking API
- user land code needs to be updated in order to leverage the new functionality (e.g. in order to address issues like in #169)

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_pubsub.py
+++ b/clients/rospy/src/rospy/impl/tcpros_pubsub.py
@@ -355,3 +355,78 @@ class TCPROSHandler(rospy.impl.transport.ProtocolHandler):
                 topic.add_connection(transport)
             
 
+class QueuedConnection(object):
+    """
+    It wraps a Transport instance and behaves like one
+    but it queues the data written to it and relays them
+    asynchronously to the wrapped instance.
+    """
+
+    def __init__(self, connection, queue_size):
+        """
+        ctor.
+        @param connection: the wrapped transport instance
+        @type  connection: Transport
+        @param queue_size: the maximum size of the queue, zero means infinite
+        @type  queue_size: int
+        """
+        super(QueuedConnection, self).__init__()
+        self._connection = connection
+        self._queue_size = queue_size
+
+        self._lock = threading.Lock()
+        self._cond_data_available = threading.Condition(self._lock)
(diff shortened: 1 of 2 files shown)


tfoote:
+1

wjwwood:
Is the expected behavior that you:
- publish
- an error occurs asynchronously
- the next publish raises
- the next publish succeeds
- an error occurs asynchronously
- the next publish raises

and so on?

Does this make more sense than having the error state persist until explicitly cleared? I don't think this is better, but I just wanted to see what you thought.

dirk-thomas:
Yes, the raise will happen on the subsequent call and only once (not repeatedly). Since non of the existing code knows about this new wrapper adding something for clearing the error is not feasible and I would prefer it not to raise forever.

As far as I have seen the code which uses this wrapper will always close the connection after an exception since the connection is in an unusable state anyway.

wjwwood:
+1

tfoote:
As it stands this is the best for hydro with minimal impact.  The other consideration is what to do going forward.  We should consider making this the default in Indigo.

mikeferguson:
+1

Tested this out on our robot, works exactly as I was hoping. @dirk-thomas @tfoote thanks to both of you for taking care this!

dirk-thomas:
Which publishers have you changed to make it work in your scenario? Then we can make sure that they get fixed in the near term.

mikeferguson:
@dirk-thomas For the most part my issues were in some of my higher level scripts that I was easily able to update. Not too much of the lower-level infrastructure that OSRF maintains uses rospy, although I know SMACH was one place where this change will make a huge difference.

mikeferguson:
I guess tf is the other place where this frequently can be a problem, since for instance, you often have tf subscribers in RVIZ on a remote machine.

PR #309, opened 2013-10-31
xmlrpcpp compile fix for libc++ (OSX)

hgaiser:
When compiling using libc++, <ios> needs to be defined for std::ios_base to be found. Compiles under OSX 10.9 with this fix.

### DIFF
--- a/utilities/xmlrpcpp/include/base64.h
+++ b/utilities/xmlrpcpp/include/base64.h
@@ -13,6 +13,10 @@
 # include <iterator>
 #endif
 
+#if defined(__clang__)
+	#include <ios>
+#endif
+
 static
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I would prefer to not do compiler specific imports here. Since the file uses `std::char_traits` I would suggest importing `string`. Can you please try this instead and provide feedback if that also solves the build issue on 10.9 for you? If yes, can you update your pull request accordingly?

hgaiser:
Including only <code>string</code> still gives me the same errors. I noticed this issue had been brought up before, but forgotten (?)

https://github.com/ros/ros_comm/pull/291

dirk-thomas:
Closing this as a duplicate of #291.

PR #313, opened 2013-11-06
roscpp: fixed name resolution of "/"

meyerj:
This patches fixes a bug in roscpp's names module, that caused the string `"/"` to be resolved as empty string because the trailing slash was removed by `ros::names::clean("/")`. I guess this is not intended.

For example, the following code works to get a XmlRpcValue of all parameters:

```
XmlRpc::XmlRpcValue all_parameters;
assert(ros::param::get("", all_parameters));
```

while the following returns an error:

```
XmlRpc::XmlRpcValue all_parameters;
assert(ros::param::get("/", all_parameters));
```

Additionally, I am not sure if the result of the `ros::names::clean()` function was well-defined for empty strings.

### DIFF
--- a/clients/roscpp/src/libros/names.cpp
+++ b/clients/roscpp/src/libros/names.cpp
@@ -107,7 +107,7 @@ std::string clean(const std::string& name)
     pos = clean.find("//", pos);
   }
 
-  if (*clean.rbegin() == '/')
+  if (clean.size() > 1 && *clean.rbegin() == '/')
   {
     clean.erase(clean.size() - 1, 1);
   }


jbohren:
@dirk-thomas Is this behavior intended?

dirk-thomas:
The intention might be that with a trailing slash you will only get the parameter - without a trailing slash you get the all parameters recursively? But I am not sure what the original intention was when it was implemented.

Another question is: is it worth potentially breaking existing code by changing the current behavior?

jbohren:
> Another question is: is it worth potentially breaking existing code by changing the current behavior?

Right, so in the current behavior, you see an error, so there shouldn't be any systems that are relying on such behavior for anything that actually has an effect.

dirk-thomas:
The proposed patch does not only change the behavior of `ros::param::get()` (which would be fine as you pointed out) but `ros::names::clean()` which is used in many more places.

meyerj:
The problem with `ros::names::clean()` is that it checks the last character of the string by using `*clean.rbegin()`, which, I think, is not well-defined for empty strings because it could result in an invalid read. So at least a check if `clean.size() > 0` should be added here without changing the resolution of empty strings or `"/"` (both will be cleaned up as empty strings).

The `ros::param::get("/")` problem could also be resolved with a modification of `ros::param::getImpl(const std::string &key, XmlRpc::XmlRpcValue &v, bool use_cache)` in [param.cpp:164](https://github.com/meyerj/ros_comm/blob/groovy-devel/clients/roscpp/src/libros/param.cpp#L164), which is less intrusive. Something like

``` cpp
std::string mapped_key = ros::names::resolve(key);
if (mapped_key.empty()) mapped_key = "/";
```

As long as the `mapped_key` is empty here, the XmlRpc call will always fail unless the master itself is patched.

dirk-thomas:
@meyerj A check for `clean.size() > 0` is not necessary. `*clean.rbegin()` will return `\0` for an empty string.

Changing the semantic of `ros::names::clean()` seems to be too risky (at least for me). Therefore I have modified param.cpp as suggested by @meyerj on the _Indigo_ branch: dbc3c3e52f4d2cad6275184a0df1ef818b3296d4

PR #316, opened 2013-11-08
allow different 127. addresses than 127.0.0.1 (fix #315)

### LINKED ISSUE #315: Multiple core using 127.* address can't run
Ros should not handle 127.0.0.0/255.0.0.0 as a unique ip: 127.0.0.1. It makes sence on specific environment to bind on 127.1.2.3 and not on 127.0.0.1 like in openshift gears.

It looks like this might be because adress starting with 127. are converted to 127.0.0.1

dirk-thomas:


tfoote:
+1

PR #321, opened 2013-11-20
support for Visual Studio 2008 and older

ToPitz:
Added fixes to support Microsoft Visual Studio prior 2010

### DIFF
--- a/clients/roscpp/include/ros/common.h.in
+++ b/clients/roscpp/include/ros/common.h.in
@@ -28,7 +28,22 @@
 #ifndef ROSCPP_COMMON_H
 #define ROSCPP_COMMON_H
 
-#include <stdint.h>
+// this is just for interoperability with visual studio, where the standard
+// integer types are not defined.
+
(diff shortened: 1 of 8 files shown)


dirk-thomas:
These typedefs should not be duplicated in multiple files. There should be only one location (probably in `cpp_common`, e.g. in `platform.h`) where those are defined.

That being said I am not sure if it makes sense to add support for Visual Studio 2008 and older for _Hydro_. _Hydro_ has been released in 2013 which is over 5 years newer than this Visual Studio version. Is there any strong argument why not to use a newer Visual Studio version?

ToPitz:
All right, learned something new - I just thought that this could be useful to someone. 
[Personally I wanted to use ros in a project where only 2008 and 2012 dlls are available and I have only a 2008 and 2010 license].

stonier:
We baselined winros support at 2010 since it was by a long margin much more cplusplus compliant than preceding microsoft offerings.

PR #324, opened 2013-12-10
Mavericks fixes

trainman419:
A few fixes that I had to make to trunk to get it to build on OS X 10.9.


dirk-thomas:
Combining the rosconsole libraries is not a great solution since it become pretty indistinguishable which backend is actually used. I chose a different approach to address the problem in #327.

PR #326, opened 2013-12-16
Fix segmentation fault on OS X 10.9 (clang / libc++)

NikolausDemmel:
Accessing front() of an empty std::deque is undefined behavior.
With gcc/libstdc++ it seems fine, on clang / libc++ it causes a segmentation fault.

### DIFF
--- a/utilities/message_filters/include/message_filters/sync_policies/approximate_time.h
+++ b/utilities/message_filters/include/message_filters/sync_policies/approximate_time.h
@@ -540,55 +540,79 @@ private:
     namespace mt = ros::message_traits;
 
     M0Event& m0 = boost::get<0>(deques_).front();
-    M1Event& m1 = boost::get<1>(deques_).front();
-    M2Event& m2 = boost::get<2>(deques_).front();
-    M3Event& m3 = boost::get<3>(deques_).front();
-    M4Event& m4 = boost::get<4>(deques_).front();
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Looks good to me +1
Thanks for the patch.

NikolausDemmel:
Thanks for merging. I should have mentioned that I only did limited testing, and that even only on OS X 10.9.

dirk-thomas:
Please mention this next time you create a pull request.

But in this case the patch only delays the fetching of the elements until they are actually needed. And since the `front()` function has not side effects I am pretty confident that it won't break anything. But the CI job on Jenkins will tell us if we are wrong.

NikolausDemmel:
Yes, sorry for not mentioning it clearly. I don't have a setup where I can easily test more than OS X 10.9 at the moment. I did test with two different filter instances (one with 2 and one with 4 parameters), but I should have run the test suite, I guess. Contributing to the core ros packages is new for me.

Does CI test open pull requests?

dirk-thomas:
No, it doesn't at the moment. It only runs on changes on the devel branch (aka after the pull request has been merged).

NikolausDemmel:
Jenkins fails, however I don't see how this is related to this patch...
http://jenkins.ros.org/job/devel-hydro-ros_comm/ARCH_PARAM=amd64,UBUNTU_PARAM=precise,label=devel/lastBuild/

PR #328, opened 2013-12-31
Transport information in SlaveAPI::getBusInfo() for `roscpp` and `rospy` nodes

mani-monaj:
With the following changes, `roscpp` and `rospy` nodes will append an extra field to `getBusInfo()` response [in [slave XMLRPC API](http://wiki.ros.org/ROS/Slave_API)] to include detailed transport information. The format is as follows:

_CONNTYPE_ connection on port _LOCALPORT_ to [_REMOTEHOST_:_PORT_ on socket  _SOCKETNUMBER_]

e.g.

```
TCPROS connection on port 56858 to [127.0.0.1:32993 on socket 10]
```
- It looks as if `UDPROS` is not completely supported by `rospy`. Therefore no response will be sent by `rospy` nodes using UDP (is it even possible?).
- For `roscpp` and UDP, I could not find a way to reliably determine `remotehost:port`  and `socketnumber` for _subscriber_ links. (_publisher_ links work fine.)
- This PR also adds an extra field to `roscpp` `getBusInfo()` responses to make the length equal to responses sent by `rospy` nodes.  

This is my first time to send pull request to core ROS packages. Any comment or hints is highly appreciated. I am also more than happy to write tests for the new feature if it is required.

### DIFF
--- a/clients/roscpp/src/libros/intraprocess_publisher_link.cpp
+++ b/clients/roscpp/src/libros/intraprocess_publisher_link.cpp
@@ -127,6 +127,12 @@ std::string IntraProcessPublisherLink::getTransportType()
   return std::string("INTRAPROCESS");
 }
 
+std::string IntraProcessPublisherLink::getTransportInfo()
+{
+  // TODO: Check if we can dump more useful information here
+  return getTransportType();
+}
+
 void IntraProcessPublisherLink::getPublishTypes(bool& ser, bool& nocopy, const std::type_info& ti)
 {
   boost::recursive_mutex::scoped_lock lock(drop_mutex_);
--- a/clients/roscpp/src/libros/publication.cpp
+++ b/clients/roscpp/src/libros/publication.cpp
@@ -282,11 +282,8 @@ XmlRpc::XmlRpcValue Publication::getStats()
   return stats;
 }
 
-// rospy returns values like this:
-// [(1, "('127.0.0.1', 62334)", 'o', 'TCPROS', '/chatter')]
-//
-// We're outputting something like this:
-// (0, (127.0.0.1, 62707), o, TCPROS, /chatter)
+// Publisher : [(connection_id, destination_caller_id, direction, transport, topic_name, connected, connection_info_string)*]
+// e.g. [(2, '/listener', 'o', 'TCPROS', '/chatter', 1, 'TCPROS connection on port 55878 to [127.0.0.1:44273 on socket 7]')]
 void Publication::getInfo(XmlRpc::XmlRpcValue& info)
 {
   boost::mutex::scoped_lock lock(subscriber_links_mutex_);
@@ -300,6 +297,8 @@ void Publication::getInfo(XmlRpc::XmlRpcValue& info)
     curr_info[2] = "o";
     curr_info[3] = (*c)->getTransportType();
     curr_info[4] = name_;
+    curr_info[5] = true; // For length compatibility with rospy
+    curr_info[6] = (*c)->getTransportInfo();
     info[info.size()] = curr_info;
   }
 }
--- a/clients/roscpp/src/libros/subscription.cpp
+++ b/clients/roscpp/src/libros/subscription.cpp
@@ -118,11 +118,8 @@ XmlRpcValue Subscription::getStats()
   return stats;
 }
 
-// rospy returns values like this:
-// (1, 'http://127.0.0.1:62365/', 'i', 'TCPROS', '/chatter')
-//
-// We're outputting something like this:
-// (0, http://127.0.0.1:62438/, i, TCPROS, /chatter)
+// [(connection_id, publisher_xmlrpc_uri, direction, transport, topic_name, connected, connection_info_string)*]
+// e.g. [(1, 'http://host:54893/', 'i', 'TCPROS', '/chatter', 1, 'TCPROS connection on port 59746 to [host:34318 on socket 11]')]
 void Subscription::getInfo(XmlRpc::XmlRpcValue& info)
 {
   boost::mutex::scoped_lock lock(publisher_links_mutex_);
@@ -136,6 +133,8 @@ void Subscription::getInfo(XmlRpc::XmlRpcValue& info)
     curr_info[2] = "i";
     curr_info[3] = (*c)->getTransportType();
     curr_info[4] = name_;
+    curr_info[5] = true; // For length compatibility with rospy
+    curr_info[6] = (*c)->getTransportInfo();
     info[info.size()] = curr_info;
   }
 }
--- a/clients/roscpp/src/libros/transport/transport_tcp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_tcp.cpp
@@ -55,6 +55,7 @@ TransportTCP::TransportTCP(PollSet* poll_set, int flags)
 , expecting_write_(false)
 , is_server_(false)
 , server_port_(-1)
+, local_port_(-1)
 , poll_set_(poll_set)
 , flags_(flags)
 {
@@ -110,6 +111,21 @@ bool TransportTCP::initializeSocket()
       std::stringstream ss;
       ss << getClientURI() << " on socket " << sock_;
       cached_remote_host_ = ss.str();
+    }
+  }
+
+  if (local_port_ < 0)
+  {
+    la_len_ = s_use_ipv6_  ? sizeof(sockaddr_in6) : sizeof(sockaddr_in);
+    getsockname(sock_, (sockaddr *)&local_address_, &la_len_);
+    switch (local_address_.ss_family)
+    {
+      case AF_INET:
+        local_port_ = ntohs(((sockaddr_in *)&local_address_)->sin_port);
+        break;
+      case AF_INET6:
+        local_port_ = ntohs(((sockaddr_in6 *)&local_address_)->sin6_port);
+        break;
     }
   }
 
@@ -217,6 +233,7 @@ bool TransportTCP::connect(const std::string& host, int port)
     sockaddr_in *address = (sockaddr_in*) &sas;
     sas_len = sizeof(sockaddr_in);
     
+    la_len_ = sizeof(sockaddr_in);
     address->sin_family = AF_INET;
     address->sin_port = htons(port);
     address->sin_addr.s_addr = ina.s_addr;
@@ -225,6 +242,7 @@ bool TransportTCP::connect(const std::string& host, int port)
   {
     sockaddr_in6 *address = (sockaddr_in6*) &sas;
     sas_len = sizeof(sockaddr_in6);
+    la_len_ = sizeof(sockaddr_in6);
     address->sin6_family = AF_INET6;
     address->sin6_port = htons(port);
     memcpy(address->sin6_addr.s6_addr, in6a.s6_addr, sizeof(in6a.s6_addr));
@@ -682,7 +700,9 @@ void TransportTCP::socketUpdate(int events)
 
 std::string TransportTCP::getTransportInfo()
 {
-  return "TCPROS connection to [" + cached_remote_host_ + "]";
+  std::stringstream str;
+  str << "TCPROS connection on port " << local_port_ << " to [" << cached_remote_host_ << "]";
+  return str.str();
 }
 
 std::string TransportTCP::getClientURI()
--- a/clients/roscpp/src/libros/transport/transport_udp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_udp.cpp
@@ -57,6 +57,7 @@ TransportUDP::TransportUDP(PollSet* poll_set, int flags, int max_datagram_size)
 , expecting_write_(false)
 , is_server_(false)
 , server_port_(-1)
+, local_port_(-1)
 , poll_set_(poll_set)
 , flags_(flags)
 , connection_id_(0)
@@ -131,7 +132,9 @@ void TransportUDP::socketUpdate(int events)
 
 std::string TransportUDP::getTransportInfo()
 {
-  return "UDPROS connection to [" + cached_remote_host_ + "]";
+  std::stringstream str;
+  str << "UDPROS connection on port " << local_port_ << " to [" << cached_remote_host_ << "]";
+  return str.str();
 }
 
 bool TransportUDP::connect(const std::string& host, int port, int connection_id)
@@ -208,6 +211,10 @@ bool TransportUDP::connect(const std::string& host, int port, int connection_id)
   Sleep(100);
 #endif
 
+  std::stringstream ss;
+  ss << host << ":" << port << " on socket " << sock_;
+  cached_remote_host_ = ss.str();
+
   if (!initializeSocket())
   {
     return false;
@@ -267,6 +274,10 @@ bool TransportUDP::initializeSocket()
       return false;
     }
   }
+
+  socklen_t len = sizeof(local_address_);
+  getsockname(sock_, (sockaddr *)&local_address_, &len);
+  local_port_ = ntohs(local_address_.sin_port);
 
   ROS_ASSERT(poll_set_ || (flags_ & SYNCHRONOUS));
   if (poll_set_)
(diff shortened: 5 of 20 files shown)


dirk-thomas:
This additional string is not machine readable and it looks like it is only usable for debugging purposes. Can you please explain the use case you need this for?

mani-monaj:
> This additional string is not machine readable and it looks like it is only usable for debugging purposes. 

You are right. I appended the information to the output of `getTransportInfo()` function that is already available for TCP/UDP transports. That was mainly because that function was being used for logging purposes and I did not want to break it. I was also did not want to add another function or change the interface. I currently use regex to parse the data, but if it is more appropriate to have machine readable version of `getTransportInfo`, I will be more than happy to contribute. 

> Can you please explain the use case you need this for?

In short: monitoring and QoS. I am currently working on a distributed monitoring system to monitor robots. The system maps the dynamic graph of processes and their inter-connections to OS level entities, then monitors those entities externally. In order to monitor the data flow between processes, it needs the low level transport information. The system is in the process of [being released](http://autonomylab.org/drums/). It currently supports ROS, but it is also designed to be able to attach to other middleware through adapters. The ROS adapter however is not [fully functional](https://github.com/AutonomyLab/drums/tree/master/drums-roscomm-patch) without low level transport information.

The other application is QoS. Knowing which ports are being used by each publisher/subscriber, ROS traffic can be shaped to meet certain QoS requirements. That is something I am currently looking into.

In addition, this information is helpful to fix current issues with running ROS nodes inside Linux containers (e.g https://github.com/jonbinney/docker_ros#publishing).

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -90,7 +91,6 @@ bool TransportTCP::setNonBlocking()
 bool TransportTCP::initializeSocket()
 {
   ROS_ASSERT(sock_ != ROS_INVALID_SOCKET);
-
```
dirk-thomas, line 93: Please avoid making any whitespace changes like this.

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -368,7 +384,6 @@ bool TransportTCP::listen(int port, int backlog, const AcceptCallback& accept_cb
 
   ::listen(sock_, backlog);
   getsockname(sock_, (sockaddr *)&server_address_, &sa_len_);
-
```
dirk-thomas, line 371: Please avoid making any whitespace changes like this.

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -378,7 +393,6 @@ bool TransportTCP::listen(int port, int backlog, const AcceptCallback& accept_cb
       server_port_ = ntohs(((sockaddr_in6 *)&server_address_)->sin6_port);
       break;
   }
-
```
dirk-thomas, line 381: Please avoid making any whitespace changes like this.

clients/roscpp/src/libros/transport/transport_udp.cpp
```diff
@@ -131,7 +132,9 @@ void TransportUDP::socketUpdate(int events)
 
 std::string TransportUDP::getTransportInfo()
 {
-  return "UDPROS connection to [" + cached_remote_host_ + "]";
+    std::stringstream str;
+    str << "UDPROS connection on port " << local_port_ << " to [" << cached_remote_host_ << "]";
+    return str.str();
```
dirk-thomas: Please keep indentation level with 2 spaces.

clients/roscpp/src/libros/transport/transport_udp.cpp
```diff
@@ -221,7 +228,6 @@ bool TransportUDP::connect(const std::string& host, int port, int connection_id)
 bool TransportUDP::createIncoming(int port, bool is_server)
 {
   is_server_ = is_server;
-
```
dirk-thomas, line 224: Please avoid making any whitespace changes like this.

clients/roscpp/src/libros/transport/transport_udp.cpp
```diff
@@ -257,7 +263,6 @@ bool TransportUDP::createIncoming(int port, bool is_server)
 bool TransportUDP::initializeSocket()
 {
   ROS_ASSERT(sock_ != ROS_INVALID_SOCKET);
-
```
dirk-thomas, line 260: Please avoid making any whitespace changes like this.

clients/roscpp/src/libros/transport/transport_udp.cpp
```diff
@@ -677,4 +686,25 @@ TransportUDPPtr TransportUDP::createOutgoing(std::string host, int port, int con
...
+{
+  ROS_ASSERT(!is_server_);
+
+  sockaddr_storage sas;
+  socklen_t sas_len = sizeof(sas);
+  getpeername(sock_, (sockaddr *)&sas, &sas_len);
+
+  sockaddr_in *sin = (sockaddr_in *)&sas;
+
+  char namebuf[128];
```
dirk-thomas, line 701: Can you please describe why you choose 128 bytes?

clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -461,6 +461,17 @@ def __init__(self, protocol, name, header=None):
...
         self.type = None 
             
+    def get_transport_info(self):
+        """
+        Get detailed connection information.
+        Similar to getTransportInfo() in 'libros/transport/transport_tcp.cpp'
+        e.g. TCPROS connection on port 41374 to [127.0.0.1:40623 on socket 6]
+        """
+        try:
+            return "%s connection on port %s to [%s:%s on socket %s]" % (self.transport_type, self.local_endpoint[1], self.remote_endpoint[0], self.remote_endpoint[1], self._fileno)
```
dirk-thomas: Which part of this line can raise a `TypeError`?

dirk-thomas:
I made a few comments but all pretty minor. Please address them when it is convenient for you.

The PR looks very promising but I will likely not merge it into _Hydro_ but only to the _Indigo_ branch for now. It could be considered for backporting in the future if there is a strong need for it though.

Btw. have you checked the compatibility between an old and new client (one without and one with this patch)?

mani-monaj:
@dirk-thomas Thank you for your feedback.

I'll fix the white-space and indention issues. I will add comments regarding your other questions as line notes.

clients/roscpp/src/libros/transport/transport_udp.cpp
mani-monaj, line 701: @dirk-thomas I used the same pattern used in `TransportTCP` to get peername : [here](https://github.com/ros/ros_comm/blob/4e8a495fdbdcf8088ae7cc569c3c0a8a45c9c9b3/clients/roscpp/src/libros/transport/transport_tcp.cpp#L248) and [here](https://github.com/ros/ros_comm/blob/4e8a495fdbdcf8088ae7cc569c3c0a8a45c9c9b3/clients/roscpp/src/libros/transport/transport_tcp.cpp#L699).

dirk-thomas, line 701: Good enough - referencing other code from a sibling file :wink:

clients/rospy/src/rospy/impl/tcpros_base.py
mani-monaj: The constructor initializes both `local_endpoint` and `remote_endpoint` to `None`. I was afraid that if for any reason `get_transport_info` get called before `set_socket` or `connect`, it would fail with `TypeError`. I think I can fix that by initializing both to `(0,0)` instead of `None`.

dirk-thomas: Yeah, I think if all values have some reasonable default the resulting string makes more sense then to have a different fallback (where it is unclear for the user why).

mani-monaj:
> I made a few comments but all pretty minor. Please address them when it is convenient for you.

Sure. I added line comments for some. I will all issues.

> The PR looks very promising but I will likely not merge it into Hydro but only to the Indigo branch for now. It could be considered for backporting in the future if there is a strong need for it though.

Thank you. I am currently working to prepare the [monitoring tool](http://autonomylab.org/drums/) that relies on this patch for release (which I hope will be some time this month). I think if _Indigo_ is going to be released on schedule, I can provide patched debs for _Hydro_ for people who want to try that before May. That is at least the case in our lab.

> Btw. have you checked the compatibility between an old and new client (one without and one with this patch)?

Not formally. But we have both the patched `rospy` and `roscpp` running on some of our robots and computers. I did experiments involving both patched and unpatched clients. I could not see any regression or incompatibility. If you could provide me with some hints about how to formally test that, I would do my best to ensure compatibility. Do you use any test suites for testing compatibility between different _ROS_ clients?

dirk-thomas:
No, there is no test suite for cross distro compatibility. The mixed usage is probably the test we can do for testing.

I wouldn't expect problems since all those args should be optional anyway. But it is always good to know that you have used them together and haven't experienced any problems yet.

mani-monaj:
I fixed the issues you mentioned (the PR is now updated). However the PR is still for _hydro-devel_ branch. Should I apply that to _indigo-devel_ and open a new pull request?

clients/roscpp/src/libros/intraprocess_publisher_link.cpp
```diff
@@ -127,6 +127,13 @@ void IntraProcessPublisherLink::handleMessage(const SerializedMessage& m, bool s
   return std::string("INTRAPROCESS");
 }
 
+
```
dirk-thomas: Please remove this extra newline.

dirk-thomas:
There are three whitespace comments pending (two old, one new). Can please address them too?

Afterwards can you squash your commits into a single commit then?

I will cherry-pick your commit to the _indigo-devel_ branch - no need to open a new PR for this.

mani-monaj:
> There are three whitespace comments pending (two old, one new). Can please address them too?

I fixed remaining white space issues.

> Afterwards can you squash your commits into a single commit then?

Done.

> I will cherry-pick your commit to the indigo-devel branch - no need to open a new PR for this.

Thank you.

clients/roscpp/src/libros/publication.cpp
```diff
@@ -283,10 +283,10 @@ void Publication::removeSubscriberLink(const SubscriberLinkPtr& sub_link)
 }
 
 // rospy returns values like this:
-// [(1, "('127.0.0.1', 62334)", 'o', 'TCPROS', '/chatter')]
+// [(1, "('127.0.0.1', 62334)", 'o', 'TCPROS', '/chatter', BOOL)]
 //
 // We're outputting something like this:
-// (0, (127.0.0.1, 62707), o, TCPROS, /chatter)
+// (0, /listener, o, TCPROS, /chatter, BOOL, TCPROS connection on port X to [Y:Z on socket S])
```
dirk-thomas: Regarding the change of the second value from `(127.0.0.1, 62707)` to `/listener`: is that updating the doc to the current behavior or does the PR change the value intentionally?

mani-monaj: I updated the doc to reflect the current behaviour.

mani-monaj: Except the last field (which was the addition of this PR), the current implementation reports transport information in the updated format.

clients/roscpp/src/libros/publication.cpp
```diff
@@ -283,10 +283,10 @@ void Publication::removeSubscriberLink(const SubscriberLinkPtr& sub_link)
 }
 
 // rospy returns values like this:
-// [(1, "('127.0.0.1', 62334)", 'o', 'TCPROS', '/chatter')]
+// [(1, "('127.0.0.1', 62334)", 'o', 'TCPROS', '/chatter', BOOL)]
```
dirk-thomas: As of this PR rospy also return another string value after the boolean, right?

clients/roscpp/src/libros/subscription.cpp
```diff
@@ -119,10 +119,10 @@ XmlRpcValue Subscription::getStats()
 }
 
 // rospy returns values like this:
-// (1, 'http://127.0.0.1:62365/', 'i', 'TCPROS', '/chatter')
+// (1, 'http://127.0.0.1:62365/', 'i', 'TCPROS', '/chatter', BOOL)
```
dirk-thomas: Same here?

mani-monaj: You are right. I should have updated these two comment lines. However the comment should not be same for publish and subscribe links (the second element is different):

```
Publisher: [CONNID, DEST_NODENAME, DIR, TRANSPORT, TOPICNAME, 1, CONNINFO_STRING]

Subscriber: [CONNID, PUB_XMLRPCURI, DIR, TRANSPORT, TOPICNAME, 1, CONNINFO_STRING]
```

Should I fix that too?

dirk-thomas: That would be great - then the comments shows exactly what to expect. Thank you for your effort.

mani-monaj: Sure, I'll do that very soon and will update this PR with a new squashed commit.

mani-monaj: @dirk-thomas Comments (docs) have been updated. _roscpp_ and _rospy_ comments are now consistent.

dirk-thomas:
@esteve @tfoote @wjwwood Please review. The PR will go into _indigo-devel_ though.

clients/rospy/src/rospy/impl/transport.py
```diff
@@ -105,6 +109,11 @@ def close(self):
     def write_data(self, data):
         raise Exception("not implemented")
 
+    ## Implements the getTransportInfo() from roscpp
+    ## Similar to getTransportInfo() in 'libros/transport/transport_tcp.cpp'
+    def get_transport_info(self):
+        return ""
```
esteve: Would it make more sense to raise an exception instead? IMHO it'd be more descriptive and force implementations to return actual transport information.

tfoote:
+1

clients/rospy/src/rospy/impl/transport.py
dirk-thomas: That sounds like a good idea:

```
raise NotImplementedError
```

@mani-monaj Can you please update the PR with this?

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -113,6 +114,21 @@ bool TransportTCP::initializeSocket()
...
+      getsockname(sock_, (sockaddr *)&local_address_, &la_len_);
+      switch (local_address_.ss_family)
+      {
+        case AF_INET:
+          local_port_ = ntohs(((sockaddr_in *)&local_address_)->sin_port);
+          break;
+        case AF_INET6:
+          local_port_ = ntohs(((sockaddr_in6 *)&local_address_)->sin6_port);
+          break;
+      }
```
dirk-thomas: Please remove two spaces from this blocks indentation to make it indented the same way as the rest of the code.

esteve:
+1

clients/rospy/src/rospy/impl/transport.py
mani-monaj: Sure.

clients/roscpp/src/libros/transport/transport_tcp.cpp
mani-monaj: Sure.

clients/rospy/src/rospy/impl/transport.py
mani-monaj: Done. 

There seems to be two classes that inherit from `Transport` class: `TCPROSTransport` and `DeadTransport`. The former has an implementation for `get_transport_info()`, but the latter does not. I don't quite understand how `DeadTransport` works. Should it implement `get_transport_info()`?

clients/roscpp/src/libros/transport/transport_tcp.cpp
mani-monaj: Done.

clients/rospy/src/rospy/impl/transport.py
dirk-thomas: The comment of DeadTransport describes briefly what it is used for:

```
## Shell class to hold stats about transport that is being killed off.
## This allows the information to stick around but the original Tranport to be gc'd
```

Please go ahead and implement `get_transport_info()` for it too. The info string could be something like "connection has been closed" or something similar.

mani-monaj: Thanks @dirk-thomas . I updated the PR. I implemented  `DeadTransport::get_transport_info()` [this way](https://github.com/ros/ros_comm/pull/328/files#diff-2968e388f9763814f35907d28b4813f0R132).

dirk-thomas:
@mani-monaj Thank you very much for your efforts!

+1 I will try to run the unit tests soon. If everything passes I will go ahead and merge it into _indigo-devel_.

mani-monaj:
@dirk-thomas You are welcome. Thank you for all your suggestions and help.

dirk-thomas:
Merged in 6980ae43fc50fee7227b5326db7b230c803777bb.

PR #329, opened 2014-01-03
No more dry packages. Remove run_depend on Lisp.

trainman419:
This fixes install problems on ARM, where lisp isn't available.


dirk-thomas:
We still do support building dry packages from source (at will support it in the future). Therefore this dependency can't be removed.

trainman419:
Ok. I will keep this in my ARM branch.

PR #336, opened 2014-01-12
readd g_level_lockup symbol for backward compatibility when log4cxx is being used

dirk-thomas:
@tfoote @wjwwood I am not sure if you have seen this:
- http://answers.ros.org/question/116436/missing-log4cxx-on-latest-ros_comm/
- https://github.com/ros-drivers/hokuyo_node/pull/8

Since I don't think that it is a good idea to pull the changes back out I would propose to readd the symbol in question (`g_level_lookup`) as done in this pull request. That backward compatible symbol can be removed for indigo again. Downstream packages will still need to explicitly state their dependency on log4cxx - I think that is reasonable since these packages are simply relying on a transitive dependency (which is gone now).

### DIFF
--- a/tools/rosconsole/CMakeLists.txt
+++ b/tools/rosconsole/CMakeLists.txt
@@ -47,7 +47,7 @@ catkin_package(
   CFG_EXTRAS rosconsole-extras.cmake
 )
 
-include(cmake/rosconsole-extras.cmake)
+include(${CATKIN_DEVEL_PREFIX}/share/${PROJECT_NAME}/cmake/rosconsole-extras.cmake)
 
 # See ticket: https://code.ros.org/trac/ros/ticket/3626
 # On mac use g++-4.2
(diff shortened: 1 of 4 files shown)


wjwwood:
Seems like the most reasonable approach to fix the API break.

trainman419:
I like it. :+1: :beers:

tfoote:
+1

130s:
@dirk-thomas Looks like your got good number of thumbs. Could you soon merge it and release? As Tully explained to me in in-person conversation, I expect with this patch that buildfarm resumes to auto-sync to shadow-fixed (I'm kind of in need of my newly built packages). Thanks!

dirk-thomas:
Related tickets for missing explicit dependencies pending:
- [x] PR2/pr2_ethercat_drivers#62
- [x] ros-drivers/freenect_stack#10
- [x] ros-drivers/hokuyo_node#7
- [x] ros-drivers/microstrain_3dmgx2_imu#3
- [x] ros-drivers/openni_camera#21
- [x] PR2/pr2_power_drivers#64

trainman419:
rosdistro PRs for pr2_ethercat_drivers and pr2_power_drivers are ready: https://github.com/ros/rosdistro/pull/2880 and https://github.com/ros/rosdistro/pull/2881

tfoote:
All packages have been patched and released.

PR #340, opened 2014-01-22
fix return value of pubUpdate() (fix #334)

### LINKED ISSUE #334: Wrong response to registerPublisher on C++ subscriber.
Looking at the network traces when bringing up a very simple ROS network we found that the `publisherUpdate` method receives a wrong answer.

The simplest case we found to reproduce this problem is launching (in different terminals):
1. $ roscore
2. $ rosrun  roscpp_tutorials listener
3. $ rosrun  roscpp_tutorials talker

When we bring up `talker` the master correctly calls the `publisherUpdate` method of `listener` (seen in the first part of the trace). Next, `listener` responds with 0 status, which means ERROR, instead of 1.

However, after this, the communication is established between the nodes, and works. The trace is found below.

This issue is not present when using rospy nodes (rostopic echo), where we see the correct response.

This may seem as a non issue, as the resulting network works. However, we are experiencing severe compatibility problems when mixing C++ node and the Java master. We started this investigation because we are also affected by the problem described in this [question](http://answers.ros.org/question/86296/rosjava_core-turtlesim-compatibility-problems/) in ROS Answers.

```
POST / HTTP/1.1
Host: 127.0.0.1:39400
Accept-Encoding: gzip
User-Agent: xmlrpclib.py/1.0.1 (by www.pythonware.com)
Content-Type: text/xml
Content-Length: 342

<?xml version='1.0'?>
<methodCall>
<methodName>publisherUpdate</methodName>
<params>
<param>
<value><string>/master</string></value>
</param>
<param>
<value><string>/chatter</string></value>
</param>
<param>
<value><array><data>
<value><string>http://127.0.0.1:52800/</string></value>
</data></array></value>
</param>
</params>
</methodCall>

HTTP/1.1 200 OK
Server: XMLRPC++ 0.7
Content-Type: text/xml
Content-length: 203

<?xml version="1.0"?>
<methodResponse><params><param>
.<value><array><data><value><i4>0</i4></value><value></value><value><i4>0</i4></value></data></array></value>
</param></params></methodResponse>
```

dirk-thomas:

### DIFF
--- a/clients/roscpp/src/libros/topic_manager.cpp
+++ b/clients/roscpp/src/libros/topic_manager.cpp
@@ -580,7 +580,7 @@ bool TopicManager::pubUpdate(const string &topic, const vector<string> &pubs)
 
   if (sub)
   {
-    sub->pubUpdate(pubs);
+    return sub->pubUpdate(pubs);
   }
   else
   {


wjwwood:
+1

tfoote:
+1

PR #341, opened 2014-01-23
Add $(shell ...) substitution (fix #102).

### LINKED ISSUE #102: roslaunch substitution based on shell commands
Migrated from https://code.ros.org/trac/ros/ticket/4072
Reported by:    raywilhe    Owned by:   kwc
Priority:   minor 
Last modified ago: Dec 2012

---

I added a shell based substitution arg (similar to how you can use environment variables within launch files). For example, the following:

```
    <param name="/my_$(shell hostname | perl -ne 'chomp and print')_and_date" value="$(shell date)" type="str" />
```

gives the following (my hostname is rabbit):

```
rosparam get /my_rabbit_and_date
Wed Dec 26 14:49:36 EST 2012
```

This is my first time contributing to ROS, so please let me know if I should submit patches through a different avenue.

RainCT:
This is an updated version of the patch by raywilhe in https://github.com/ros/ros_comm/issues/102.

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -38,6 +38,7 @@ by roslaunch and xacro, but it is not yet a top-level ROS feature.
 """
 
 import os
+import subprocess
 
 try:
     from cStringIO import StringIO # Python 2.x
@@ -256,6 +257,15 @@ def _arg(resolved, a, args, context):
         raise ArgException(arg_name)
 
 
+def _shell(resolved, a, args, context):
+    """
+    process $(shell) arg
+    :return: updated resolved argument, ``str``
+    """
+    value = subprocess.check_output(' '.join(args), shell=True, universal_newlines=True)
+    return resolved.replace("$(%s)"%a, value)
+
+
 def resolve_args(arg_str, context=None, resolve_anon=True):
     """
     Resolves substitution args (see wiki spec U{http://ros.org/wiki/roslaunch}).
@@ -290,6 +300,7 @@ def resolve_args(arg_str, context=None, resolve_anon=True):
         'optenv': _optenv,
         'anon': _anon,
         'arg': _arg,
+        'shell': _shell,
     }
     resolved = _resolve_args(arg_str, context, resolve_anon, commands)
     # than resolve 'find' as it requires the subsequent path to be expanded already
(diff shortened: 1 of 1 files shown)


po1:
Are there practical use cases for this? I know the `<arg>` tag is far from perfect, but you could very well use it to pass the output of a shell command to launch arguments.
It looks to me that it would add more complexity to launch files, and thus increase the loading time.
Also, I am afraid that this would lead to more incompatibilities between platforms if people started to use this feature extensively.

RainCT:
I'm sure there are, although I'm not too interested in it, what I really want is supporting eg. string comparisons in ifs (may submit a patch for that sometime). That would allow us to combine many duplicate launch files where only little details change (eg. depending on the robot).

dirk-thomas:
The shell substitution this PR adds is conceptionally problematic since it encourages platform specific code in launch files. Since you already mentioned that you actually don't need that "power" for your use case a more specific solution should be developed. Depending on how severe the change / extension to roslaunch should be it might be worth to go through a community review via the REP process (http://www.ros.org/reps/rep-0000.html).

RainCT:
Well, launch files _are_ platform specific.

One problem we have here is that we have a copy of all launch files for each robot/model, because there are mostly subtle differences between them; this is not at all optimal, since there's a lot of duplication and files can get out of sync.

This could be avoided if roslaunch implemented: a) string comparison, and b) integer comparisons (<, >, =). Since those could be combined with "env" it would let us unify most launch files.

Would this (string and integer comparisons) require a REP or should I just submit a patch?

dirk-thomas:
Launch files are not by design platform specific. And actually a lot of launch files are definitely not.

Current launch files only support simple conditions (see http://wiki.ros.org/roslaunch/XML#if_and_unless_attributes). If you would like to work on a pull request which adds the proposed comparisons that would be great and imo does not require a REP. It might be good to discuss the way you want to realize it before diving into the code to avoid work in the "wrong" direction.

Please also consider providing the corresponding documentation for the wiki page once it the pull request is done.

RainCT:
Hi Dirk,

Thanks for your response. My idea was adding substitutions, for instance: "$(streq a b)", "$(eq a b)", "$(lt a b)" and "$(le a b)" (or maybe "$(cmp a OP b)" with OP \in {<,>,<=,>=,==}).

This would require support for substitution nesting, which I've implemented here: https://github.com/RainCT/ros_comm/commit/779e83fb465303fabb4aca934decd5044bae90df (I'll open a pull request for this once I've tested it a bit).

dirk-thomas:
This is replaced by #382 and a to-be-written pull request adding the comparison functions. Therefore closing this ticket.

PR #349, opened 2014-02-04
modify rostest to wait when other instances are running

dirk-thomas:
Until now rostest just fails when another instance is running. With this modification it actually waits (with a message) until other instances have finished. Than allows catkin to drop it's special handling of running unit tests single threaded.

For a workspace with the base ROS packages the time necessary to run the unit tests is reduced by roughly 15%. The result might vary significantly bases on the number and kind of tests and their order in CMake. (E.g. it does not prevent CMake to schedule N rostests where N-1 of them will ust be waiting.)

### DIFF
--- a/tools/rostest/scripts/rostest
+++ b/tools/rostest/scripts/rostest
@@ -31,5 +31,23 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+import socket
+import time
 from rostest import rostestmain
+
+message_printed = False
+while True:
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
+        s.bind(('127.0.0.1', 22421))  # choose port close to test roscore
+        break
+    except socket.error as e:
+        if e.errno != socket.errno.EADDRINUSE:
+            raise
+        if not message_printed:
+            print('Another rostest instance is currently running, waiting until it finished...')
+            message_printed = True
+        time.sleep(0.25)
+        continue
+
 rostestmain()


tfoote:
+1

wjwwood:
Is there a contingency for Windows? I know we are not extremely concerned about supporting it, but it would be better to gracefully fail in the case the `fcntl` does not exist.

dirk-thomas:
Even if I don't like the socket solution it is indeed better for cross platform.

@tfoote @wjwwood Should we use this rather then the fcntl approach?

wjwwood:
That actually seems like an elegant solution, though it does waste a port, but I don't really think that is a problem. Are there any major draw backs to using this method?

Another option would be to checkout how this module works:

https://pypi.python.org/pypi/lockfile

It provides portable file locking using linking on Unix and `mkdir` on Windows, and it is MIT licensed so you could conceivably just steal the classes/functions you need from it.

wjwwood:
https://github.com/smontanaro/pylockfile

dirk-thomas:
There are two difficulties with lock files:
- you need to find a place which is writable
- the lock file must be 100% guaranteed removed in any case (sig kill, power out, etc.)

Since especially the last is pretty difficult to achieve and requires quite some custom code I went for the solution with the socket. And the `rostest` later will anyway use a socket so it's not such a big issue (if not exactly that socket is in use by something else).

We could also use the `fcntl` approach if available and fall back to the socket approach on Windows. But the `fcntl` approach is also not perfect in case of NFS.

wjwwood:
I would vote for the socket method on all platforms then.

The only other reason to not use the socket that I can think of is that it might not be as performant as `fcntl`, but I think that would only ever be a problem if you were locking and unlocking very fast which we will not be doing.

tfoote:
I agree use the socket instead of the fnctl

PR #350, opened 2014-02-05
Simplify code: map.erase() can take a string directly

RainCT:
Removes a few lines...

Out of curiosity, what's the reason for having g_subscribed_params in the first place? Does it make a speed difference (by requiring fewer memory accesses when the key is not in the subscriber list)?


RainCT:
Actually, the "g_subscribed_params.erase(sub_it);" line looks wrong, you're still subscribed since (the parameter server's) unsubscribeParam isn't called anywhere.

RainCT:
OK, I've addressed the later issue in my fix for #351 (in a separate pull request).

RainCT:
I'm closing this since it got mixed up with pull request #352.

Issue #351, opened 2014-02-05
[roscpp] Parameter caching doesn't work correctly for namespaces


RainCT:
Currently using getParamCached on a namespace results in:
 a) The parameter not being updated correctly when one of its childs changes.
 b) A memory leak where the value of the child is blindly stored into the g_params map (and never removed).

I'll submit a patch.

dirk-thomas:
Can you please implement a unit test for this issue which fails before and passes after your patch?

RainCT:
Sure, I've added one to my pull request.

RainCT:
Do you have any more feedback on this?

dirk-thomas:
The changes PR #352 does are pretty significant. It combines bug fixes (the missing invalidation of the cached parent parameters) with new features (adding caching of non-existing parameters) which makes it more difficult to judge the impact and potential issues.

I have separated the bug fix into a separate PR which should address **a)** of your original comment: #384 

Can you please point out where exactly the memory leak is which you mentioned in **b)**?

RainCT:
"g_params[clean_key] = v;" in update() is adding a new key to the cache dictionary, which isn't tracked anywhere  (note that if you subscribe to X and then store a tree in X, update will be called with X/foo whenever a child foo of X changes).

dirk-thomas:
`g_params` stores the cached parameters until they are `del()`eted. I don't understand what you mean with "isn't tracked anywhere" and why that would be a memory leak.

Regarding the update notification: based on your short description I would say it is intended that the subscriber of X continues to be notified. May be I don't get your use case correctly. Could you create a unit test for this (which fails currently and passes with the patch proposed in this PR)?

RainCT:
$ rosparam set /foo 'hi'
c++: Code subscribes to /foo

$ rosparam set /foo '{sub: 1}'
c++: updates g_params['/foo'], ok
$ rosparam set /foo/sub 2
c++: since you are subscribed to /foo, update('/foo/sub', ...) is called: it creates g_params['/foo/sub'] (which will never be accessed)
$ rosparam delete /foo
c++: g_params['/foo'] is erased, but g_params['/foo/sub'] is leaked

IIRC, with my patch the behaviour will be:
  a) if update('/foo', ...) is called, it is updated
[...]

PR #352, opened 2014-02-05
Fix parameter caching (esp. in regards to namespaces)

RainCT:
Hi,

This pull request addresses issue #351:

> Currently using getParamCached on a namespace results in:
> a) The parameter not being updated correctly when one of its childs changes.
> b) A memory leak where the value of the child is blindly stored into the g_params map (and never removed).

### DIFF
--- a/clients/roscpp/src/libros/param.cpp
+++ b/clients/roscpp/src/libros/param.cpp
@@ -221,12 +221,7 @@ bool del(const std::string& key)
     if (sub_it != g_subscribed_params.end())
     {
       g_subscribed_params.erase(sub_it);
-
-      M_Param::iterator param_it = g_params.find(mapped_key);
-      if (param_it != g_params.end())
-      {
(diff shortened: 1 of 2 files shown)


po1:
I really appreciate that you addressed the issue, since this is a problem in `play_motion`, for example.
That being said, I do not feel comfortable with the API change to non-const. I would favor a solution that does not involve changing function signatures.

RainCT:
It's an internal function, not API, so it shouldn't really be a problem (otherwise a const-cast could be used).

po1:
I would feel more at peace with a `const_cast` with proper documentation as to why we need it.

dirk-thomas:
I would propose to close this in favor of #384. Would that be fine with you?

RainCT:
Yeah

dirk-thomas:
Thank you for providing the fixes for these bugs!

PR #355, opened 2014-02-11
remove rosgraph_msgs and std_srvs (moved to separate repo)

dirk-thomas:
The message packages `rosgraph_msgs` and `std_srvs` must be moved to a separate repo (https://github.com/ros/ros_comm_msgs) since currently there is a circular dependency which prevents releasing both repos into _Indigo_. `ros_comm` depends on `roslisp` which depends on those two packages inside `ros_comm`.

While this also applies to _Groovy_ and _Hydro_ I am not planning to split it for them since the repos are already released and therefore the circular dependency does not pose a problem anymore.


tfoote:
+1

wjwwood:
+1

PR #356, opened 2014-02-12
lz4 for rosbag

bcharrow:
I'd really like rosbag to have a fast compression option.  The use case is for larger bag files that take forever to process when you use bz2; while bz2 is great for long term storage, it can be unreasonably slow when you want to repeatedly process a bag.  lz4 (https://code.google.com/p/lz4/) seems like a pretty good option for this use case, so I made a prototype in roscpp.  The results are encouraging, though my implementation isn't complete so the changes shouldn't be merged just yet.

To measure performance, I used this [big bagfile](http://infinity.csail.mit.edu/data/2011/2011-01-18-06-37-58.bag).  Info about it is [here](http://infinity.csail.mit.edu/data/2011/utilities/2011-01-18-06-37-58/2011-01-18-06-37-58.bag.info).  See http://projects.csail.mit.edu/stata/downloads.php for details.

My machine: i7 sandy bridge, 16GB DDR3 RAM, intel SSD, compiled everything scratch using Release build type.

Create compressed bag file

| Type | Real (sec) | User (sec) | Size (%) |
| --- | --: | --: | --: |
| bz2 (python) | 2986.8 | 2954.2 | 17.16 |
| bz2 (C++) | 3050.0 | 3009.4 | 17.16 |
| lz4 (C++) | 51.2 | 21.6 | 25.83 |

Create decompressed bag file from compressed one

| Type | Real (sec) | User (sec) |
| --- | --: | --: |
| bz2 (python) | 333.1 | 303.0 |
| bz2 (C++) | 351.9 | 307.3 |
| lz4 (C++) | 88.8 | 7.5 |

Real / user are from time command.  bz2 (python) is current "rosbag compress" where as bz2 (c++) is the new executable I added.  These results are pretty consistent with what the lz4 page has.

A couple of notes:
1. As far as I can tell, lz4 isn't available as a system dependency, so I hackishly added the files to rosbag_storage.
2. I added C++ executables for compress/decompress so I wouldn't have to add things in python, but given the difference between c++/python for bz2, it looks like compress/decompress for lz4 should also be done in python.
3. I used the [lz4 streaming format](https://docs.google.com/document/d/1gZbUoLw5hRzJ5Q71oPRN6TO4cRMTZur60qip-TE7BhQ) to encode multiple chunks, although no checksums are currently present.
4. I forked form 1.9.54

If there's a chance this can eventually get merged, I'll finish the implementation, add support for python and update the rosbag command.

Thoughts?

### DIFF
--- a/tools/rosbag_storage/src/lz4_stream.cpp
+++ b/tools/rosbag_storage/src/lz4_stream.cpp
@@ -1,0 +1,204 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2014, Ben Charrow
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of Willow Garage, Inc. nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+*
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+********************************************************************/
+
+#include "rosbag/chunked_file.h"
+
+#include <iostream>
+#include <cstring>
+#include "console_bridge/console.h"
+
+using std::string;
+
+namespace rosbag {
+
+LZ4Stream::LZ4Stream(ChunkedFile* file)
+    : Stream(file), block_size_id_(6) {
+    buff_size_ = roslz4_blockSizeFromIndex(block_size_id_) + 64;
+    buff_ = new char[buff_size_];
+}
+
+LZ4Stream::~LZ4Stream() {
+    delete[] buff_;
+}
+
+CompressionType LZ4Stream::getCompressionType() const {
+    return compression::LZ4;
+}
+
+void LZ4Stream::startWrite() {
+    setCompressedIn(0);
+
+    int ret = roslz4_compressStart(&lz4s_, block_size_id_);
+    switch(ret) {
+    case ROSLZ4_OK: break;
+    case ROSLZ4_MEMORY_ERROR: throw BagIOException("ROSLZ4_MEMORY_ERROR: insufficient memory available"); break;
+    case ROSLZ4_PARAM_ERROR: throw BagIOException("ROSLZ4_PARAM_ERROR: bad block size"); break;
+    default: throw BagException("Unhandled return code");
+    }
+    lz4s_.output_next = buff_;
+    lz4s_.output_left = buff_size_;
+}
+
+void LZ4Stream::write(void* ptr, size_t size) {
+    lz4s_.input_left = size;
+    lz4s_.input_next = (char*) ptr;
+
+    writeStream(ROSLZ4_RUN);
+    setCompressedIn(getCompressedIn() + size);
+}
+
+void LZ4Stream::writeStream(int action) {
+    int ret = ROSLZ4_OK;
+    while (lz4s_.input_left > 0 ||
+           (action == ROSLZ4_FINISH && ret != ROSLZ4_STREAM_END)) {
+        ret = roslz4_compress(&lz4s_, action);
+        switch(ret) {
+        case ROSLZ4_OK: break;
+        case ROSLZ4_OUTPUT_SMALL:
+            if (lz4s_.output_next - buff_ == buff_size_) {
+                throw BagIOException("ROSLZ4_OUTPUT_SMALL: output buffer is too small");
+            } else {
+                // There's data to be written in buff_; this will free up space
+                break;
+            }
+        case ROSLZ4_STREAM_END: break;
+        case ROSLZ4_PARAM_ERROR: throw BagIOException("ROSLZ4_PARAM_ERROR: bad block size"); break;
+        case ROSLZ4_ERROR: throw BagIOException("ROSLZ4_ERROR: compression error"); break;
+        default: throw BagException("Unhandled return code");
+        }
+
+        // If output data is ready, write to disk
+        int to_write = lz4s_.output_next - buff_;
+        if (to_write > 0) {
+            if (fwrite(buff_, 1, to_write, getFilePointer()) != to_write) {
+                throw BagException("Problem writing data to disk");
+            }
+            advanceOffset(to_write);
+            lz4s_.output_next = buff_;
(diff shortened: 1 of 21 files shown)


forouher:
+1

Rosbagging topics like pointclouds was always a hassle for me. Uncompressed rosbag maxed out my non-SSD hard drive, whereas bz2 compression didn't go well with my CPU (which had other things to do ;)

So this looks like a promising middle ground.

po1:
What is the motivation for `lz4` in particular?
It is my understanding that it performs really fast, sacrificing compression ratio.

`bzip2` is notoriously slow for both compression _and_ decompression. Compression usually only happens once per bagfile, but decompression will take place whenever the bag file is opened, and this is usually when we need speed.

Are there any benchmarks with other algorithms, such as `zlib`? The latter has the advantage of being quite ubiquitous, and does not require duplicating its source in the ROS source tree.

Aside for the specific algorithm used, +1 for using anything else than `bzip2`

bcharrow:
I picked lz4 because it's substantially faster for both compression and decompression than the alternatives I found, zlib included.  https://code.google.com/p/lz4/ has the comparison you asked for; lz4's decompression speed is about 6 times faster than zlib.  I'm inclined to suggest that this difference in speed merits the duplication of lz4's source, but I can see both sides.

Although it's still duplication, would it be better to create a separate pure CMake project (a la console_bridge) for compressing / decompressing lz4 streams?  rosbag_storage would then depend on that.

Google's [snappy](https://code.google.com/p/snappy/) or [LZO](http://www.oberhumer.com/opensource/lzo/) are alternatives that are available and faster than zlib.  However, lz4 still seems to faster than both with no real difference in compression ratio.

dirk-thomas:
Is the (now available) Ubuntu package (http://packages.ubuntu.com/source/trusty/lz4 / http://packages.ubuntu.com/source/saucy/lz4) new enough to be used rather then the embedded version? If yes, can you please update the PR to use that package (including a rosdep rule for it)?

dirk-thomas:
Regarding the added executables `compress` and `decompress`: how are they different to the existing commands `rosbag compress` / `rosbag decompress`? Wouldn't it make sense to extend the existing commands with options to choose the compression type instead of adding new executables?

dirk-thomas:
Could you please also address the `todos` mentioned in the new code and implement the additional checks?

bcharrow:
It looks like the new ubuntu package is recent enough, I'll update the PR accordingly.  This doesn't address python bindings, though.  There's a [pip package](https://pypi.python.org/pypi/lz4), but it doesn't use the system version of lz4.  How this sort of thing is handled elsewhere?

I added the compress / decompress executables to see if they'd be faster than the compress / decompress options in python.  It looks like there's no performance improvement, so I agree it makes more sense to extend the existing rosbag commands to choose the compression type.

In my next update, I'll also address the TODOs, and add support for python.

dirk-thomas:
Great, thank you.

I am not sure if there is a Python implementation around - or a ready to use binding. But we should aim to keep the Python version of rosbag feature equal to the C++ version.

dirk-thomas:
@bcharrow It would be great if you could address the pending details so that it can be released into _Indigo_.

bcharrow:
@dirk-thomas I plan on getting everything finished shortly (i.e., a week or two).  Will this work for indigo?

dirk-thomas:
That sounds good. It can definitely go into _indigo-devel_. I just want to make sure it gets in early enough to have some soak time before the release. Thanks for working on it.

bcharrow:
I've now implemented C++/python bindings for the [lz4 streaming format](https://docs.google.com/a/seas.upenn.edu/document/d/1gZbUoLw5hRzJ5Q71oPRN6TO4cRMTZur60qip-TE7BhQ/edit), incorporated them into C++/python rosbag, and added options to the relevant command line tools.  To make the commit history a little cleaner, I squashed, and rebased off of indigo-devel.  This set of changes uses the system lz4 library for block compression.

The new thing in these commits is the roslz4 package which contains python/C++ APIs for lz4 streaming similar to those for bzip2.  This includes breaking up input into chunks, computing checksums to validate data, and a stream descriptor header.  I made the package because I figured having the headers/checksums was a good idea for long term data storage in bags, and as far as I can tell there isn't an open source library which implements the streaming format (though it is a requested feature https://code.google.com/p/lz4/issues/detail?id=42).  I had to add the source for the xxhash library because it's used to compute the checksums, and not available as a system dependence.  roslz4 does follow the lz4 stream spec, so if at some point a system library becomes available, the package could be removed.

Using roslz4, the changes to rosbag are much simpler; I don't think there's anything too complicated going on there.

Although I didn't commit unit tests, I have done some testing.  One example: https://gist.github.com/bcharrow/11016627 is a file compressor/decompressor implemented with roslz4.  I've successfully compressed/decompressed files produced by the lz4c tool (apt-get install libz4-tool).  This gives some confidence that the standard is being followed.

Two things I still need to do:
- add the read streaming functions for LZ4Stream (i.e., startRead(), read(), stopRead())
- Add a rosdep rulre for lz4; for now you can test with: apt-get install liblz4-dev

bcharrow:
That last couple of actions completes the TODOs.

One question: where are the streaming read functions actually called?  Grepping through ros_comm, it looks like Stream::startRead() is only called in ChunkedFile::setReadMode(), and that that's only called in ChunkedFile:seek() which always uses the uncompressed option.  Are startRead() for BZ2Stream and LZ4Stream actually unused or am I missing something?

bcharrow:
It looks like the build failed due to a git checkout error: "FATAL: Could not checkout null with start point dc400a09d95f4af3e25d961cd209e60ece1d0147"?  Not sure what that hash refers to.

bcharrow:
Never mind, that's the merge commit.  

@dirk-thomas Do you know how I can fix the build?

dirk-thomas:

tools/rosbag_storage/src/lz4_stream.cpp
```diff
@@ -0,0 +1,204 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2008, Willow Garage, Inc.
```
dirk-thomas: That should be your copyright?

utilities/roslz4/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
+
+install(TARGETS roslz4_py
+  LIBRARY DESTINATION ${CMAKE_INSTALL_PREFIX}/${PYTHON_INSTALL_DIR}/roslz4)
+
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION}
```
dirk-thomas, line 50: Without having tries the install this rule looks problematic. The trailing slash of `include/${PROJECT_NAME}/` only installs the content of that folder but the destination is the global include folder? I think it should either remove the slash to install the folder itself or change the destination to the package specific include destination.

utilities/roslz4/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+catkin_package(
+  INCLUDE_DIRS include
+  LIBRARIES roslz4)
+
+include_directories(include ${lz4_INCLUDE_DIRS} ${catkin_INCLUDE_DIRS})
+add_library(roslz4 src/lz4s.c src/xxhash.c)
+target_link_libraries(roslz4 ${lz4_LIBRARY} ${catkin_LIBRARIES})
+
+# Python bindings
+set(Python_ADDITIONAL_VERSIONS 2.7)
```
dirk-thomas: Do not use a hard coded Python version here.

Instead use the following two lines:

```
set(Python_ADDITIONAL_VERSIONS "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}")
set(PythonLibs_FIND_VERSION "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}")
```

utilities/roslz4/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(roslz4)
+
+find_package(catkin COMPONENTS)
+
+find_path(lz4_INCLUDE_DIR NAMES lz4.h)
+find_library(lz4_LIBRARY NAMES lz4)
```
dirk-thomas: If those two resources are in non standard locations you will also need to export those via catkin_package().

dirk-thomas:
Beside my tiny comments the PR looks great and I am looking forward to merge it. Thank you, Ben.

utilities/roslz4/CMakeLists.txt
bcharrow: The C module currently uses python 2.  Should I follow tf2_py's lead and use these lines?

```
find_package(PythonLibs 2 REQUIRED)
include_directories(${PYTHON_INCLUDE_PATH})
```

dirk-thomas: It should definitely also compile again Python 3.x. We made huge efforts in the last half year to get the code base to support both and it would be said to introduce new stuff which makes it not build again.

bcharrow:
I think that should do it.

Thanks for the helpful feedback, Dirk.

dirk-thomas:
Please rebase the PR to make it mergable. Also consider squashing the commits into one or a few commits.

bcharrow:
Rebased and condensed into 3 commits.

dirk-thomas:
Thank you very much for this significant contributions and the effort to iterate on it.

PR #359, opened 2014-02-15
roscpp: support for /clock remapping

xqms:
/clock was the only topic which cannot be remapped, which hurt flexibility
without reason.

Admittedly, my use case for remapping /clock is somewhat weird (controlling rviz time from outside while also running gazebo in parallel), but hey, more flexibility can't hurt, right?

Also applies to hydro-devel and indigo-devel without problems.


dirk-thomas:
Merged into _indigo-devel_ for now (0002b419cfbad574447a49bf141d4d6661f54023). Might be backported for future releases of _hydro_ and/or _groovy_.

PR #365, opened 2014-02-22
Add hostname lookup failure to xmlrpc. Fixes #364

### LINKED ISSUE #364: xmlrpc should print errors
When xmlrpc encounters an error trying to connect to a node, for example when trying to subscribe to a topic, it generates a reasonable error message and then calls XmlRpcUtil::error() ... which silently swallows it.

This should instead at least print an error message telling the user that a connection could not be established. I suspect this would help debug 90-95% of the network problems in ROS caused by incorrect ROS_IP/ROS_HOSTNAME settings.

I think we need to add an appropriate print statement here:
https://github.com/ros/ros_comm/blob/indigo-devel/utilities/xmlrpcpp/src/XmlRpcUtil.cpp#L48-L67

trainman419:
Now we get  nice error message if we cannot resolve the hostname of the hostname of the remote machine:

```
$ rostopic info /chatter
Type: std_msgs/String

Publishers: 
 * /talker (http://foo:58605/)

Subscribers: None

$ rosrun roscpp_tutorials listener
Couldn't find an AF_INET address for [foo]
```

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcSocket.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcSocket.cpp
@@ -209,6 +209,7 @@ XmlRpcSocket::connect(int fd, std::string& host, int port)
   hints.ai_family = AF_UNSPEC;
   if (getaddrinfo(host.c_str(), NULL, &hints, &addr) != 0)
   {
+    fprintf(stderr, "Couldn't find an %s address for [%s]\n", s_use_ipv6_ ? "AF_INET6" : "AF_INET", host.c_str());
     return false;
   }
 


utilities/xmlrpcpp/src/XmlRpcSocket.cpp
```diff
@@ -209,6 +209,7 @@ static void initWinSock()
   hints.ai_family = AF_UNSPEC;
   if (getaddrinfo(host.c_str(), NULL, &hints, &addr) != 0)
   {
+    printf("Couldn't find an %s address for [%s]\n", s_use_ipv6_ ? "AF_INET6" : "AF_INET", host.c_str());
```
dirk-thomas: Please update the code so that the error message goes to _stderr_.

trainman419:
Fixed error message to go to stderr.

dirk-thomas:
Thx.

trainman419:
This is probably small enough and useful enough to go into Groovy and Hydro as well. Should I submit those as separate PRs?

dirk-thomas:
No, I will cherry-pick them manually when I consider new releases into _groovy_ and _hydro_.

PR #366, opened 2014-02-27
make rostest in CMakeLists optional (ros/rosdistro#3010)

bulwahn:
I was unsure if I should add this commit to hydro-devel and/or indigo-devel. So for the moment, I just added this patch to the hydro-devel branch, because the latest release 1.10.1 was on this branch.


dirk-thomas:
Merged into _indigo-devel_: a7dd5d5db18cd9a30f7c7e257c1259c40104a83d

tools/topic_tools/CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ install(PROGRAMS
 
 #Testing
 if(CATKIN_ENABLE_TESTING)
+  find_package(rostest rosunit)
```
dirk-thomas, line 65: This does not work. It must use two `find_package()` calls. Fixed in e3eea61e51a49a6edbab79c4ee0b8c02b0ab5199

Issue #373, opened 2014-03-06
Using queue_size in a rospy publisher causes slowdown when there are multiple subscribers


jonbinney:
When a rospy publisher uses a queue_size, if there are multiple subscribers (more than 2), I see lots of dropped messages. This happens even when the messages are small, and the CPU load is nowhere near maxed. It happens even if the queue_size is very large.  To reproduce, use this `pub` script:

``` python
#!/usr/bin/env python
import rospy
import sys
from std_msgs.msg import String

pub_rate = int(sys.argv[1])
queue_size = None
if len(sys.argv) > 2:
    queue_size = int(sys.argv[2])

rospy.init_node('pub', anonymous=True)
pub = rospy.Publisher('foo', String, queue_size = queue_size)

r = rospy.Rate(pub_rate)
while not rospy.is_shutdown():
    pub.publish(String('helloooooooo'))
    r.sleep()
```

In one terminal, do `./pub 150 1000` to publish messages at 150 Hz on topic "foo" with a queue_size of 1000. Then in open 3 or 4 other terminals and run `rostopic hz /foo -w200`. For each rostopic hz you add, the rate drops. On my computer, with 3 of them, I get 100-130 hz instead of 150hz.  Checking top shows only 12% CPU usage for the publisher node.

If you run the same experiment, but start the publisher with `./pub 150`, it will create the publisher without queue_size, and you should see all subscribers getting 150 hz.

Since the publisher doesn't seem to be CPU limited or network limited, I'm not sure what is going on. Maybe lock contention somewhere?

jonbinney:
(i'm running this all on one machine, with everything talking to localhost)

dirk-thomas:
This looks like bad locking / signaling strategy. Please try PR #374 which should address this issue.

@mikeferguson Since you were the user trying this feature can you please also try the referenced PR and confirm that it works for your use case (the same or likely even better then before)?

mikeferguson:
@dirk-thomas I can test this, but I'm not sure if I will get to it today. Worst case, I will test over the weekend.

Issue #369, opened 2014-03-03
Condition variable doesn't exist


jonbinney:
Inside of rospy's QueuedConnection, there is an exception handler which tries to grab a condition variable called "self._cond": https://github.com/ros/ros_comm/blob/indigo-devel/clients/rospy/src/rospy/impl/tcpros_pubsub.py#L431

But that class has no member variable called self._cond, it has only self._cond_data_available and self._cond_queue_swapped. As a result, I see the following error output:

```
Exception in thread Thread-23:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 504, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/opt/ros/hydro/lib/python2.7/dist-packages/rospy/impl/tcpros_pubsub.py", line 431, in _run
    with self._cond:
  File "/opt/ros/hydro/lib/python2.7/dist-packages/rospy/impl/tcpros_pubsub.py", line 390, in __getattr__
    return getattr(self._connection, name)
AttributeError: 'TCPROSTransport' object has no attribute '_cond'
```

jonbinney:
@dirk-thomas this looks like it was added along with queue_size to handle disconnects

dirk-thomas:
That looks definitely like a bug. I guess the line should be:

```
with self._lock:
```

Can you please describe what you did in order to reproduce the problem? And try of changing the referenced line with my guess fixed the problem for you?

jonbinney:
I added queue_size to a rospy publisher. It may have had a client which disconnected which then triggered this. I'll try and create a small test case

jonbinney:
Here's a script that will reproduce it:

```
import rospy
from std_msgs.msg import String

rospy.init_node('pub_disconnect')
pub = rospy.Publisher('foo', String, queue_size=100)

r = rospy.Rate(100)
while not rospy.is_shutdown():
    pub.publish(String('foooooo'))
    r.sleep()
```

Start that script, then start `rostopic echo foo` in a separate window. Then kill rostopic with a sigterm (ctrl-\) and you should see the publisher throw an exception.

jonbinney:
Thanks @dirk-thomas !

jonbinney:
Can you backport this fix to hydro-devel?

jonbinney:
I'll make a pull-request to hydro-devel to backport it

[...]

PR #374, opened 2014-03-06
improve asynchonous publishing performance

dirk-thomas:

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_pubsub.py
+++ b/clients/rospy/src/rospy/impl/tcpros_pubsub.py
@@ -376,9 +376,7 @@ class QueuedConnection(object):
 
         self._lock = threading.Lock()
         self._cond_data_available = threading.Condition(self._lock)
-        self._cond_queue_swapped = threading.Condition(self._lock)
         self._queue = []
-        self._waiting = False
         self._error = None
 
         self._thread = threading.Thread(target=self._run)
@@ -401,11 +399,8 @@ class QueuedConnection(object):
                 del self._queue[0]
             self._queue.append(data)
             self._cond_data_available.notify()
-            # ensure that thread has actually swapped the queues and is processig them
-            # if it was waiting for being notified
-            # to enforce behavior to be as close as possible to blocking
-            if self._waiting:
-                self._cond_queue_swapped.wait()
+        # effectively yields the rest of the thread quantum
+        time.sleep(0)
         return True
 
     def _run(self):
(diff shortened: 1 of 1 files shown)


wjwwood:
Reading briefly about using `time.sleep(0)` as a yield reveals that it may not work that way on OS X or Windows. What are you trying to accomplish with that? It seems like there should be a better way to do that.

wjwwood:
Maybe not, that seems to be the recommended way to do it.

wjwwood:
At least this recommends calling a `delayfunc` with `0`:

http://docs.python.org/2/library/sched.html

And later mentions that `delayfunc` is usually `time.sleep`.

dirk-thomas:
Please try the above example script with e.g. three listening `rostopic`s on OS X in order to figure out how it behaves on that system.

I was able to get to 1kHz without a problem - and I think that is a restriction of the above script. If you e.g. invoke `publish` five times per loop I could easily go beyond that frequency.

wjwwood:
When disconnecting a `rostopic hz` I get this traceback:

```
Exception in thread Thread-14:
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 808, in __bootstrap_inner
    self.run()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 761, in run
  File "/Users/william/hydro/install/lib/python2.7/site-packages/rospy/impl/tcpros_pubsub.py", line 422, in _run
    with self._cond:
  File "/Users/william/hydro/install/lib/python2.7/site-packages/rospy/impl/tcpros_pubsub.py", line 388, in __getattr__
    return getattr(self._connection, name)
AttributeError: 'TCPROSTransport' object has no attribute '_cond'
```

From what I can see, this should never have worked? Maybe it was always broken, but this change exposed it?

wjwwood:
Other than the traceback from above, this patch does indeed resolve the issue described in #373.

dirk-thomas:
You are running old code: #369.

jonbinney:
I tested this on my system, and it does fix #373 for me.

jonbinney:
Assuming this patch gets merged, will it get backported to hydro-devel?

dirk-thomas:
Yes, it will definitely be backported due to its significant impact and the fact that nearly no code is actually using the queuing yet and therefore the chance of a regression a very slim.

dirk-thomas:
Cherry-picked to _hydro-devel_: f577021a701163951e3ee42e0136535f0659060e

jonbinney:
Thanks @dirk-thomas !

PR #375, opened 2014-03-10
Custom error handler

esteve:
This branch allows users to define a custom behavior for handling exceptions raised from within a service.

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_service.py
+++ b/clients/rospy/src/rospy/impl/tcpros_service.py
@@ -540,7 +540,7 @@ class ServiceImpl(_Service):
     Implementation of ROS Service. This intermediary class allows for more configuration of behavior than the Service class.
     """
     
-    def __init__(self, name, service_class, handler, buff_size=DEFAULT_BUFF_SIZE):
+    def __init__(self, name, service_class, handler, buff_size=DEFAULT_BUFF_SIZE, error_handler=None):
         super(ServiceImpl, self).__init__(name, service_class)
 
         if not name or not isstring(name):
@@ -552,6 +552,8 @@ class ServiceImpl(_Service):
 
         
         self.handler = handler
+        if error_handler is not None:
+            self.error_handler = error_handler
         self.registered = False
         self.seq = 0
         self.done = False
@@ -567,7 +569,10 @@ class ServiceImpl(_Service):
         logdebug("[%s]: new Service instance"%self.resolved_name)
 
     # TODO: should consider renaming to unregister
-    
+
+    def error_handler(self, e, exc_type, exc_value, tb):
+        logerr("Error processing request: %s\n%s" % (e, traceback.format_exception(exc_type, exc_value, tb))
+
     def shutdown(self, reason=''):
         """
         Stop this service
@@ -624,9 +629,10 @@ class ServiceImpl(_Service):
             rospy.core.rospydebug("handler raised ServiceException: %s"%(e))
             self._write_service_error(transport, "service cannot process request: %s"%e)
         except Exception as e:
-            logerr("Error processing request: %s\n%s"%(e,traceback.print_exc()))
+            (exc_type, exc_value, tb) = sys.exc_info()
+            self.error_handler(e, exc_type, exc_value, tb)
             self._write_service_error(transport, "error processing request: %s"%e)
-    
+
     def handle(self, transport, header):
         """
         Process incoming request. This method should be run in its
@@ -671,7 +677,8 @@ class Service(ServiceImpl):
       s = Service('getmapservice', GetMap, get_map_handler)
     """
     
-    def __init__(self, name, service_class, handler, buff_size=DEFAULT_BUFF_SIZE):
+    def __init__(self, name, service_class, handler,
+                 buff_size=DEFAULT_BUFF_SIZE, error_handler=None):
         """
         ctor.
 
@@ -690,10 +697,16 @@ class Service(ServiceImpl):
         is always considered a failure.
         
         @type  handler: fn(req)->resp
+
         @param buff_size: size of buffer for reading incoming requests. Should be at least size of request message
         @type  buff_size: int
+
+        @param error_handler: callback function for handling errors
+        raised in the service code.
+        @type  error_handler: fn(exception, exception_type, exception_value, traceback)->None
         """
-        super(Service, self).__init__(name, service_class, handler, buff_size)
+        super(Service, self).__init__(name, service_class, handler, buff_size,
+                                      error_handler)
 
         #TODO: make service manager configurable
         get_service_manager().register(self.resolved_name, self)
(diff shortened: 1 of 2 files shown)


wjwwood:
This looks ok to me. The default behavior shouldn't change and the other part has been tested against capabilities.

wjwwood:
+1

clients/rospy/src/rospy/impl/tcpros_service.py
```diff
@@ -540,7 +540,7 @@ class ServiceImpl(_Service):
     Implementation of ROS Service. This intermediary class allows for more configuration of behavior than the Service class.
     """
     
-    def __init__(self, name, service_class, handler, buff_size=DEFAULT_BUFF_SIZE):
+    def __init__(self, name, service_class, handler, buff_size=DEFAULT_BUFF_SIZE, error_handler=None):
```
dirk-thomas, line 543: Since the new argument is anything but trivial it should get a description in the doc block. Or at least mention which other constructor it relates to (since it is documented there).

clients/rospy/src/rospy/impl/tcpros_service.py
```diff
@@ -671,7 +677,8 @@ class Service(ServiceImpl):
       s = Service('getmapservice', GetMap, get_map_handler)
     """
     
-    def __init__(self, name, service_class, handler, buff_size=DEFAULT_BUFF_SIZE):
+    def __init__(self, name, service_class, handler,
+        buff_size=DEFAULT_BUFF_SIZE, error_handler=None):
```
dirk-thomas: This kind of indentation does not follow the PEP guideline.
Either keep it in one line or indent it with eight space to distinguish it from the indented block.

clients/rospy/src/rospy/impl/tcpros_service.py
wjwwood, line 543: He has some doc stuff in the `Service` class, I think this is slim on docs because it is the `Impl` class.

clients/rospy/src/rospy/impl/tcpros_service.py
wjwwood: Yup, according to PEP8 it should either be:

``` python
def __init__(self, name, service_class, handler,
             buff_size=DEFAULT_BUFF_SIZE, error_handler=None):
```

or 

``` python
def __init__(
    self,
    name,
    service_class, handler,
    buff_size=DEFAULT_BUFF_SIZE,
    error_handler=None,
):
```

test/test_rospy/test/rostest/test_basic_services.py
```diff
@@ -115,7 +132,7 @@ def services():
 
     s5 = rospy.Service(STRING_CAT_SERVICE_NAKED, StringString, string_cat_naked)
     s6 = rospy.Service(STRING_CAT_SERVICE_WRAPPED, StringString, string_cat_wrapped)
-    
+
```
dirk-thomas: Please avoid any non-related white space changes in pull requests in the future. You can keep this for now since it only is in a test file and will like not affect backporting patches in the future.

clients/rospy/src/rospy/impl/tcpros_service.py
```diff
@@ -690,10 +697,16 @@ def __init__(self, name, service_class, handler, buff_size=DEFAULT_BUFF_SIZE):
...
         @param buff_size: size of buffer for reading incoming requests. Should be at least size of request message
         @type  buff_size: int
+
+        @param error_handler: callback function for handling errors
+        raised in the service code.
+        @type  error_handler: fn(exception, exception_type, exception_value, traceback)->None
         """
-        super(Service, self).__init__(name, service_class, handler, buff_size)
+        super(Service, self).__init__(name, service_class, handler, buff_size,
+            error_handler)
```
wjwwood: Same indention problem here.

test/test_rospy/test/rostest/test_basic_services.py
```diff
@@ -222,7 +239,16 @@ def test_constants(self):
             self.assertEquals(Req.INT32_Z, resp.ret_int32)
             self.assertEquals(Req.UINT32_Z, resp.ret_uint32)
             self.assert_(math.fabs(Req.FLOAT32_Z - resp.ret_float32) < 0.001)
-        
+
+    def test_faulty_service(self):
+        s = rospy.Service(FAULTY_SERVICE, FaultyService, faulty_handler.call_handler,
+            faulty_handler.custom_error_handler)
```
wjwwood: And here.

clients/rospy/src/rospy/impl/tcpros_service.py
dirk-thomas: Another option I have seen was to indent the wrapped line with eight spaces. Imo that it better then these both options.

dirk-thomas: The following example passes at least all the linters I have in my editor (pep8, pep257, frosted):

```
def func(
        'foo', 'bar',
        keyword1=None,
        keyword2=False, keyword3=None,):
    pass
```

wjwwood:
``` python
def func(
        'foo', 'bar',  # <-- This is a syntax error for me...
        keyword1=None,
        keyword2=False, keyword3=None,):
    pass
```

wjwwood:
``` python
def func(
        foo, bar,
        keyword1=None,
        keyword2=False, keyword3=None,):  # <-- This also gives an error
    pass
```

The error is `continuation line does not distinguish itself from next logical line`, but this is fine:

``` python
def func(
        foo, bar,
        keyword1=None,
        keyword2=False, keyword3=None,
):
    pass
```

wjwwood: I am using the `flake8` linter as well and I believe this is where these errors are coming from, but the eight spaces indention is weird to me because I wouldn't do that for a dict or list:

``` python
a = {
    'foo': 1,
    'bar': 2,
}
a = dict(
    foo=1,
    bar=2,
}
```

dirk-thomas: But a dict or list is different since it does not open a new block. When using only four spaces for indentation I actually get the warning which says something like: "indentation level is not distinguishable from block".

test/test_rospy/test/rostest/test_basic_services.py
dirk-thomas: Since the change is actually touching this line I would be okay with cleaning up the trailing spaces.

dirk-thomas:
Please address the one remaining comment and finally squash your commits.

esteve:
@dirk-thomas done!

dirk-thomas:
Cherry-picked to _hydro-devel_.

Issue #376, opened 2014-03-11
rostopic.get_topic_type returns wrong type


jihoonl:
The following issue describes it.

https://github.com/RobotWebTools/rosbridge_suite/issues/85#issuecomment-37260395

dirk-thomas:
I tried the following sequence (all in separate shells) based on the referenced ticket:

```
rostopic pub /ns/string std_msgs/String "str"
rostopic pub /ns/string/int std_msgs/Int16 5
rostopic list
rostopic echo /ns/string
rostopic echo /ns/string/int
rostopic type /ns/string
rostopic type /ns/string/int
```

All of the commands return the correct values (the data and types of that specific publisher).

Please clarify what exact sequence of commands (preferably just `rostopic`) you are running, what results you get and what you expect.

jihoonl:
I was testing by the following

```
rostopic type /should_be_string   # expected to be unknown type
rostopic pub /should_be_string std_msgs/String "str"
rostopic type /should_be_string  # expected to be std_msgs/String

rostopic type /should_be_string/should_be_int # expected to be unknown type but returned std_msgs/String
```

My expected output from the last code was unknown topic since it is unpublished yet. To be specific, I expected `rostopic.get_topic_type` to return (None, None, None). However, it returns the type of its namespace instead. 

So I was wondering if it is expected behavior. And if it is desired behavior, I wondered if you have any suggestion to detect the presence of `topic` as well as its type.

`rostopic.get_topic_type` is used to detect the topic presence and its type in rosbridge in the following code. What the code does is that it checks the presence of topic and sets up a publisher when it received `publish()` request from rosbridge client.

https://github.com/RobotWebTools/rosbridge_suite/blob/hydro-devel/rosbridge_library/src/rosbridge_library/internal/publishers.py#L125-L160

So far, we found the workaround solution suggested by @garaemon in https://github.com/RobotWebTools/rosbridge_suite/issues/85#issuecomment-37260827, which is checking whether the third return value of `rostopic.get_topic_type` is presented or not.

I would like to know if there is more efficient way to achieve this.

dirk-thomas:
Yes, that is currently the expected behavior which is also documented like this in the API: https://github.com/ros/ros_comm/blob/indigo-devel/tools/rostopic/src/rostopic/__init__.py#L320

The problem is that `rostopic` uses the same separator (`/`) withing the topic name as for separating the field inside the topic. E.g. the following works with rostopic:

```
[...]

PR #378, opened 2014-03-12
add check for message fields when determining message type

dirk-thomas:
This addresses #376 - or at least the feasible part of it.

@esteve @tfoote @wjwwood Please provide feedback.

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -307,6 +307,27 @@ def _get_topic_type(topic):
         matches = [(t, t_type) for t, t_type in val if topic.startswith(t+'/')]
         # choose longest match
         matches.sort(key=itemgetter(0), reverse=True)
+
+        # try to ignore messages which don't have the field specified as part of the topic name
+        while matches:
+            t, t_type = matches[0]
+            msg_class = roslib.message.get_message_class(t_type)
+            if not msg_class:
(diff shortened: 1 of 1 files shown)


tools/rostopic/src/rostopic/__init__.py
```diff
@@ -307,6 +307,27 @@ def _get_topic_type(topic):
...
+            msg_class = roslib.message.get_message_class(t_type)
+            if not msg_class:
+                # if any class is not fetchable skip ignoring any message types
+                break
+            msg = msg_class()
+            pattern = topic[len(t):].rstrip('/')
+            if pattern == '':
+                break
+            try:
+                eval('msg'+'.'.join(pattern.split('/')))
```
tfoote: Is this sanitized sufficiently at this point to be safe?

dirk-thomas: The same logic is already used in line 287 but I do agree some kind of sanity checking would be nice.

wjwwood:
+1 code review

jihoonl:
This addresses the issue https://github.com/RobotWebTools/rosbridge_suite/issues/85

esteve:
+1

dirk-thomas:
In my recent commit I replaced both uses of `eval()` with a loop of `getattr()` calls. That is indeed safer - I was able to craft some really interesting `rostopic ...` command lines :wink:

tfoote:
+1

Issue #294, opened 2013-10-09
roslaunch: Add arg description


liborw:
It would be nice to have a simple description attached to the launchfile args so that the `roslaunch --ros-args` would print something more descriptive.

tlind:
+1

jbohren:
Possible implementation plan for arguments in a top-level launchfile:

---

Modify [`XmlLoader::_arg_tag`](https://github.com/ros/ros_comm/blob/indigo-devel/tools/roslaunch/src/roslaunch/xmlloader.py#L257) to add an optional "description" attribute.

Modify [`ContextLoader::add_arg`](https://github.com/ros/ros_comm/blob/indigo-devel/tools/roslaunch/src/roslaunch/loader.py#L212) to accept an optional "description" argument, and to store that description in the context. 

Update [`arg_dump.py`](https://github.com/ros/ros_comm/blob/indigo-devel/tools/roslaunch/src/roslaunch/arg_dump.py#L60) to output each argument's default value and documentation string.

---

dirk-thomas:
That sounds like a good additional. Could someone please put together a pull request with this?

jbohren:
Writing a PR now.

jbohren:
PR submitted: https://github.com/ros/ros_comm/pull/379

PR #358, opened 2014-02-15
Python 3 compatibility

dirk-thomas:

PR #379, opened 2014-03-14
Adding 'doc' attribute to roslaunch <arg> tags

jbohren:
This is implemented as described in https://github.com/ros/ros_comm/issues/294#issuecomment-37079708

Additionally, it brings `arg_dump.py` into python3 compliance with the `print()` function.

This should also be a viable modification for release into hydro.

### DIFF
--- a/tools/roslaunch/src/roslaunch/arg_dump.py
+++ b/tools/roslaunch/src/roslaunch/arg_dump.py
@@ -45,7 +45,7 @@ from roslaunch.config import load_config_default
 def get_args(roslaunch_files):
     loader = roslaunch.xmlloader.XmlLoader(resolve_anon=False)
     config = load_config_default(roslaunch_files, None, loader=loader, verbose=False, assign_machines=False)
-    return loader.root_context.resolve_dict['arg']
+    return loader.root_context.resolve_dict.get('arg_doc', {})
 
 def dump_args(roslaunch_files):
     """
@@ -55,9 +55,26 @@ def dump_args(roslaunch_files):
     @param roslaunch_files: list of launch files to load
     @type  roslaunch_files: str
     """
+
     try:
-        for arg in sorted(get_args(roslaunch_files).items()):
-        	print arg[0]
+        args = get_args(roslaunch_files)
+
+        if len(args) == 0:
+            print("No arguments.")
+        else:
+            required_args = [(name, (doc or 'undocumented', default)) for (name, (doc, default)) in args.items() if not default]
+            optional_args = [(name, (doc or 'undocumented', default)) for (name, (doc, default)) in args.items() if default]
+
+            if len(required_args) > 0:
+                print("Required Arguments:")
+                for (name, (doc, _)) in sorted(required_args):
+                    print("  %s: %s" % (name, doc))
+
+            if len(optional_args) > 0:
+                print("Optional Arguments:")
+                for (name, (doc, default)) in sorted(optional_args):
+                    print("  %s (default \"%s\"): %s" % (name, default, doc))
+
     except RLException as e:
         print >> sys.stderr, str(e)
         sys.exit(1)
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -209,7 +209,7 @@ class LoaderContext(object):
             self._remap_args.remove(m)
         self._remap_args.append(remap)
         
-    def add_arg(self, name, default=None, value=None):
+    def add_arg(self, name, default=None, value=None, doc=None):
         """
         Add 'arg' to existing context. Args are only valid for their immediate context.
         """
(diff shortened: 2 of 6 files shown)


jbohren:
The output for the following roslaunch file is shown below:

``` xml
<launch>
  <arg name="foo" default="true" doc="I pity the foo'."/>
  <arg name="bar" doc="Someone walks into this."/>
  <arg name="baz" default="false"/>
  <arg name="nop"/>
</launch>
```

```
[moldy-crow:...mm/test/test_roslaunch/test]$ roslaunch --ros-args ros_args.launch
  bar: (default: None) Someone walks into this.
  baz: (default: "false") 
  foo: (default: "true") I pity the foo'.
  nop: (default: None) 
```

jbohren:
Actually, I just had another idea on the output, so hold off on this PR.

jbohren:
With the revised commit, it outputs the following:

```
[moldy-crow:...mm/test/test_roslaunch/test]$ roslaunch --ros-args ros_args.launch
Required Arguments:
  bar: Someone walks into this.
  nop: None
Optional Arguments:
  baz (default "false"): None
  foo (default "true"): I pity the foo'.
```

dirk-thomas:
Please remove all Python related changes from the PR. There is a separate PR pending for Python 3 compatibility (#358).

jbohren:
Will do.

jbohren:
Updated to remove python3 future stuff.

tools/roslaunch/src/roslaunch/arg_dump.py
```diff
@@ -45,7 +45,8 @@
 def get_args(roslaunch_files):
     loader = roslaunch.xmlloader.XmlLoader(resolve_anon=False)
     config = load_config_default(roslaunch_files, None, loader=loader, verbose=False, assign_machines=False)
-    return loader.root_context.resolve_dict['arg']
+    resolve_dict = loader.root_context.resolve_dict
+    return resolve_dict['arg_doc'] if 'arg_doc' in resolve_dict else dict()
```
dirk-thomas: Please use the following instead:

```
return getattr(loader.root_context.resolve_dict, 'arg_doc', {})
```

tools/roslaunch/src/roslaunch/arg_dump.py
```diff
@@ -55,9 +56,26 @@ def dump_args(roslaunch_files):
...
     try:
-        for arg in sorted(get_args(roslaunch_files).items()):
-        	print arg[0]
+        args = get_args(roslaunch_files)
+
+        if len(args) == 0:
+            print("No arguments.")
+        else:
+            required_args = [(name,(doc,default)) for (name,(doc,default)) in args.items() if not default]
+            optional_args = [(name,(doc,default)) for (name,(doc,default)) in args.items() if default]
```
dirk-thomas: Please add spaces after the `,` following PEP8.

Also multiple times later...

tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -239,6 +239,16 @@ def add_arg(self, name, default=None, value=None):
             # no value or default: appending to arg_names is all we
             # need to do as it declares the existence of the arg.
             pass
+
+        # add arg documentation string dict if it doesn't exist yet and if it can be used
+        if not 'arg_doc' in resolve_dict:
+            resolve_dict['arg_doc'] = {}
+        arg_doc_dict = resolve_dict['arg_doc']
+
+        if not value:
```
dirk-thomas, line 248: Without have read the full PR - why is it checking for `not value` instead of for `doc`?

jbohren, line 248: Because if an `<arg>` tag has a `value` attribute, then it can't be overriden from the command line, so it's not really part of the interface to the launchfile.

tools/roslaunch/src/roslaunch/arg_dump.py
jbohren: Replaced with `return loader.root_context.resolve_dict.get('arg_doc', {})`

tools/roslaunch/src/roslaunch/arg_dump.py
jbohren: Done in 204594f

jbohren:
The current PR has the following behavior:

``` xml
<launch>
  <!-- ros_args.launch -->
  <arg name="foo" default="true" doc="I pity the foo'."/>
  <arg name="bar" doc="Someone walks into this."/>
  <arg name="baz" default="false"/>
  <arg name="nop"/>
  <arg name="fix" value="true"/>
</launch>
```

```
$> roslaunch --ros-args ros_args.launch
Required Arguments:
  bar: Someone walks into this.
  nop: undocumented
Optional Arguments:
  baz (default "false"): undocumented
  foo (default "true"): I pity the foo'.
```

tools/roslaunch/src/roslaunch/loader.py
dirk-thomas, line 248: Oh, I see, thanks for clarifying.

dirk-thomas:
Could you also create a simple test which invoked the command line / API on the example launch file you already added? That would ensure that the functionality will keep working in the future.

jbohren:
Unit test added.

dirk-thomas:
Thank you for that great PR. The CI job passes and the output of `--ros-args` is now much more helpful.

One minor thing before the merge, can please also update the test script to use spaces after commas as well as after colons in a dict (to follow PEP8).

jbohren:
> One minor thing before the merge, can please also update the test script to use spaces after commas as well as after colons in a dict (to follow PEP8).

Done.

130s:
[Wiki page is updated for this change](http://wiki.ros.org/action/info/roslaunch/XML/arg?action=diff&rev2=11&rev1=10). Please add any correction if necessary.

Issue #267, opened 2013-08-03
Change from Boost::signal to Boost::signals2::signal


tfoote:
In newer versions of Boost Boost.Signal has been deprecated.  

William identified it here: https://github.com/ros/geometry/issues/23

But in changing that I found that the underlying message_filter class exposes boost::signal in it's API.

dirk-thomas:
Since this is an API breaking change it is not happening for Groovy/Hydro but earliest for Indigo.

trainman419:
+1 to putting this into Indigo before boost removes it completely in a future version. This could affect compatibility with future versions of Ubuntu

Also +1 to any solution that makes the warning messages go away.

dirk-thomas:
Done in #381.

PR #381, opened 2014-03-28
update API to use boost::signals2 (#267)

dirk-thomas:


dirk-thomas:
@ros-pull-request-builder Please run CI on this pull request...

dirk-thomas:
This is breaking API between _Hydro_ and _indigo_ (and I can't imagine a way to avoid it).

@tfoote @wjwwood Please review and vote on API breaking change for _Indigo_.

tfoote:
+1 Boost is changing we'll need to do it anyway. If using the typedef it should keep working.

utilities/message_filters/include/message_filters/sync_policies/exact_time.h
```diff
@@ -45,15 +45,7 @@
...
-#ifndef BOOST_SIGNALS_NO_DEPRECATION_WARNING
-#define BOOST_SIGNALS_NO_DEPRECATION_WARNING
-#define ros_BOOST_SIGNALS_NO_DEPRECATION_WARNING
-#endif
-#include <boost/signals.hpp>
-#ifdef ros_BOOST_SIGNALS_NO_DEPRECATION_WARNING
-#undef ros_BOOST_SIGNALS_NO_DEPRECATION_WARNING
-#undef BOOST_SIGNALS_NO_DEPRECATION_WARNING
-#endif
+#include <boost/signals2.hpp>
```
wjwwood: Does this header (and similar headers here) even need this include?

wjwwood:
+1 from me, I think this change needs to go in, we just need to update the Migration guide. My guess is that very few people actually use this API directly.

dirk-thomas:
I removed the unnecessary includes from most of the message_filters headers.

wjwwood:
+1

dirk-thomas:
@ros-pull-request-builder

PR #382, opened 2014-03-31
Add substitution nesting

RainCT:
Rationale:
- Needed for supporting eg. comparison of params (see last comment in https://github.com/ros/ros_comm/pull/341#issuecomment-37585632)
- It enables stuff like eg. &lt;include file="$(find $(arg robot)_description)/robots/upload_$(arg robot).launch" /&gt;.
- Why shouldn't we have this? :)

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -69,7 +69,7 @@ def _env(cmd, args, context):
     @raise SubstitutionException: if arg invalidly specified
     """
     if len(args) != 1:
-        raise SubstitutionException("$(env var) command only accepts one argument [%s]"%a)
+        raise SubstitutionException("$(env var) command only accepts one argument [%s]"%cmd)
     try:
         return os.environ[args[0]]
     except KeyError as e:
@@ -124,9 +124,9 @@ def _find(resolved, a, args, context):
 
     :param resolved: full string
     :type resolved: str
-    :param a: subtitution string
+    :param a: substitution string
     :type a: str
-    :param args: subtitution arguments
+    :param args: substitution arguments
     :type a: dict
     :param context: see resolve_args()
     :type context: dict
@@ -432,7 +432,7 @@ def _parse_args(arg_str):
             i = j = j + 2
         elif arg_str[j] == ')' and not t.is_root():
             t._add_child(arg_str[i:j])
-            # End of the current substituion
+            # End of the current substitution
             t = t.parent
             i = j = j + 1
         else:
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -240,7 +240,7 @@ def _arg(resolved, a, args, context):
     :raises: :exc:`ArgException` If arg invalidly specified
     """
     if len(args) == 0:
-        raise SubstitutionException("$(arg var) must specify an environment variable [%s]"%(a))
+        raise SubstitutionException("$(arg var) must specify an <arg> tag name [%s]"%(a))
     elif len(args) > 1:
         raise SubstitutionException("$(arg var) may only specify one arg [%s]"%(a))
     
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -38,6 +38,7 @@ by roslaunch and xacro, but it is not yet a top-level ROS feature.
 """
 
 import os
+import re
 
 try:
     from cStringIO import StringIO # Python 2.x
@@ -60,7 +61,7 @@ class ArgException(SubstitutionException):
     """
     pass
 
-def _env(resolved, a, args, context):
+def _env(cmd, args, context):
     """
     process $(env) arg
     @return: updated resolved argument
@@ -70,11 +71,11 @@ def _env(resolved, a, args, context):
     if len(args) != 1:
         raise SubstitutionException("$(env var) command only accepts one argument [%s]"%a)
     try:
-        return resolved.replace("$(%s)"%a, os.environ[args[0]])
+        return os.environ[args[0]]
     except KeyError as e:
         raise SubstitutionException("environment variable %s is not set"%str(e))
 
-def _optenv(resolved, a, args, context):
+def _optenv(cmd, args, context):
     """
     process $(optenv) arg
     @return: updated resolved argument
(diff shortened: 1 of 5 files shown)


dirk-thomas:
@ros-pull-request-builder

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -120,14 +121,37 @@ def _find(resolved, a, args, context):
...
+    :param context: see resolve_args()
+    :type context: dict
+
     :returns: updated resolved argument, ``str``
     :raises: :exc:SubstitutionException: if PKG invalidly specified
     :raises: :exc:`rospkg.ResourceNotFound` If PKG requires resource (e.g. package) that does not exist
     """
-    if len(args) != 1:
-        raise SubstitutionException("$(find pkg) command only accepts one argument [%s]" % a)
+    assert len(args) == 1
```
dirk-thomas, line 138: Why did you replace the exceptions with assertions (here and several more times)?

tools/roslaunch/test/unit/test_roslaunch_depends.py
```diff
@@ -63,7 +63,7 @@ def fakestdout():
 def test_roslaunch_deps_main():
     from roslaunch.depends import roslaunch_deps_main
     roslaunch_d = rospkg.RosPack().get_path('roslaunch')
-    rosmaster_d = rospkg.RosPack().get_path('rosmaster')
+    rosmaster_d = rospkg.RosPack().get_path('test_rosmaster')
```
dirk-thomas, line 66: Why did you change the referenced package? Was that necessary for some reason? Since `roslaunch` does not depend on `test_rosmaster` it should not use that.

tools/roslaunch/test/unit/test_substitution_args.py
```diff
@@ -49,7 +49,7 @@ def test__arg():
     
     # test invalid
     try:
-        _arg('$(arg)', 'arg', [], ctx)
+        _arg('arg', [], ctx)
```
dirk-thomas, line 52: Why did you have to change this test?

tools/roslaunch/test/unit/test_substitution_args.py
```diff
@@ -60,29 +60,24 @@ def test__arg():
         ('', ('$(arg empty)', 'arg empty', ['empty'], ctx)),
         ('sim', ('$(arg baz)', 'arg baz', ['baz'], ctx)),
         
-        # test with other args present, should only resolve match
-        ('1234512345', ('$(arg foo)$(arg foo)', 'arg foo', ['foo'], ctx)),
-        ('12345$(arg baz)', ('$(arg foo)$(arg baz)', 'arg foo', ['foo'], ctx)),            
-        ('$(arg foo)sim', ('$(arg foo)$(arg baz)', 'arg baz', ['baz'], ctx)),            
-
```
dirk-thomas, line 67: Why did you remove these?

tools/roslaunch/test/unit/test_substitution_args.py
```diff
@@ -142,7 +138,7 @@ def test_resolve_args():
             
     # test against strings that should not match
     noop_tests = [
-        '$(find roslaunch', '$find roslaunch', '', ' ', 'noop', 'find roslaunch', 'env ROS_ROOT', '$$', ')', '(', '()',
+        '$find roslaunch', '', ' ', 'noop', 'find roslaunch', 'env ROS_ROOT', '$$', ')', '(', '()',
```
dirk-thomas, line 141: Why did you remove this?

tools/roslaunch/test/unit/test_roslaunch_depends.py
RainCT, line 66: iirc it was necessary for the test to pass (this is before any of the other changes)

tools/roslaunch/test/unit/test_substitution_args.py
RainCT, line 52: The signature of _arg changed, it no longer takes "resolved" as first parameter.

tools/roslaunch/test/unit/test_substitution_args.py
RainCT, line 67: _arg no longer gets called with the entire string (only the relevant entry), so these tests are now meaningless.

tools/roslaunch/src/roslaunch/substitution_args.py
RainCT, line 138: It is already checked in the wrapper function _find_hack(). Similarly, _find_executable()/etc are only called by _find(), not directly.

tools/roslaunch/test/unit/test_substitution_args.py
RainCT, line 141: I moved '$(find roslaunch' to the line below, ie. from no-op to parse error.

This is because the new parser will throw an "Unbalanced parentheses in substitution args" error. If there's some value in keeping the old behaviour, this can be removed by just removing the two lines at the end of _parse_args(), but I found it made more sense to complain about this probable typo.

RainCT:
Hi Dirk,

Thanks for the review!

Sorry about the delay, I was travelling when I got the notification and completely forgot to go back to it.

RainCT:
Friendly ping.

efernandez:
@dirk-thomas What is preventing this to be merged? I think this is a good feature and I'd be happy to help with the missing part if any/needed.

dirk-thomas:
While the patch implements an interesting feature it significantly changes the internal structure of roslaunch. The current tests are not sufficient to ensure that this big change does not break existing launch files. Therefore the PRs hasn't been merged.

There are a few other PRs regarding new features in roslaunch and they are all pending for the same reason.

I am aware that this is not a satisfactory answer. But stability for Indigo is very important.

To move forward on this (and other similar PRs) I think one option would be to get more user with plenty of (complex) roslaunch files to try these changes and confirm that it doesn't break anything for them.

gerkey:
Could we accomplish the same thing using the `pyparsing` module?  E.g., this simple parser:

```
from pyparsing import *
d =  '$(find $(arg robot)_description)/robots/upload_$(arg robot).launch'
OneOrMore(nestedExpr(opener='$(') + Word(printables, excludeChars=['$'])).parseString(d)
```

yields:

```
([(['find', (['arg', 'robot'], {}), '_description'], {}), '/robots/upload_', (['arg', 'robot'], {}), '.launch'], {})
```

rethink-rlinsalata:
+1, to the nested substitution feature. This would make a big difference in our use cases (very much along the same lines as the '$(robot_model)_description' examples discussed above). Not having this has meant we have had to make launch files (robot) platform specific, and install/devel specific, which really shouldn't be.

@dirk-thomas Considering the original substitution xml test launch file is still part of the tests, I'm curious what else you see is missing to ensure this feature doesn't break existing launch files?

(also, to @gerkey 's point - I agree it would be good improvement to leverage a dedicated Parsing library for parsing in roslaunch. However, at this point I think it would be better to add this valuable feature first, incrementally upon the implementation that already exists. Then, doing a library refactor could be tested against stably reproducing the same set of features / behavior.)

rethink-imcmahon:
+1 from me as well. We spent a while searching for any documentation or ROS answers regarding nested args today, and I was surprised this hasn't come up more often.

The python evaluation addition to `xacro` was a huge boost to its usefulness, which @rhaschke appears to have brought to `roslaunch` with https://github.com/ros/ros_comm/pull/707 . I was able to construct an incredibly configurable gripper `xacro` with 8x different swappable fingers and fingertips with this feature https://github.com/ros/xacro/pull/54

dirk-thomas:
I will close this in favor if the current PR #762 (which is a follow up of #707 mentioned in the previous comment).

PR #383, opened 2014-04-01
Topic/Connection statistic measurements

forouher:
These patches add an enhancement to roscpp and rospy that monitors some metrics (period, traffic, message delay, dropped messages) over every ROS connection and periodically sends out aggregated statistics data over a common topic /statistics.

The rationale behind this is that it is currently hard to measure these things without actually subscribing to the topics. E.g., if you want to know at which frequency your Kinect camera publishes images, you have to check with "rostopic hz /foo". But that will create quite a noticable load on the system and in any case might show a misleading number.  These patches add a method to measure those things in a lightweight way for the whole system.

I've extended rqt_graph to annotate the edges of the ROS graph with the statistics measured:

stats_viewer

For backward compatibility these statistics are disabled by default. They can be enabled via "rosparam set enable_statistics true".

I believe this enhancement might of useful to the ROS community at large. If there are any questions, concerns, remarks, feel free to share them.

cheers
Dariush

### DIFF
--- a/clients/roscpp/src/libros/statistics.cpp
+++ b/clients/roscpp/src/libros/statistics.cpp
@@ -1,0 +1,202 @@
+/*
+ * Copyright (C) 2013-2014, Dariush Forouher
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *   * Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *   * Neither the names of Stanford University or Willow Garage, Inc. nor the names of its
+ *     contributors may be used to endorse or promote products derived from
+ *     this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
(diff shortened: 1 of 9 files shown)


clients/rospy/src/rospy/impl/statistics.py
```diff
@@ -0,0 +1,303 @@
...
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+
+import struct
+import select
+
+try:
+    from cStringIO import StringIO #Python 2.x
+    import thread as _thread # Python 2
```
dirk-thomas: What is these two imports needed for? They seem to be unused.

clients/rospy/src/rospy/impl/statistics.py
```diff
@@ -0,0 +1,303 @@
...
+    python3 = 0
+    def isstring(s):
```
dirk-thomas: This function seem to be unused either.

dirk-thomas:
Please update the code to conform to PEP8 (e.g. spaces instead of tabs, spaces around operators, spaces after commas etc.).

It would be great if the parameters / constraints like (window size etc.) could also be configured using parameters.

clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -728,7 +731,7 @@ def receive_loop(self, msgs_callback):
                     if self.socket is not None:
                         msgs = self.receive_once()
                         if not self.done and not is_shutdown():
-                            msgs_callback(msgs)
+                            msgs_callback(msgs,self)
```
dirk-thomas: Please add a space after the comma (PEP8).

dirk-thomas:
Please also update the C++ code to comply with the code style: e.g. add spaces after commas as well as add spaces around operators. Also keep curly braces on separate lines. See http://wiki.ros.org/CppStyleGuide for details.

clients/roscpp/src/libros/statistics.cpp
```diff
@@ -0,0 +1,208 @@
...
+        msg.period_mean = ros::Duration(0);
+        msg.period_stddev = ros::Duration(0);
+        msg.period_max = ros::Duration(0);
+    }
+
+      if (!pub_.getTopic().length()) {
+        ros::NodeHandle n("~");
+        // creating the publisher in the constructor results in a deadlock. so do it here.
+        pub_ = n.advertise<rosgraph_msgs::TopicStatistics>("/statistics",1);
+      }
```
dirk-thomas: The indentation of this block seems to be wrong.

dirk-thomas:

clients/rospy/src/rospy/impl/statistics.py
```diff
@@ -0,0 +1,255 @@
...
+
+        # Calculate how many bytes of traffic did this message need?
+        self.stat_bytes_window_ = stat_bytes - self.stat_bytes_last_
+        self.stat_bytes_last_ = stat_bytes
+
+        # rospy has the feature to subscribe a topic with AnyMsg which aren't deserialized.
+        # Those subscribers won't have a header. But as these subscribers are rather rare
+        # ("rostopic hz" is the only one I know of), I'm gonna ignore them.
+        if msg._has_header:
+            self.delay_list_.append(arrival_time - msg.header.stamp)
```
tfoote, line 245: This is not valid. Data can be from the past.

tfoote, line 245: Playing back a bag file is a canonical use case which could give you several months of "delay".

This is more a measure of age of data, and the data should be timestamped as when it was observed. If there's multiple stages in a pipeline this number will increase at each stage.

dirk-thomas:
Closing in favor of #398.

PR #385, opened 2014-04-05
rosmsg does not show messages existing only in devel space

dirk-thomas:
Initially reported here: http://answers.ros.org/question/147586/unable-to-locate-custom-action-using-rosmsg-with-catkin

This effects especially messages generated for actions.

### DIFF
--- a/tools/rosmsg/package.xml
+++ b/tools/rosmsg/package.xml
@@ -19,6 +19,7 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
+  <run_depend version_gte="0.6.4">catkin</run_depend>
   <run_depend>genmsg</run_depend>
   <run_depend>python-rospkg</run_depend>
   <run_depend>rosbag</run_depend>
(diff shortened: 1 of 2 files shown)


dirk-thomas:
This will effectively double the search time in case of a _devel_ space since `rosmsg` has to look through both spaces (_source_ and _devel_). In _install_ space the patch should be a no-op.

tfoote:
+1

dirk-thomas:
Since the previous search results might utilize the rospkg cache but the new additional search using catkin_pkg does not the performance might actually be worse. In http://answers.ros.org/question/147586/unable-to-locate-custom-action-using-rosmsg-with-catkin/ it is mentioned that `rosmsg list` goes from nearly instant result to 15-30s runtime.

dirk-thomas:
Based on ros/catkin@cb1e8febcc6cb819e25a9fc6d3235379ac97e5e1 the repeated crawling can be avoided which speeds up this use case significantly.

dirk-thomas:
In my testing the speed of `rosmsg list` was only slightly slower (from 0.45s to 0.55s) in a workspace with ros base built in isolation. Before the performance patch it took about 6s.

wjwwood:
+1

PR #386, opened 2014-04-29
adds op tool (for python operations)

efernandez:
Allows to do python operations between message fields taken from several topics.

A typical use case consist on calculating the norm of a vector, or converting a quaternion to Euler angles for visualization; some examples are included in the `op` python script header comment.

### DIFF
--- a/tools/topic_tools/scripts/op
+++ b/tools/topic_tools/scripts/op
@@ -51,6 +51,10 @@ class TopicOp:
         self.expression = args.expression
 
         input_class, input_topic, self.input_fn = rostopic.get_topic_class(args.input)
+        if input_topic == None:
+            print 'ERROR: Wrong input topic (or topic field): %s'%args.input
+            exit()
+
         output_class = roslib.message.get_message_class(args.output_type)
 
         self.sub = rospy.Subscriber(input_topic, input_class, self.callback)
--- a/tools/topic_tools/scripts/op
+++ b/tools/topic_tools/scripts/op
@@ -13,59 +13,28 @@ after performing a valid python operation (including numpy).
 The operations are done on the message, which is taken in the variable 'm'.
 
 * Some examples:
-$ rosrun topic_tools op /imu/orientation/x /x std_msgs/Float64 '-rad2deg(m)'
+$ rosrun topic_tools op /imu/orientation/x /x_in_degrees std_msgs/Float64 '-rad2deg(m)'
 $ rosrun topic_tools op /imu/orientation /norm std_msgs/Float64 'sqrt(sum(array([m.x, m.y, m.z, m.w])))'
-
-* Interactive mode:
-$ rosrun topic_tools op <in> <out> <type> <exp> True
-
-This goes into an ipdb debugger session inside the callback for the input (in)
-topic, so you can redefine the self.expression (exp) to use, or create new logic.
-Indeed you can create new attributes and def functions. However, you cannot:
-- self.expression cannot use an assignment
-- cannot create local variables
-
-In order to create a new function and variable, you can follow this example,
-which computes a weighted mean:
-ipdb> import types
-ipdb> self.mean = 0
-ipdb> def compute_mean(self, x): self.mean = 0.9 * self.mean + 0.1 * x; return self.mean
-ipdb> self.compute_mean = types.MethodType( compute_mean, self )
-ipdb> self.expression = 'self.compute_mean( m.data )'
-
-Once you are done, to avoid entering again in the callback, just do:
-ipdb> self.interactive = False
-before hitting 'c' to continue.
-
 """
 
 import roslib
-
 import rospy
 import rostopic
+
 import sys
 
-import ipdb
-
 from numpy import *
-
-def set_trace():
-    ip = ipapi.get()
-    def_colors = ip.options.colors
-    Pdb(def_colors).set_trace(sys._getframe().f_back)
-
 
 class TopicOp:
 
     def __init__( self ):
         if len( sys.argv ) < 4:
-            sys.exit( 'Usage: %s <input> <output topic> <output type> [<expression on m>] [<interactive mode>]' % sys.argv[0] )
+            sys.exit( 'Usage: %s <input> <output topic> <output type> [<expression on m>]' % sys.argv[0] )
 
         self.input        = sys.argv[1]
         self.output_topic = sys.argv[2]
         self.output_type  = sys.argv[3]
         self.expression   = sys.argv[4] if len( sys.argv ) > 4 else 'm'
-        self.interactive  = sys.argv[5] == 'True' if len( sys.argv ) > 5 else False
 
         # We need the class, rather than the type,
         # so we use get_topic_class instead of get_topic_type.
@@ -80,9 +49,6 @@ class TopicOp:
             m = self.input_fn( msg )
         else:
             m = msg
-
-        if self.interactive:
-            ipdb.set_trace()
 
         res = eval( self.expression )
         self.pub.publish( res )
--- a/tools/topic_tools/scripts/op
+++ b/tools/topic_tools/scripts/op
@@ -31,7 +31,7 @@ class TopicOp:
     def __init__(self):
         parser = argparse.ArgumentParser(
             formatter_class=argparse.RawTextHelpFormatter,
-            description='Make a python operation on a topic.\n\n'
+            description='Apply a Python operation to a topic.\n\n'
                         'Usage:\n\t<input> <output topic> <output type> [<expression on m>]\n\n'
                         'Example:\n\trosrun topic_tools op /imu/orientation /norm std_msgs/Float64 \'sqrt(sum(array([m.x, m.y, m.z, m.w])))\'')
         parser.add_argument('input')
--- a/tools/topic_tools/scripts/op
+++ b/tools/topic_tools/scripts/op
@@ -30,12 +30,17 @@ class TopicOp:
         parser = argparse.ArgumentParser(
             formatter_class=argparse.RawTextHelpFormatter,
             description='Apply a Python operation to a topic.\n\n'
-                        'Usage:\n\t<input> <output topic> <output type> [<expression on m>]\n\n'
-                        'Example:\n\trosrun topic_tools op /imu/orientation /norm std_msgs/Float64 \'sqrt(sum(array([m.x, m.y, m.z, m.w])))\'')
-        parser.add_argument('input')
-        parser.add_argument('output_topic')
-        parser.add_argument('output_type')
-        parser.add_argument('expression', default='m')
+                        'Usage:\n\trosrun topic_tools op '
+                        '<input> <output topic> <output type> '
+                        '[<expression on m>]\n\n'
+                        'Example:\n\trosrun topic_tools op /imu/orientation '
+                        '/norm std_msgs/Float64 '
+                        '\'sqrt(sum(array([m.x, m.y, m.z, m.w])))\'')
+        parser.add_argument('input', help='Input topic or topic field.')
+        parser.add_argument('output_topic', help='Output topic.')
+        parser.add_argument('output_type', help='Output topic type.')
+        parser.add_argument('expression', default='m',
+                help='Python expression to apply on the input message \'m\'.')
 
         args = parser.parse_args()
 
--- a/tools/topic_tools/scripts/op
+++ b/tools/topic_tools/scripts/op
@@ -43,8 +43,6 @@ class TopicOp:
 
         self.expression = args.expression
 
-        # We need the class, rather than the type,
-        # so we use get_topic_class instead of get_topic_type.
         input_class, input_topic, self.input_fn = rostopic.get_topic_class(args.input)
         output_class = roslib.message.get_message_class(args.output_type)
 
@@ -65,7 +63,7 @@ if __name__ == '__main__':
 
     rospy.init_node('op', anonymous=True)
 
-    x = TopicOp()
+    app = TopicOp()
 
     while not rospy.is_shutdown():
         rospy.spin()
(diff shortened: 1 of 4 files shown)


tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,100 @@
...
+@author: enriquefernandez
+@todo: support non-scalar output types (requires defining a syntax for the expression, maybe YAML)
+
+Allows to take a topic or one of it fields and output it on another topic
+after performing a valid python operation (including numpy).
+
+The operations are done on the message, which is taken in the variable 'm'.
+
+* Some examples:
+$ rosrun pal_topic_tools op /imu/orientation/x /x std_msgs/Float64 '-rad2deg(m)'
```
RainCT: - remove pal_
- "/imu/orientation/x /x" <- this is rather confusing, maybe "/x_in_degrees"?

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,100 @@
...
+ipdb> self.expression = 'self.compute_mean( m.data )'
+
+Once you are done, to avoid entering again in the callback, just do:
+ipdb> self.interactive = False
+before hitting 'c' to continue.
+
+"""
+
+import roslib
+roslib.load_manifest('pal_topic_tools')
```
RainCT: catkin?

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,100 @@
...
+$ rosrun pal_topic_tools op /imu/orientation /norm std_msgs/Float64 'sqrt(sum(array([m.x, m.y, m.z, m.w])))'
+
+* Interactive mode:
+$ rosrun pal_topic_tools op <in> <out> <type> <exp> True
+
+This goes into an ipdb debugger session inside the callback for the input (in)
+topic, so you can redefine the self.expression (exp) to use, or create new logic.
+Indeed you can create new attributes and def functions. However, you cannot:
+- self.expression cannot use an assignment
+- cannot create local variables
```
RainCT: - double negation (cannot cannot)

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,99 @@
...
+after performing a valid python operation (including numpy).
+
+The operations are done on the message, which is taken in the variable 'm'.
+
+* Some examples:
+$ rosrun topic_tools op /imu/orientation/x /x std_msgs/Float64 '-rad2deg(m)'
+$ rosrun topic_tools op /imu/orientation /norm std_msgs/Float64 'sqrt(sum(array([m.x, m.y, m.z, m.w])))'
+
+* Interactive mode:
+$ rosrun topic_tools op <in> <out> <type> <exp> True
```
RainCT: This True isn't very nice, {-i, --interactive} would make more sense.

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,99 @@
...
+def set_trace():
+    ip = ipapi.get()
+    def_colors = ip.options.colors
+    Pdb(def_colors).set_trace(sys._getframe().f_back)
+
+
+class TopicOp:
+
+    def __init__( self ):
+        if len( sys.argv ) < 4:
```
RainCT: Use argparse?

RainCT:
Honestly I don't really see the point for this, at least all this interactive part. It's easier to just create a script with a rospy.Subscriber than to remember how it works.

I'd maybe keep this script, but only for very easy operations (like the rad2deg). For anything else, just write a script.

efernandez:
I've simplified the code, removing the interactive mode as advised and using argparse, apart from some clean up.

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,72 @@
...
+
+# Allow for numpy operations in the given expression
+from numpy import *
+
+class TopicOp:
+
+    def __init__(self):
+        parser = argparse.ArgumentParser(
+            formatter_class=argparse.RawTextHelpFormatter,
+            description='Make a python operation on a topic.\n\n'
```
RainCT: Nitpick: Sound a bit weird to me. Maybe something like "Apply a Python transformation to a topic"?

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,72 @@
...
+        parser.add_argument('output_topic')
+        parser.add_argument('output_type')
+        parser.add_argument('expression', default='m')
+
+        args = parser.parse_args()
+
+        self.expression = args.expression
+
+        # We need the class, rather than the type,
+        # so we use get_topic_class instead of get_topic_type.
```
RainCT: Why is this comment about get_topic_type relevant?

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,72 @@
...
+
+        res = eval(self.expression)
+        self.pub.publish(res)
+
+
+if __name__ == '__main__':
+
+    rospy.init_node('op', anonymous=True)
+
+    x = TopicOp()
```
RainCT: app = TopicOp()

RainCT:
LGTM.

tools/topic_tools/scripts/op
efernandez: I'd like to make explicit that it perform and operation, because the name of the script is `op`.
But I could use this instead:
"Apply a Python operation to a topic."
Any comment?

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Created on Fri Aug  9 12:13:00 2013
```
dirk-thomas: Please remove this line.

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self):
+        parser = argparse.ArgumentParser(
+            formatter_class=argparse.RawTextHelpFormatter,
+            description='Apply a Python operation to a topic.\n\n'
+                        'Usage:\n\t<input> <output topic> <output type> [<expression on m>]\n\n'
+                        'Example:\n\trosrun topic_tools op /imu/orientation /norm std_msgs/Float64 \'sqrt(sum(array([m.x, m.y, m.z, m.w])))\'')
+        parser.add_argument('input')
+        parser.add_argument('output_topic')
+        parser.add_argument('output_type')
+        parser.add_argument('expression', default='m')
```
dirk-thomas: Please add help attributes for the arguments.

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,70 @@
...
+        self.sub = rospy.Subscriber(input_topic, input_class, self.callback)
+        self.pub = rospy.Publisher(args.output_topic, output_class)
+
+    def callback(self, msg):
+        if self.input_fn is not None:
+            m = self.input_fn(msg)
+        else:
+            m = msg
+
+        res = eval(self.expression)
```
dirk-thomas: You should make sure that the local variables are only the ones you want to make available. Currently the user could also access `self` and `msg`.

tools/topic_tools/scripts/op
```diff
@@ -0,0 +1,70 @@
...
+
+# Allow for numpy operations in the given expression
+from numpy import *
+
+class TopicOp:
+
+    def __init__(self):
+        parser = argparse.ArgumentParser(
+            formatter_class=argparse.RawTextHelpFormatter,
+            description='Apply a Python operation to a topic.\n\n'
```
dirk-thomas: I think this needs a bit more documentation, e.g. that the message object is made available in the variable `m`.

dirk-thomas:
The script name `op` seems to be not very clear. Just one idea would be naming it `transform`? But other suggestions are welcome.

wjwwood:
Would "transform" make more sense than "op"? Data comes in, what is essentially a transform function gets applied to it, and data goes out again.

tfoote:
Since this tool is actually transforming while relaying I would suggest that this not be a separate tool but be done through additional arguments to the relay command line tool. Such that you can "relay INTOPIC OUTTOPIC --transform OUTPUT_TYPE EXPRESSION

tools/topic_tools/scripts/op
efernandez: I've added a more detailed description below.

efernandez:
@dirk-thomas @wjwwood  `op` stands for operation, but if you find it unintuitive, I'm happy with renaming everything to `transform`. If you agree, I'll make the change.

@tfoote AFAIK `relay` operates at a lower level, where we don't have access to the message data (like the message fields), so we can't transform it with regular mathematical operations. Moreover, it's C++ code, so IMHO it won't be easy to provide this functionality, which is this Python code is done by means of the `eval` function.

dirk-thomas:
+1 for calling it `transform` and keeping it separate from `relay`.

tools/topic_tools/scripts/transform
```diff
@@ -0,0 +1,81 @@
...
+        self.sub = rospy.Subscriber(input_topic, input_class, self.callback)
+        self.pub = rospy.Publisher(args.output_topic, output_class)
+
+    def callback(self, msg):
+        if self.input_fn is not None:
+            m = self.input_fn(msg)
+        else:
+            m = msg
+
+        res = eval(self.expression)
```
dirk-thomas: My previous comment still applies here:

You should make sure that the local variables are only the ones you want to make available. Currently the user could also access `self` and `msg`.

efernandez:
Now only `m` should be available in the given expression.

I'm also trying to add env-hooks so we have auto-completion.

efernandez:
With the help of @po1, now there's also bash completion for `rosrun topic_tools transform`.
This is self-contained in the single last commit, in case you don't like how it's done.

tools/topic_tools/scripts/transform
```diff
@@ -0,0 +1,86 @@
...
+        self.sub = rospy.Subscriber(input_topic, input_class, self.callback)
+        self.pub = rospy.Publisher(args.output_topic, output_class, queue_size=1)
+
+    def callback(self, m):
+        if self.input_fn is not None:
+            m = self.input_fn(m)
+
+        try:
+            res = eval(self.expression, {}, {"m":m})
+            self.pub.publish(res)
```
po1, line 72: Shouldn't this line be outside of the `try` block? Or in its own?

tools/topic_tools/scripts/transform
```diff
@@ -0,0 +1,86 @@
...
+        except NameError as e:
+            print "Expression using variables other than 'm': %s"%e.message
+        except UnboundLocalError as e:
+            print "Wrong expression: %s"%e.message
+        except:
```
po1, line 77: I don't feel very comfortable with the naked `except`. If the tool fails, it will hide the real reason. I don't really see why we should hide it. Python stack frames outputs are fairly explicit and informative.

dirk-thomas:
The code is not yet Python 3 compatible. Please make it work with both Python 2 as well as 3. See http://wiki.ros.org/python_2_and_3_compatible_code for some common things.

tools/topic_tools/scripts/transform
efernandez, line 72: In that case we'll have to check whether `res` has been set or not, so I think this is cleaner.

efernandez:
I've tried to made the code Python 3 compatible, but I cannot test it on my machine because my ROS installation uses Python 2 and some packages are not found when I try to run the code with Python 3. @dirk-thomas Could you verify it runs with Python 3?

dirk-thomas:
I did some minor changes and merged it: ede12dbb4fd353e8ef4f62d95d1e0c2aa42526c6

efernandez:
Thanks Dirk!

BTW, the description says:

<pre>
A typical use case consist on calculating the norm of
a vector, or converting a quaternion to Euler angles
for visualization
</pre>

For the norm, we must do something like this:

<pre>
rosrun topic_tools transform /base_imu/orientation /norm std_msgs/Float32 'linalg.norm([m.x, m.y, m.z, m.w])
</pre>

However, to convert to Euler angles, it's not trivial. I have one example, but it requires to import `tf` and with the current implementation it might not be possible to has an expression that resolves to a `geometry_msgs/Vector3`. I can try to think of a solution and make another PR, but on the meantime, I'll prefer to remove the text that mentions the conversion to Euler angles. And I think the example for the `norm` should be included.
The solution could imply using `exec` instead of `eval`, and something like `out = in` instead of `m`.

dirk-thomas:
Please make another PR to update the usage description.

efernandez:
@dirk-thomas The description is only wrong in ede12db, but in the script is already OK.

Anyway, I'll try to give support for non-scalar output types; indeed, @po1 has told me an idea to do it using only `m`. If it works, I'll create a PR for that.

po1:
@efernandez  you can use something like the following:

``` python
msg = message_type(*eval("[{}]".format(input_str)))
```

With this, you can return multiple comma-separated values and they will constitute the fields of your message.

A yet better option would be to parse a key-value structure and run each value through `eval()` to generate the message.

I would see something like:

```
rosrun topic_tools transform /topic_in /topic_out geometry_msgs/Twist "{linear: {x: sqrt(m.x)}}"
```

It might be a bit overkill though.

efernandez:
@po1 I've implemented your first approach in #404 

BTW, I had to use `{}` instead of `[{}]` because otherwise this two examples wouldn't be possible:

<pre>
rosrun topic_tools transform /imu/orientation /euler geometry_msgs/Vector3 'tf.transformations.euler_from_quaternion([m.x, m.y, m.z, m.w])' --import tf
rosrun topic_tools transform /imu/orientation/x /x_in_degrees std_msgs/Float64 -- '-numpy.rad2deg(m)'
</pre>

The first one requires `{}`, and the second one requires the check for `Iterable` because `[{}]` cannot be used.

PR #395, opened 2014-05-02
fixes day comparison

efernandez:
I think this should also be sync'd downstream to hydro-devel.


tfoote:
+1
@ros-pull-request-builder please test this

dirk-thomas:
The patch will be _backported_ to older ROS distros.

PR #397, opened 2014-05-02
print parameter values in roslaunch (fix #89)

### LINKED ISSUE #89: roslaunch sorted and more verbose output
Migrated from https://code.ros.org/trac/ros/ticket/3905
Reported by:    kruset  Owned by:   kwc
Priority:   minor 
Attachment available: No
Last modified ago: Apr 2012

---

Hi,

working on navigation I notice the output of roslaunch could be improved IMO:

```
$ roslaunch navigation_stage move_base_amcl_10cm.launch
PARAMETERS
 * /use_sim_time
 * /move_base_node/global_costmap/observation_sources
 * /amcl/gui_publish_rate
 * /move_base_node/local_costmap/origin_y
```

The parameter list is not sorted and does not tell the parameters. i know you can get the parameters with --dump-params, but for that you need to run the roslaunch command in another shell, which can be a hassle, or stop that launch and start again (hassle again). I see no benefit in not showing the param values by default.

So I suggest both sorting outputs and showing params. Diff rosinstall/config.py:

```
@@ -222,11 +223,11 @@
         """
         summary = '\nSUMMARY\n========'
         if self.clear_params:
-            summary += '\n\nCLEAR PARAMETERS\n' + '\n'.join([' * %s'%p for p in self.clear_params])
+            summary += '\n\nCLEAR PARAMETERS\n' + '\n'.join(sorted([' * %s'%p for p in self.clear_params]))
         if self.params:
-            summary += '\n\nPARAMETERS\n' + '\n'.join([' * %s'%k for k in self.params])
+            summary += '\n\nPARAMETERS\n' + '\n'.join(sorted([' * %s  \t%s'%(k, v.value) for k, v in self.params.items()]))
         if not local:
-            summary += '\n\nMACHINES\n' + '\n'.join([' * %s'%k for k in self.machines if k])
+            summary += '\n\nMACHINES\n' + '\n'.join(sorted([' * %s'%k for k in self.machines if k]))
         summary += '\n\nNODES\n'
         namespaces = {}
         if local:
@@ -240,7 +241,7 @@
             else:
                 namespaces[ns].append(n)
         for k,v in namespaces.iteritems():
-            summary += '  %s\n'%k + '\n'.join(['    %s'%_summary_name(n) for n in v])
+            summary += '  %s\n'%k + '\n'.join(sorted(['    %s'%_summary_name(n) for n in v]))
             summary += '\n'
         return summary
```

Result:

```
$ roslaunch navigation_stage move_base_amcl_10cm.launch
...
PARAMETERS
 * /amcl/gui_publish_rate       10.0
 * /amcl/kld_err        0.05
 * /amcl/kld_z          0.99
 * /amcl/laser_lambda_short     0.1
```

Also I notice dump-params produces very ugly output for floats:

```
/amcl/odom_alpha1: 0.20000000000000001
/amcl/odom_alpha2: 0.20000000000000001
/amcl/odom_alpha3: 0.80000000000000004
/amcl/odom_alpha4: 0.20000000000000001
```

Not sure where that comes from, not sure how to fix it.

dirk-thomas:

### DIFF
--- a/tools/roslaunch/src/roslaunch/config.py
+++ b/tools/roslaunch/src/roslaunch/config.py
@@ -229,7 +229,17 @@ class ROSLaunchConfig(object):
         if self.clear_params:
             summary += '\n\nCLEAR PARAMETERS\n' + '\n'.join(sorted([' * %s'%p for p in self.clear_params]))
         if self.params:
-            summary += '\n\nPARAMETERS\n' + '\n'.join(sorted([' * %s'%k for k in self.params]))
+            def strip_string(value):
+                max_length = 20
+                if len(value) > max_length:
+                    value = value[:max_length - 3] + '...'
+                # if non printable characters are present return replacement
+                for char in value:
+                    i = ord(char)
+                    if i < 32 or i > 126:
+                        return '<...>'
+                return value
+            summary += '\n\nPARAMETERS\n' + '\n'.join(sorted([' * %s: %s' % (k, strip_string(str(v.value))) for k, v in self.params.items()]))
         if not local:
             summary += '\n\nMACHINES\n' + '\n'.join(sorted([' * %s'%k for k in self.machines if k]))
         summary += '\n\nNODES\n'


tools/roslaunch/src/roslaunch/config.py
```diff
@@ -229,7 +229,17 @@ def summary(self, local=False):
...
+                max_length = 20
+                if len(value) > max_length:
+                    value = value[:max_length - 3] + '...'
+                # if non printable characters are present return replacement
+                for char in value:
+                    i = ord(char)
+                    if i < 32 or i > 126:
+                        return '<...>'
+                return value
+            summary += '\n\nPARAMETERS\n' + '\n'.join(sorted([' * %s \t%s' % (k, strip_string(str(v.value))) for k, v in self.params.items()]))
```
wjwwood: This formatting `' * %s \t%s'` is going to produce weird results in my opinion, I would opt for something like this:

``` python
summary += "\n\nPARAMETERS\n"
template = " * {0<" + str(max([len(x) for x in self.params])) + "} {1}"
param_strs = [template.format(k, strip_string(str(v.value))) for k, v in self.params.items()]
summary += '\n'.join(sorted(param_strs))
```

dirk-thomas: If one of the parameters is very long the distance between short names and values would be significant which might make keeping track of lines difficult. But I am fine with any approach. Lets see what @tfoote is preferring.

dirk-thomas: If the max length will result in a line wrap on the terminal that might be bad if it is applied to all parameters.

wjwwood: Yeah, that's all true, but in that case I would just put a little bit of white space, using one `\t` isn't going to align the output and isn't going to give consistent space between. I think we should choose one or the other. So I guess my alternative suggestion is to just drop the `\t`.

tfoote: I think sticking with the simple solution of a fixed space offset is good enough to read and is simple to implement. Possibly with a colon delimter ": "

dirk-thomas: Done as @tfoote suggested: `* %s: %s`

wjwwood: +1

PR #398, opened 2014-05-05
Topic/Connection statistic measurements

dirk-thomas:
@forouher FYI

This PR replaces #383. I have:
- squashed your commits
- cleaned up some more code
- and fixed a compilation error with the unit tests in order to run CI

Please provide further patches against this branch.

### DIFF
--- a/clients/roscpp/CMakeLists.txt
+++ b/clients/roscpp/CMakeLists.txt
@@ -87,6 +87,7 @@ add_library(roscpp
   src/libros/topic_manager.cpp
   src/libros/poll_manager.cpp
   src/libros/publication.cpp
+  src/libros/statistics.cpp
   src/libros/intraprocess_subscriber_link.cpp
   src/libros/intraprocess_publisher_link.cpp
   src/libros/callback_queue.cpp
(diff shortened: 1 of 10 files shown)


dirk-thomas:
As @tfoote mentioned on the other PR the name `delay` seems to be not fitting very well. The timestamp in the header might be much older is e.g. data is played back from a bag file. We would suggest renaming the variable and everything related to something different and would propose `age` as a better term. @forouher What do you think about that?

forouher:
"Age" fits better. "Delay" was what I originally wanted to measure, but I couldn't figure out a way to do it without modifying the TCPROS protocol. Taking the header stamp was the next best thing I could think of. But it's not the same, obviously.

I'll rename the variables etc. to "age".

dirk-thomas:
Can you please add some documentation about this feature to the wiki? Either under ros_comm / roscpp / rospy or the rqt plugin or both - I am not sure where it makes more sense.

forouher:
Sure, I will amend the wiki (probably at both at ros_comm and rqt_graph with references to each other).

Thanks for merging this!

PR #399, opened 2014-05-05
fix rosbag record --node (fix #357)

### LINKED ISSUE #357: `rosbag record --node` crashes if monitored node dies violently
To reproduce, compile the following two programs:

```
// talker.cpp
#include "ros/ros.h"
#include "std_msgs/Int32.h"

#include <sstream>

{
  ros::init(argc, argv, "talker");
  ros::NodeHandle n;
  ros::Publisher chatter_pub = n.advertise<std_msgs::Int32>("chatter", 1000);
  ros::Rate loop_rate(10);

  int count = 0;
  while (ros::ok())
  {
    std_msgs::Int32 msg;
    msg.data = count;
    chatter_pub.publish(msg);
    ros::spinOnce();
    loop_rate.sleep();
    ++count;
  }

  return 0;
}
```

```
// listener.cpp
#include "ros/ros.h"
#include "std_msgs/Int32.h"

void chatterCallback(const std_msgs::Int32::ConstPtr& msg)
{
    ROS_INFO("I heard: [%d]", msg->data);
    if (msg->data == 30) {
        _exit(-1);
    }
}

{
  ros::init(argc, argv, "listener");
  ros::NodeHandle n;
  ros::Subscriber sub = n.subscribe("chatter", 1000, chatterCallback);
  ros::spin();
  return 0;
}
```

Now, in four separate terminals:

```
$ roscore
```

```
$ rosbag record --node /listener -O test.bag
```

```
$ ./listener
```

```
$ ./talker
```

The rosbag process will exit with the exception "terminate called after throwing an instance of 'XmlRpc::XmlRpcException'" and leave a broken bag 'test.bag.active' that cannot be fixed by `rosbag reindex`.

The problem can be traced to the Recorder::doCheckMaster() function in ros_comm/tools/rosbag/src/recorder.cpp.

Changing line 570 fixes it:

```
- if (!c.isFault() && resp.size() > 0 && static_cast<int>(resp[0]) == 1)
+ if (!c.isFault() && resp.valid() && resp.size() > 0 && static_cast<int>(resp[0]) == 1)
```

Note also that there is an error in line 578:

```
ROS_ERROR("Node at: [%s] failed to return subscriptions.", static_cast<std::string>(resp[2]).c_str());
```

Here, `resp` is meant to refer to the response to the XmlRpc "lookupNode" call to master, but it actually refers to the "getSubscriptions" call. Thus, the error messsage does not print a URI, as intended.

dirk-thomas:

### DIFF
--- a/tools/rosbag/src/recorder.cpp
+++ b/tools/rosbag/src/recorder.cpp
@@ -562,17 +562,17 @@ void Recorder::doCheckMaster(ros::TimerEvent const& e, ros::NodeHandle& node_han
         } else {
 
           XmlRpc::XmlRpcClient c(peer_host.c_str(), peer_port, "/");
-          XmlRpc::XmlRpcValue req;
-          XmlRpc::XmlRpcValue resp;
-          req[0] = ros::this_node::getName();
-          c.execute("getSubscriptions", req, resp);
+          XmlRpc::XmlRpcValue req2;
+          XmlRpc::XmlRpcValue resp2;
+          req2[0] = ros::this_node::getName();
+          c.execute("getSubscriptions", req2, resp2);
           
-          if (!c.isFault() && resp.size() > 0 && static_cast<int>(resp[0]) == 1)
+          if (!c.isFault() && resp2.valid() && resp2.size() > 0 && static_cast<int>(resp2[0]) == 1)
           {
-            for(int i = 0; i < resp[2].size(); i++)
+            for(int i = 0; i < resp2[2].size(); i++)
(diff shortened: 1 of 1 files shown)


wjwwood:
Does this address the comment about line 578?

dirk-thomas:
Yes, the inner variables got a different name so that the error message actually shows the information from the outside response.

wjwwood:
+1

tfoote:
+1

PR #402, opened 2014-05-07
uses else clause

efernandez:
Following the recommendations in https://docs.python.org/2/tutorial/errors.html#handling-exceptions


dirk-thomas:
Related to #386.

PR #404, opened 2014-05-07
Adds support for non-scalar output types

efernandez:
And also adds `--import` arg.

### DIFF
--- a/tools/topic_tools/scripts/transform
+++ b/tools/topic_tools/scripts/transform
@@ -3,16 +3,17 @@
 
 """
 @author: enriquefernandez
-@todo: support non-scalar output types (requires defining a syntax for the expression, maybe YAML)
 
 Allows to take a topic or one of it fields and output it on another topic
-after performing a valid python operation (including numpy).
+after performing a valid python operation.
 
 The operations are done on the message, which is taken in the variable 'm'.
 
-* Some examples:
-$ rosrun topic_tools transform /imu/orientation/x /x_in_degrees std_msgs/Float64 '-rad2deg(m)'
+* Examples:
+$ rosrun topic_tools transform /imu/orientation/x /x_in_degrees std_msgs/Float64 -- '-numpy.rad2deg(m)'
 $ rosrun topic_tools transform /imu/orientation /norm std_msgs/Float64 'sqrt(sum(array([m.x, m.y, m.z, m.w])))'
+$ rosrun topic_tools transform /imu/orientation /norm std_msgs/Float64 'numpy.linalg.norm([m.x, m.y, m.z, m.w])' --import numpy
+$ rosrun topic_tools transform /imu/orientation /euler geometry_msgs/Vector3 'tf.transformations.euler_from_quaternion([m.x, m.y, m.z, m.w])' --import tf
 """
 
 from __future__ import print_function
@@ -22,10 +23,8 @@ import rospy
 import rostopic
 
 import argparse
+import importlib
 import sys
-
-# Allow for numpy operations in the given expression
-from numpy import *
 
 from collections import Iterable
 
@@ -41,7 +40,7 @@ class TopicOp:
                         'through another topic.\n\n'
                         'Usage:\n\trosrun topic_tools transform '
                         '<input> <output topic> <output type> '
-                        '[<expression on m>]\n\n'
+                        '[<expression on m>] [--import numpy tf]\n\n'
                         'Example:\n\trosrun topic_tools transform /imu/orientation '
                         '/norm std_msgs/Float64 '
                         '\'sqrt(sum(array([m.x, m.y, m.z, m.w])))\'')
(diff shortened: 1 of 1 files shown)


po1:
I fear this is going to fail with a string. Can you test it?
If it does fail, maybe you should add a check for `not isinstance(res, basestring)`

dirk-thomas:
Please rebase to head in order to make the PR mergable.

efernandez:
@po1 I changed the check to use `isinstance(res, (list, tuple))`.
@dirk-thomas rebased on top of `indigo-devel`.

tools/topic_tools/scripts/transform
```diff
@@ -51,9 +50,22 @@ class TopicOp:
...
+        )
 
         args = parser.parse_args()
 
+        self.modules = {}
+        for module in args.modules:
+            try:
+                mod = importlib.import_module(module)
+            except ImportError:
+                print('Failed to import module:', module)
```
dirk-thomas, line 65: Please print error messages to stderr, like this:

```
print('...', file=sys.stderr)
```

tools/topic_tools/scripts/transform
```diff
@@ -61,17 +73,17 @@ class TopicOp:
...
 
     def callback(self, m):
         if self.input_fn is not None:
             m = self.input_fn(m)
 
         try:
-            res = eval(self.expression, {}, {'m': m})
+            res = eval("{}".format(self.expression), self.modules, {'m': m})
         except NameError as e:
             print("Expression using variables other than 'm':", e.message)
```
dirk-thomas, line 88: Same here.

dirk-thomas:
Wouldn't it be more flexible to allow the user to use `import` within its expression? Then you wouldn't need the import option at all.

dirk-thomas:
Since we are releasing ros_comm _now_ I will merge this PR with my print-related comments: b8e7b987b0621c7a126c176027a768232da28ac7

I would still suggest considering my other comment regarding the imports. A potential change can be merged for the next patch release then.

efernandez:
@dirk-thomas  Regarding the imports, since I use `eval` to evaluate the given `expression`, which must be a Python expression and not a statement (as `exec` will support). For that reason, we cannot use `import module` inside `eval`, and that's why I added the `--import` argument. IMHO, using `exec` will require an output variable to assign the given expression to, so the user must do that.

These are the current examples using `eval`:

<pre>
$ rosrun topic_tools transform /imu/orientation/x /x_str std_msgs/String 'str(m)'
$ rosrun topic_tools transform /imu/orientation/x /x_in_degrees std_msgs/Float64 -- '-numpy.rad2deg(m)'
$ rosrun topic_tools transform /imu/orientation /norm std_msgs/Float64 'numpy.sqrt(numpy.sum(numpy.array([m.x, m.y, m.z, m.w])))'
$ rosrun topic_tools transform /imu/orientation /norm std_msgs/Float64 'numpy.linalg.norm([m.x, m.y, m.z, m.w])'
$ rosrun topic_tools transform /imu/orientation /euler geometry_msgs/Vector3 'tf.transformations.euler_from_quaternion([m.x, m.y, m.z, m.w])' --import tf
</pre>

Using `exec` they should be something like (using `out` and `in`, instead of simply `m`):

<pre>
$ rosrun topic_tools transform /imu/orientation/x /x_str std_msgs/String 'out = str(in)'
$ rosrun topic_tools transform /imu/orientation/x /x_in_degrees std_msgs/Float64 -- 'import numpy; out = -numpy.rad2deg(in)'
$ rosrun topic_tools transform /imu/orientation /norm std_msgs/Float64 'from numpy import *; out = sqrt(sum(array([in.x, in.y, in.z, in.w]))))'
$ rosrun topic_tools transform /imu/orientation /norm std_msgs/Float64 'from numpy.linalg import norm; out = norm([in.x, in.y, in.z, in.w])'
$ rosrun topic_tools transform /imu/orientation /euler geometry_msgs/Vector3 'from tf.transformations import euler_from_quaternion; out = euler_from_quaternion([in.x, in.y, in.z, in.w])'
</pre>

Basically, we need `out =`, but an advantage is that we can do `from module import *`.

@dirk-thomas @po1 What do you think? It might actually be better to use `exec`, since it's more clear and flexible for the user: just write a Python statement (or many).

po1:
We have to consider the point where it will simply be easier for the user to fire ipython and to type a few lines in an interactive interpreter rather than to fiddle with the syntax of this tool.

I like what the tool does for simple cases, but the syntax can already be cumbersome. My opinion is that it better stay simple.

dirk-thomas:
I am okay keeping the imports as-is. I was just throwing in the idea. Bit the rational to keep it simpler is a good point.

efernandez:
Alright.
BTW, could this tool be back-ported to hydro-devel?

dirk-thomas:
Since it is newly introduced I will consider it for back porting for the next Hydro patch release.

Can you please add documentation about it to the wiki so that users know about the tool?

efernandez:
@dirk-thomas wiki doc added to:
http://wiki.ros.org/topic_tools
http://wiki.ros.org/topic_tools/transform

dirk-thomas:
Great - thank you!

PR #407, opened 2014-05-08
Add Mux and Demux to hydro

Squelsh:
I was missing a Demux tool. So I brought the old code back to life.
While working on that, I also fixed the service-servers of the Mux tool, so more than one Mux can be used.

Code is tested for hydro. But I think it can also be merged into newer versions. As I have no newer version running, I could not test. ==> Would be nice, if someone could do that.

If my request gets merged, I am happy to update the ROS WIKI page.

### DIFF
--- a/tools/topic_tools/src/demux.cpp
+++ b/tools/topic_tools/src/demux.cpp
@@ -2,8 +2,8 @@
 // demux is a generic ROS topic demultiplexer: one input topic is fanned out
 // to 1 of N output topics. A service is provided to select between the outputs
 //
+// Copyright (C) 2009, Morgan Quigley
 // Copyright (C) 2014, Andreas Hermann
-// Code copied and adapted from the "toppic mux" by Morgan Quigley
 //
 // Redistribution and use in source and binary forms, with or without
 // modification, are permitted provided that the following conditions are met:
@@ -124,7 +124,7 @@ bool sel_srv_cb( topic_tools::DemuxSelect::Request  &req,
 
 void in_cb(const boost::shared_ptr<ShapeShifter const>& msg)
 {
-  ROS_INFO("Received an incoming msg ...");
+  ROS_DEBUG("Received an incoming msg ...");
   // when a message is incoming, check, if the requested publisher is already existing.
   // if not, create it with the information available from the incoming message.
   if(!g_selected->pub)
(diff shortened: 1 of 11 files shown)


tools/topic_tools/src/demux.cpp
```diff
@@ -1,9 +1,9 @@
 ///////////////////////////////////////////////////////////////////////////////
 // demux is a generic ROS topic demultiplexer: one input topic is fanned out
-// to 1 of N output topics. A service and topic subscription are provided
-// to select between the outputs.
+// to 1 of N output topics. A service is provided to select between the outputs
 //
-// Copyright (C) 2009, Morgan Quigley
+// Copyright (C) 2014, Andreas Hermann
```
tfoote, line 6: The copyright should be side by side not replace, even though it's mostly new. 

```
Copyright (C) 2009, Morgan Quigley
Copyright (C) 2014, Andreas Hermann
```

dirk-thomas:
This comment (https://github.com/Squelsh/ros_comm/blob/hydro-devel/tools/topic_tools/src/mux.cpp#L301) states explicitly that the services are provided on the node "mux" and suggests using remapping in order to use multiple mux instances.

This patch would effectively change the public API of the tool which is not a feasible change for Hydro. Even for Indigo it is questionable if that should be changed.

The demux part is definitely a good additional. Thank you for porting this functionality.

Squelsh:
Hi Tully, hi Dirk,
sorry for the quirks. I totally agree with your comments.
@tfoote : I corrected the Copyright as suggested.
@dirk-thomas : I reverted my changes and adapted the demux to behave the same way as the Mux.

dirk-thomas:
I will add the missing install rule (https://github.com/ros/ros_comm/pull/407/files#diff-ae340f410bae427e1853101dfdbade7eR55) when merging it into _indigo-devel_.

dirk-thomas:
For `mux` there are some Python scripts to ease the operations (add, delete, list, select). Wouldn't it make sense to have similar scripts for `demux`?

Squelsh:
I did not notice these scripts till now. Will have a look on monday.

Squelsh:
Wrong button )):

However: I don't see the point in a script, where the copyright is three times the lenghts of the script itself, but I committed the according scripts for the demux.

dirk-thomas:
Merged to indigo-devel in 4edc26b3ab23f39073d901bf1c69066c98277e81. It will be backported to hydro-devel before the next patch release. Thank you for the PR.

PR #410, opened 2014-05-14
rospy: shutdown hooks to be any callable object

jbohren:
This allows rospy shutdown hooks to be any callable object, instead of requiring they be actual functions or methods.


dirk-thomas:
Nice - thanks for the patch.

Issue #411, opened 2014-05-16
master::execute with wait_for_master=false is hiding comunication errors


v-lopez:
Some methods such as get param in https://github.com/ros/ros_comm/blob/indigo-devel/clients/roscpp/src/libros/param.cpp#L312 are implemented with calls to master::execute with wait_for_master=false.

In this situation, a failure to get a parameter because it is not set cannot be distinguished from a failure due to connection errors with the master.

Some feedback with regards to the type of error should be provided, maybe by printing the error message in https://github.com/ros/ros_comm/blob/indigo-devel/clients/roscpp/src/libros/master.cpp#L206 even if wait_for_master is false.

po1:
I don't know at all if this is feasible, but I'd favor a feedback in the logic rather than in the logs. I see three possible ways of doing this:
- the classical C approach with `errno`, which would be the least invasive but would mean keeping a global variable around, and ensuring that it is thread-local,
- changing the interface of `validateXmlrpcResponse` and `master::execute` so that we can get some feedback about what went wrong,
- and C++ exceptions, of which I am not really in favor here.

dirk-thomas:
The problem with the proposed ways (2. and 3.) is that they would change the public interface which is not feasible to be applied in an already released ROS distro. And even for an upcoming distro it must be carefully done since that implies that everyone using the API needs to update its code - either to adapt it to the changed interface or catch the additional exception.

The first approach would be possible without changing the API. But currently ROS does not use that kind of error signaling which would make it kind of special.

I can't think of a different way right now which would not require an API change.

po1:
> The problem with the proposed ways (2. and 3.) is that they would change the public interface which is not feasible to be applied in an already released ROS distro. And even for an upcoming distro it must be carefully done since that implies that everyone using the API needs to update its code - either to adapt it to the changed interface or catch the additional exception.

The API can be extended without necessarily implying any breaks. Parts of it can be overloaded or have default arguments, so that old code should stay compatible.
Before I spend too much time on this, would a patch that extends (but does not break) API be acceptable?

dirk-thomas:
Sure.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #412, opened 2014-05-16
Always display error message on communication error

v-lopez:
Fixes https://github.com/ros/ros_comm/issues/411


dirk-thomas:
The problem with this patch is that there are other use cases where the code path is invoked with the same arguments which should _not_ output the warning message since that would pollute the output in these cases.

dirk-thomas:
I will close this ticket since I don't see a way to achieve the desired goal of seeing a message but only in some specific cases.

PR #414, opened 2014-05-21
fix rostopic echo for subfields of time and duration

dirk-thomas:


tfoote:
+1

dirk-thomas:
Replaced with ros/genpy#29.

PR #429, opened 2014-06-04
Python 3 compatibility

dirk-thomas:


dirk-thomas:
Zero regressions with Python 2.

@wjwwood Please review.

tools/rosmaster/test/test_rosmaster_paramserver.py
```diff
@@ -82,7 +83,12 @@ def test_compute_param_updates(self):
 
             ]
         for correct, args in tests:
-            val = compute_param_updates(*args)
+            reg = Registrations(Registrations.PARAM_SUBSCRIPTIONS)
+            reg.map = args[0]
+            param_key = args[1]
+            param_val = args[2]
+
+            val = compute_param_updates(reg, param_key, param_val)
```
wjwwood, line 91: This change doesn't seem to be equivalent, it seems as though `reg` would have not been passed before. Is this right?

wjwwood:
Other than the comment +1

tools/rosmaster/test/test_rosmaster_paramserver.py
dirk-thomas, line 91: Yeah, looks pretty weird. But actually the test was not using the API as specified before and did not pass in a `Registrations` instance as it should be. So this patch should be a cleanup.

PR #430, opened 2014-06-05
Python 3 compatibility

dirk-thomas:


dirk-thomas:
@wjwwood please review.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -1274,12 +1287,39 @@ def __str__(self):
...
+        return self._compare(other, lambda s, o: s != o)
+
+class _IndexEntry(ComparableMixin):
     def __init__(self, time):
         self.time = time
 
-    def __cmp__(self, other):
-        return self.time.__cmp__(other.time)
+    def _cmpkey(self):
+        return self.time
```
wjwwood, line 1322: I had never seen this before, but after looking into it, it seems reasonable.

wjwwood:
+1

tools/rosbag/src/rosbag/bag.py
dirk-thomas, line 1322: :wink:

Issue #434, opened 2014-06-07
roscpp: race conditions in ros::service::call for zombie services


xqms:
ROS nodes can leave "zombie" services behind when they don't exit cleanly. I have found several issues that can pop up when you try to call such a zombie service.

The main issue is that `TransportTCP::connect()` is called before a disconnect callback is registered on the transport (problem is [here](https://github.com/ros/ros_comm/blob/a2ca01ecf31d677dd5c1b08cc6852bd0dac02b97/clients/roscpp/src/libros/service_manager.cpp#L266)). Because roscpp does connects asynchronously, the information that the connect failed may be lost if the signal is raised before callback registration.

Furthermore, `TransportTCP` tries to [signal](https://github.com/ros/ros_comm/blob/6980ae43fc50fee7227b5326db7b230c803777bb/clients/roscpp/src/libros/transport/transport_tcp.cpp#L467) in later reads that the connection is closed, but since most read callbacks silently ignore errors ([example](https://github.com/ros/ros_comm/blob/231a457bb7f3d19ca2e5eb3d3e7b0b0f73babea3/clients/roscpp/src/libros/connection.cpp#L386)), it does not help at all. The callback simply exits (and TransportTCP removes the socket from the PollSet). `ServiceServerLink::call()` waits forever for call completion.

I have prepared a unit test for the issue, which I will attach as a pull request immediately. I'm not sure about the fix. I think it would be best to fix both issues, even if fixing one would be enough to prevent the lockup.

PR #435, opened 2014-06-07
test_roscpp: added unit test for calling zombie services, see #434

xqms:
Zombie services are left behind by crashed nodes. Currently roscpp can
lock up when you call such a zombie service. This adds a unit test (which currently fails).

See #434 for details.

### DIFF
--- a/clients/roscpp/src/libros/connection.cpp
+++ b/clients/roscpp/src/libros/connection.cpp
@@ -329,13 +329,12 @@ void Connection::drop(DropReason reason)
     {
       dropped_ = true;
       did_drop = true;
-
-      drop_signal_(shared_from_this(), reason);
     }
   }
 
   if (did_drop)
   {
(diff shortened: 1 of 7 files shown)


dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
The CI fails for the new test which is intentional. I will wait to merge this until the PR also includes a fix for it.

But I don't expect to have time to work on this soon. So if someone would like to look into this that would be very welcome.

dirk-thomas:
Thank you for the patch as well as the test!

xqms:
Thanks for merging. Do you have an estimate on when this will hit the repositories?

dirk-thomas:
`ros_comm` has been released into _Indigo_ due to a different patch necessary. So it should be in the shadow-fixed repo within one day. Public depends fully on if there are regressions in any packages - if not it might be a couple of days.

PR #441, opened 2014-06-17
Added missing header needed for Android

garyservin:
This enables the use of readv() and writev() on Android.


dirk-thomas:
@ros-pull-request-builder

PR #443, opened 2014-06-17
rename variables within rosconsole macros

### LINKED ISSUE #442: rosconsole: cannot print variables named "enabled" or "loc"
I've got a funny one for you:

```
bool enabled = false;
ROS_ERROR("enabled: %d", enabled);
```

What would you expect? Certainly not this:

```
enabled: 1
```

The culprit is the `ROSCONSOLE_DEFINE_LOCATION()` macro [here](https://github.com/ros/ros_comm/blob/d6076920c99918d22580ddc26d7d4f8d0228ac59/tools/rosconsole/include/ros/console.h#L327). It defines local variables `enabled` and `loc`, which then shadow parameters which have the same name.

I propose changing the names of those variables to something less general (`__rosconsole__enabled`?).

Bug was found by my colleague Philipp Allgeuer (@pallgeuer).

dirk-thomas:
Resolves #442.

### DIFF
--- a/tools/rosconsole/include/ros/console.h
+++ b/tools/rosconsole/include/ros/console.h
@@ -326,20 +326,20 @@ ROSCONSOLE_DECL std::string formatToString(const char* fmt, ...);
 
 #define ROSCONSOLE_DEFINE_LOCATION(cond, level, name) \
   ROSCONSOLE_AUTOINIT; \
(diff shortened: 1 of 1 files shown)


wjwwood:
+1 for indigo and not hydro, don't forget to update the migration guide.

dirk-thomas:
Good point - done: http://wiki.ros.org/action/diff/indigo/Migration?action=diff&rev1=13&rev2=14

PR #446, opened 2014-06-19
Respawn delay

contradict:
Sometimes, especially for hardware drivers, it helps to pause before restarting a failed node.

This branch adds a configurable delay to the roslaunch respawn feature. If the delay is left unspecified, it defaults to 0, giving the same behavior as before. The value is configured in the node tag:
<node
     name="myname"
     pkg="flaky_hardware"
     type="unreliable_node"
    respawn="True"
    respawn_delay="5.0"
/>
respawn_delay has units seconds.

### DIFF
--- a/tools/roslaunch/src/roslaunch/core.py
+++ b/tools/roslaunch/src/roslaunch/core.py
@@ -426,7 +426,8 @@ class Node(object):
                  'filename']
 
     def __init__(self, package, node_type, name=None, namespace='/', \
-                 machine_name=None, args='', respawn=False, respawn_delay=0.0, \
+                 machine_name=None, args='', \
+                 respawn=False, respawn_delay=0.0, \
                  remap_args=None,env_args=None, output=None, cwd=None, \
                  launch_prefix=None, required=False, filename='<unknown>'):
         """
@@ -620,7 +621,8 @@ class Test(Node):
         to what it was initialized with, though the properties are the same
         """
         attrs = Node.xmlattrs(self)
-        attrs = [(a, v) for (a, v) in attrs if a not in ['respawn', 'respawn_delay']]
+        attrs = [(a, v) for (a, v) in attrs if a not in ['respawn', \
+                                                         'respawn_delay']]
         attrs.append(('test-name', self.test_name))
 
         if self.retry:
--- a/tools/roslaunch/src/roslaunch/core.py
+++ b/tools/roslaunch/src/roslaunch/core.py
@@ -419,13 +419,14 @@ class Node(object):
     """
     __slots__ = ['package', 'type', 'name', 'namespace', \
                  'machine_name', 'machine', 'args', 'respawn', \
+                 'respawn_delay', \
                  'remap_args', 'env_args',\
                  'process_name', 'output', 'cwd',
                  'launch_prefix', 'required',
                  'filename']
 
     def __init__(self, package, node_type, name=None, namespace='/', \
-                 machine_name=None, args='', respawn=False, \
+                 machine_name=None, args='', respawn=False, respawn_delay=0.0, \
                  remap_args=None,env_args=None, output=None, cwd=None, \
                  launch_prefix=None, required=False, filename='<unknown>'):
         """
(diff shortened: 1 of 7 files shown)


dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Please update the PR so that added/changed lines comply with PEP 8 and avoid any non-necessary whitespace changes in order to keep the diff minimal.

contradict:
Can you be more specific? I took a crack at this and found that most of what I could do was breaking long lines. Most of the existing code uses very long lines, so I'm not sure this is a good change. I found one line where I unnecessarily changed the indentation and a couple where I used space inside parens in a non-pep 8 way. Is there some aspect of the formatting you would particularly like me to fix?

tools/roslaunch/src/roslaunch/pmon.py
```diff
@@ -538,15 +558,15 @@ def _run(self):
             for p in procs:
                 try:
                     if not p.is_alive():
-                        logger.debug("Process[%s] has died, respawn=%s, required=%s, exit_code=%s",p.name, p.respawn, p.required, p.exit_code)
+                        logger.debug("Process[%s] has died, respawn=%s, required=%s, exit_code=%s",
+                                p.name,
+                                "True(%f)"%p.respawn_delay if p.respawn else p.respawn,
```
dirk-thomas: Please add spaces around `%`.

tools/roslaunch/src/roslaunch/pmon.py
```diff
@@ -582,18 +604,23 @@ def _run(self):
...
+            _respawn=[]
             for r in respawn: 
                 try:
                     if self.is_shutdown:
                         break
-                    printlog("[%s] restarting process"%r.name)
-                    # stop process, don't accumulate errors
-                    r.stop([])
-                    r.start()
+                    if r.should_respawn()<=0.0:
```
dirk-thomas: Please add spaces around `<=`.

tools/roslaunch/src/roslaunch/pmon.py
```diff
@@ -582,18 +604,23 @@ def _run(self):
...
+                        printlog("[%s] restarting process"%r.name)
```
dirk-thomas: Spaces around `%`.

tools/roslaunch/test/unit/test_roslaunch_pmon.py
```diff
@@ -427,6 +462,12 @@ def f():
         
         self.failIf(marker.marked, "pmon had to be externally killed")        
 
+        self.failIf(p3.spawn_count < 2, "process did not respawn")
+
+        self.failIf(p4.respawn_interval < p4.respawn_delay,
+                "Respawn delay not respected: %s %s"%(p4.respawn_interval,
```
dirk-thomas: Spaces around `%`.

tools/roslaunch/src/roslaunch/core.py
```diff
@@ -514,6 +518,7 @@ def xmlattrs(self):
             ('output', self.output),
             ('cwd', cwd_str), 
             ('respawn', self.respawn), #not valid on <test>
+            ('respawn_delay', self.respawn_delay), #not valid on <test>
```
dirk-thomas: A trailing comment should have two spaces before the `#` and one after.

dirk-thomas:
Please rebase the PR to the tip of the branch.

contradict:
I have so far been unable to duplicate this build failure on my system. I tried running the tests using the jenkins_tools package, setting up my own containerized environment, and running the test in my development environment. In all cases, the test which fails here runs successfully. The raw log of the test run above indeed looks different than a successful run, it emits the same error message as a successful run but much more frequently. What can I do to more accurately replicate the test environment?

dirk-thomas:
You can only run the tests locally (`make run_tests`). There is currently no good way to resemble the exact environment from the build farm.

I will check if I see anything obvious but might not have enough time to get to a useful result.

dirk-thomas:
@ros-pull-request-builder

wjwwood:
+1

dirk-thomas:
Squashed and merged in c171283e548cfdbe5bdafeb5559462beff529a89

dirk-thomas:
@contradict Can you please update the documentation in the wiki to mention the new attribute introduced in _Indigo_ (and post the link to the wiki changes here)?

contradict:
Update added here, I hope this is the correct place: http://wiki.ros.org/roslaunch/XML/node

dirk-thomas:
Thanks, I have updated the page to mention that the attribute has been introduced in Indigo: http://wiki.ros.org/action/diff/roslaunch/XML/node?action=diff&rev1=20&rev2=21

PR #448, opened 2014-06-23
fix typo in variable name set by param

andreasBihlmaier:


dirk-thomas:
Thank you for the pull request.

PR #450, opened 2014-06-26
roscpp: handle queue lengths of multiple publishers on the same topic better

xqms:
Currently, all `ros::Publisher` on a single topic share a message queue with the length of the queue determined by the first publisher. On advertising, `TopicManager` tries to find a previous `Publication` for the advertised topic and only adds the callbacks to it, discarding any other options the caller provided (see [here](https://github.com/ros/ros_comm/blob/c4acde3171368313ad72c559f1d8f0035217f23a/clients/roscpp/src/libros/topic_manager.cpp#L355)).

There are often cases in which this behavior does not make sense. Each "advertiser" of the topic probably provides a reasonable queue length for **his** messages. So the internal queue length of `Publication` should probably be the **sum** of all requested queue lengths.

This PR currently contains a unit test for that case (which currently fails). If you agree that the current behavior is not desired, I will try to bring in a patch for discussion.

### DIFF
--- a/clients/roscpp/include/ros/publication.h
+++ b/clients/roscpp/include/ros/publication.h
@@ -73,7 +73,7 @@ public:
   /**
    * \brief returns the max queue size of this publication
    */
-  inline size_t getMaxQueue() { return max_queue_; }
+  size_t getMaxQueue();
   /**
    * \brief Get the accumulated stats for this publication
    */
@@ -168,6 +168,8 @@ private:
 
   typedef std::vector<SubscriberCallbacksPtr> V_Callback;
   V_Callback callbacks_;
+
+  // Also protects max_queue_, since it depends on the registered callbacks
   boost::mutex callbacks_mutex_;
 
   V_SubscriberLink subscriber_links_;
--- a/clients/roscpp/include/ros/forwards.h
+++ b/clients/roscpp/include/ros/forwards.h
@@ -103,10 +103,12 @@ struct SubscriberCallbacks
   SubscriberCallbacks(const SubscriberStatusCallback& connect = SubscriberStatusCallback(),
                       const SubscriberStatusCallback& disconnect = SubscriberStatusCallback(),
                       const VoidConstPtr& tracked_object = VoidConstPtr(),
-                      CallbackQueueInterface* callback_queue = 0)
+                      CallbackQueueInterface* callback_queue = 0,
+                      unsigned int queue_size = 0)
   : connect_(connect)
   , disconnect_(disconnect)
   , callback_queue_(callback_queue)
+  , queue_size_(queue_size)
   {
     has_tracked_object_ = false;
     if (tracked_object)
@@ -121,6 +123,9 @@ struct SubscriberCallbacks
   bool has_tracked_object_;
   VoidConstWPtr tracked_object_;
   CallbackQueueInterface* callback_queue_;
+
+  // Keep track of requested queue size per callback
+  uint32_t queue_size_;
 };
 typedef boost::shared_ptr<SubscriberCallbacks> SubscriberCallbacksPtr;
 
--- a/clients/roscpp/src/libros/node_handle.cpp
+++ b/clients/roscpp/src/libros/node_handle.cpp
@@ -296,7 +296,7 @@ Publisher NodeHandle::advertise(AdvertiseOptions& ops)
   }
 
   SubscriberCallbacksPtr callbacks(new SubscriberCallbacks(ops.connect_cb, ops.disconnect_cb, 
-							   ops.tracked_object, ops.callback_queue));
+							   ops.tracked_object, ops.callback_queue, ops.queue_size));
 
   if (TopicManager::instance()->advertise(ops, callbacks))
   {
--- a/clients/roscpp/src/libros/publication.cpp
+++ b/clients/roscpp/src/libros/publication.cpp
@@ -102,8 +102,17 @@ void Publication::addCallbacks(const SubscriberCallbacksPtr& callbacks)
 {
   boost::mutex::scoped_lock lock(callbacks_mutex_);
 
+  // If a publisher requested an infinite queue size (max_queue=0),
+  // respect that
+  if(callbacks_.empty() || max_queue_ != 0)
+  {
+    if(callbacks->queue_size_ == 0)
+      max_queue_ = 0;
+    else
+      max_queue_ += callbacks->queue_size_;
+  }
+
   callbacks_.push_back(callbacks);
-  max_queue_ += callbacks->queue_size_;
 
   // Add connect callbacks for all current subscriptions if this publisher wants them
   if (callbacks->connect_ && callbacks->callback_queue_)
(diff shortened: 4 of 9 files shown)


xqms:
The second commit (4c0035d) makes the unit test pass for me.

dirk-thomas:
I am not sure if the sum of all publisher queue length arguments is the final solution for the problem. It definitely make the queue bigger and problems less likely to occur though.

But consider the following simple example:
- one publisher specifying a queue length of 1 and publishing with 10Hz
- one publisher specifying a queue length of 10 and publishing with 1 Hz

Using a single queue with a length of 11 will not give you the result both publishers would expect.

xqms:
> one publisher specifying a queue length of 1 and publishing with 10Hz

What exactly do you expect to happen here? The publishing is done in the `PollSet` thread, as fast as it can. So your example will probably publish at 11Hz from both Publishers, and I would say that is expected.

The only thing you do with the queue length in my experience is turning it up when the publisher thread can't keep up. You never turn it _down_ to limit the rate, since it can't guarantee that. It's not message frequency, it's more like a "load expectation" for the `PollSet` thread.

Perhaps we would have to change the fixed queue length to something like checking the message frequency and dropping messages if the frequency is too high, but that is a big API change. In my opinion quite impossible at this point ;-)

clients/roscpp/include/ros/publication.h
```diff
@@ -162,12 +162,14 @@ class ROSCPP_DECL Publication
   std::string datatype_;
   std::string md5sum_;
   std::string message_definition_;
-  size_t max_queue_;
   uint32_t seq_;
   boost::mutex seq_mutex_;
 
   typedef std::vector<SubscriberCallbacksPtr> V_Callback;
   V_Callback callbacks_;
+  size_t max_queue_;
```
dirk-thomas: Please avoid any non-crucial changes like moving code around.

dirk-thomas:
One way or the other: neither the max or sum of queue sizes will make the actual behavior match the publishers expectations.

The first publisher "expects" that he can queue messages if the subscriber is several seconds behind. But exactly that is not going to happen when the second publisher is present.

Increasing the queue size to the sum of all sizes will make it less likely to be a problem though. As long as we agree that this is only a partial workaround everything is good.

xqms:
> Increasing the queue size to the sum of all sizes will make it less likely to be a problem though. As long as we agree that this is only a partial workaround everything is good.

Yes, that's what I meant in my last comment. The queue size constraint does not mix well with the optimization of placing all messages of a topic in one queue and sending them out together.

I still can't think of a better solution if we do not want to get rid of this optimization and keep separate queues (and connections!) for each publisher. But I have removed the claim of doing it 'properly' from the PR title ;-)

dirk-thomas:
I will wait for our CI server to catch up and when it passes all tests it can be merged.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@xqms One case this patch seems to not address is that the `queue_size` might be _zero_ which indicates an unlimited queue. Can you pleae update the PR to address that?

dirk-thomas:
@jonbinney @mikeferguson FYI Can you imagine any cases where this patch might have negative side effects (especially for high level packages)?

wjwwood:
I see the benefit to doing a sum of the queue sizes over the current behavior or doing a max of the queue sizes. However, I am weary of this this change causing increased resource usage in very particular scenarios. I think we should bring in some outside review on this one.

I am +0 to this since the result is still a flawed concept at the expense of a considerable behavior change. This is a good example of why we need to break API with ROS 2.0.

jonbinney:
@dirk-thomas this could do funny things if people rely on the current behavior (of the queue-size being set with the first advertise). Consider the case where I have 10 different filter classes, each of which take an image and publish a filtered image. Then in some main loop of my node i might have:

```
switch (some_case) {
case 1:
   filter_class_1.publish_filtered_image(img)
   break;
case 2:
   filter_class_2.publish_filtered_image(img)
   break
. . ..
case 10:
   filter_class_10.publish_filtered_image(img)
}
```

To make things easy, someone may create a separate publisher in each filter class, but for the same topic and with the same queue_size. If that queue_size is 100, this PR would make the total queue size 1000, which could be a LOT of memory. That said, I don't know of any specific existing ROS package which follows this pattern.

As has been pointed out above, having a separate queue for each publisher seems to be the "Right" thing to do. This would mean that if I have two classes which publish on the same topic, they will behave the same whether i start them in separate roscpp nodes, or if i combine them into a single running roscpp node. (I realize there's an implementation effort with that)

I would consider switching to either sum of queue sizes or switching to separate queues to be pretty big changes, not just quick workarounds.

dirk-thomas:
@jonbinney This PR is exactly proposing this: using the sum of queue sizes. I am not sure if I understand you correctly - do you consider that to be a quick workaround which you don't like to see?

The increased memory usage is indeed a consequence. I think the increases memory usage is an acceptable trade off compared to a better behavior for a use case like this:
- publisher A has a queue size of 1 and publishes messages with 1 Hz
- publisher B has a queue size of 1000 and publishes messages with 1 kHz

Both work well when being used independently but when used together they don't work at all - at least when A is registered before B.

Having separate queues would indeed be the clean solution but due to the necessary refactoring that will very likely be out of scope.

jonbinney:
My point was just that the increase in memory usage when switching to the sum of queue sizes could be extreme. In my example, switching to sum of queue sizes would multiply memory usage by 10.

jonbinney:
(factor of 10)

wjwwood:
The other half of the behavior which isn't clear is that if the Publisher's are destroyed then the queue size will go back down, (@dirk-thomas will have to confirm that) so it might have fluctuating queue size which may also be bad if the queue is staying full.

I agree that changing the implementation to have a queue for each publisher is not in the scope of this issue nor something we could do for Indigo.

jonbinney:
Agreed, we should not try to do the split queues for indigo. My hunch is that it is best to stick with the current behavior for indigo, since even switching to sum of queue sizes could break things (cause lots of memory usage) and since sum of queue sizes isn't the "right" fix anyway.

Edit: "not try"

dirk-thomas:
The queue size should go down again when publishers are removed: https://github.com/ros/ros_comm/pull/450/files#diff-7e9728328ffc5c67a41abae998836bb2R126

The pending question is to handle `queue_size=0` correctly - see my previous comment for this (https://github.com/ros/ros_comm/pull/450#issuecomment-48651207).

dirk-thomas:
Whenever the queue size is reduced I think the rest of the code base won't correctly reduce the size of the queue: e.g. TransportSubscriberLink::enqueueMessage assumes to only remove one message before adding a new one while it will now need to remove old messages until the length of the queue is smaller then the updated max size.

xqms:
I seem to have messed up the commit anyway, sorry. Please wait a moment while I fix this mess.

xqms:
Okay, now it's ready for review. Sorry for the noise.

@dirk-thomas: I integrated your suggestion with the local variable.

> Whenever the queue size is reduced I think the rest of the code base won't correctly reduce the size of the queue: e.g. TransportSubscriberLink::enqueueMessage assumes to only remove one message before adding a new one while it will now need to remove old messages until the length of the queue is smaller then the updated max size.

Do you think throwing away old messages until the queue has the correct size is okay? I was kind of wondering about this and decided not to do it since I wasn't sure. Of course, the 'correct' solution is to throw away any messages published by _this_ publisher, but that's not possible without separate queues.

xqms:
Here's a commit for dropping messages until the queue size is okay again, in case we decide that is the better approach.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
We have discussed this intensively and do see a big chance for regressions. The problem is that the API is designed to allow the Publisher being used as a kind of singleton (per topic) so different code parts access the same instance - sometime just to get access to the publishers properties.

This patch would change the behavior of the API dramatically and will likely cause hard-to-detect regressions in several higher level packages. Also the Python client library would need be updated to provide similar behavior. Due to this uncertainty and big impact the PR won't be merged for the upcoming release of Indigo.

I do understand that the answer is not great for you since it means the current behavior (which is far from being good) will stick around in ROS for longer. But as the maintainer and having to guarantee stability of these core functionalities I am too uncomfortable to merge this change into Indigo.

I am not sure how the upgrade path can look like. We could consider to apply it for the next ROS distribution (j-turtle). Since the development cycle for it is very long it would give us the chance to identify problems and try to address them.

xqms:
I see, I understand your point about the regressions.

Is the consensus still that the best solution would be separate queues for each publisher? Maybe we can make that happen in time for j-turtle?

dirk-thomas:
Even the separate queue approach would break the currently documented behavior and how packages use the API. We could go for it but I am not sure if it is worth the effort.

For the new ROS 2 related development this is being addresses already so we might not spend a lot of time implementing this change for ROS 1.

xqms:
Alright, then I'm fine with closing this PR. The issue is easy to work around once you understand what is going on. And we can wait with a proper solution until ROS 2.

dirk-thomas:
Thank you for your effort anyway.

PR #452, opened 2014-06-29
implement ROS_IP=localhost behavior for roscpp, and for outbound rospy connections

codebot:
The wiki doc for ROS_IP:  http://wiki.ros.org/ROS/EnvironmentVariables    says that setting ROS_IP=localhost will "prevent remote components from being able to talk to your local component." This can be useful in some settings. However, this behavior was previously only implemented for inbound rospy topic connections and services. This pull request implements it for inbound and outbound roscpp connections, and for outbound rospy connections.

### DIFF
--- a/clients/roscpp/include/ros/transport/transport.h
+++ b/clients/roscpp/include/ros/transport/transport.h
@@ -141,6 +141,11 @@ protected:
    */
   bool isHostAllowed(const std::string &host);
 
+  /**
+   * \brief returns true if this transport is only allowed to talk to localhost
+   */
+  bool isOnlyLocalhostAllowed() { return only_localhost_allowed_; }
+
 private:
   bool only_localhost_allowed_;
   std::vector<std::string> allowed_hosts_;
--- a/clients/roscpp/include/ros/transport/transport.h
+++ b/clients/roscpp/include/ros/transport/transport.h
@@ -39,6 +39,7 @@
 #include <boost/function.hpp>
 #include <boost/shared_ptr.hpp>
 #include <boost/enable_shared_from_this.hpp>
+#include <vector>
 
 namespace ros
 {
@@ -54,7 +55,7 @@ class Header;
 class Transport : public boost::enable_shared_from_this<Transport>
 {
 public:
-  Transport() {}
+  Transport();
   virtual ~Transport() {}
 
   /**
@@ -134,8 +135,18 @@ protected:
   Callback disconnect_cb_;
   Callback read_cb_;
   Callback write_cb_;
+
+  /**
+   * \brief returns true if the transport is allowed to connect to the host passed to it.
+   */
+  bool isHostAllowed(const std::string &host);
+
+private:
+  bool only_localhost_allowed_;
+  std::vector<std::string> allowed_hosts_;
 };
 
 }
 
 #endif // ROSCPP_TRANSPORT_H
+
--- a/clients/roscpp/include/ros/transport/transport.h
+++ b/clients/roscpp/include/ros/transport/transport.h
@@ -139,12 +139,12 @@ protected:
   /**
    * \brief returns true if the transport is allowed to connect to the host passed to it.
    */
-  bool isHostAllowed(const std::string &host);
+  bool isHostAllowed(const std::string &host) const;
 
   /**
    * \brief returns true if this transport is only allowed to talk to localhost
    */
-  bool isOnlyLocalhostAllowed() { return only_localhost_allowed_; }
+  bool isOnlyLocalhostAllowed() const { return only_localhost_allowed_; }
 
 private:
   bool only_localhost_allowed_;
--- a/clients/roscpp/include/ros/transport/transport.h
+++ b/clients/roscpp/include/ros/transport/transport.h
@@ -154,4 +154,3 @@ private:
 }
 
 #endif // ROSCPP_TRANSPORT_H
-
--- a/clients/roscpp/src/libros/transport/transport.cpp
+++ b/clients/roscpp/src/libros/transport/transport.cpp
@@ -59,8 +59,6 @@ Transport::Transport()
   #endif
   if (ros_hostname_env && !strcmp(ros_hostname_env, "localhost"))
     only_localhost_allowed_ = true;
-  else if (ros_ip_env && !strcmp(ros_ip_env, "localhost"))
-    only_localhost_allowed_ = true;
   else if (ros_ip_env && !strncmp(ros_ip_env, "127.", 4))
     only_localhost_allowed_ = true;
 
--- a/clients/roscpp/src/libros/transport/transport.cpp
+++ b/clients/roscpp/src/libros/transport/transport.cpp
@@ -96,7 +96,7 @@ Transport::Transport()
   }
 }
 
-bool Transport::isHostAllowed(const std::string &host)
+bool Transport::isHostAllowed(const std::string &host) const
 {
   if (!only_localhost_allowed_)
     return true; // doesn't matter; we'll connect to anybody
@@ -111,7 +111,7 @@ bool Transport::isHostAllowed(const std::string &host)
       return true; // hooray
   }
   ROS_WARN("ROS_HOSTNAME / ROS_IP is set to only allow local connections, so "
-           "a requested connection to %s is being rejected.", host.c_str());
+           "a requested connection to '%s' is being rejected.", host.c_str());
   return false; // sadness
 }
 
--- a/clients/roscpp/src/libros/transport/transport.cpp
+++ b/clients/roscpp/src/libros/transport/transport.cpp
@@ -1,0 +1,117 @@
+/*
+ * Software License Agreement (BSD License)
+ *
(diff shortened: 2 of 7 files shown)


gerkey:
Anywhere that we check for `ROS_IP=='localhost'`, we should also check for `ROS_HOSTNAME=='localhost'`.  They're used interchangeably (which is confusing, but then it's always been confusing to support two variables for this purpose).

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -211,6 +211,28 @@ void TransportTCP::setKeepAlive(bool use, uint32_t idle, uint32_t interval, uint
...
+  #ifdef _MSC_VER
+    _dupenv_s(&ros_ip_env, NULL, "ROS_IP");
+  #else
+    ros_ip_env = getenv("ROS_IP");
+  #endif
+  if (ros_ip_env && !strcmp(ros_ip_env, "localhost"))
+  {
+    // we must verify that the requested connection is to localhost
+    char our_hostname[256] = {0};
+    gethostname(our_hostname, sizeof(our_hostname)-1);
```
gerkey: What about the case where we're asked to connect to a node via one of our own non-loopback IP addresses?  E.g., we're in `'localhost'` mode and we hear about a publisher at `'192.168.1.100'`, which is our own address?  That connection should be allowed.  I think that we need to iterate over all of our own IP addresses, plus our hostname (maybe fully-qualified and not?), plus the special cases of `'localhost'`, `'127.*'`, etc.  Basically, we need to match on any host/address string that legitimately matches our machine.

Looking at the Python version, it seems that we need some form of `rosgraph.network.is_local_address()`.

Same for the UDP case below.

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -211,6 +211,28 @@ void TransportTCP::setKeepAlive(bool use, uint32_t idle, uint32_t interval, uint
 
 bool TransportTCP::connect(const std::string& host, int port)
 {
+  char *ros_ip_env = NULL;
```
gerkey: It seems wasteful to redo this lookup logic on every attempted connect.  Perhaps somewhere else, during initialization, we can build up the list of host/address strings that we're willing to match, then here we just match against that list.  

Same for the UDP case below.

dirk-thomas:
According to the docs (http://wiki.ros.org/ROS/EnvironmentVariables#ROS_IP.2BAC8-ROS_HOSTNAME) `ROS_IP` should only be set to an IP address:

> Use ROS_IP if you are specifying an IP address, and ROS_HOSTNAME if you are specifying a host name.

So I don't think that setting `ROS_IP=localhost` is valid. The patch should only check the `ROS_HOSTNAME` instead.

clients/roscpp/src/libros/transport/transport_tcp.cpp
codebot: The tradeoff with caching the valid hostnames is that (I assume) it will
break ABI compatibility. Maybe that's OK. I wasn't sure how much that
mattered.

In any event, this branch only executes during the connect phase, and only
on nodes with ROS_HOSTNAME set to localhost, so I don't see it as much of a
performance issue, but i agree it is a code ugliness issue.

dirk-thomas:
@ros-pull-request-builder

clients/roscpp/src/libros/transport/transport.cpp
```diff
@@ -0,0 +1,119 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2014, Open Source Robotics Foundation
```
dirk-thomas: `Open Source Robotics Foundation, Inc.`

clients/roscpp/include/ros/transport/transport.h
```diff
@@ -134,8 +135,23 @@ class Transport : public boost::enable_shared_from_this<Transport>
...
+
+  /**
+   * \brief returns true if the transport is allowed to connect to the host passed to it.
+   */
+  bool isHostAllowed(const std::string &host);
+
+  /**
+   * \brief returns true if this transport is only allowed to talk to localhost
+   */
+  bool isOnlyLocalhostAllowed() { return only_localhost_allowed_; }
```
dirk-thomas: Should those two methods be `const`?

clients/roscpp/src/libros/transport/transport.cpp
```diff
@@ -0,0 +1,119 @@
...
+  #ifdef _MSC_VER
+    _dupenv_s(&ros_ip_env, NULL, "ROS_IP");
+    _dupenv_s(&ros_hostname_env, NULL, "ROS_HOSTNAME");
+  #else
+    ros_ip_env = getenv("ROS_IP");
+    ros_hostname_env = getenv("ROS_HOSTNAME");
+  #endif
+  if (ros_hostname_env && !strcmp(ros_hostname_env, "localhost"))
+    only_localhost_allowed_ = true;
+  else if (ros_ip_env && !strcmp(ros_ip_env, "localhost"))
```
dirk-thomas: Since `ROS_IP` must contain a valid IP address when being set I think we should not check it for `localhost`.

Also applies to Python code below.

clients/roscpp/src/libros/transport/transport.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return true; // ipv4 localhost
+  // now, loop through the list of valid hostnames and see if we find it
+  for (std::vector<std::string>::iterator it = allowed_hosts_.begin(); 
+       it != allowed_hosts_.end(); ++it)
+  {
+    if (host == *it)
+      return true; // hooray
+  }
+  ROS_WARN("ROS_HOSTNAME / ROS_IP is set to only allow local connections, so "
+           "a requested connection to %s is being rejected.", host.c_str());
```
dirk-thomas: Please put single quotes around the %s to make it more readable:

> .. connection to '%s' is being ...

codebot: Thank you; these changes are now incorporated into the PR.

clients/roscpp/include/ros/transport/transport.h
```diff
@@ -134,8 +135,23 @@ class Transport : public boost::enable_shared_from_this<Transport>
...
+
+private:
+  bool only_localhost_allowed_;
+  std::vector<std::string> allowed_hosts_;
 };
 
 }
 
 #endif // ROSCPP_TRANSPORT_H
+
```
dirk-thomas: The file already has a trailing newline. Please remove the additional one.

codebot: done

dirk-thomas:
@ros-pull-request-builder

clients/roscpp/src/libros/transport/transport.cpp
```diff
@@ -0,0 +1,117 @@
...
+    _dupenv_s(&ros_ip_env, NULL, "ROS_IP");
+    _dupenv_s(&ros_hostname_env, NULL, "ROS_HOSTNAME");
+  #else
+    ros_ip_env = getenv("ROS_IP");
+    ros_hostname_env = getenv("ROS_HOSTNAME");
+  #endif
+  if (ros_hostname_env && !strcmp(ros_hostname_env, "localhost"))
+    only_localhost_allowed_ = true;
+  else if (ros_ip_env && !strncmp(ros_ip_env, "127.", 4))
+    only_localhost_allowed_ = true;
```
dirk-thomas, line 63: This line also needs to work with IPv6 addresses.

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -377,7 +384,6 @@ bool TransportTCP::listen(int port, int backlog, const AcceptCallback& accept_cb
     return false;
   }
 
-
```
dirk-thomas, line 380: Please revert the whitespace change.

clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -518,6 +518,16 @@ def connect(self, dest_addr, dest_port, endpoint_id, timeout=None):
         @type  timeout: float
         @raise TransportInitError: if unable to create connection
         """
+        # first make sure that if ROS_IP=localhost, we will not attempt
```
dirk-thomas: Sorry, for bother about more details... but the comment should refer to ROS_HOSTNAME.

dirk-thomas:
@ros-pull-request-builder

wjwwood:
+1 for Indigo

dirk-thomas:
Squashed and merged in 8713a1e22f69535a88b1a2f9c3b6a4c73abf6cf1

dirk-thomas:
@codebot9000 Can you please update the documentation in the wiki to clarify the previous behavior as well as the new behavior as of _Indigo_ (and post the link to the wiki changes here)?

PR #453, opened 2014-06-30
fix finding specific version of PythonLibs with CMake 3

dirk-thomas:
@wjwwood Please review.


wjwwood:
+1

NikolausDemmel:
Can we backport this to groovy/hyrdo?

dirk-thomas:
Yes, it will be backported.

PR #456, opened 2014-07-03
Possibility to add multiple tests with different arguments

ct2034:
I needed to add multiple tests with different arguments in one CMakeLists.txt. This led to a conflict as these had the same target name, due to the identical test-file-name. This includes arguments to the target name, if those are stated.

### DIFF
--- a/tools/rostest/cmake/rostest-extras.cmake.em
+++ b/tools/rostest/cmake/rostest-extras.cmake.em
@@ -38,9 +38,8 @@ function(add_rostest file)
 
   string(REPLACE "/" "_" _testname ${_testname})
   if(ARGN) # inserting parameters in name, if set
-    string(REPLACE ";" "__" argnstring "${ARGN}")
-    string(REGEX REPLACE "ARGS" "" argnstring ${argnstring})
-    string(REPLACE ":=" "_" argnstring ${argnstring})
+    string(REGEX REPLACE "[;:=.]|ARGS" "_" argnstring "${ARGN}")
+    string(REPLACE " " "-" argnstring "${argnstring}")
     string(REPLACE "." "_" _testname ${_testname})
     set(_testname "${_testname}_${argnstring}.test")
     message("-- Added testname WITH arguments: ${_testname}")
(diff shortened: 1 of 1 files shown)


tools/rostest/cmake/rostest-extras.cmake.em
```diff
@@ -37,6 +37,14 @@ function(add_rostest file)
   rostest__strip_prefix(_testname "${PROJECT_BINARY_DIR}/")
 
   string(REPLACE "/" "_" _testname ${_testname})
+  if(ARGN) # inserting parameters in name, if set
```
dirk-thomas: Why you are using `ARGN` here. It will contain all arguments beside the test file (including `WORKING_DIRECTORY`).

dirk-thomas:
Please provide some example CMake lines how you intend to use this to help understanding your patch.

ct2034:
#### About ARGN:

ARGN is only set if arguments are passed. This is the output of catkin_make on the test file mentioned below:

```
-- +++ processing catkin package: 'test_cmake'
-- ==> add_subdirectory(my_navigation_test/test_cmake)
ARGN: 
ARGN: ARGS;test1:=true
-- Added testname WITH arguments: launch_the_test___test1__true.test
ARGN: ARGS;test1:=true;test2:=true;test3:=true
-- Added testname WITH arguments: launch_the_test___test1__true_test2__true_test3__true.test
```

This line was added to rostest-extras.cmake.em: 'message("ARGN: ${ARGN}")'

#### Example code:

The following code is from my testpackage: https://github.com/ct2034/test_cmake/blob/master/CMakeLists.txt

```
add_rostest(launch/the.test) # <- this leads to an empty ARGN - see above
add_rostest(launch/the.test ARGS  test1:=true)
add_rostest(launch/the.test  ARGS test1:=true test2:=true  test3:=true)
```

This produces the output above. Without this patch it would fail, beacuse all the targets would be called 'launch_the_test'.

dirk-thomas:
The function uses `cmake_parse_arguments` to parse the passed arguments. What you want to use is `_rostest_ARGS`. Otherwise when passing `WORKING_DIRECTORY somepath` it would be using that too.

ct2034:
By the way: the thing I actually need the change for is this one: https://github.com/ct2034/my_navigation_test/blob/hydro_dev/CMakeLists.txt

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
The patch should be applied to the latest branch which is _indigo-devel_. Therefore closing this PR in favor of #462. It might be cherry-picked to previous distro but for new features that depends on the need and chance of introducing regressions.

PR #462, opened 2014-07-10
Rostest target with args

dirk-thomas:
Replaces #456.

@ct2034 FYI

### DIFF
--- a/tools/rostest/cmake/rostest-extras.cmake.em
+++ b/tools/rostest/cmake/rostest-extras.cmake.em
@@ -37,6 +37,19 @@ function(add_rostest file)
   rostest__strip_prefix(_testname "${PROJECT_BINARY_DIR}/")
 
   string(REPLACE "/" "_" _testname ${_testname})
+
+  # to support registering the same test with different ARGS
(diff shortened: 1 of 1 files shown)


tfoote:
+1

wjwwood:
+1 for Indigo and for backports to Hydro

dirk-thomas:
I updated the patch to actually only consider the passed `ARGS` rather then `ARGN`. It now also passes the correct xunit file. See ff1de571b61d0e0653e5297f49158789fa5b87f1.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ct2034 Please try the latest version of the patch and verify that it works for you as expected.

ct2034:
@dirk-thomas I tried it with a test project. It is working as expected. :+1: :)
But is it possible to include these changes also to hydro? Because the main project I need this for is still running hydro.

dirk-thomas:
Cherry-picked in 8761d8141b1abebe1b5fac9941390ec8b9160f4c. Will be available as 1.10.11 soonish.

PR #465, opened 2014-07-16
Fixed bug when ifa_addr is NULL

jczarnowski:
For example where the interface is socketcan.


dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Regression of #452.

PR #466, opened 2014-07-18
Make rostest use a random master port

### LINKED ISSUE #141: enable rostest to find unused rosmaster port  so that multiple test can run at the same time  (ros ticket #3826)
It would be nice if rostest is able to find unused rosmaster port if 22422 is used by another rostest process, so that multiple test can run at the same time. Currently we got following error.

  File "/opt/ros/electric/stacks/ros_comm/tools/roslaunch/src/roslaunch/launch.py", line 375, in _launch_master
      raise RLException("roscore cannot run as another roscore/master is already running. \nPlease kill other roscore/master processes before relaunching.\nThe ROS_MASTER_URI is %s"%(m.uri))

---

trac data:
- Owner: **kwc**
- Reporter: **k-okada**
- Reported at: **Sun Jan 29 23:24:32 2012**
- URL: https://code.ros.org/trac/ros/ticket/3826

trainman419:
Modify rostest so that it uses the rosmaster API to start an in-process master with a random port number, and then use the resulting master URI for the following launches.

This allows me to run tests in parallel (with the appropriate make -j flag), and speeds up tests on my quad-core machine by a factor of 3.5.

This fixes #141

### DIFF
--- a/tools/rostest/scripts/rostest
+++ b/tools/rostest/scripts/rostest
@@ -31,23 +31,6 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-import socket
-import time
 from rostest import rostestmain
-
-message_printed = False
-while True:
-    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    try:
-        s.bind(('127.0.0.1', 22421))  # choose port close to test roscore
-        break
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Please recreate the PR against the _indigo-devel_ branch.

Any feature addition like this will first be merged to the latest development branch. Considering it for back port might be an option but in this specific case is not very likely.

trainman419:
Opened #467

dirk-thomas:
Closing in favor of #467.

PR #467, opened 2014-07-18
Make rostest use a random master port

### LINKED ISSUE #141: enable rostest to find unused rosmaster port  so that multiple test can run at the same time  (ros ticket #3826)

trainman419:
Opening #466 against indigo

Modify rostest so that it uses the rosmaster API to start an in-process master with a random port number, and then use the resulting master URI for the following launches.

This allows me to run tests in parallel (with the appropriate make -j flag), and speeds up tests on my quad-core machine by a factor of 3.5.

This fixes #141

### DIFF
--- a/tools/rostest/src/rostest/rostest_parent.py
+++ b/tools/rostest/src/rostest/rostest_parent.py
@@ -44,14 +44,16 @@ import roslaunch.xmlloader
 
 import roslaunch.parent
 
+from rosmaster.master import Master
+
 class ROSTestLaunchParent(roslaunch.parent.ROSLaunchParent):
 
-    def __init__(self, config, roslaunch_files, port):
+    def __init__(self, config, roslaunch_files):
         if config is None:
             raise Exception("config not initialized")
         # we generate a run_id for each test
         run_id = roslaunch.core.generate_run_id()
-        super(ROSTestLaunchParent, self).__init__(run_id, roslaunch_files, is_core=True, port=port, is_rostest=True)
+        super(ROSTestLaunchParent, self).__init__(run_id, roslaunch_files, is_core=False, is_rostest=True)
         self.config = config
         
     def _load_config(self):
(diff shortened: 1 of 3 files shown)


trainman419:
I've tested this against hydro; I haven't tested it against indigo.

tools/rostest/src/rostest/rostest_parent.py
```diff
@@ -44,14 +44,16 @@
 
 import roslaunch.parent
 
+from rosmaster.master import Master
+
 class ROSTestLaunchParent(roslaunch.parent.ROSLaunchParent):
 
-    def __init__(self, config, roslaunch_files, port):
+    def __init__(self, config, roslaunch_files):
```
dirk-thomas, line 51: Please keep the `port` as an optional keyword argument which defaults to `0` or `None`. That still allows it to pass in a custom port.

tools/rostest/src/rostest/rostest_parent.py
```diff
@@ -63,6 +65,9 @@ def setUp(self):
         initializes self.config and xmlrpc infrastructure
         """
         self._start_infrastructure()
+        self.master = Master(port=0)
```
dirk-thomas, line 68: Use the passed in `port` from the constsructor here.

tools/rostest/src/rostest/rostest_parent.py
```diff
@@ -44,14 +44,16 @@
 
 import roslaunch.parent
 
+from rosmaster.master import Master
```
dirk-thomas, line 47: Since this package now includes code directly from the `rosmaster` package please add it as a `run_depend` to the `package.xml` file.

dirk-thomas:
Besides the small comments the patch locks great and works for me.

The performance gain was much smaller for me - but that will highly depend on the content of the workspace.

dirk-thomas:
Closing in favor of #468 in order to get it released before the upcoming last rebuild before Indigo.

PR #468, opened 2014-07-18
make rostest use a random master port to allow parallel testing

dirk-thomas:
Supersedes #467.


trainman419:
I've been thinking about this some more since I opened it. It is probably cleaner to add a self.master.stop() in the test teardown fubction, so that we don't leak open ports and masters if there are multiple tests in a file.

Issue #472, opened 2014-07-22
roslaunch: Allow roslaunch XML to be passed on stdin


mikepurvis:
The intent would be enabling templated, xacro'd, or otherwise generated roslaunch, without a temporary file in between. Key design decision would be whether a sentinel is required as the file argument, or if it assumes to look on stdin in the absence of a supplied filename. Eg,

```
rosrun my_pkg create_launch | roslaunch  # assume to use stdin
rosrun my_pkg create_launch | roslaunch -  # sentinel value for file argument
```

Happy to implement this, with the maintainer's blessing to either of the above approaches.

dirk-thomas:
I would think that having a sentinel value makes sense. I lot of users might be confused when calling it without arguments - currently they get a useful usage message.

It would be great if you can provide a PR for this feature.

PR #473, opened 2014-07-23
Support roslaunch XML on stdin.

mikepurvis:
This addresses #472.

My apologies for the whitespace modifications; if you feel strongly, I can revise the PR to reverse them. pep8 hates trailing whitespace, so my editor is set to strip it on save. Functionally, I believe this where it needs to be. Some example tests:

``` bash
$ roslaunch

roslaunch: error: you must specify at least one input file

$ roslaunch -
... logging to /home/administrator/.ros/log/edd1c682-1272-11e4-8714-001c42910917/roslaunch-vm-precise2-3063.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

Passed '-' as file argument, attempting to read roslaunch XML from stdin.
^CTraceback (most recent call last):
  File "/home/administrator/roslaunch_ws/devel/bin/roslaunch", line 5, in <module>
    exec(fh.read())
  File "<string>", line 35, in <module>
  File "<string>", line 279, in main
KeyboardInterrupt

$ roslaunch foo
[foo] is not a launch file name
The traceback for the exception was written to the log file

$ roslaunch rosserial_server socket.launch
[ launch occurs successfully ]

$ echo '<blah></blah>' | roslaunch -
... logging to /home/administrator/.ros/log/35e04d4a-1273-11e4-bb24-001c42910917/roslaunch-vm-precise2-3250.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

Passed '-' as file argument, attempting to read roslaunch XML from stdin.
... 14 bytes read successfully.

... loading XML
Launch string: <blah></blah>

Exception: Invalid roslaunch XML syntax: no root <launch> tag
The traceback for the exception was written to the log file

$ echo '<launch><node pkg="rosserial_server" type="socket_node" name="foo" /></launch>' | roslaunch -
... logging to /home/administrator/.ros/log/50488b34-1273-11e4-8288-001c42910917/roslaunch-vm-precise2-3262.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

Passed '-' as file argument, attempting to read roslaunch XML from stdin.
... 79 bytes read successfully.

... loading XML
Added node of type [rosserial_server/socket_node] in namespace [/]
started roslaunch server http://vm-precise2:53924/
[ launch continues successfully ]
```

I've avoided modifying the `--help` output to mention this functionality, since it's a bit esoteric and the command-line help should stay simple and focused. I'll add a comment to the roslaunch wiki page once we're merged.

### DIFF
--- a/tools/roslaunch/src/roslaunch/parent.py
+++ b/tools/roslaunch/src/roslaunch/parent.py
@@ -67,7 +67,7 @@ class ROSLaunchParent(object):
     and then starting up any remote processes. The __main__ method
     delegates most of runtime to ROSLaunchParent.
 
-    This must be called from the Python Main thread due to signal registration.
+    This must be called from the Python Main thread due to signal registration.    
     """
 
     def __init__(self, run_id, roslaunch_files, is_core=False, port=None, local_only=False, process_listeners=None,
--- a/tools/roslaunch/src/roslaunch/parent.py
+++ b/tools/roslaunch/src/roslaunch/parent.py
@@ -67,10 +67,11 @@ class ROSLaunchParent(object):
     and then starting up any remote processes. The __main__ method
     delegates most of runtime to ROSLaunchParent.
 
-    This must be called from the Python Main thread due to signal registration.    
+    This must be called from the Python Main thread due to signal registration.
     """
 
-    def __init__(self, run_id, roslaunch_files, is_core=False, port=None, local_only=False, process_listeners=None, verbose=False, force_screen=False, is_rostest=False):
+    def __init__(self, run_id, roslaunch_files, is_core=False, port=None, local_only=False, process_listeners=None,
+            verbose=False, force_screen=False, is_rostest=False, roslaunch_strs=None):
         """
         @param run_id: UUID of roslaunch session
         @type  run_id: str
@@ -101,6 +102,7 @@ class ROSLaunchParent(object):
         self.process_listeners = process_listeners
         
         self.roslaunch_files = roslaunch_files
+        self.roslaunch_strs = roslaunch_strs
         self.is_core = is_core
         self.is_rostest = is_rostest
         self.port = port
@@ -118,7 +120,8 @@ class ROSLaunchParent(object):
         self.config = self.runner = self.server = self.pm = self.remote_runner = None
 
     def _load_config(self):
-        self.config = roslaunch.config.load_config_default(self.roslaunch_files, self.port, verbose=self.verbose)
+        self.config = roslaunch.config.load_config_default(self.roslaunch_files, self.port,
+                roslaunch_strs=self.roslaunch_strs, verbose=self.verbose)
 
         # #2370 (I really want to move this logic outside of parent)
         if self.force_screen:
--- a/tools/roslaunch/src/roslaunch/__init__.py
+++ b/tools/roslaunch/src/roslaunch/__init__.py
@@ -194,8 +194,9 @@ def _validate_args(parser, options, args):
 
     elif len(args) == 0:
         parser.error("you must specify at least one input file")
-    elif [f for f in args if not os.path.exists(f)]:
-        parser.error("The following input files do not exist: %s"%f)
+    elif args != ['-']:
+        if [f for f in args if not os.path.exists(f)]:
+            parser.error("The following input files do not exist: %s"%f)
 
     if len([x for x in [options.node_list, options.find_node, options.node_args, options.ros_args] if x]) > 1:
         parser.error("only one of [--nodes, --find-node, --args --ros-args] may be specified")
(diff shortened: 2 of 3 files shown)


dirk-thomas:
The patch should indeed avoid white space changes - otherwise merging changes between branches becomes more painful. Please remove those from the PR.

Also please add the option to the help. Otherwise it is not discoverable by users.

mikepurvis:
Okay, updated to remove the remaining whitespace changes. I've also noted the stdin support in the usage string, as well as making it clearer how the utility supports being passed multiple roslaunch files. The help output now appears as so:

``` bash
$ roslaunch -h
       roslaunch [options] <filename> [<filename>...] [arg_name:=value...]

If <filename> is a single dash ('-'), launch XML is read from standard input.

Options:
  --files               Print list files loaded by launch file, including
                        launch file itself
  --args=NODE_NAME      Print command-line arguments for node
  --nodes               Print list of node names in launch file
  --find-node=NODE_NAME
                        Find launch file that node is defined in
  -c NAME, --child=NAME
                        Run as child service 'NAME'. Required with -u
  --local               Do not launch remote nodes
  --screen              Force output of all local nodes to screen
  -u URI, --server_uri=URI
                        URI of server. Required with -c
  --run_id=RUN_ID       run_id of session. Required with -c
  --wait                wait for master to start before launching
  -p PORT, --port=PORT  master port. Only valid if master is launched
  --core                Launch core services only
  --pid=PID_FN          write the roslaunch pid to filename
  -v                    verbose printing
  --dump-params         Dump parameters of all roslaunch files to stdout
  --skip-log-check      skip check size of log folder
  --ros-args            Display command-line arguments for this launch file
  --disable-title       Disable setting of terminal title
```

dirk-thomas:
Please add new arguments to the end of the list since otherwise it breaks code (which might not use keyword but positional arguments).

tools/roslaunch/src/roslaunch/parent.py
```diff
@@ -67,10 +67,11 @@ class ROSLaunchParent(object):
     and then starting up any remote processes. The __main__ method
     delegates most of runtime to ROSLaunchParent.
 
-    This must be called from the Python Main thread due to signal registration.    
+    This must be called from the Python Main thread due to signal registration.
```
dirk-thomas: Please revert this white space change.

tools/roslaunch/src/roslaunch/parent.py
```diff
@@ -70,7 +70,8 @@ class ROSLaunchParent(object):
     This must be called from the Python Main thread due to signal registration.    
     """
 
-    def __init__(self, run_id, roslaunch_files, is_core=False, port=None, local_only=False, process_listeners=None, verbose=False, force_screen=False, is_rostest=False):
+    def __init__(self, run_id, roslaunch_files, is_core=False, port=None, local_only=False, process_listeners=None,
+            verbose=False, force_screen=False, is_rostest=False, roslaunch_strs=None):
```
mikepurvis, line 74: This is the only place where a function signature has changed; the new parameter has been appended. Would you like the keyword ordering in the function calls to match the signatures?

mikepurvis:
I believe that should be good to go now; let me know if you'd like further changes.

dirk-thomas:
I wasn't aware that `roslaunch` could be invoked with multiple `filenames`. In this case using `-` for one of them seems to not work - it is only supported when passing a single file.

This should be either clarified in the usage line or the code should be extended to support it even together with other filenames.

mikepurvis:
All right, now works alongside other files. The only proviso is that the stdin-supplied XML always comes at the end, regardless of where the dash appeared in the args list. Fixing this would be a significant undertaking, and roslaunch makes no guarantees about launch order anyway.

When this is satisfactory, would you like the commits crushed?

dirk-thomas:
@ros-pull-request-builder

tools/roslaunch/src/roslaunch/parent.py
dirk-thomas, line 74: I just expected the keywords to be in order and didn't check. Its good as it is.

dirk-thomas:
LGTM +1.

@tfoote @wjwwood Please review too.

tfoote:
+1

wjwwood:
+1

dirk-thomas:
@mikepurvis Can you please update the wiki page for roslaunch to include a description of this new feature (and mark the addition with "new-in-Indigo") and post the diff link here?

mikepurvis:
Done! http://wiki.ros.org/roslaunch/Commandline%20Tools#Launch_syntax

dirk-thomas:
Awesome - thank you for the pull request, the iterations on it and the fast doc update!

Issue #474, opened 2014-07-25
Node is remapped but parameter is not remapped


tarukosu:
Node name can be remapped, but parameter is not remapped.
That's why parameter does not work.

This causes trouble when remapping a node whose node name and topic name are the same.

dirk-thomas:
Can you please provide a minimal reproducible example which does not work for you and describe which parameter you expect and what you actually get?

tarukosu:
This is a sample launch file.

```
<launch>
  <remap from="/joy" to="/joy_remap"/>
  <node pkg="joy" type="joy_node" name="joy" output="screen">
    <param name="dev" type="string" value="/dev/input/js1" />
  </node>
</launch>
```

The node name is /joy_remap but the parameter name is /joy/dev.
So I expect the parameter name is also remaped to /joy_remap/dev or node name is not remaped.

dirk-thomas:
The behavior is consistent across all current ROS distros: Groovy, Hydro, Indigo.

Therefore it does not look like a recent regression. Also all remapping related unit tests pass. It might be that you expect a behavior which is not implemented that way.

Anyway I won't have time in the near future to look into in depth. Please consider investigating it  and if necessary provide a pull request.

tarukosu:
I see.
But I think this is a bag of roslaunch because node name and parameter should be remapped in the same way.

However, we can avoid this problem by changing node name , so it is not urgent.
If I have time to improve, I will send a pull request.
Thank you.

PR #478, opened 2014-07-29
workaround for issue #474

k-okada:
this is workaround for issue #474
However I wrote this patch without knowing what is the expected behavior of using grooup/remap/param , is there any specification on this?


dirk-thomas:
Please create patches against the latest development branch only (if it applied to that): indigo-devel

The roslaunch wiki page contains all the information available.

k-okada:
thanks create new PR on #489

PR #479, opened 2014-07-29
workaround for issue #474

k-okada:
this is workaround for issue #474
However I wrote this patch without knowing what is the expected behavior of using grooup/remap/param , is there any specification on this?

> The roslaunch wiki page contains all the information available.

yes, but it is not explicitly written that remap changes node name but not affect param , so if you remap node name like follows:

```
<launch>
  <remap from="/joy" to="/joy_remap"/>
  <node pkg="joy" type="joy_node" name="joy" output="screen">
    <param name="dev" type="string" value="/dev/input/js1" />
  </node>
</launch>
```

you have to change param name setting as 

```
    <param name="/joy_remap/dev" type="string" value="/dev/input/js1" />
```

### DIFF
--- a/tools/roslaunch/src/roslaunch/xmlloader.py
+++ b/tools/roslaunch/src/roslaunch/xmlloader.py
@@ -40,6 +40,7 @@ from __future__ import print_function
 
 import itertools
 import sys
+import re
 import traceback
 import logging
 
@@ -419,6 +420,8 @@ class XmlLoader(loader.Loader):
                     # strip leading ~, which is optional/inferred
                     pkey = pkey[1:]
                 pkey = param_ns.ns + pkey
+                for remap in remap_context.remap_args():
+                    pkey = re.sub('^/?'+remap[0], remap[1],pkey)
                 ros_config.add_param(Param(pkey, p.value), verbose=verbose)
                     
             if not is_test:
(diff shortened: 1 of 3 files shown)


dirk-thomas:
@ros-pull-request-builder

tools/roslaunch/src/roslaunch/xmlloader.py
```diff
@@ -419,6 +420,8 @@ def _node_tag(self, tag, context, ros_config, default_machine, is_test=False, ve
                     # strip leading ~, which is optional/inferred
                     pkey = pkey[1:]
                 pkey = param_ns.ns + pkey
+                for remap in remap_context.remap_args():
+                    pkey = re.sub('^/?'+remap[0], remap[1],pkey)
```
dirk-thomas, line 424: I don't think this is the right place to modify the behavior. The namespace of the surrounding node `param_ns.ns` should already contain the remapped name instead.

dirk-thomas:
According to the docs (http://wiki.ros.org/roslaunch/XML/remap) it must also be patched for the group tag.

k-okada:
Can you write correct behavior for remap/group tag as test code?

tools/roslaunch/test/xml/test-params-valid.xml
tools/roslaunch/test/unit/test_xmlloader.py

dirk-thomas:
I am sorry, but I currently don't have time to work on this feature.

dirk-thomas:

Issue #480, opened 2014-07-29
rostopic echo of array indices fails


trainman419:
This is part of the root cause for https://github.com/ros-visualization/rqt_common_plugins/issues/261

When publishing an array message (on topic `/array`):

```
Header header
float32[] values
```

`rostopic echo /array/values[0]` fails, but `rostopic echo /array/values` succeeds. 

This worked on Hydro, and fails on Indigo. It looks like the underlying bug here is a change somewhere in rostopic's `_get_topic_type()`

trainman419:
I'm able to reproduce this with the `test_pub.py` node in this package: https://github.com/trainman419/test_publishers

dirk-thomas:
Please try #481 and check if it also fixes the problem you have with rqt_plot.

trainman419:
I'm not near my Indigo machine at the moment; I'll try to test this when I get home tonight.

trainman419:
Confirmed; this fixes rostopic echo for array members, and rqt_plot for the same cases, although I do get some rather ugly stacktraces whlie typing that don't happen in the hydro version: 

```
Traceback (most recent call last):
  File "/home/htpc/catkin_ws/src/rqt_common_plugins/rqt_plot/src/rqt_plot/plot_widget.py", line 208, in on_topic_edit_textChanged
    plottable, message = is_plottable(topic_name)
  File "/home/htpc/catkin_ws/src/rqt_common_plugins/rqt_plot/src/rqt_plot/plot_widget.py", line 108, in is_plottable
    fields, message = get_plot_fields(topic_name)
  File "/home/htpc/catkin_ws/src/rqt_common_plugins/rqt_plot/src/rqt_plot/plot_widget.py", line 49, in get_plot_fields
    topic_type, real_topic, _ = topic_helpers.get_topic_type(topic_name)
  File "<string>", line 421, in get_topic_type
  File "<string>", line 404, in _get_topic_type
  File "<string>", line 305, in msgevalgen
AssertionError: Topic name 'values[0' contains '[' but does not end with ']'
```

dirk-thomas:
What exactly did you try when you got the stack trace? From the assertion it looks like you passed a topic name with an open bracket but without a closing one.

trainman419:
Yep. I got that stack while typing the full topic name.

dirk-thomas:
Can you please post a reproducible example?

trainman419:
[...]

PR #481, opened 2014-07-29
fix handling of array index when being part of the topic

dirk-thomas:
Addresses #480.

Also handles splice arguments.


tfoote:
+1

PR #484, opened 2014-08-04
Fix C++11 compatibility issue in topic.h

### LINKED ISSUE #483: roscpp: topic.h is not compatible with c++11
[This line](https://github.com/ros/ros_comm/blob/indigo-devel/clients/roscpp/include/ros/topic.h#L64) gives the following error when compiled with `c++11` option:

```
/opt/ros/indigo/include/ros/topic.h|64 col 12| error: no viable conversion from 'MConstPtr' (aka 'shared_ptr<const stdr_msgs::RobotIndexedVectorMsg_<std::allocator<void> > >') to 'bool'
         return message_;
```

The reason is that in C++11, `shared_ptr` has an explicit `operator bool` which means that a `shared_ptr` can't be implicitly converted to a `bool`. A trivial fix could be applied to make the code compatible with both the new and the old stardards:

``` cpp
return static_cast<bool>(message_);
```

Alternatively, one could write:

``` cpp
return message_.get() != 0;
```

Please let me know which version is preferable and I will submit a pull request.

taketwo:
Resolves #483.

### DIFF
--- a/clients/roscpp/include/ros/topic.h
+++ b/clients/roscpp/include/ros/topic.h
@@ -61,7 +61,7 @@ public:
 
   bool hasMessage()
   {
-    return message_;
+    return static_cast<bool>(message_);
   }
 
   MConstPtr getMessage()


dirk-thomas:
@ros-pull-request-builder

PR #486, opened 2014-08-07
Implements delivered_msgs datafield to TopicStatistics.msg

mrdanbrooks:
Populates the proposed delivered_msgs field in rosgraph_msgs/TopicStatistics.


dirk-thomas:
This depends on ros/ros_comm_msgs#2 being merged and released.

Since this patch requires a newer version of the message please add an explicit version dependency for the upcoming patch release of the message package.

PR #491, opened 2014-08-19
add specific exception for time jumping backwards

### LINKED ISSUE #485: rospy.sleep(): request for specific exception if time moved backwards
I kindly request for the TODO of rospy.sleep() in [rospy.timer.py](https://github.com/ros/ros_comm/blob/hydro-devel/clients/rospy/src/rospy/timer.py) to be implemented and use a specific exception if ros time moved backwards.
I had trouble to find the cause of a python node shutting down, when using 'rosbag play -l --clock bagfile'. When the implementation follows the ros tutorials on python the ROSInterruptException is always catched:

``` python
if __name__ == '__main__':
    try:
        talker()
    except rospy.ROSInterruptException: pass
```

And there is no error message telling you, that time moved backwards.

jowave:
reason: ROS shutdown and time jumps should be distinguishable.
ROSTimeMovedBackwardsException inherits from ROSInterruptException
to preserve the API.
fixes #485

### DIFF
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -44,6 +44,12 @@ import rospy.rostime
 # author: tfield (Timers)
 # author: kwc (Rate, sleep)
 
+class ROSTimeMovedBackwardsException(rospy.exceptions.ROSInterruptException): 
+    """
+    Exception if time moved backwards
+    """
+    pass
+
 class Rate(object):
     """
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please recreate your pull request against _indigo-devel_. No patches will be merge into Hydro until:
- they have been tested, merged and released into Indigo
- are necessary bug fixes
- except if they are specific to Hydro and don't apply to Indigo

PR #492, opened 2014-08-19
add specific exception for time jumping backwards

### LINKED ISSUE #485: rospy.sleep(): request for specific exception if time moved backwards

jowave:
reason: ROS shutdown and time jumps should be distinguishable.
ROSTimeMovedBackwardsException inherits from ROSInterruptException
to preserve the API.
fixes #485

### DIFF
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -65,8 +65,10 @@ class Rate(object):
         account the time elapsed since the last successful
         sleep().
         
-        @raise ROSInterruptException: if ROS time is set backwards or if
-        ROS shutdown occurs before sleep completes
+        @raise ROSInterruptException: if ROS shutdown occurs before
+        sleep completes
+        @raise ROSTimeMovedBackwardsException: if ROS time is set
+        backwards
         """
         curr_time = rospy.rostime.get_rostime()
         # detect time jumping backwards
@@ -83,7 +85,6 @@ class Rate(object):
         if curr_time - self.last_time > self.sleep_dur * 2:
             self.last_time = curr_time
 
-# TODO: may want more specific exceptions for sleep
 def sleep(duration):
     """
     sleep for the specified duration in ROS time. If duration
@@ -91,8 +92,10 @@ def sleep(duration):
     
     @param duration: seconds (or rospy.Duration) to sleep
     @type  duration: float or Duration
-    @raise ROSInterruptException: if ROS time is set backwards or if
-    ROS shutdown occurs before sleep completes
+    @raise ROSInterruptException: if ROS shutdown occurs before sleep
+    completes
+    @raise ROSTimeMovedBackwardsException: if ROS time is set
+    backwards
     """
     if rospy.rostime.is_wallclock():
         if isinstance(duration, genpy.Duration):
@@ -128,7 +131,9 @@ def sleep(duration):
                 rostime_cond.wait(0.5)
 
         if rospy.rostime.get_rostime() < initial_rostime:
-            raise rospy.exceptions.ROSInterruptException("ROS time moved backwards")
+            time_jump = (initial_rostime - rospy.rostime.get_rostime()).to_sec()
+            rospy.core.logerr("ROS time moved backwards: %ss", time_jump)
+            raise rospy.exceptions.ROSTimeMovedBackwardsException(time_jump)
         if rospy.core.is_shutdown():
             raise rospy.exceptions.ROSInterruptException("ROS shutdown request")
 
--- a/test/test_rospy/test/unit/test_rospy_exceptions.py
+++ b/test/test_rospy/test/unit/test_rospy_exceptions.py
@@ -62,3 +62,16 @@ class TestRospyExceptions(unittest.TestCase):
             raise ROSInterruptException("test")
         except KeyboardInterrupt:
             pass
+
+    def test_ROSTimeMovedBackwardsException(self):
+        from rospy.exceptions import ROSTimeMovedBackwardsException, ROSInterruptException
+        try:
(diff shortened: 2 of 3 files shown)


jowave:
Recreated the pull request against indigo-devel.
Sorry for the inconvenience.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Can you please move the exception class to `clients/rospy/src/rospy/exceptions.py` where most other exceptions are defined.

Beside this minor detail the patch lgtm.

jowave:
I moved the exception and updated the pull request.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Please update your patch to pass all unit tests. It currently breaks 145 tests.

dirk-thomas:
Please update your patch to pass all unit tests. It currently breaks 145 tests.

jowave:
I have updated the patch. Apparently it passed the tests now.

wjwwood:
+1

clients/rospy/src/rospy/timer.py
```diff
@@ -128,7 +131,8 @@ def sleep(duration):
                 rostime_cond.wait(0.5)
 
         if rospy.rostime.get_rostime() < initial_rostime:
-            raise rospy.exceptions.ROSInterruptException("ROS time moved backwards")
+            rospy.core.logerr("ROS time moved backwards")
```
dirk-thomas: Can you please include the time offset in the error message so that the user has a better idea how much the time jumped.

jowave:
I added the time offset in the error message and a small test for the exception.

tlind:
I am worried that commit d934e53 might break existing code which is trying to catch the exception. Since there was previously no specific exception for this case, the only way of specifically catching the time-moved-backwards issue was to check the exception message, e.g.:

```
try:
    r.sleep()
except rospy.exceptions.ROSInterruptException as e:
    if e.message == "ROS time moved backwards":
        rospy.logwarn("Time has moved backwards (probably ROS clock has been reset)!")
```

However, it seems that supplying the time offset as an additional argument to the exception constructor will break that message:

```
>>> print rospy.exceptions.ROSInterruptException("ROS time moved backwards").message
ROS time moved backwards
>>> print rospy.exceptions.ROSInterruptException("ROS time moved backwards", -10.0).message
[empty string]
```

Might it be a better idea to only rospy.logerr() the time offset, and leave the exception message as-is? Or add an additional field to the exception class which contains the time offset?

EDIT: I just read that Exception.message seems to be deprecated since 2.6, though I've been using it a few times and never saw a warning anywhere. Not sure how much other code depends on it. The correct way seems to be to use e.args[0], which returns the same result in both cases.

dirk-thomas:
@tlind That is a good point.

@jowave Can you please update the new exception class to store the time offset in a member variable and keep the message of the exception as-is.

jowave:
I modified the code to work with Exception.message and added a test to check for the right message.

test/test_rospy/test/unit/test_rospy_exceptions.py
```diff
@@ -62,3 +62,17 @@ def test_ROSInterruptException(self):
...
             pass
+
+    def test_ROSTimeMovedBackwardsException(self):
+        from rospy.exceptions import ROSTimeMovedBackwardsException, ROSInterruptException
+        try:
+            raise ROSTimeMovedBackwardsException(1.0)
+        except ROSInterruptException as e:
+            # ensure the message is not changed, because old code may check it
+            self.assertEqual("ROS time moved backwards", e.message)
+            pass
```
dirk-thomas: This line should be obsolete.

dirk-thomas:
Please squash your commits after removing the unnecessary `pass` line.

dirk-thomas:
Thank you for you efforts on this. Merging...

jowave:
You're welcome. Thank you too.

Edit: Hmm, the ros-pull-request-builder is answering. Should i not have replied to a closed pull request?

dirk-thomas:
No worries. The Jenkins check is still experimental. We hopefully can improve that in the future.

clients/rospy/src/rospy/timer.py
```diff
@@ -128,7 +131,9 @@ def sleep(duration):
                 rostime_cond.wait(0.5)
 
         if rospy.rostime.get_rostime() < initial_rostime:
-            raise rospy.exceptions.ROSInterruptException("ROS time moved backwards")
+            time_jump = (initial_rostime - rospy.rostime.get_rostime()).to_sec()
+            rospy.core.logerr("ROS time moved backwards: %ss", time_jump)
```
peci1, line 135: Is it really needed to log the exception? I use ROS nodes together with Gazebo, where restarting the time is quite normal, and these messages start being annoying. The previous behavior also did not log the exception.

dirk-thomas: I agree with you that the log message is not strictly necessary. On the other hand this code has been integrated over two years ago and since then been released in three ROS distributions. Since no concerns were raised in the meantime I don't see the log message as a problem either.

Please feel free to create a pull request against the latest development branch (currently *lunar-devel*).

PR #498, opened 2014-09-17
Fix roslaunch compatibility with paramiko 1.10.0

mvollrath:
Older versions of paramiko return the IdentityFile as a string from the SSH config.

Newer versions return a list.


mvollrath:
Some reference material related to this issue:

https://bugs.launchpad.net/duplicity/+bug/1156746

https://github.com/paramiko/paramiko/blob/master/paramiko/client.py#L204

https://github.com/paramiko/paramiko/blob/master/paramiko/config.py#L80

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Thanks!

Issue #495, opened 2014-09-10
rosmaster doesn't close sockets properly, hangs


lrasinen:
Steps to reproduce:

1) Start `roscore`
2) Launch rosout repeatedly: `while :; do rosrun rosout rosout; done`
3) Monitor CLOSE_WAIT sockets: `lsof | grep rosmaster | grep -c CLOSE_WAIT`

The number of sockets keeps steadily climbing until the open file limit is hit. rosmaster then becomes unresponsive (e.g. rostopic list or rosnode list both just hang) and starts consuming 100% CPU.

Versions:
Ubuntu 12.04.5 LTS
ros-hydro-desktop-full  1.0.2-0precise-20140807-2038-+0000

Background:
While starting rosout in a loop isn't the most sensible thing, it's just the minimal test case to reproduce this. Trying "rosnode list" or "rostopic list" doesn't leak sockets in the same way.

I've had a few instances of this happening and it probably hasn't been rosout (but due to other pressures I haven't had the time to figure out the flaky component)

dirk-thomas:
The master is not garbaging ServerProxy instances in the case when it notifies nodes to shut down. You likely restarted a non-anonymous node several times which leaked connections for every shut down node.

Fixed with #499.

PR #499, opened 2014-09-19
garbage ServerProxy after telling a node to shutdown

dirk-thomas:
Addresses #495.


wjwwood:
+1

tfoote:
+1 verified locally to fix the reproducible issue.

PR #500, opened 2014-09-19
Added option to specify the minimum disk space at which recording is stopped

cedricpradalier:
We record on a small-ish disk (10GB available) and we've been bitten several time by the hard-coded 1GB limit at which recording stops. On our system, is is 10% of the capacity that is lost. This option makes this parameter manually configurable for a given system.

### DIFF
--- a/tools/rosbag/include/rosbag/recorder.h
+++ b/tools/rosbag/include/rosbag/recorder.h
@@ -106,6 +106,8 @@ struct ROSBAG_DECL RecorderOptions
     uint32_t        max_size;
     ros::Duration   max_duration;
     std::string     node;
+    unsigned long long   min_space;
+    std::string     min_space_str;
 
     std::vector<std::string> topics;
 };
--- a/tools/rosbag/include/rosbag/recorder.h
+++ b/tools/rosbag/include/rosbag/recorder.h
@@ -87,27 +87,27 @@ struct ROSBAG_DECL RecorderOptions
 {
     RecorderOptions();
 
-    bool            trigger;
-    bool            record_all;
-    bool            regex;
-    bool            do_exclude;
-    bool            quiet;
-    bool            append_date;
-    bool            snapshot;
-    bool            verbose;
-    CompressionType compression;
-    std::string     prefix;
-    std::string     name;
-    boost::regex    exclude_regex;
-    uint32_t        buffer_size;
-    uint32_t        chunk_size;
-    uint32_t        limit;
-    bool            split;
-    uint32_t        max_size;
-    ros::Duration   max_duration;
-    std::string     node;
-    unsigned long long   min_space;
-    std::string     min_space_str;
+    bool               trigger;
+    bool               record_all;
+    bool               regex;
+    bool               do_exclude;
+    bool               quiet;
+    bool               append_date;
+    bool               snapshot;
+    bool               verbose;
+    CompressionType    compression;
+    std::string        prefix;
+    std::string        name;
+    boost::regex       exclude_regex;
+    uint32_t           buffer_size;
+    uint32_t           chunk_size;
+    uint32_t           limit;
+    bool               split;
+    uint32_t           max_size;
+    ros::Duration      max_duration;
+    std::string        node;
+    unsigned long long min_space;
+    std::string        min_space_str;
 
     std::vector<std::string> topics;
 };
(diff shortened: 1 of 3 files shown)


dirk-thomas:
@ros-pull-request-builder

tools/rosbag/src/record.cpp
```diff
@@ -140,6 +141,38 @@ rosbag::RecorderOptions parseOptions(int argc, char** argv) {
...
+                default:
+                    opts.min_space = value;
+                    break;
+            }
+        } else {
+            value = 1073741824ull;
+            opts.min_space_str = "1G";
+            opts.min_space = value;
+        }
+        // ROS_INFO("Rosbag using minimum space of %lld bytes, or %s",opts.min_space,opts.min_space_str.c_str());
```
tfoote: This would be better as a ROS_DEBUG instead of a commented ROS_INFO

tools/rosbag/src/record.cpp
```diff
@@ -140,6 +141,38 @@ rosbag::RecorderOptions parseOptions(int argc, char** argv) {
...
+    }
```
tfoote, line 173: Is the min_space set if the option is not passed? I see an else if the value fails to parse, but not for the option is not specified.

cedricpradalier, line 173: I thought that was the point of the default value in the option description, but over-initializing does not hurt.

tools/rosbag/src/record.cpp
cedricpradalier: Will do. Thanks

tools/rosbag/src/record.cpp
tfoote, line 173: Sorry, I was looking for a ull number for the initialization. The 1G value will be fine.

cedricpradalier:
Integrated latest comments. Should I reissue a merge request?

dirk-thomas:
Please update the patch to follow similar code style than in the surrounding code, e.g. spaces around operators, after commas, etc.

cedricpradalier:
OK, will do.

tfoote:
You don't need to reopen the pull-request. When you update your public branch for the source (in this case cedricpradalier:indigo-devel), this pull-request automatically receives the updates. Though do comment when you push code updates as we don't get a notification email when code changes.

cedricpradalier:
I've updated what I found, but if there are other spaces and commas you want me to move, please be specific. It takes me a lot of time to compare format character per character, when this type of prettyfying should really be done by an automatic parser in my opinion.

cedricpradalier:
Following up on this little patch. Anything I should do to push it in?
Regards.

tools/rosbag/include/rosbag/recorder.h
```diff
@@ -87,25 +87,27 @@ struct ROSBAG_DECL RecorderOptions
...
-    std::string     prefix;
-    std::string     name;
-    boost::regex    exclude_regex;
-    uint32_t        buffer_size;
-    uint32_t        chunk_size;
-    uint32_t        limit;
-    bool            split;
-    uint32_t        max_size;
-    ros::Duration   max_duration;
-    std::string     node;
```
dirk-thomas, line 108: Please avoid reindenting the existing lines.

tools/rosbag/src/record.cpp
```diff
@@ -140,6 +141,36 @@ rosbag::RecorderOptions parseOptions(int argc, char** argv) {
     {
       opts.limit = vm["limit"].as<int>();
     }
+    if (vm.count("min-space"))
+    {
+        std::string ms = vm["min-space"].as<std::string>();
+        long long int value=1073741824ull;
+        char mul=0;
```
dirk-thomas, line 148: Whitespaces around `=`.

tools/rosbag/src/record.cpp
```diff
@@ -140,6 +141,36 @@ rosbag::RecorderOptions parseOptions(int argc, char** argv) {
     {
       opts.limit = vm["limit"].as<int>();
     }
+    if (vm.count("min-space"))
+    {
+        std::string ms = vm["min-space"].as<std::string>();
+        long long int value=1073741824ull;
```
dirk-thomas, line 147: Whitespaces around `=`.

dirk-thomas:
Please consider updating the PR according to the comments in order to be merged. Thank you.

dirk-thomas:
Manually merged with modification in f8fa420e54772cddb07c0d7d6693c9b59f94cc6e.

PR #510, opened 2014-10-09
ros_bag 508

csherstan:
Reworked bag.py to provide various methods for getting info about the bag as a programmatic object rather than in string or yaml format.

### DIFF
--- a/test/test_rosbag/test/test_bag.py
+++ b/test/test_rosbag/test/test_bag.py
@@ -263,12 +263,117 @@ class TestRosbag(unittest.TestCase):
                 bag.write("/test_bag", String(data='also'))
                 bag.write("/test_bag/more", String(data='alone'))
                 
-        bag = rosbag.Bag(fn)
+        with rosbag.Bag(fn) as bag:
+            self.assertEquals(bag.get_message_count(), 300)
+            self.assertEquals(bag.get_message_count(topic_filters='/test_bag'), 200)
+            self.assertEquals(bag.get_message_count(topic_filters=['/test_bag', '/test_bag/more']), 300)
+            self.assertEquals(bag.get_message_count(topic_filters=['/none']), 0)
         
-        self.assertEquals(bag.get_message_count(), 300)
-        self.assertEquals(bag.get_message_count(topic_filters='/test_bag'), 200)
-        self.assertEquals(bag.get_message_count(topic_filters=['/test_bag', '/test_bag/more']), 300)
-        self.assertEquals(bag.get_message_count(topic_filters=['/none']), 0)
+    def test_get_compression_info(self):
+        fn = '/tmp/test_get_compression_info.bag'
+        
+        # No Compression
+        with rosbag.Bag(fn, mode='w') as bag:
+            for i in xrange(100):
+                bag.write("/test_bag", Int32(data=i))
+                
+        with rosbag.Bag(fn) as bag:
+            info = bag.get_compression_info()
+            self.assertEquals(info.compression, rosbag.Compression.NONE)
+            # 167 Bytes of overhead, 50 Bytes per Int32.
+            self.assertEquals(info.uncompressed, 5167)
+            self.assertEquals(info.compressed, 5167)
+        
+        with rosbag.Bag(fn, mode='w', compression=rosbag.Compression.BZ2) as bag:
+            for i in xrange(100):
+                bag.write("/test_bag", Int32(data=i))
+                
+        with rosbag.Bag(fn) as bag:
+            info = bag.get_compression_info()
+            self.assertEquals(info.compression, rosbag.Compression.BZ2)
+            self.assertEquals(info.uncompressed, 5167)
+            
+            # the value varies each run, I suspect based on rand, but seems
+            # to generally be around 960 to 980 on my comp
+            self.assertLess(info.compressed, 1000)
+            self.assertGreater(info.compressed, 900)
+        
+    def test_get_time(self):
+        fn = '/tmp/test_get_time.bag'
+        
+        with rosbag.Bag(fn, mode='w') as bag:
+            for i in xrange(100):
+                bag.write("/test_bag", Int32(data=i), t=genpy.Time.from_sec(i))
+                
+        with rosbag.Bag(fn) as bag:
+            start_stamp = bag.get_start_time()
+            end_stamp = bag.get_end_time()
+            
+            self.assertEquals(start_stamp, 0.0)
+            self.assertEquals(end_stamp, 99.0)
+
+    def test_get_type_and_topic_info(self):
+        fn = '/tmp/test_get_type_and_topic_info.bag'
+        topic_1 = "/test_bag"
+        topic_2 = "/test_bag/more"
+        with rosbag.Bag(fn, mode='w') as bag:
+            for i in xrange(100):
+                bag.write(topic_1, Int32(data=i))
+                bag.write(topic_1, String(data='also'))
+                bag.write(topic_2, String(data='alone'))
+                
+        with rosbag.Bag(fn) as bag:
+            msg_types, topics = bag.get_type_and_topic_info()
+            self.assertEquals(len(msg_types), 2)
+            self.assertTrue("std_msgs/Int32" in msg_types)
+            self.assertTrue("std_msgs/String" in msg_types)
+            self.assertEquals(len(topics), 2)
+            self.assertTrue(topic_1 in topics)
+            self.assertTrue(topic_2 in topics)
+            
+            self.assertEquals(topics[topic_1].message_count, 200)
+            self.assertEquals(topics[topic_1].msg_type, "std_msgs/Int32")
+            self.assertEquals(topics[topic_2].message_count, 100)
+            self.assertEquals(topics[topic_2].msg_type, "std_msgs/String")
+            
+            #filter on topic 1
+            msg_types, topics = bag.get_type_and_topic_info(topic_1)
+            
+            # msg_types should be unaffected by the filter
+            self.assertEquals(len(msg_types), 2)
+            self.assertTrue("std_msgs/Int32" in msg_types)
+            self.assertTrue("std_msgs/String" in msg_types)            
+            
+            self.assertEquals(len(topics), 1)
+            self.assertTrue(topic_1 in topics)
+            
+            self.assertEquals(topics[topic_1].message_count, 200)
+            self.assertEquals(topics[topic_1].msg_type, "std_msgs/Int32")
+            
+            #filter on topic 2
+            msg_types, topics = bag.get_type_and_topic_info(topic_2)
+            
+            # msg_types should be unaffected by the filter
+            self.assertEquals(len(msg_types), 2)
+            self.assertTrue("std_msgs/Int32" in msg_types)
+            self.assertTrue("std_msgs/String" in msg_types)            
+            
+            self.assertEquals(len(topics), 1)
+            self.assertTrue(topic_2 in topics)
+            
+            self.assertEquals(topics[topic_2].message_count, 100)
+            self.assertEquals(topics[topic_2].msg_type, "std_msgs/String")
+            
+            #filter on missing topic
+            msg_types, topics = bag.get_type_and_topic_info("/none")
+            
+            # msg_types should be unaffected by the filter
+            self.assertEquals(len(msg_types), 2)
+            self.assertTrue("std_msgs/Int32" in msg_types)
+            self.assertTrue("std_msgs/String" in msg_types)            
+            
+            # topics should be empty
+            self.assertEquals(len(topics), 0)
         
     def _print_bag_records(self, fn):
         with open(fn) as f:
--- a/test/test_rosbag/test/test_bag.py
+++ b/test/test_rosbag/test/test_bag.py
@@ -47,6 +47,7 @@ from rosbag import bag
 import rospy
 from std_msgs.msg import Int32
 from std_msgs.msg import ColorRGBA
+from std_msgs.msg import String
 
 class TestRosbag(unittest.TestCase):
     def setUp(self):
@@ -253,7 +254,22 @@ class TestRosbag(unittest.TestCase):
         with rosbag.Bag(fn) as b:
             for topic, msg, t in b:
                 pass
-
+            
+    def test_get_message_count(self):
+        fn = '/tmp/test_get_message_count.bag'
+        with rosbag.Bag(fn, mode='w') as bag:
+            for i in xrange(100):
+                bag.write("/test_bag", Int32(data=i))
+                bag.write("/test_bag", String(data='also'))
+                bag.write("/test_bag/more", String(data='alone'))
+                
+        bag = rosbag.Bag(fn)
+        
+        self.assertEquals(bag.get_message_count(), 300)
+        self.assertEquals(bag.get_message_count(topic_filters='/test_bag'), 200)
+        self.assertEquals(bag.get_message_count(topic_filters=['/test_bag', '/test_bag/more']), 300)
+        self.assertEquals(bag.get_message_count(topic_filters=['/none']), 0)
+        
     def _print_bag_records(self, fn):
         with open(fn) as f:
             f.seek(0, os.SEEK_END)
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -460,21 +460,30 @@ class Bag(object):
                 
         return collections.namedtuple("CompressionTuple", ["compression", "uncompressed", "compressed"])(compression=compression, uncompressed=uncompressed, compressed=compressed)
     
-    def get_message_count(self):
+    def get_message_count(self, topic_filters=None, type_filters=None):
         """
             Returns the number of messages in the bag.
+            
+            topic_filters - One or more topics to filter by
         """
         
         num_messages = 0
-        if self._chunks:
-            num_messages = 0
-            for c in self._chunks:
-                for counts in c.connection_counts.values():
-                    num_messages += counts
+        
+        if topic_filters is not None:
+            info = self.get_type_and_topic_info(topic_filters=topic_filters)
+            for topic in info.topics.itervalues():
+                num_messages+=topic.message_count
         else:
-            num_messages = sum([len(index) for index in self._connection_indexes.values()])
+            if self._chunks:
+                num_messages = 0
+                for c in self._chunks:
+                    for counts in c.connection_counts.values():
+                        num_messages += counts
+            else:
+                num_messages = sum([len(index) for index in self._connection_indexes.values()])
             
         return num_messages
+    
     def get_timing_info(self):
         if self._chunks:
             start_stamp = self._chunks[ 0].start_time.to_sec()
@@ -488,17 +497,22 @@ class Bag(object):
         return collections.namedtuple("TimingTuple", ["duration",
                                                       "start_stamp", "end_stamp"])(duration=duration,
                                                                                    start_stamp=start_stamp, end_stamp=end_stamp)
-    def get_type_and_topic_info(self):
+    def get_type_and_topic_info(self, topic_filters=None):
         """
             coallates info about the type and topics in the bag.
-
+            
+            topic_filters - specify one or more topic to filter by.
+            
+            Note, it would be nice to filter by type as well, but there appear to be some limitations in the current architecture
+            that prevent that from working when more than one message type is written on the same topic.
+            
             returns (TypesAndTopicsTuple):
                 - types(dict):
                     key: type name
                     val: md5hash
                 - topics(TopicTuple):
                     type: msg type (Ex. "std_msgs/String")
-                    message_count: the number of messages of the topic
+                    message_count: the number of messages of the particular type
                     connections: the number of connections
                     frequency: the data frequency
         """
@@ -506,8 +520,6 @@ class Bag(object):
         datatypes = set()
         datatype_infos = []
         
-        # loop through all the connections and collect the associated message types. FYI, if a topic has more than one type
-        # only the first type for that topic is listed
         for connection in list(self._connections.values()):
             if connection.datatype in datatypes:
                 continue
@@ -515,7 +527,15 @@ class Bag(object):
             datatype_infos.append((connection.datatype, connection.md5sum, connection.msg_def))
             datatypes.add(connection.datatype)
             
-        topics = sorted(set([c.topic for c in self._get_connections()]))
+        topics = []
+        # load our list of topics and optionally filter
+        if topic_filters is not None:
+            if not isinstance(topic_filters, list):
+                topic_filters = [topic_filters]
+            topics = sorted(set(topic_filters))
+        else:
+            topics = sorted(set([c.topic for c in self._get_connections()]))
+
             
         topic_datatypes    = {}
         topic_conn_counts  = {}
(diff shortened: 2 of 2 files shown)


tools/rosbag/src/rosbag/bag.py
```diff
@@ -94,6 +95,24 @@ class Compression:
     NONE = 'none'
     BZ2  = 'bz2'
     LZ4  = 'lz4'
+    
+class BagInfo(object):
+    def __init__(self, path=None, version=None, size=None, indexed=None, 
+                 duration=None, start=None, end=None, messages=None, 
+                 compression=None, uncompressed=None, compressed=None, types={}, topics={}):
```
wjwwood: You should use `None` for the `types` and `topics` argument defaults and then do something like this:

``` python
self.types = types or {}
self.topics = topics or {}
```

See: http://stackoverflow.com/questions/1132941/least-astonishment-in-python-the-mutable-default-argument

csherstan: Thanks! I'd never run into that before.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,6 +420,205 @@ def close(self):
...
+            returns (CompressionTuple):
+                compression - Type of compression used
+                uncompressed - Number of uncompressed Bytes
+                compressed - Number of compressed Bytes
+        """
+        compression = self.compression
+        uncompressed = self.size
+        compressed = 0
+        
+        if len(self._chunk_headers) != 0:
```
wjwwood: You don't have to change this, but for future reference, this could just be `if self._chunk_headers:`

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,6 +420,205 @@ def close(self):
...
+                    compression_uncompressed[chunk_header.compression] = chunk_header.uncompressed_size
+                    compression_compressed[chunk_header.compression] = chunk_header.compressed_size
+                else:
+                    compression_counts[chunk_header.compression] += 1
+                    compression_uncompressed[chunk_header.compression] += chunk_header.uncompressed_size
+                    compression_compressed[chunk_header.compression] += chunk_header.compressed_size
+
+            chunk_count = len(self._chunk_headers)
+
+            main_compression_count, main_compression = list(reversed(sorted([(v, k) for k, v in compression_counts.items()])))[0]
```
wjwwood: The outside `list(...)` is probably not needed since `sorted(...)` returns a sorted copy of the list, and `reversed(sorted(...))` is totally fine, but you can also do `sorted(..., reverse=True)`: https://docs.python.org/2/library/functions.html#sorted

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,6 +420,205 @@ def close(self):
...
+            
+            topic_filters - One or more topics to filter by
+        """
+        
+        num_messages = 0
+        
+        if topic_filters is not None:
+            info = self.get_type_and_topic_info(topic_filters=topic_filters)
+            for topic in info.topics.itervalues():
+                num_messages+=topic.message_count
```
wjwwood: PEP8 (http://legacy.python.org/dev/peps/pep-0008/) says to put whitespace around operators, i.e. `num_messages += topic.message_count`. It might not be like this all throughout the file, but given the choice on new work you should prefer this style.

test/test_rosbag/test/test_bag.py
```diff
@@ -253,7 +254,22 @@ def test_future_version_works(self):
         with rosbag.Bag(fn) as b:
             for topic, msg, t in b:
                 pass
-
+            
+    def test_get_message_count(self):
+        fn = '/tmp/test_get_message_count.bag'
```
wjwwood, line 259: Where does this file come from? Is it generated by the tests or is it a hard coding that accidentally got committed? If it is generated it should use the `tempfile` module and not assume `/tmp/test_get_message_count.bag` is an available path.

wjwwood:
Thanks for the contribution, other than some questions and nit picks the code review looks ok.

tools/rosbag/src/rosbag/bag.py
csherstan: Yeah, lot's of this is left over from the original code I grabbed from **str**. I'll change it though.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

test/test_rosbag/test/test_bag.py
csherstan, line 259: This is the only comment I haven't made a change for. The behaviour here is consistent with the rest of the test file. Note that I am first writing this file and then reading it back in. The file does not need to exist already, but we do need access to write to tmp.

You may be right about using the tempfile module, but if that's the case, could we open a separate issue to refactor this whole file to use it?

wjwwood, line 259: It is fine to leave it as is if the rest of the file is doing it, and it would be good to update everything to use tempfile, but I agree that should go in a different issue.

csherstan, line 259: Created issue: https://github.com/ros/ros_comm/issues/511

tools/rosbag/src/rosbag/bag.py
```diff
@@ -48,6 +48,7 @@
 import threading
 import time
 import yaml
+import collections
```
dirk-thomas: Please move this import to be in alphabetical order (PEP8).

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,6 +420,205 @@ def close(self):
...
+            Returns the number of messages in the bag.
+            
+            topic_filters - One or more topics to filter by
+        """
+        
+        num_messages = 0
+        
+        if topic_filters is not None:
+            info = self.get_type_and_topic_info(topic_filters=topic_filters)
+            for topic in info.topics.itervalues():
```
dirk-thomas: Please use Python 3 compatible code: `itervalues()` --> `values()`.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,6 +420,205 @@ def close(self):
...
+                    type: msg type (Ex. "std_msgs/String")
+                    message_count: the number of messages of the particular type
+                    connections: the number of connections
+                    frequency: the data frequency
+        """
+        
+        datatypes = set()
+        datatype_infos = []
+        
+        for connection in list(self._connections.values()):
```
dirk-thomas: Why is `list()` necessary here?

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,6 +420,205 @@ def close(self):
...
+            datatypes.add(connection.datatype)
+            
+        topics = []
+        # load our list of topics and optionally filter
+        if topic_filters is not None:
+            if not isinstance(topic_filters, list):
+                topic_filters = [topic_filters]
+            topics = sorted(set(topic_filters))
+        else:
+            topics = sorted(set([c.topic for c in self._get_connections()]))
```
dirk-thomas: Please extract the `sorted(set(..))` calls outside of the condition - since it is performed in both branches.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,6 +420,205 @@ def close(self):
...
+
+        # process datatypes       
+        types = {}
+        for i, (datatype, md5sum, msg_def) in enumerate(sorted(datatype_infos)):
+            types[datatype] = md5sum
+            
+        # process topics
+        topics_t = {}
+        TopicTuple = collections.namedtuple("TopicTuple", ["type", "message_count", "connections", "frequency"])
+        for i, topic in enumerate(topics):
```
dirk-thomas: What is `i` used for in this loop?

dirk-thomas:
Overall the newly introduced class `BagInfo` seems to only bundle together information already available in the existing API. Wouldn't it make more sense to provide these helper functions in a separate Python module rather then adding more and more stuff to the existing package?

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected

tools/rosbag/src/rosbag/bag.py
csherstan: Don't know, most of this is code copied from the existing **str** or _get_yaml_info. Removed it.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

csherstan:
Hmmm, well, personally I think it makes more sense to keep these helpers internal to Bag. Bag should be able to describe itself, rather than have someone else do it. Additionally, many of these functions use internal methods and members (ones that start with _), so I wouldn't want to be accessing those from outside the Bag class out of principle. As far as the BagInfo class goes, I actually wonder if it would make more sense to tuck it inside Bag itself: Bag.BagInfo.

Additionally, I intend to follow this pull request with another one for refactoring the code in _get_yaml_info and __str__ to make use of these helper functions.

Recall that I am new to python and the ROS, so if I have missed something I'm happy to be corrected.

dirk-thomas:
The new `BagInfo` class contains the following members which are all available through the public `Bag` API as far as I can see:
- path: `filename` property
- version: `version` property
- size: `size` property
- duration: `get_timing_info()[0]`
- start: `get_timing_info()[1]`
- end: `get_timing_info()[2]`
- messages: `get_message_count()`
- compression: `get_compression_info()[0]`
- uncompressed: `get_compression_info()[1]`
- compressed: `get_compression_info()[2]`
- types: `get_type_and_topic_info()[0]`
- topics: `get_type_and_topic_info()[1]`

The only attribute currently not cover by the public API is the `indexed` flag.

Since this looks like a simple duplication of API I would rather not introduce these duplicate interfaces. If there is a need for the `indexed` flag that can be simply added to the `Bag` API.

wjwwood:
First off, thanks for taking the time to put a pull request together, I understand it takes a non-trivial amount of effort.

I guess what I'm missing from this pull request is a better description of the issue you are trying to solve and how this pull request solves it. Which has to do with justification for each of the changes.

As @dirk-thomas has pointed out, some of the stuff in your BagInfo class simply duplicates information which is available in the Bag class itself. Now I agree that some of the new accessors have semantic value, for instance I would be +1 for having `Bag.duration`, `Bag.start`, and `Bag.end` added to the Bag API as syntactic sugar on top of `Bag.get_timing_info` and as @dirk-thomas pointed out, the `indexed` flag was not previously public information. I think both of these changes could be merged here.

However, I don't see the need for a new class, BagInfo, nor a reason for some of the other additions, e.g. `BagInfo.messages` just replaces `Bag.get_message_count`. Now there may exist good reasons and justifications for this new BagInfo class and some of its members, but I'd like you to articulate those before I would be in favor of adding them to the rospy core code base.

If you simply want to represent the same Bag information in a different way, then I would say it should start in an external package, committing here only the changes which expose previously private information about bags. Then if that external package and it's API are widely used then we can consider merging it into the core system. I think this is a much easier and lower risk method for trying out changes like this.

csherstan:
Oh, I see, so your only real issue is that BagInfo seems redundant?

Let me first address what this pull request was meaning to address. What I was trying to do was to expose the information that was being processed internally for **str** and _get_yaml_info programmatically, which I have done with the various "get" methods I added. The addition of the get_info method and the BagInfo class was simply to provide a convenient wrapper to represent all the info in one object.

While I don't understand why adding BagInfo is an issue or what risk it represents, I'm fine with removing it, it just seemed like a convenient way to encapsulate the data.

csherstan:
Removed BagInfo and Bag.get_info.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -94,7 +95,7 @@ class Compression:
     NONE = 'none'
     BZ2  = 'bz2'
     LZ4  = 'lz4'
-
+
```
dirk-thomas: Please remove the trailing white space.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,177 @@ def close(self):
...
+    
+    def get_timing_info(self):
+        if self._chunks:
+            start_stamp = self._chunks[ 0].start_time.to_sec()
+            end_stamp = self._chunks[-1].end_time.to_sec()
+        else:
+            start_stamp = min([index[ 0].time.to_sec() for index in self._connection_indexes.values()])
+            end_stamp = max([index[-1].time.to_sec() for index in self._connection_indexes.values()])
+        
+        duration = end_stamp - start_stamp
```
dirk-thomas: I am not sure if it is a good idea to compute derived values in the API. Anyone can trivially do this in user land based on `end` and `start`.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,177 @@ def close(self):
...
+                    periods = [s1 - s0 for s1, s0 in zip(stamps[1:], stamps[:-1])]
+                    med_period = _median(periods)
+                    if med_period > 0.0:
+                        topic_freqs_median[topic] = 1.0 / med_period
+
+        topics = sorted(topic_datatypes.keys())
+
+        # process datatypes       
+        types = {}
+        for i, (datatype, md5sum, msg_def) in enumerate(sorted(datatype_infos)):
```
dirk-thomas: This should not use `enumerate` since it doesn't use the index `i`.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,177 @@ def close(self):
...
+            types[datatype] = md5sum
+            
+        # process topics
+        topics_t = {}
+        TopicTuple = collections.namedtuple("TopicTuple", ["type", "message_count", "connections", "frequency"])
+        for topic in topics:
+            topic_msg_count = topic_msg_counts[topic]
+            topics_t[topic] = TopicTuple(type =
```
dirk-thomas: The keyword arguments should not have spaces around the equal sign (PEP8).

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,177 @@ def close(self):
...
+                num_messages = 0
+                for c in self._chunks:
+                    for counts in c.connection_counts.values():
+                        num_messages += counts
+            else:
+                num_messages = sum([len(index) for index in self._connection_indexes.values()])
+            
+        return num_messages
+    
+    def get_timing_info(self):
```
dirk-thomas: Wouldn't it be easier to implement a `get_start_time` and `get_end_time` function instead?

csherstan:
What about the use of list() on line 542?

connections = list(self._get_connections(topic))

tools/rosbag/src/rosbag/bag.py
csherstan: Yeah, that's reasonable if we're to remove "duration".

tools/rosbag/src/rosbag/bag.py
csherstan: Ok.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

csherstan:
Just wondering if there's any more progress on this?

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,186 @@ def close(self):
...
+            
+    def get_compression_info(self):
+        """
+            Returns information about the compression of the bag
+            
+            returns (CompressionTuple):
+                compression - Type of compression used
+                uncompressed - Number of uncompressed Bytes
+                compressed - Number of compressed Bytes
+        """
```
dirk-thomas, line 413: Please change the docblock to follow the style in this file. Use the same way of formatting, mention all arguments and return values as well as document the types of each. (This applies to all docblocks in your patch.).

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,186 @@ def close(self):
...
+            compression_compressed = {}
+            for chunk_header in self._chunk_headers.values():
+                if chunk_header.compression not in compression_counts:
+                    compression_counts[chunk_header.compression] = 1
+                    compression_uncompressed[chunk_header.compression] = chunk_header.uncompressed_size
+                    compression_compressed[chunk_header.compression] = chunk_header.compressed_size
+                else:
+                    compression_counts[chunk_header.compression] += 1
+                    compression_uncompressed[chunk_header.compression] += chunk_header.uncompressed_size
+                    compression_compressed[chunk_header.compression] += chunk_header.compressed_size
```
dirk-thomas: Please refactor this conditional block to something like this which does not repeat the assigned values twice:

```
if chunk_header.compression not in compression_counts:
    compression_counts[chunk_header.compression] = 0
if chunk_header.compression not in compression_uncompressed:
    compression_uncompressed[chunk_header.compression] = 0
if chunk_header.compression not in compression_compressed:
    compression_compressed[chunk_header.compression] = 0

compression_counts[chunk_header.compression] += 1
compression_uncompressed[chunk_header.compression] += chunk_header.uncompressed_size
compression_compressed[chunk_header.compression] += chunk_header.compressed_size
```

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,186 @@ def close(self):
...
 
+            chunk_count = len(self._chunk_headers)
+
+            main_compression_count, main_compression = sorted([(v, k) for k, v in compression_counts.items()], reverse=True)[0]
+            compression = str(main_compression)
+
+            all_uncompressed = (sum([count for c, count in compression_counts.items() if c != Compression.NONE]) == 0)
```
dirk-thomas: Please simplify this logic a bit.

`all_uncompressed` is `True` if all non-compression counts are 0. And then you check for `not all_uncompressed`.

What about writing it like this:

```
any_compressed = any([count > 0 for c, count in compression_counts.items() if c != Compression.NONE])
if any_compressed
```

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,186 @@ def close(self):
...
+        
+        num_messages = 0
+        
+        if topic_filters is not None:
+            info = self.get_type_and_topic_info(topic_filters=topic_filters)
+            for topic in info.topics.values():
+                num_messages += topic.message_count
+        else:
+            if self._chunks:
+                num_messages = 0
```
dirk-thomas: This seems redundant with line 453.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,186 @@ def close(self):
...
+            
+        return num_messages
+    
+    def get_start_time(self):
+        """
+            Returns the start time of the bag.
+        """
+        
+        if self._chunks:
+            start_stamp = self._chunks[ 0].start_time.to_sec()
```
dirk-thomas: Please remove the extra space before the `0` inside the brackets. (Same again two lines down.)

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,186 @@ def close(self):
...
+
+            if self._connection_indexes_read:
+                stamps = [entry.time.to_sec() for entry in self._get_entries(connections)]
+                if len(stamps) > 1:
+                    periods = [s1 - s0 for s1, s0 in zip(stamps[1:], stamps[:-1])]
+                    med_period = _median(periods)
+                    if med_period > 0.0:
+                        topic_freqs_median[topic] = 1.0 / med_period
+
+        topics = sorted(topic_datatypes.keys())
```
dirk-thomas: Please remove this line and move the content where it is actually being used (line 575):

```
for topic in sorted(topic_datatypes.keys()):
```

tools/rosbag/src/rosbag/bag.py
```diff
@@ -401,7 +402,186 @@ def close(self):
...
+        topic_datatypes = {}
+        topic_conn_counts = {}
+        topic_msg_counts = {}
+        topic_freqs_median = {}
+        
+        for topic in topics:
+            connections = list(self._get_connections(topic))
+            
+            if len(connections) == 0:
+                continue
```
dirk-thomas, line 554: The existing usage of `self._get_connections()` implies that it must always return at least a single element which makes this check unnecessary.

dirk-thomas:
Beside the small comments on the code itself I will have to read and understand how the new functions compute their results. May be it would be good to also add tests for them.

tools/rosbag/src/rosbag/bag.py
csherstan, line 554: Sorry, I've been staring at this and don't see how that's the case. The generator in _get_connections could fail to ever yield if the topic used to filter is not actually in the connections, which would produce an empty list. This makes the check necessary. If there is no connection with the correct topic type then we skip over that connection. Also, in this case the previous comment above about removing the line 565 seems to not be correct either.

Am I missing something?

tools/rosbag/src/rosbag/bag.py
csherstan: Sorry, move the 'content'? Did you mean 'comment'? If so, that seems to me to be in the right place.

dirk-thomas: I meant "move the logic of this line" further down - into line 580 now.

tools/rosbag/src/rosbag/bag.py
dirk-thomas, line 554: You are right - when the `topic_filters` are not in the list of topics the function would yield nothing.

Please update the check to this then:

```
if not connections:
    continue
```

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected, but instead I simply moved the summations into the for loop above that was already traversing the same objects.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

tools/rosbag/src/rosbag/bag.py
csherstan: Corrected.

tools/rosbag/src/rosbag/bag.py
csherstan, line 554: corrected

csherstan:
Also added additional unit tests.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Manually squashed and merged in f3f1d07596cb65f1fe14a689a2d73653af55b8e2.

Thank you,

PR #515, opened 2014-10-24
Fix exceptions if python is built without ipv6 support.

aballier:
See:
http://bugs.python.org/issue16208
http://bugs.python.org/issue7735

### DIFF
--- a/tools/rosgraph/src/rosgraph/network.py
+++ b/tools/rosgraph/src/rosgraph/network.py
@@ -167,7 +167,10 @@ def is_local_address(hostname):
     :returns True: if hostname maps to a local address, False otherwise. False conditions include invalid hostnames.
     """
     try:
-        reverse_ips = [host[4][0] for host in socket.getaddrinfo(hostname, 0, 0, 0, socket.SOL_TCP)]
+        if use_ipv6():
+            reverse_ips = [host[4][0] for host in socket.getaddrinfo(socket.gethostname(), 0, 0, 0, socket.SOL_TCP)]
+        else:
+            reverse_ips = [host[4][0] for host in socket.getaddrinfo(socket.gethostname(), 0, socket.AF_INET, 0, socket.SOL_TCP)]
     except socket.error:
         return False
     local_addresses = ['localhost'] + get_local_addresses()
--- a/tools/roslaunch/src/roslaunch/core.py
+++ b/tools/roslaunch/src/roslaunch/core.py
@@ -90,7 +90,8 @@ def is_machine_local(machine):
     :returns: True if machine is local and doesn't require remote login, ``bool``
     """    
     try:
-        machine_ips = [host[4][0] for host in socket.getaddrinfo(machine.address, 0, 0, 0, socket.SOL_TCP)]
+        # If Python has ipv6 disabled but machine.address can be resolved somehow to an ipv6 address, then host[4][0] will be int
+        machine_ips = [host[4][0] for host in socket.getaddrinfo(machine.address, 0, 0, 0, socket.SOL_TCP) if isinstance(host[4][0],str)]
     except socket.gaierror:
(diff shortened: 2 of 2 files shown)


tools/rosgraph/src/rosgraph/network.py
```diff
@@ -167,7 +167,10 @@ def is_local_address(hostname):
     :returns True: if hostname maps to a local address, False otherwise. False conditions include invalid hostnames.
     """
     try:
-        reverse_ips = [host[4][0] for host in socket.getaddrinfo(hostname, 0, 0, 0, socket.SOL_TCP)]
+        if use_ipv6():
+            reverse_ips = [host[4][0] for host in socket.getaddrinfo(socket.gethostname(), 0, 0, 0, socket.SOL_TCP)]
+        else:
+            reverse_ips = [host[4][0] for host in socket.getaddrinfo(socket.gethostname(), 0, socket.AF_INET, 0, socket.SOL_TCP)]
```
dirk-thomas, line 173: This is the same way as it is already used in other parts of the code (https://github.com/ros/ros_comm/blob/4d2b423239452da9e6cc7bb4b1bbad22bd0376fa/tools/rosgraph/src/rosgraph/network.py#L234). The change looks good - was probably just forgotten in the original patch for IPv6.

tools/roslaunch/src/roslaunch/core.py
```diff
@@ -90,7 +90,7 @@ def is_machine_local(machine):
     :returns: True if machine is local and doesn't require remote login, ``bool``
     """    
     try:
-        machine_ips = [host[4][0] for host in socket.getaddrinfo(machine.address, 0, 0, 0, socket.SOL_TCP)]
+        machine_ips = [host[4][0] for host in socket.getaddrinfo(machine.address, 0, 0, 0, socket.SOL_TCP) if isinstance(host[4][0],str)]
```
dirk-thomas, line 94: What type is `host[4][0]` in your case? Are you using Python 2 or 3?

aballier, line 94: I tried with both 2.7.8 and 3.3.5 on my desktop (ipv6 enabled): they're 'str' with both; 

however, on the board where i disabled ipv6, I get this:

Python 2.7.8 (default, Oct 24 2014, 15:09:35) 
[GCC 4.8.3] on linux2

> > > import socket
> > > socket.has_ipv6
> > > False
> > > socket.getaddrinfo("localhost", 0, 0, 0, socket.SOL_TCP)
> > > [(10, 1, 6, '', (10, '\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')), (2, 1, 6, '', ('127.0.0.1', 0))]

so here, host[4][0] is int, causing the exception.

it can be fixed on my side by removing the '::1 localhost' line in /etc/hosts though, so that python doesn't even try to look that up, but i thought it would be useful to make the code more robust too.

dirk-thomas, line 94: Can you please add a comment above the line describing that with IPv6 disabled the value type is int.

aballier, line 94: done

dirk-thomas:
Manually merged to address PEP8 issue: 29265816d8774f87c12d90dd54dedf5f6815f8c8 f283fa94d0a10d6a207a4fe9eefc5b841fb85022

dirk-thomas:
This change breaks a unit test: http://jenkins.ros.org/job/devel-indigo-ros_comm/125/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/testReport/unit.test_roslaunch_launch/TestRoslaunchLaunch/test_validate_master_launch/

@aballier Why does you patch use `socket.gethostname()` instead of `hostname`?

And changed it back to use `hostname` again: bb134395915633484bf84bd3f153e4243d391ef6

PR #516, opened 2014-10-25
Fix: Failing to launch python scripts fails with different error

manzato:
If you forget to add the launch-prefix in a script, roslaunch notices it, but when wants to report the problem, fails to correctly use the exception object. This simple PR fixes this.


dirk-thomas:
Regression of 24e45419bdd4b0d588321e3b376650c7a51bf11c

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
LGTM

tfoote:
+1

wjwwood:
+1

PR #518, opened 2014-10-31
Do not use ifaddrs on Android as it is not natively supported

garyservin:


dirk-thomas:
With this patch the `allowed_hosts_` variable will not contain the IP addresses of the various network interfaces and therefore reject connections coming from those. Have you tested that this actually still works on Android with connections coming from other processes on the local host?

dirk-thomas:
@ros-pull-request-builder

garyservin:
@dirk-thomas sorry for the delay.

I've tested running two nodes (talker and listener) on the same phone, running the master both on the phone and on the pc and they worked.

I've also tested running one node on the pc and one on the phone and they were able to communicate.

Is this what you meant?

Issue #517, opened 2014-10-28
rospy: console output formatting using ROSCONSOLE_FORMAT


mjyc:
ROS C++ has [a feature](http://wiki.ros.org/rosconsole#Console_Output_Formatting) that controls output formatting using ROSCONSOLE_FORMAT environment variable. However, it seems like this feature is not in rospy and [this issue](https://github.com/osrf/capabilities/issues/24) shows possible update on rospy to provide the feature.

Is this feature still being considered to be implemented in rospy?
If not, I'm interested in implementing this feature myself & make PR, however, rospy seems quite complex and not sure where to implement this functionality. Any guidelines & suggestion to implement this would be very helpful.

dirk-thomas:
It is not planned to add this feature to rospy (hence there is no open ticket for yet). Please feel free to look into it but consider that any patch which requires changing API / default behavior will likely only be applied for future ROS versions.

`rospy` uses the Python `logging` package to handle most of the logging logic. You can find the entry points for logging in the `rospy.core` module: https://github.com/ros/ros_comm/blob/fab4840ef79f5334bcd043e057a1909183842159/clients/rospy/src/rospy/core.py#L119

I will mark this ticket with the milestone `untargeted` since the maintainers are not likely not spend any time on this feature.

wjwwood:
Reference: https://github.com/osrf/capabilities/issues/24

dirk-thomas:
The corresponding PR has been merged and released. Thanks!

PR #519, opened 2014-10-31
rosconsole format for rospy

mjyc:
quick & dirty solution for issue:
https://github.com/ros/ros_comm/issues/517

Since in python it uses [logging module](https://docs.python.org/2/library/logging.html) and in C++ it uses [log4cxx](http://logging.apache.org/log4cxx/index.html), I could not find 1 to 1 translation for some outputs. For example, `${thread}` and `${logger}` does not match well with that of C++.

Also I was not sure if this file was a good place to write this functionality. @dirk-thomas recommended to check out [this file](https://github.com/ros/ros_comm/blob/fab4840ef79f5334bcd043e057a1909183842159/clients/rospy/src/rospy/core.py#L119) but this file was easier to make change. Let me know if you have suggestions,

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -60,7 +60,7 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
 
     logname = logname or 'unknown'
     log_dir = rospkg.get_log_dir(env=env)
-
+    
     # if filename is not explicitly provided, generate one using logname
     if not filename:
         log_filename = os.path.join(log_dir, '%s-%s.log'%(logname, os.getpid()))
@@ -99,7 +99,7 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
         sys.stderr.write("WARNING: cannot load logging configuration file, logging is disabled\n")
         logging.getLogger(logname).setLevel(logging.CRITICAL)
         return log_filename
-
+    
     # pass in log_filename as argument to pylogging.conf
     os.environ['ROS_LOG_FILENAME'] = log_filename
     # #3625: disabling_existing_loggers=False
@@ -112,7 +112,7 @@ def makedirs_with_parent_perms(p):
     (existing) parent directory. This is useful for logging, where a
     root process sometimes has to log in the user's space.
     :param p: directory to create, ``str``
-    """
+    """    
     p = os.path.abspath(p)
     parent = os.path.dirname(p)
     # recurse upwards, checking to make sure we haven't reached the
@@ -127,7 +127,7 @@ def makedirs_with_parent_perms(p):
         if s.st_uid != s2.st_uid or s.st_gid != s2.st_gid:
             os.chown(p, s.st_uid, s.st_gid)
         if s.st_mode != s2.st_mode:
-            os.chmod(p, s.st_mode)
+            os.chmod(p, s.st_mode)    
 
 _logging_to_rospy_names = {
     'DEBUG': ('DEBUG', '\033[32m'),
@@ -162,11 +162,8 @@ class RosStreamHandler(logging.Handler):
             msg = msg.replace('${file}', str(record.pathname))
             msg = msg.replace('${line}', str(record.lineno))
             msg = msg.replace('${function}', str(record.funcName))
-            try:
-                from rospy import get_name
-                msg = msg.replace('${node}', get_name())
-            except ImportError:
-                msg = msg.replace('${node}', 'unavailable')
+            from rospy import get_name
+            msg = msg.replace('${node}', get_name())
             if self._get_time is not None and not self._is_wallclock():
                 msg += ' [%f]' % self._get_time()
             msg += '\n'
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -60,7 +60,7 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
 
     logname = logname or 'unknown'
     log_dir = rospkg.get_log_dir(env=env)
-    
+
     # if filename is not explicitly provided, generate one using logname
     if not filename:
         log_filename = os.path.join(log_dir, '%s-%s.log'%(logname, os.getpid()))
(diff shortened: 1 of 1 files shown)


tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -152,11 +152,26 @@ def __init__(self, colorize=True):
...
+            msg = os.environ['ROSCONSOLE_FORMAT']
+            msg = msg.replace('${severity}', level)
+            msg = msg.replace('${message}', str(record.getMessage()))
+            msg = msg.replace('${time}', str(time.time()))
+            msg = msg.replace('${thread}', str(record.thread))
+            msg = msg.replace('${logger}', str(record.name))
+            msg = msg.replace('${file}', str(record.pathname))
+            msg = msg.replace('${line}', str(record.lineno))
+            msg = msg.replace('${function}', str(record.funcName))
+            from rospy import get_name
```
dirk-thomas: This is not possible. You can't import `rospy` here - `ros_graph` does _not_ depend on `rospy`. It can also not add that dependency since `rospy` actually depends on `rosgraph`.

mjyc: What if I wrap the code using rospy with try & catch block? Similar to the [line 145-151 of this file](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rosgraph/src/rosgraph/roslogging.py#L145-L151)? Is there some special problem with using `get_name` function of rospy? I tried to modify [core.py](https://github.com/ros/ros_comm/blob/indigo-devel/clients/rospy/src/rospy/core.py) file, for example somehow try to overwrite this `emit` function from `core.py`.. would that be right direction? any other suggestions?

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -152,11 +152,29 @@ def __init__(self, colorize=True):
...
+            msg = msg.replace('${file}', str(record.pathname))
+            msg = msg.replace('${line}', str(record.lineno))
+            msg = msg.replace('${function}', str(record.funcName))
+            try:
+                from rospy import get_name
+                msg = msg.replace('${node}', get_name())
+            except ImportError:
+                msg = msg.replace('${node}', 'unavailable')
+            if self._get_time is not None and not self._is_wallclock():
+                msg += ' [%f]' % self._get_time()
```
dirk-thomas: I find it rather unintuitive to append the time at the end of the string. Would an additional replacement of something like `walltime` make more sense?

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -152,11 +152,29 @@ def __init__(self, colorize=True):
...
+            msg = msg.replace('${thread}', str(record.thread))
+            msg = msg.replace('${logger}', str(record.name))
+            msg = msg.replace('${file}', str(record.pathname))
+            msg = msg.replace('${line}', str(record.lineno))
+            msg = msg.replace('${function}', str(record.funcName))
+            try:
+                from rospy import get_name
+                msg = msg.replace('${node}', get_name())
+            except ImportError:
+                msg = msg.replace('${node}', 'unavailable')
```
dirk-thomas: Please reduce the redundant code part and only do the minimal stuff inside the try block:

```
try:
    from rospy import get_name
    node_name = get_name()
except ImportError:
    node_name = '<unknown_node_name>'
msg = msg.replace('${node}', node_name)
```

tools/rosgraph/src/rosgraph/roslogging.py
mjyc: @dirk-thomas  I agree with you on the time at the end of the string. I kept that to be consistent with [the original `msg` string generated by `emit` function](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rosgraph/src/rosgraph/roslogging.py#L153-L158).

Do we need to see both [time.time()](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rosgraph/src/rosgraph/roslogging.py#L153-L158) and [self._get_time()](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rosgraph/src/rosgraph/roslogging.py#L157) If not, how about this emit function?

```
    def emit(self, record):
        level, color = _logging_to_rospy_names[record.levelname]
        if self._get_time is not None and not self._is_wallclock():
            walltime = '[%f]' % self._get_time()
        else:
            walltime = '[%f]' % time.time()
        if 'ROSCONSOLE_FORMAT' in os.environ.keys():
            msg = os.environ['ROSCONSOLE_FORMAT']
            msg = msg.replace('${severity}', level)
            msg = msg.replace('${message}', str(record.getMessage()))
            msg = msg.replace('${time}', walltime)
            msg = msg.replace('${thread}', str(record.thread))
            msg = msg.replace('${logger}', str(record.name))
            msg = msg.replace('${file}', str(record.pathname))
            msg = msg.replace('${line}', str(record.lineno))
            msg = msg.replace('${function}', str(record.funcName))
            try:
                from rospy import get_name
                node_name = get_name()
            except ImportError:
                node_name = '<unknown_node_name>'
            msg = msg.replace('${node}', node_name)
            msg += '\n'
        else:
            msg = '[%s] [WallTime: %f]' % (level, walltime)
            msg += ' %s\n' % record.getMessage()
        if record.levelno < logging.WARNING:
            self._write(sys.stdout, msg, color)
        else:
            self._write(sys.stderr, msg, color)
```

Basically, it tries to use the timestamp from `self._get_time()` first, and if it is not available it resorts to `time.time()`.

dirk-thomas: I think it would be more flexible to expose both times through different variables. That way the user can choose what he wants to see (e.g. even both).

mjyc:
@dirk-thomas introduced `${walltime}` variable in addition to the exisiting `${time}` variable as you recommended. FYI, `${walltime}` is `time.time()` to be consistent with [the original code](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rosgraph/src/rosgraph/roslogging.py#L155).

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
The one failing test is due to the dependency `genpy` not yet being rereleased.

PR #521, opened 2014-11-03
Do not use Python when building for Android

garyservin:


dirk-thomas:
@ros-pull-request-builder

Issue #522, opened 2014-11-04
rospy: TCPROS, header `callerid` optional field


akru:
According to http://wiki.ros.org/ROS/TCPROS:
- **callerid** field is NOT required.

But when publisher response header does not contain this field, messages from publisher will not be received, because _tcpros_base.py_ has crashed.

```
[rospy.client][INFO] 2014-11-04 15:44:37,097: init_node, name[/rostopic_28955_1415105077037], pid[28955]
[xmlrpc][INFO] 2014-11-04 15:44:37,098: XML-RPC server binding to 0.0.0.0:0
[xmlrpc][INFO] 2014-11-04 15:44:37,098: Started XML-RPC server [http://krakov:44930/]
[rospy.init][INFO] 2014-11-04 15:44:37,098: ROS Slave URI: [http://krakov:44930/]
[rospy.impl.masterslave][INFO] 2014-11-04 15:44:37,098: _ready: http://krakov:44930/
[rospy.registration][INFO] 2014-11-04 15:44:37,099: Registering with master node http://krakov:11311
[xmlrpc][INFO] 2014-11-04 15:44:37,099: xml rpc node: starting XML-RPC server
[rospy.init][INFO] 2014-11-04 15:44:37,199: registered with master
[rospy.rosout][INFO] 2014-11-04 15:44:37,199: initializing /rosout core topic
[rospy.rosout][INFO] 2014-11-04 15:44:37,201: connected to core topic /rosout
[rospy.simtime][INFO] 2014-11-04 15:44:37,203: /use_sim_time is not set, will not subscribe to simulated time [/clock] topic
[rospy.internal][INFO] 2014-11-04 15:44:37,265: topic[/chatter] adding connection to [http://krakov:39475], count 0
[rospy.internal][WARNING] 2014-11-04 15:44:37,268: Unknown error initiating TCP/IP socket to krakov:41717 (http://krakov:39475): Traceback (most recent call last):
  File "/opt/ros/indigo/lib/python2.7/site-packages/rospy/impl/tcpros_base.py", line 557, in connect
    self.read_header()
  File "/opt/ros/indigo/lib/python2.7/site-packages/rospy/impl/tcpros_base.py", line 618, in read_header
    self._validate_header(read_ros_handshake_header(sock, self.read_buff, self.protocol.buff_size))
  File "/opt/ros/indigo/lib/python2.7/site-packages/rospy/impl/tcpros_base.py", line 585, in _validate_header
[...]

PR #525, opened 2014-11-04
Fix: TCPROS header validation crash when `callerid` header is not set

akru:
Pull request by issue https://github.com/ros/ros_comm/issues/522


dirk-thomas:
As mentioned in the referenced issue it seems that the same code is also used for the service connections where the header field is actually required. So the patch will likely need some conditional logic.

dirk-thomas:
The specific line was recently introduced (https://github.com/ros/ros_comm/commit/b64458c7e79719cbf48b4088326c7925182c5314#diff-a90cf003a88553002fc767ae33135f31R575) and is only used for the statistics stuff.

The service class is already ensuring that it must receive a caller id (e.g. https://github.com/ros/ros_comm/blob/6fac86e39fefaad58db92389098901cdd1a7bf48/clients/rospy/src/rospy/impl/tcpros_service.py#L222).

PR #530, opened 2014-11-09
add test code for multibyte string

k-okada:
I'm not sure if this is what you expected, In python, unicode data is transferred to str data.

### DIFF
--- a/test/test_rospy/test/rostest/test_basic_services.py
+++ b/test/test_rospy/test/rostest/test_basic_services.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# -*- coding: utf-8 -*-
 # Software License Agreement (BSD License)
 #
 # Copyright (c) 2008, Willow Garage, Inc.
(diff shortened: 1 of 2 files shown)


dirk-thomas:

k-okada:
```
19:13:15 Build timed out (after 120 minutes). Marking the build as failed.
19:13:15 Build was aborted
19:13:15 [WARNINGS] Skipping pub
```

the built was failed due to time out.

test/test_rospy/test/rostest/test_basic_services.py
```diff
@@ -213,6 +214,28 @@ def test_String_String(self):
...
+    def test_String_String_unicode(self):
+        from std_msgs.msg import String
+        from test_rospy.srv import StringString, StringStringRequest
+        from test_rospy.msg import Val
+        import sys
+        Cls = StringString
+        Req = StringStringRequest
+
+        for name in [STRING_CAT_SERVICE_NAKED, STRING_CAT_SERVICE_WRAPPED]:
+            if sys.version < '3':
```
dirk-thomas: This is a very fragile way to check the major Python version. Please use `sys.version_info.major` for a numeric comparison.

mikepurvis: I'm not sure the version check is even required. The `u` prefix syntax is valid in both Python 2 and 3. In 2, it produces a `unicode`, and in 3, a `str`. So I think the first block would be adequate for both versions.

k-okada: @dirk-thomas changed to use sys.version_info.major
@mikepurvis as far as I remember, interpreting unicode has been changed between python 2 vs python 3
- http://lucumr.pocoo.org/2014/1/5/unicode-in-2-and-3/
- https://docs.python.org/3/howto/unicode.html#python-s-unicode-support
- https://docs.python.org/2/howto/unicode.html#python-2-x-s-unicode-support

```
k-okada@p40-yoga:/tmp/ros_comm/test/test_rospy/test/rostest$ python
Python 2.7.6 (default, Oct 26 2016, 20:30:19)
[GCC 4.8.4] on linux2
>>> u'オープンソース' == 'オープンソース'
__main__:1: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them\
 as being unequal
False
```
and
```
k-okada@p40-yoga:~$ python3
Python 3.4.3 (default, Nov 17 2016, 01:08:31) 
[GCC 4.8.4] on linux
>>> u'オープンソース' == 'オープンソース'
True
```

dirk-thomas:

test/test_rospy/test/rostest/test_basic_services.py
mikepurvis: Right, but look at what happens in the two cases:

```
if sys.version_info.major < '3':
    # Python 2, returns a <unicode>.
    resp_req = self._test(name, Cls, Req(String(u'ロボット'), Val(u'机器人')))
else:
    # Python 3, returns a <str>.
    resp_req = self._test(name, Cls, Req(String('ロボット'), Val('机器人')))
```

But you can use the `u` prefix in both, so the following can replace the above:

```
# Returns <unicode> in Python 2, <str> in Python 3.
resp_req = self._test(name, Cls, Req(String(u'ロボット'), Val(u'机器人')))
```

mikepurvis:
Closing in favour of #985.

PR #532, opened 2014-11-13
Fix delay on detecting a running rosmaster with use_sim_time set

cmansley:


cmansley:
rosbag record will wait for a full second before recording data if the use_sim_time is true. Since, the timer uses  rostime, it will not begin executing until a clock is detected, then a full second will pass until the timer fires and checks for the master. This is an issue for scripts that run rosbag record. By checking first for the rosmaster, we can avoid the one second startup delay.

dirk-thomas:
Can you please provide a minimal example which I can run to reproduce the behavior before and after this patch?

cmansley:
Yes. Here is the setup.

You can generate a bag file like this, running each command in a separate terminal:

```
roscore
rostopic pub /message std_msgs/String "Hello" -r 10
rosbag record --all -O temp
```

You can tear down the setup and create a new setup, running each in a terminal in sequence:

```
roscore
rosparam set use_sim_time true
rosbag record --all -O delayed
```

rosbag record is waiting for a clock signal.

Then, we can 

```
rosbag play --clock temp.bag
```

A `rosbag info` on `delayed.bag` will show that approximately 1 second of `/message` is missing (10 msgs)

With the patch, the 10 messages are not missing.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
I also tried it locally and it works as advertized. Thank you.

Issue #533, opened 2014-11-19
Reconnect publisher-based logic can cause non-terminating leaked threads on bad subscribers


rethink-rlinsalata:
The `robust_connect_subscriber()` can get in a state where the local _subscriber_ side of the connection becomes invalid in some way, but the publisher still exists - which leads to never-ending attempts to reconnect.

The reconnect logic in `robust_connect_subscriber()` of `tcpros_pubsub.py` overwrites the flag to attempt reconnect based solely on the state of the target publisher, which can hide "FATAL" `Exception` errors from "Unknown errors" in the function's call to `TCPROSTransport().connect()` of `tcpros_base.py` that close the TCPROSTransport instance and according to the Exception handling shouldn't reconnect ("`# FATAL: no reconnection as error is unknown`"). This leads to a 'leaked' thread that starts a never-ending loop of always failing to connect, sleeping for an exponentially increasing amount of time, and then over-riding the 'no reconnection' state and always trying to reconnect - and the thread is never cleaned up, even when the parent Subscriber that started it goes away.

---

The chain of events/calls goes like this:
- 1) In `tcpros_pubsub.py`, `TCPROSHandler().create_transport()` starts the thread with the target of `robust_connect_subscriber(...)`, at ~L250.
- 2) `robust_connect_subscriber()` starts a while loop based on if the TCPROSTransport instance (`conn`) is `not done`, and then does a try/except wrapped call to `connect()`, at ~L168

``` python
    interval = 0.5
    while conn.socket is None and not conn.done and not rospy.is_shutdown():
        try:
            conn.connect(dest_addr, dest_port, pub_uri, timeout=60.)
        except rospy.exceptions.TransportInitError as e:
            rospyerr("unable to create subscriber transport: %s.  Will try again in %ss", e, interval)
            interval = interval * 2
            time.sleep(interval)

            # check to see if publisher state has changed
            conn.done = not check_if_still_publisher(resolved_topic_name, pub_uri)

```
- 3) In `tcpros_base.py`, if the `connect()` function throws a (standard) `Exception` type exception, it is caught in the following lines, logged as an "Unknown error" and then the TCPROSTransport instance is closed (which sets the `done` variable to `True` and various children to `None`), with a comment indicating there should be no reconnection. ~L563-569
  
  ``` python
  def connect(..):
      ...
      except Exception as e:
[...]

PR #538, opened 2014-12-15
fix infinitely retrying subscriber

dirk-thomas:
Addresses #533.


dirk-thomas:
Both failing tests are already the same on the _indigo-devel_ branch (e.g. because the dependency genpy has not yet being rereleased).

Issue #539, opened 2014-12-22
roslaunch: Support for $(empty arg) substitution


javanegmond:
I was working with openni2_launch when I came across this issue:
https://github.com/ros-drivers/openni2_launch/issues/19

I figured the best way to solve it would be a sustitution command `empty` that checks wether an arg has been set, such that one could use this kind of lines:
`<arg if="$(empty tf_prefix)" ... >`

I'm no python hero, but I decided to go down the rabbit hole and implement it; pull request will be added soon. I don't know how to properly test these kind of changes, so forgive me that I have only just tested it with simple example launch files as well as a changed openni2.launch.

There is a small point of discussion I guess: what to do with args that were not set at all. I have let them behave the same as with `$(arg arg)`, but perhaps it would be ok to say an non-set arg is actually empty.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #540, opened 2014-12-22
Added "empty" substitution command

javanegmond:
Implementation of my solution to #539.

Adds `_empty` and is really similar to `_arg`.

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -256,6 +256,31 @@ def _arg(resolved, a, args, context):
         raise ArgException(arg_name)
 
 
+def _empty(resolved, a, args, context):
+    """
+    process $(empty arg)
+    
+    :returns: "1" if arg is empty, "0" otherwise, ``str``
+    :raises: :exc:`ArgException` If arg invalidly specified
+    """
+    if len(args) == 0:
+        raise SubstitutionException("$(empty var) must specify an environment variable [%s]"%(a))
+    elif len(args) > 1:
+        raise SubstitutionException("$(empty var) may only specify one arg [%s]"%(a))
+    
+    if 'arg' not in context:
+        context['arg'] = {}
+    arg_context = context['arg']
+
+    arg_name = args[0]
+    if arg_name in arg_context:
+        arg_value = arg_context[arg_name]
+        is_empty = len(arg_value) == 0
+        return resolved.replace("$(%s)"%a, "1" if is_empty else "0")
+    else:
+        raise ArgException(arg_name)
+
+
 def resolve_args(arg_str, context=None, resolve_anon=True):
     """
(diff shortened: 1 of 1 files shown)


tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -290,6 +315,7 @@ def resolve_args(arg_str, context=None, resolve_anon=True):
         'optenv': _optenv,
         'anon': _anon,
         'arg': _arg,
+		'empty': _empty,
```
gavanderhoorn, line 318: Indentation.

mikeferguson:
+1 to concept, looks like there is a fix needed to indentation though.

dirk-thomas:
The `roslaunch` implementation with all the recursion and context switching is fragile at best. Other pull requests adding features like this (#296) have not been merged since it is difficult to judge if changes break the existing functionality.

Beside the actual code change it also requires significant documentation effort as well as adding unit tests to ensure that a new feature works correctly (and doesn't break existing functionality).

May be it would also help to "plan" what exact features are missing and should be added. `empty` seems to be a very specific test. Probably implementing a generic `compare` would be more flexible). May this could be discussed on the SIG roslaunch but it hasn't been active recently (http://wiki.ros.org/sig/roslaunch).

piyushk:
@dirk-thomas @gavanderhoorn What's the correct way to proceed to get this feature (or equivalent) in. Here's what I gathered from Dirk's comment:
1. Implementing `compare` might be more powerful than `empty`. I agree with this.
2. The PR needs to also add unit tests for the new feature.
3. What's the best place to discuss this feature. The SIG looks like it's been dead for a while, and I'll assume that more recent contributors are not subscribed to it. I feel like this feature would be extremely useful.

dirk-thomas:
There are two major concerns with this kind of PRs (there are multiple for roslaunch: e.g. #296, #382):
- Without having a clear "roadmap" what features are considered it is difficult to decide if they make sense the way they work, if they should be split into separate features or combined in a more generic feature and how they will interact with all the other functionality. This PR and the just closed #593 are a good example of that.
- Since the roslaunch code is rather fragile the chances for regressions are pretty big. And roslaunch is a core component for ROS so any regression would have a huge impact. On the other side there are not many tests (especially to ensure that more complex use cases are working). Therefore I hesitate to merge more complex changes (which sadly has the downside that interesting new features are not becoming available).

In order to move forward both concerns need to be worked on. But my major focus is currently on ROS 2 and therefore I can not spend any significant time on ROS 1 feature development. Therefore a group of people would need to actively discuss the "roadmap" aspect and determine which features should be integrated, which shouldn't and how they fit together. Additionally increasing the amount of tests which cover more use cases especially complex ones would help to ensure that changes don't break the existing behavior.

piyushk:
@dirk-thomas Thanks for the detailed response!

I'll come back and take a look at some features I'm interested in when I have more time.

dirk-thomas:
This feature should be possible once #707 gets merged. Therefore I will close this ticket.

Issue #544, opened 2015-01-08
rospy: Connection / Memory Leak, adding conn. to closed _SubscriberImpl


rethink-rlinsalata:
This is the third of a 3 piece series of memory/thread leak issues (#520, #533) I've been investigating, that mostly arise from race conditions during intensive barrages of rapidly creating new subscriptions and unsubscribing in a rospy node to topics coming from an already heavily-loaded (robot) node.

A problem can occur where a subscriber / connection is "leaked", such that the objects that run the connection are orphaned and never cleaned up, and the socket connection is still held open by the leaked thread, which is continuing to run the TCP receive / callback loop, even though the originating "Subscriber" is "closed" or gone.  Furthermore, new Subscribers to the same Topic and Publisher will not re-use those leaked, underlying connections like they normally would (since they are orphaned/leaked), thus leading to an ever increasing process memory in the rospy process that is Subscribing **_AND**_ in the process of the Publisher Node on the other end of the TCP socket connection (even a roscpp-based node) -- until one or both are killed by the kernel oom_killer.

The more technical description:  The two objects in charge of the connection are the underlying, per-topic `rospy.topics._SubscriberImpl` and the TCP connection object `rospy.impl.tcpros_base.TCPROSTransport`.  A major part of the problem _seems to be_ that somehow (race condition?) there is a new `TCPROSTransport` connection object being created and added to the `<_SubImpl>.connections[]` array of a `_SubImpl` _that has already been closed_ (`_SubImpl.closed==True`).

For the most direct evidence, we can see the bad/bug state by adding two log debug checks for the `_SubImpl.closed` flag in the `_SubImpl.add_connections()` and `_SubImpl.receive_callback()` functions, as seen in the diff [here](https://gist.github.com/rethink-rlinsalata/6b4c41e1580e00471107#file-leakadd_dbg-diff):  
**_SubscriberImpl.add_connection():**

``` diff
diff --cc clients/rospy/src/rospy/topics.py
index 43fab7a,43fab7a..803c23c
--- a/clients/rospy/src/rospy/topics.py
+++ b/clients/rospy/src/rospy/topics.py
@@@ -365,6 -365,6 +386,15 @@@
          """
          rospyinfo("topic[%s] adding connection to [%s], count %s"%(self.resolved_name, c.endpoint_id, len(self.connections)))
          with self.c_lock:
++            #LEAK-DBG
++            if self.closed:
++                rospyerr((
[...]

PR #545, opened 2015-01-08
protect against add_connection leak on already closed SubImpl

rethink-rlinsalata:
Protect against race condition where a new `TCPROSTransport`
connection is created and added to a `_SubscriberImpl` that is
already closed. Checks and returns False if closed so transport
creator can shutdown socket connection. Fixes issue #544.

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_pubsub.py
+++ b/clients/rospy/src/rospy/impl/tcpros_pubsub.py
@@ -261,8 +261,9 @@ class TCPROSHandler(rospy.impl.transport.ProtocolHandler):
         if sub.add_connection(conn): #pass tcp connection to handler
             return 1, "Connected topic[%s]. Transport impl[%s]"%(resolved_name, conn.__class__.__name__), dest_port
         else:
+            # _SubscriberImpl already closed or duplicate subscriber created
             conn.close()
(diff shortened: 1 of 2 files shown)


dirk-thomas:
@ros-pull-request-builder

rethink-rlinsalata:
Self-Notes:
- This doesn't solve the complete problem because it does not address the source of the race condition, just tries to add a catch to prevent the bad leak effects on the symptoms.
- It is unclear what the side-effects are to the caller (of `create_transport()`) that was attempting to create the new connection on the closing `_SubImpl` in the first place, if the connection is not added and returns False / an error.
- I originally tested this fix worked when I changed the return value at the end of `add_connection()` from `return True` to->> `return not self.closed`; however, I thought it would be "more correct" according to the description of the function's return value to do the check before adding to the connections array or the pollers, etc.  This does skip the "garbage collection" steps, but more than that, I'm worried there still may be no lock preventing the race condition of the _SubImpl closing while the connection is being added later in the function, and no one knowing about it....

Also, I modified the error message because I couldn't see how the `sub.add_connection()` could actually return False any other way (it used to say "duplicate subscriber" but there were no breaks nor return Falses nor even catches for exceptions that would let it return nothing).

dirk-thomas:
Closing in favor of #603.

PR #549, opened 2015-01-15
reset 'g_initialized' variable in shutdown routine

Karsten1987:
Resetting the variable `g_initialized` in shutdown routine.

By this, we can implement a re-initialization with different remapping options (such as `ROS_MASTER_URI`), since then the remap key `__master` is reconsidered.

https://github.com/ros/ros_comm/blob/indigo-devel/clients/roscpp/src/libros/init.cpp#L444

Thus, a simple change of a ros master in code could be done via:

``` c++
ros::shutdown();
std::string new_uri = "http://xxx.xxx.xxx.xxx:11311";
std::map< std::string, std::string > remap;
remap["__master"] = new_uri;
ros::init( remap, "node_name" );
```

While googling I found already existing requests for that:

http://answers.ros.org/question/131546/set-ros_master_uri-directly-in-code/

https://code.ros.org/trac/ros/ticket/3601

Is there any reason for not set this variable to `false`?

### DIFF
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -589,6 +589,7 @@ void shutdown()
 
   g_started = false;
   g_ok = false;
+  g_initialized = false;
   Time::shutdown();
 }
 


dirk-thomas:
The cpp client library of ROS was not designed to support re-initializing itself after a shutdown.
The initialization process involves many different parts (https://github.com/ros/ros_comm/blob/858c79a713847d28c8d772441a9b909d9b8153e3/clients/roscpp/src/libros/init.cpp#L444) and some of them are not side-effect free.

While resetting `g_initialized` might work in your specific use case it will very likely fail in general - especially when changing the input to the second `init()` invocation.
Therefore I think this patch should not be merged as-is (and changing `roscpp` to support this cleanly would require a very significant amount of effort which is not likely to be addressed in the near future).

I would recommend you to simply set the global variable back to `false` if that is what you want to do for your use case.

Karsten1987:
I understand that this might not be side-effect free, but would this then not indicate issues in the shutdown process? 

Further, since the variable `g_initialized` is declared as static inside the .cpp file, I am not aware of any c++ way to reset that. Could you help me how to do that?

vrabaud:
oh well, we'll wait for ROS 2 and branch for now then ....

wjwwood:
I would not recommend "reinitializing" because roscpp wasn't designed for it, but if it works for their use case and they understand that there be dragons this way, I don't see any great harm. The only drawback I can imagine is that people calling `init` a second time after `shutdown` currently get an exception, but after this pull request they would not.

---

So another idea is to add a setter for the `g_initialized` variable, so that you could do something like this:

``` c++
ros::shutdown();
ros::set_initialized_state(false);
std::string new_uri = "http://xxx.xxx.xxx.xxx:11311";
std::map< std::string, std::string > remap;
remap["__master"] = new_uri;
ros::init( remap, "node_name" );
```

The problem with this is idea is that if someone, incorrectly calls this function before doing `shutdown`, it may cause problems for `roscpp`. We would need to see how `g_initialized` is used to make sure that's ok. At any rate this would be an API addition, so it wouldn't affect any existing code, but it would let @Karsten1987 try what they will at their own risk.

wjwwood:
Also, @Karsten1987 what are you trying to achieve? I feel like the example you gave would be much easier to implement using a wrapper script which sets the master and runs the C++ node, then when you want to change, it stops the C++ node, changes the env and runs it again. `rviz` has a script like this:

https://github.com/ros-visualization/rviz/blob/indigo-devel/scripts/choose-master.py

Karsten1987:
Thanks for the replies. When I see the python script correctly, it starts the node with a subprocess, having it run in a separate context.
https://github.com/ros-visualization/rviz/blob/indigo-devel/scripts/choose-master.py#L237

I would like to avoid that, since the execution of the 'ros-client' is integrated in a bigger program/process, and should provide services depending on which rosmaster connects to it. 
The idea here would be to have a constant program flow and publish the information to whatever rosmaster. For this, I am looking for a method to on-the-fly change the rosmaster.

But as @vrabaud said, I can fork it for now and work with this. I was more interested why this variable is not set back to false.

dirk-thomas:
I looked into the init functions a bit more:
- `network::init` (https://github.com/ros/ros_comm/blob/35e5964da7b02a76a1bffb2dfcb08eb275594060/clients/roscpp/src/libros/network.cpp#L186): would not set `g_host` correctly if the initialization arguments change
- `master::init` (https://github.com/ros/ros_comm/blob/35e5964da7b02a76a1bffb2dfcb08eb275594060/clients/roscpp/src/libros/master.cpp#L50): would not set `g_uri` correctly if the initialization arguments change
- `this_node::init` (https://github.com/ros/ros_comm/blob/35e5964da7b02a76a1bffb2dfcb08eb275594060/clients/roscpp/src/libros/this_node.cpp#L75):  would not set `g_namespace` and `g_name` correctly if the initialization arguments change
- `file_log::init` (https://github.com/ros/ros_comm/blob/35e5964da7b02a76a1bffb2dfcb08eb275594060/clients/roscpp/src/libros/file_log.cpp#L60): would not set `g_log_directory` correctly if the initialization arguments change
- `param::init` (https://github.com/ros/ros_comm/blob/35e5964da7b02a76a1bffb2dfcb08eb275594060/clients/roscpp/src/libros/param.cpp#L781): would register the parameter update callback multiple times

Because of these side effects I am skeptical to allow users to re-initialize.

An alternative option would be to introduce a new function to reset the init flag.
This can then also print some warnings about the potentially bad side effects and the user can decide to ignore them.

But at least the repeated callback registration would need to be addressed since it even happens without changing any initialization arguments.

dirk-thomas:
I will close this ticket since it requires many more changes to allow reinitializing cleanly. If someone would like to work on the related parts above I would be happy to review pull requests which enable this cleanly.

Issue #548, opened 2015-01-15
Control maximum logfile size


adrienbarral:
Hello, 

I wrote a question on ros::answer some days ago about how to keep control of the logfile size. http://answers.ros.org/question/200611/is-it-possible-to-control-the-size-of-log-file/

I read carefully source code of roslaunch, and I understoud how the log file are managed. 

For me the problem, is that log file are generated by redirecting stdout and stderr of a subprocess in a file.

I propose to patch roslaunch to redirect stdout and stderr of the node subprocess in a python.logging.logger object that will be created with a RotatingFile handler. Paramters of this RotatingFile handlers are "maximum log file size" and "maximum log file count". These paramters will be two new attribute of the XML tag "node".

I will do a pull request very soon, thanks to review it and tell me what do you think about it.

Regards,

dirk-thomas:
I can see this feature being useful. You might also want to look at the pending PR #418. Since for this feature the flow to stdout / stderr needs to be controlled the same way it might make sense to build on-top of it.

adrienbarral:
Ok, I will integrate the feature of PR #418 in my patch. It's quite easy, I just have to add a logging.handler to the logger if output=="both" ....

I Think I will be ready to commit it today.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #551, opened 2015-01-16
This is the pull request for the patch of issue #548 and the already existing PR #418

adrienbarral:
I propose to patch roslaunch to redirect stdout and stderr of the node subprocess in a python.logging.logger object that will be created with a RotatingFile handler. Parameters of this RotatingFile handlers are "maximum log file size" and "maximum log file count". These paramters will be two new attributes of the XML tag "node".

As suggested by dirk-thomas, I did a "cherry pick" of the feature introduced by PR #418 . With the use of logging.logger this was quite fastforward...

As I said in the commit comments I added some unit test, and I did some tests by hand because I don't find an "easy" way to build automatic unit test for file Rotation etc ...

I am open to any comments ...

### DIFF
--- a/tools/roslaunch/src/roslaunch/core.py
+++ b/tools/roslaunch/src/roslaunch/core.py
@@ -444,7 +444,7 @@ class Node(object):
         :param remap_args: list of [(from, to)] remapping arguments, ``[(str, str)]``
         :param env_args: list of [(key, value)] of
         additional environment vars to set for node, ``[(str, str)]``
-        :param output: where to log output to, either Node, 'screen' or 'log', ``str``
+        :param output: where to log output to, 'screen', 'log' or both, ``str``
         :param cwd: current working directory of node, either 'node', 'ROS_HOME'. Default: ROS_HOME, ``str``
         :param launch_prefix: launch command/arguments to prepend to node executable arguments, ``str``
         :param required: node is required to stay running (launch fails if node dies), ``bool``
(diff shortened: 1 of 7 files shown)


wjwwood:
Patches issue #548 and incorporates #418.

(hash tags don't auto link when in the subject line)

IanTheEngineer:
Wow, this PR seems to address many of the issues I've encountered with `roslaunch` logging. Is there any reason it wasn't fully incorporated into `ros_comm` (other than the merge conflicts)?

IanTheEngineer:
Out of curiosity, has this not been merged due to failing tests? The tests themselves aren't entirely clear on what is wrong. The benefits of this PR are tangible - `stdout` and `stderr` in launch node logfiles, auto rotating of logs that get too big, etc. - so I'd prefer not to see this PR continue to languish.

dirk-thomas:
I can't comment on why it hasn't been reviewed earlier but I see the following issues with the current patch:
- It adds a required argument to the `LocalProcess` constructor which is not feasible. It should instead add an optional keyword argument to the end of the argument list in order to not break compatibility.
- Since the destination (`stdout` and `stderr`) is now always split (even when all the output should just go to the screen) I would expect the output order to get messed up (as commented on previous PRs: https://github.com/ros/ros_comm/pull/413#discussion_r12960601). I think it is reasonable to expect that with any change to the logging it shouldn't regress in cases which are currently supported.
- Last time it was able to run the tests it failed several tests.
- It has merge conflicts. When being updated it should be made against _kinetic-devel_ since that is the current release.

harakas:
Just a random comment: please add support for asynchronous logging into a file. As in prevent the node from blocking when file io is slow.

mohiz:
Does a plan exist for merging this PR in the near future? Log rotation is an important feature in a production level programs and I am wonder why this feature does not exist in ros log already. I am looking for a clean and standard way to use log rotation in my programs.

dirk-thomas:
Please see the [one year old comment](https://github.com/ros/ros_comm/pull/551#issuecomment-238383328) what needs to happen to get the patch merged. The PR will also need to retarget the latest ROS distro (currently *lunar-devel*).

jacknlliu:
@dirk-thomas any update?

dirk-thomas:

Issue #546, opened 2015-01-13
rostest: add_rostest function should expose the underlying target


czalidis:
`add_rostest` function should expose the underlying target so we can add dependencies between targets explicitly. This could be useful when we want to add a dependency between a rostest and a target defined by a `catkin_download_test_data` call.

Currently, when running `catkin_make run_tests_my_pkg` catkin scans the dependencies of rostest targets in `my_pkg` but does not know that a file needs to be downloaded in order to run the test, so the test fails.

On the other hand, we can run `catkin_make tests` before we call the `run_tests_my_pkg` and the `catkin_download_test_data` executes properly because it has a dependency to `tests` target, so maybe this is the way to do it.

wjwwood:
Hi @czalidis, I agree with you that it should expose the target somehow. The only question is how.

The target name is created deterministically in this code block and stored as `_target_name`:

https://github.com/ros/ros_comm/blob/indigo-devel/tools/rostest/cmake/rostest-extras.cmake.em#L34-51

Ever since https://github.com/ros/ros_comm/pull/456 this target name has been based on the arguments to rostest as well as the file name and test type and package, etc... I believe this target name namespacing is a good thing to preserve.

One option would be to allow the user to specify a target name, something like this:

``` cmake
add_rostest(... TARGET_NAME my_target_name ...)
```

But this would make the test names inconsistent and more likely to collide with another when being used with `catkin_make` (not a problem with `catkin_make_isolated`).

Therefore I would suggest just exposing the automatically generated name, which you could do with another free function like `get_rostest_name(output_var_name add_rostest_args)`. But if you did that you would need to redundantly pass the arguments to both this function and `add_rostest`.

So finally I would recommend adding an optional argument to `add_rostest` like `STORE_TARGET_NAME_IN` which then takes a variable name to store the target name in, so you would end up with something like this:

``` cmake
add_rostest(path/to/foo.test ... STORE_TARGET_NAME_IN foo_target_name ...)
add_dependencies(${foo_target_name} ...)
```

I don't think we have time to dig in and create a pull request for this, complete with tests and documentation. However, if you wanted to try and do so, hopefully this description of the issue will help you get started.

dirk-thomas:
Imo the API should allow the user to optionally pass a custom target name.
While the default makes sense the API should not restrict the user.

But the last proposal from William (enabling to retrieve the target name) is a good idea, too.
But we should name the argument `TARGET_VARIABLE` which is more in line with other CMake arguments.

[...]

PR #555, opened 2015-01-21
Add option DEPENDENCIES to add_rostest function

czalidis:
Discussion about this feature initially started at #546.

@wjwwood suggested that `add_rostest()` should have an option so that it could store the automatically generated target in a variable, in order to be used later by the user. I found that this was not possible, because `add_rostest()` computes a fraction of the underlying target that finally gets created by a call to a helper function, `catkin_run_tests_target()`, provided by the `catkin` package. The `_testname` that `add_rostest()` computes gets prefixed by various strings like `run_tests_${PROJECT_NAME}` or `run_tests_${PROJECT_NAME}_${type}`, so `add_rostest()` has no information about the the underlying target.
- https://github.com/ros/ros_comm/blob/indigo-devel/tools/rostest/cmake/rostest-extras.cmake.em#L34-51
- https://github.com/ros/catkin/blob/indigo-devel/cmake/test/tests.cmake#L92-144

@dirk-thomas suggested that the user should be able to pass a custom target name. This could be done the same way `catkin_add_gtest()` does, defining a custom target (passed by the user) as the target of the executable and calling the `catkin_run_tests_target()` to produce a target in the form of `run_tests_${PROJECT_NAME}_${type}...`, defining dependencies between the two targets.

https://github.com/ros/catkin/blob/indigo-devel/cmake/test/gtest.cmake#L25-59

In order to implement either of the suggestions, we need to change the `catkin_run_tests_target()` function, since it actually deals with the underlying targets.

My suggestion is to just add an option to `add_rostest()` to define dependencies and pass these dependencies to `catkin_run_tests_target()` as it can define dependencies between the targets that creates and user defined dependencies.

I can not imagine a use case where the user would need the target name, except for defining dependencies, since there is no executable or library to build. In my mind this solution addresses all potential use cases but of course I could be wrong.

If that PR gets merged I could add some documentation [here](http://docs.ros.org/indigo/api/catkin/html/howto/format2/rostest_configuration.html). Also as a note, the option to define rostests with multiple arguments is not documented anywhere, so could add that too.

### DIFF
--- a/tools/rostest/cmake/rostest-extras.cmake.em
+++ b/tools/rostest/cmake/rostest-extras.cmake.em
@@ -15,7 +15,7 @@ function(add_rostest file)
   set(ROSTEST_EXE "${rostest_DIR}/../../../@(CATKIN_GLOBAL_BIN_DESTINATION)/rostest")
 @[end if]@
 
-  cmake_parse_arguments(_rostest "" "WORKING_DIRECTORY" "ARGS" ${ARGN})
+  cmake_parse_arguments(_rostest "" "WORKING_DIRECTORY" "ARGS;DEPENDENCIES" ${ARGN})
 
   # Check that the file exists, #1621
   set(_file_name _file_name-NOTFOUND)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
LGTM.

@wjwwood ?

Probably http://wiki.ros.org/rostest/Writing should also be updated since it is the package specific documentation?

wjwwood:
+1, I agree that he should add a note about it on that wiki page.

PR #556, opened 2015-01-22
Crash in topic statistics with unset header timestamps

ericperko:
While trying to enable topic statistics on an existing ROS system, I encountered a crash that can take down pretty much any roscpp node (maybe rospy too, but I haven't tested that yet) if `enable_statistics` is set to true. With statistics enabled, any node that receives a message that includes a `std_msgs/Header` where the `header.stamp` is sufficiently older than the received time will generate an abort due an an uncaught exception inside roscpp.  This is trivially achievable while running with wall clock time and receiving a message with an unset `header.stamp` (defaulting to 0.0); this is actually how I stumbled on this issue, as some publishers in the system were not setting `header.stamp`.

The underlying issue is more complex than simply ignoring messages with a `header.stamp` of 0, as the real issue are messages where the message's stamp is far enough away from the computed mean that a ros::Duration cannot hold that difference (it's exacerbated by the square when accumulating to calculate the std dev, but I did encounter it on just `ros::Duration t = msg.stamp_age_mean - *it`).

I've got a minimal test case as well as a fix for the actual crash in roscpp in this PR so far. There's probably a better fix that might be able to actually preserve the statistics by using a larger size duration for intermediates, but regardless roscpp shouldn't start crashing on an existing system just because we turned on statistics :-) .

I haven't gotten a chance to look into whether rospy is affected by a similar issue, but if it uses the same pair of int32's to represent Duration objects I expect it is.

cc @forouher

### DIFF
--- a/clients/roscpp/src/libros/statistics.cpp
+++ b/clients/roscpp/src/libros/statistics.cpp
@@ -147,8 +147,16 @@ void StatisticsLogger::callback(const boost::shared_ptr<M_string>& connection_he
       msg.stamp_age_stddev = ros::Duration(0);
       for(std::list<ros::Duration>::iterator it = stats.age_list.begin(); it != stats.age_list.end(); it++)
       {
-        ros::Duration t = msg.stamp_age_mean - *it;
-        msg.stamp_age_stddev += ros::Duration(t.toSec() * t.toSec());
+        try
+        {
+          ros::Duration t = msg.stamp_age_mean - *it;
+          msg.stamp_age_stddev += ros::Duration(t.toSec() * t.toSec());
+        }
+        catch(std::runtime_error& e)
+        {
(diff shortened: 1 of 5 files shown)


forouher:
Yeah, I guess all that conversions back and forth between int32 and double time might cause problems...

Is it the "Duration is out of dual 32-bit range" runtime error from duration.cpp that you are seeing?

ericperko:
Yes, that's the error that I've been seeing.

dirk-thomas:
Thank you for the PR. It will take some time until we will can review and merge it.

scpeters:
I can confirm the test added in 09413aa6842a892a50a1c242ce8c318750f11371 fails, and that it is fixed by  5712356. I don't know if there are any other side-effects not captured in the test suite.

dirk-thomas:
Superseded by #600.

PR #563, opened 2015-02-02
roslz4: Improve py3k compatibility.

aballier:
Use explicit relative import, py3.3 and 3.4 refuse to import roslz4 otherwise.

This is trivial and backward compatible, but unfortunately very far from being sufficient for py3k support. I'm filling this to avoid letting it rot in my local tree.

### DIFF
--- a/utilities/roslz4/src/roslz4/__init__.py
+++ b/utilities/roslz4/src/roslz4/__init__.py
@@ -30,7 +30,7 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-from _roslz4 import *
+from ._roslz4 import *
 
 def compress(data):
     compressor = LZ4Compressor()


dirk-thomas:
Thanks, the modified import is indeed necessary.

@bcharrow Could you look into making the shared library also importable with Python 3?

bcharrow:
Thanks for the fix!  I'll take a look at improving Python 3 compatibility soon.

aballier:
Note: When I was talking about "very far from being sufficient for py3k support" I was talking about the whole ros_comm: I have like half the testsuite failing with it.
roslz4, on the other hand, seems perfectly ok after this patch.

dirk-thomas:
The import fails for me due to a missing symbol:

```
ImportError: dynamic module does not define init function (PyInit__roslz4)
```

@aballier Are you sure that you have setup the Python 3 environment correctly? Especially you either need a virtual env with all Python packages installed for Python 3 or the python3-\* Debian packages (which are not side-by-side installable with the Python 2 ones).

Most of ROS core should work perfectly fine nowadays - as well as most unit tests (with some issue left though).

aballier:
You're probably building with python2 headers: See https://github.com/ros/ros_comm/blob/indigo-devel/utilities/roslz4/src/_roslz4module.c#L411 ; that symbol is defined but under #if PY_MAJOR_VERSION >= 3

I'm not using debian nor virtualenv but gentoo where all sane python-x.y can be installed in parallel (if you think about it, this can quickly be a complete bloat but it is very useful for development and python porting); apart from setting python to the correct version, I also feed cmake with '-DPYTHON_EXECUTABLE="path to versionned python executable"' so maybe that's the reason it finds python2 headers for you.

As for the tests, "half the testsuite" was a bit exagerated :) I had much more failures without the above patch, now it is down to 8/129 failures. It's possible I'm just frightened but that kind of failures might be hard to fix:

```
ERROR: test_std_msgs_MultiArray (unit.test_genmsg_py.TestGenmsgPy)
  File "/usr/lib64/python3.3/site-packages/std_msgs/msg/_UInt8MultiArray.py", line 115, in serialize
    buff.write(struct.pack('<I%ss'%length, length, _x))
struct.error: argument for 's' must be a bytes object

During handling of the above exception, another exception occurred:

  File "/var/tmp/portage/dev-ros/ros_comm-1.11.10/work/ros_comm-1.11.10/test/test_rospy/test/unit/test_genmsg_py.py", line 356, in test_std_msgs_MultiArray
    UInt8MultiArray().serialize(buff)
  File "/usr/lib64/python3.3/site-packages/std_msgs/msg/_UInt8MultiArray.py", line 116, in serialize
    except struct.error as se: self._check_types(struct.error("%s: '%s' when writing '%s'" % (type(se), str(se), str(_x))))
  File "/usr/lib64/python3.3/site-packages/genpy/message.py", line 333, in _check_types
    raise SerializationError(str(exc))
genpy.message.SerializationError: <class 'struct.error'>: 'argument for 's' must be a bytes object' when writing ''

ERROR: test_std_msgs_SignedInt (unit.test_genmsg_py.TestGenmsgPy)
  File "/var/tmp/portage/dev-ros/ros_comm-1.11.10/work/ros_comm-1.11.10/test/test_rospy/test/unit/test_genmsg_py.py", line 266, in test_std_msgs_SignedInt
    maxp = long(math.pow(2, w-1)) - 1
NameError: global name 'long' is not defined
```

dirk-thomas:
You are right, I didn't switch to a full Python 3 build. I guess this PR is good to be merged then.

Most of the remaining Python 3 issues are related to ros/genpy#26. But we decided to not spend more time on looking into it for now.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Sorry for the multiple triggers. First was a flaky test. Second was the job failing...

dirk-thomas:
@ros-pull-request-builder

PR #569, opened 2015-02-11
Pause playback for topics

tdenewiler:
Added support for pausing when specified topics are about to be published. Can be turned on and off during playback.

It would be a nice feature of rosbag play if you had the ability to pause when specific topics are published. This way you could let a bag run until a particular sensor reading occurs and then see what the immediate effect of that data is on the rest of the system.

### DIFF
--- a/tools/rosbag/include/rosbag/player.h
+++ b/tools/rosbag/include/rosbag/player.h
@@ -90,6 +90,7 @@ struct ROSBAG_DECL PlayerOptions
 
     std::vector<std::string> bags;
     std::vector<std::string> topics;
+    std::vector<std::string> pausetopics;
 };
 
 
@@ -182,6 +183,8 @@ private:
     ros::NodeHandle node_handle_;
 
     bool paused_;
+
+    bool pause_for_topics_;
 
     ros::WallTime paused_time_;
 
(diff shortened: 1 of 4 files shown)


tdenewiler:
I ran unit tests twice, once after checking out `hydro-devel` branch and once after checking out `pause_topics` branch. The steps:
1. `catkin_make`
2. `catkin_make run_tests` so that test bag files are downloaded
3. `catkin_make run_tests_test_rosbag`

This resulted in all of the tests passing except for one (the same test fails in both branches):

```
[ RUN      ] rosbag.read_indexed_1_2_succeeds unknown file: Failure C++ exception with description "Error reading version line" thrown in the test body.
[  FAILED  ] rosbag.read_indexed_1_2_succeeds (0 ms)
```

The changes made in `pause_topics` do not appear to have caused any regressions based on existing unit tests.

dirk-thomas:
Thank you for the PR. It will take some time until we will can review and merge it.

tools/rosbag/include/rosbag/player.h
```diff
@@ -90,6 +90,7 @@ struct ROSBAG_DECL PlayerOptions
 
     std::vector<std::string> bags;
     std::vector<std::string> topics;
+    std::vector<std::string> pausetopics;
```
dirk-thomas: Can you please update this name and all its usage including the command line to something including a separator: either `pause_topics` or `paused_topics`.

gerkey:
Thanks again for the patch.

I'm replacing this PR by #595, targeting Indigo.  We're only putting critical bug fixes into Hydro these days.

Issue #416, opened 2014-05-22
roscpp: More failure-tolerant API for persistent services


adolfo-rt:
This issue is inspired by [this ros-answers question](http://answers.ros.org/question/167626/persistent-service-reconnection/).

To robustly use persistent service clients, one must currently add some boilerplate code for verifying client validity and reconnect if necessary. To avoid code duplication and to favor code brevity, it would be desirable to include this in roscpp itself.

Preserving a behavior that does not reconnect by default might be valuable for backwards-compatibility or for power users that wish to implement custom reconnection policies. With this in mind, a reconnect option could be specified on construction.

On a somewhat related note, it might also be relevant to expose an accessor that allows to query if a client is persistent or not. I opened #415 to track this separately.

dirk-thomas:
I will mark the issue with the milestone "untargeted" for now since I won't have time to work on it.

But please consider providing a pull request to add this feature. I will be happy to review and merge it.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #574, opened 2015-02-26
Introduce auto-reconnect logic for persistent services that lose connection

ntenenz:
Allows users to pass an optional two flags to have persistent services automatically attempt to reconnect. Default behavior remains unchanged.

See https://github.com/ros/ros_comm/issues/416

### DIFF
--- a/clients/roscpp/include/ros/service_client.h
+++ b/clients/roscpp/include/ros/service_client.h
@@ -49,16 +49,31 @@ public:
 
   /**
    * @brief Call the service aliased by this handle with the specified request/response messages.
+   * @param req The service request
+   * @param res The service response
+   * @param reconnect If true, persistent services will attempt to automatically reconnect if their connection is lost.
+   * @param timeout The time to wait for the service to become available after a disconnect. If timeout is -1 (default), the ndoe waits until it is shutdown.
    * @note The request/response message types must match the types specified in the templated call to NodeHandle::serviceClient()/service::createClient()
    */
   template<class MReq, class MRes>
-  bool call(MReq& req, MRes& res)
+  bool call(MReq& req, MRes& res, bool reconnect = false, ros::Duration timeout = ros::Duration(-1))
   {
     namespace st = service_traits;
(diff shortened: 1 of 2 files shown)


wjwwood:
Thanks for the pull. We're a little behind right now, but we'll get to this as soon as we are able.

ntenenz:
Any status update on this? It's a fairly simple patch.

wjwwood:
Looks straight forward to me, I haven't had time to build it locally or test it.

This changes the public interface, so as simple as it is, it isn't trivial. I'd recommend making a note about this addition on the wiki and we have to consider what version(s) of ROS to put it in. Also, it would be best if there were a test to show that it works. Ideally in the form of an automated test but otherwise in the form of an example package which with instructions on how to observer the feature in action.

@dirk-thomas can trigger a build on the farm at least to see if it breaks any code or tests off the bat.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:

dirk-thomas:
This seems to break several unit tests. Can you please look into them and update the PR accordingly.

mgrrx:
@ros/ros_comm-maintainers this PR seems interesting to me but I see some problems here.
In #738 we have a second PR for the roscpp ServiceClient. Both introduce timeouts that are set in two different ways and have different meanings: in this PR it's the timeout for the reconnect whereas in #738 it's the timeout for blocking service calls.
I'd prefer a different solution:
- a setter to specify the reconnect timeout for persistent connections
- an argument for the service call which specifies how long to wait for the message being sent and the response. This timeout should also include the time spent for reconnects

What do you think @dirk-thomas ?
We might also need a similar solution for rospy, but there it's not that straightforward, especially since service clients are called with *args, **kwargs arguments.

dirk-thomas:
Having two separate timeout arguments sounds good to me. Both could be passed as constructor arguments though or is there a reason why one should be passed to the constructor and the other via a setter?

PR #576, opened 2015-03-05
Add rosconsole command to change logger levels from command line

cmansley:
There is only the rqt_logger_level command for changing the logger
level for node. If you are doing remote work without an X session, you
have to send service commands directly or change the logger
configuration. This command line tool allows you to query and set the
logger level.

### DIFF
--- a/tools/rosconsole/CMakeLists.txt
+++ b/tools/rosconsole/CMakeLists.txt
@@ -121,3 +121,8 @@ if(CATKIN_ENABLE_TESTING)
     target_link_libraries(${PROJECT_NAME}-thread_test ${PROJECT_NAME})
   endif()
 endif()
+
+catkin_python_setup()
+
+catkin_install_python(PROGRAMS scripts/rosconsole
+  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
--- a/tools/rosconsole/CMakeLists.txt
+++ b/tools/rosconsole/CMakeLists.txt
@@ -123,6 +123,3 @@ if(CATKIN_ENABLE_TESTING)
 endif()
 
 catkin_python_setup()
-
-catkin_install_python(PROGRAMS scripts/rosconsole
-  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
--- a/tools/rosconsole/setup.py
+++ b/tools/rosconsole/setup.py
@@ -1,0 +1,13 @@
+#!/usr/bin/env python
+
+from distutils.core import setup
+from catkin_pkg.python_setup import generate_distutils_setup
+
+d = generate_distutils_setup(
+    packages=['rosconsole'],
+    package_dir={'': 'src'},
+    scripts=['scripts/rosconsole'],
+    requires=['genmsg', 'genpy', 'roslib', 'rospkg']
+)
+
+setup(**d)
--- a/tools/rosconsole/src/rosconsole/__init__.py
+++ b/tools/rosconsole/src/rosconsole/__init__.py
@@ -1,8 +1,5 @@
 #!/usr/bin/env python
 
-# make sure we aren't using floor division
-from __future__ import division, print_function
-
 NAME = 'rosconsole'
 
 import os
--- a/tools/rosconsole/src/rosconsole/__init__.py
+++ b/tools/rosconsole/src/rosconsole/__init__.py
@@ -1,0 +1,144 @@
+#!/usr/bin/env python
+
+# make sure we aren't using floor division
+from __future__ import division, print_function
+
+NAME = 'rosconsole'
+
+import os
+import sys
+import socket
+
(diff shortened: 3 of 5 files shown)


tools/rosconsole/CMakeLists.txt
```diff
@@ -121,3 +121,8 @@ if(CATKIN_ENABLE_TESTING)
     target_link_libraries(${PROJECT_NAME}-thread_test ${PROJECT_NAME})
   endif()
 endif()
+
+catkin_python_setup()
```
dirk-thomas, line 125: The package needs a `setup.py` file for this to work. You might have missed to add it to the repo.

tools/rosconsole/CMakeLists.txt
```diff
@@ -121,3 +121,8 @@ if(CATKIN_ENABLE_TESTING)
...
+
+catkin_install_python(PROGRAMS scripts/rosconsole
+  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
```
dirk-thomas: It probably makes sense to install this into the global bin folder rather then the package-specific libexec folder to be able to run it without `rosrun`. Instead of using `catkin_install_python` you can list it as a script in the setup.py file.

tools/rosconsole/src/rosconsole/__init__.py
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env python
+
+# make sure we aren't using floor division
+from __future__ import division, print_function
```
dirk-thomas: It looks like you need neither of these two.

tools/rosconsole/src/rosconsole/__init__.py
```diff
@@ -0,0 +1,144 @@
...
+
+Type rosconsole <command> -h for more detailed usage, e.g. 'rosconsole list -h'
+""")
+    sys.exit(getattr(os, 'EX_USAGE', 1))
+
+def main(argv=None):
+    if argv is None:
+        argv=sys.argv
+    # filter out remapping arguments in case we are being invoked via roslaunch
+    argv = rospy.myargv(argv)
```
dirk-thomas, line 127: Since all nodes will "race" when being invoked through `roslaunch` it might not be a good idea to use this script to try to change the logger level since the targeted node might not have started fast enough.

dirk-thomas:
This is a great new feature.

I made a couple of small comments inline. The "biggest" suggestion would be to use `argparse` instead of custom argv handling. This will likely also automatically improve the usage information for users.

tools/rosconsole/src/rosconsole/__init__.py
cmansley, line 127: Do you have a suggestion to fix this in the tool? Because it would be a shame to lose out on the functionality of the tool just because of the possibility of abuse.

cmansley:
I fixed some of the copy and paste errors and I added the setup.py file. Thanks for the feedback. 

With regards to argparse, I simply copied what exists in the other default ros command line tools. And at least on older systems, argparse is not part of the default python install.

tools/rosconsole/src/rosconsole/__init__.py
dirk-thomas, line 127: You can keep this line in the patch. I just wanted to state that using it in a launch file is imo not recommended. Therefore the comment is a bit misleading.

roslaunch is not designed to handle this and will very likely not be modified to do so. We are aware of that limitation and ROS 2 will for sure address those but that won't help in ROS 1.

dirk-thomas:
Actually nowadays it is part of it. Since this patch targets Indigo which implies Ubuntu Trusty which comes with Python 2.7 which includes argparse by default.

Anyway you don't have to use `argparse` - it was only meant as a suggestion. A lot of existing tools still use `optparse`.

tools/rosconsole/src/rosconsole/__init__.py
cmansley, line 127: Another copy and paste artifact. I can remove the comment.

tools/rosconsole/setup.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+
+from distutils.core import setup
+from catkin_pkg.python_setup import generate_distutils_setup
+
+d = generate_distutils_setup(
+    packages=['rosconsole'],
+    package_dir={'': 'src'},
+    scripts=['scripts/rosconsole'],
+    requires=['genmsg', 'genpy', 'roslib', 'rospkg']
```
dirk-thomas, line 10: Can you please explain why you used these four dependencies?

dirk-thomas:
Please make new Python code comply  with PEP8.

tools/rosconsole/setup.py
cmansley, line 10: I copied from all other default tools.

gerkey:
Thanks for the contribution.  I made a few small fixes and created #594 to replace this PR.

PR #578, opened 2015-03-11
force clock publishing in case of to heavy load on rosbag play (fix #82)

### LINKED ISSUE #82: rosbag: Thread publishing time should not block on a read operation
While playing back a bag, rosbag stops publishing clock for variable durations. It's likely that these pauses occur because rosbag cannot read the bag file from NFS quickly enough. Even so, rosbag should continue publishing clock so that other nodes can tell that the bag is still playing.

migration of trac ticket 3321:
  https://code.ros.org/trac/ros/ticket/3321

jennan:
resolves #82

### DIFF
--- a/test/test_rosbag/CMakeLists.txt
+++ b/test/test_rosbag/CMakeLists.txt
@@ -25,6 +25,8 @@ if(CATKIN_ENABLE_TESTING)
     FILENAME test/test_rosbag_latched_pub.bag MD5 ed70b8cbb8a53220434aeb24bccf7342)
   catkin_download_test_data(download_data_test_clock.bag http://tutux.web4all.fr/ros/clock.bag
     FILENAME test/clock.bag MD5 d81aca4472ad4220b282ecf9e2e557e7)
+  catkin_download_test_data(download_data_test_clock_alive.bag http://tutux.web4all.fr/ros/clock_alive.bag
+    FILENAME test/clock_alive.bag MD5 8b1b690b93161eec88396aa72957c5a8)
 
   catkin_add_nosetests(test/test_bag.py)
 
@@ -45,6 +47,9 @@ if(CATKIN_ENABLE_TESTING)
   configure_file(test/clock_sub.test.in 
                  ${PROJECT_BINARY_DIR}/test/clock_sub.test)
   add_rostest(${PROJECT_BINARY_DIR}/test/clock_sub.test)
+  configure_file(test/clock_sub_alive.test.in 
+                 ${PROJECT_BINARY_DIR}/test/clock_sub_alive.test)
+  add_rostest(${PROJECT_BINARY_DIR}/test/clock_sub_alive.test)
 
   add_rostest(test/latched_pub.test)
   configure_file(test/latched_sub.test.in ${PROJECT_BINARY_DIR}/test/latched_sub.test)
--- a/test/test_rosbag/CMakeLists.txt
+++ b/test/test_rosbag/CMakeLists.txt
@@ -26,7 +26,7 @@ if(CATKIN_ENABLE_TESTING)
   catkin_download_test_data(download_data_test_clock.bag http://download.ros.org/data/rosbag/test_clock.bag
     FILENAME test/clock.bag MD5 d81aca4472ad4220b282ecf9e2e557e7)
   catkin_download_test_data(download_data_test_clock_alive.bag http://tutux.web4all.fr/ros/clock_alive.bag
-    FILENAME test/clock_alive.bag MD5 8b1b690b93161eec88396aa72957c5a8)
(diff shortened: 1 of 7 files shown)


dirk-thomas:
Can you please provide a reproducible example for this which fails before the patch and works with it being applied. It would be best to create the example in the form of a unit test.

dirk-thomas:
ping

jennan:
et voilà ! :) Please let me know if/when I could remove the test bag from my personal server.

jennan:
This fix reveals another bug in _rosbag play_ with **--keep-alive** mode : _/clock_ is not published after all messages (of a heavy rosbag) have been published.

Indeed, fix https://github.com/jennan/ros_comm/commit/adf63f6b1fa158683c64bd48d0695fd352e73081 forces publishing _/clock_ when the player is overwhelmed by messages to publish (and don't find time to publish some _/clock_ messages). The problem is that, after playing all bag messages, the simulated time is so much shifted (compared to what it should be if everything went well, without heavy load) that the Wall Clock time corresponding to _simulated time + 10 sec._ (see [player.cpp:L391](https://github.com/ros/ros_comm/blob/b7796fea666aac7b330aad870f7e7b994eda8daf/tools/rosbag/src/player.cpp#L391)) is already passed and thus no _/clock_ message is published. But if no _/clock_ message is published, no update of the simulated time is performed, and so on... leading to no _/clock_ message at all.

This situation does not happen often, but it happened to me with some rosbags of kinect data. I provide a test with an handcrafted rosbag than should reveal this issue (if it is not to much computer-dependent).

dirk-thomas:
I have uploaded the test bagfile to http://download.ros.org/data/rosbag/test_clock.bag Please update the PR to use that.

jennan:
There is a second test bagfile to move for the second fix about --keep-alive option.

dirk-thomas:
Does the second bag file need to be 250 MB? Could you please try to reproduce the problem with a much smaller bag file?

jennan:
Well... in fact it's a compressed bag of about 50GB (don't remember exactly) :-). As I said, the second bug only appears when playback accumulates a delay for the clock which is more than 10sec. It appears with my experimental data (kinect data), at the end of playing back a bag of 40s of kinect data (about 700 MB with LZ4 compression). I tried to make a "minimal" testcase, but one has to overwhelm the player with data for more than 30sec to get this bad behaviour happening... As the second bag is artificially created, would it be better if I add the python script to make it (slow) instead of the resulting bag ?

dirk-thomas:
If you could create a "smaller" example which reproduces the problem that would be highly appreciated. Otherwise every test build needs to download these test data which slows it down.

jennan:
The new bag should be ok, thanks to bz2 compression (15Mo).

jennan:
Hello, I would like to know if a can remove the file... and what to expect concerning the integration of this fix :).

dirk-thomas:
@ros-pull-request-builder

test/test_rosbag/CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ if(CATKIN_ENABLE_TESTING)
     FILENAME test/test_future_version_2.1.bag MD5 ddcb5e8711b16514eb640330ed93c332)
   catkin_download_test_data(download_data_test_rosbag_latched_pub.bag http://download.ros.org/data/rosbag/test_rosbag_latched_pub.bag
     FILENAME test/test_rosbag_latched_pub.bag MD5 ed70b8cbb8a53220434aeb24bccf7342)
+  catkin_download_test_data(download_data_test_clock.bag http://download.ros.org/data/rosbag/test_clock.bag
+    FILENAME test/clock.bag MD5 d81aca4472ad4220b282ecf9e2e557e7)
+  catkin_download_test_data(download_data_test_clock_alive.bag http://tutux.web4all.fr/ros/clock_alive.bag
```
dirk-thomas, line 28: Sorry, the file is finally available on ros.org: see http://download.ros.org/data/rosbag/clock_alive.bag

Can you please update the PR to use that URL.

dirk-thomas:

dirk-thomas:
This seems to make several tests fail. Please look into updating the patch.

mikepurvis:
This seems like a great change, but it would have to retarget **kinetic-devel** and pass tests.

@jennan I don't think I have write access to your PR branch, and I'll totally understand if you're fed up with waiting around. Do you mind if I create a new PR to carry this forward?

mikepurvis:
Closing in favour of #982.

jennan:
Hi, thank you for taking over this pull request. Unfortunately it is unlikely that I will have time to look at it again so...  good luck ;-).

mikepurvis:
Sounds good. Thanks again for the contribution, and sorry for the long turn around time here.

PR #580, opened 2015-03-11
Support non-standard GTest location.

### LINKED ISSUE #579: Rostest with non-standard GTest location and no catkin_add_gtest does not compile
`add_rostest_gtest` currently relies upon Catkin's gtest.cmake file to define `GTEST_FOUND`. If that variable is found, rostest will create the new test target. However, what is perhaps unintended is that `add_rostest_gtest` also relies on Catkin's `catkin_add_gtest` function to include any non-standard gtest include and lib directories (note that Catkin supports gtest installs found via `find_package`, so this is a likely scenario).

If one were to create a package with only `add_rostest_gtest` tests and _no_ `catkin_add_gtest` calls, along with using find_package to get a GTest installed in a non-standard location, the rostest compilation will fail due to an inability to find GTest headers and libs.

kyrofa:
This PR resolves #579 simply by making use of the include and lib paths given by Catkin's gtest.cmake file.


dirk-thomas:
This patch should be obsolete with catkin providing a function to build an executable linked against gtest but not registering it as a test: https://github.com/ros/catkin/pull/726

Issue #571, opened 2015-02-19
ROS_DEBUG_DTHROTTLE


galou:
Sometimes I want to show a message that a node is waiting for a message on a specific topic. For that, I use ROS_DEBUG_THROTTLE.

However, ROS_DEBUG_THROTTLE shows a message on first call. I would like to have a Delayed (hence DTHROTTLE, name to be discussed), that would first show the message after the specified rate elapsed.

I'm OK to write a PR after the name has been discussed.

dirk-thomas:
I am not sure I understand the expected behavior. Can you please clarify. If the macro is only being called once but not again later you want the message to be sent after a delay?

galou:
The behavior would be the same as ROS_DEBUG_THROTTLE with the exception of the time of the first message:
ROS_DEBUG_THROTTLE(1, ...) sends at t = 0, 1, 2, 3, ...
ROS_DEBUG_DTHROTTLE(1, ...) would send at t = 1, 2, 3, ...
where t = 0 is the time where the macro is called for the first time with this message.

Does it make sense?

dirk-thomas:
Ok, I think I get the idea.

The current `THROTTLE` macros are described with:

```
Throttled output will print a message at most once per "period".
```

The implementation is pretty straight forward. For each message the current time is compared to the last printed message and if the duration exceeds the interval the message is actually printed and the last timestamp updated.

How would you described the new macro and how would you imagine the implementation to work?

galou:
The implementation is straight forward, I think,

in include/ros/console.h, copy ROS_LOG_THROTTLE to ROS_LOG_DTHROTTLE and replace

``` c++
    static double last_hit = 0.0; \
    ::ros::Time now = ::ros::Time::now(); \
```

with

``` c++
    ::ros::Time now = ::ros::Time::now(); \
    static double last_hit = now.toSec(); \
```

For the description, I would write something like:

```
DThrottled output will print a message at most once per "period" and at first after "period".
```

galou:
I didn't really replied one question: "If the macro is only being called once but not again later you want the message to be sent after a delay?"
No the user would need to repeatedly call the macro, either the macro outputs something right at the moment when it's called or nothing.
Typically the macro would be called inside a loop where no message need to be sent before a first "period" elapsed, in the case that the problem solves by itself in this time interval.

dirk-thomas:
[...]

PR #581, opened 2015-03-17
Add DELAYED_THROTTLE versions of log macros

galou:
Add for example the ROS_DEBUG_DELAYED_THROTTLE to print a throttled message where the first message will not appear before a time delay of `rate` has elapsed.

Cf. discussion #571.


wjwwood:
Looks reasonable to me.

tfoote:
It looks good, but we should target it for jade.  This is a large low level API change.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@tfoote Since the PR does not change any existing API but only introduces a new group of macros I would merge this to _Indigo_. Worst case which can happen is it doesn't build (since the tests passed that not the case) or the new feature does not work as expected.

Issue #582, opened 2015-03-18
Confusing documentation NodeHandle::subscribe(...) and ::publisher(...)


Bardo91:
In [node_handle.h](https://github.com/ros/ros_comm/blob/indigo-devel/clients/roscpp/include/ros/node_handle.h), publisher's and subscriber's creation documentation sais that:

```
  On failure, an empty Publisher/Subscriber which can be checked with:
  if (handle)
  {
  ...
  }
```

`handle` seems to be a NodeHandle, however is the subscriber's or publisher's handle. That is a bit confusing. I Suggest change `handle`to `subscriber`, `subscriber_handle` or similar.

dirk-thomas:
That sounds like a good clarification. The doc could also just include the definition of the variable. Then it would be even more obvious what it stands for.

Please consider providing a PR with the doc change.

PR #583, opened 2015-03-18
Clarified documentation about checking Publishers, Subscribers and AdvertiseServices.

Bardo91:
As suggested in [https://github.com/ros/ros_comm/issues/582], clarified methodology of checking if Publisher, Subscribers and AdvertiseServices are valid. Added definition of objects in order not to confuse old "handle" variable of type Publisher/Subscriber/AdvertiseSevice with NodeHandle type.

### DIFF
--- a/clients/roscpp/include/ros/node_handle.h
+++ b/clients/roscpp/include/ros/node_handle.h
@@ -855,7 +855,7 @@ bool Foo::callback(std_srvs::Empty& request, std_srvs::Empty& response)
 ros::NodeHandle nodeHandle;
 Foo foo_object;
 ros::ServiceServer service = nodeHandle.advertiseService("my_service", &Foo::callback, &foo_object);
-if (service)  // Enter if advertise service is valid
+if (service)  // Enter if advertised service is valid
 {
 ...
 }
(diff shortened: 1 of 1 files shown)


clients/roscpp/include/ros/node_handle.h
```diff
@@ -289,7 +289,9 @@ namespace ros
    * \return On success, a Publisher that, when it goes out of scope, will automatically release a reference
    * on this advertisement.  On failure, an empty Publisher which can be checked with:
 \verbatim
-if (handle)
+ros::NodeHandle nodeHandle;
+ros::publisher pub = nodeHandle.advertise<std_msgs::Empty>("my_topic", 1, (ros::SubscriberStatusCallback)callback);
+if (pub)	// Enter if publisher is valid
```
dirk-thomas: Please replace the tab before the `//` with two spaces. Same for all the other blocks.

Bardo91: Done in next commit.

clients/roscpp/include/ros/node_handle.h
```diff
@@ -810,7 +848,14 @@ ros::ServiceServer service = handle.advertiseService("my_service", &Foo::callbac
...
 \verbatim
-if (handle)
+bool Foo::callback(std_srvs::Empty& request, std_srvs::Empty& response)
+{
+  return true;
+}
+ros::NodeHandle nodeHandle;
+Foo foo_object;
+ros::ServiceServer service = nodeHandle.advertiseService("my_service", &Foo::callback, &foo_object);
+if (service)  // Enter if advertise service is valid
```
dirk-thomas: Spelling: `Enter if advertised service is valid` (adding a `d` after `advertise`).

dirk-thomas:
Thank you, squashed and merged in cc9d91a3533d027621160ea2e23b55b6ed1f44f8.

PR #588, opened 2015-03-25
Remaining time for Rate

Adirio:
Offer the remaining time of the Rate class so that actions can be taken in critic applications

### DIFF
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -59,6 +59,19 @@ class Rate(object):
         self.last_time = rospy.rostime.get_rostime()
         self.sleep_dur = rospy.rostime.Duration(0, int(1e9/hz))
 
+    def remaining(self):
+        """
+        Get the time ramaining for rate to sleep
+        """
+        curr_time = rospy.rostime.get_rostime()
+        # detect time jumping backwards
+        if self.last_time > curr_time:
+            self.last_time = curr_time
+
+        # calculate remaining time
+        elapsed = curr_time - self.last_time
+        return self.sleep_dur - elapsed
+
     def sleep(self):
         """
         Attempt sleep at the specified rate. sleep() takes into
@@ -70,14 +83,7 @@ class Rate(object):
         @raise ROSTimeMovedBackwardsException: if ROS time is set
         backwards
         """
-        curr_time = rospy.rostime.get_rostime()
-        # detect time jumping backwards
-        if self.last_time > curr_time:
-            self.last_time = curr_time
-
-        # calculate sleep interval
-        elapsed = curr_time - self.last_time
-        sleep(self.sleep_dur - elapsed)
+        sleep(self.remaining())
         self.last_time = self.last_time + self.sleep_dur
 
         # detect time jumping forwards, as well as loops that are
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -85,6 +85,7 @@ class Rate(object):
         @raise ROSTimeMovedBackwardsException: if ROS time is set
         backwards
         """
+        curr_time = rospy.rostime.get_rostime()
         sleep(self.remaining())
         self.last_time = self.last_time + self.sleep_dur
 
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -59,13 +59,14 @@ class Rate(object):
         self.last_time = rospy.rostime.get_rostime()
         self.sleep_dur = rospy.rostime.Duration(0, int(1e9/hz))
 
-    def remaining(self):
+    def _remaining(self, curr_time):
         """
-        Get the time remaining for rate to sleep.
+        Calculate the time remaining for rate to sleep.
+        @param curr_time: current time
+        @type  curr_time: L{Time}
         @return: time remaining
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@ros-pull-request-builder

clients/rospy/src/rospy/timer.py
```diff
@@ -59,6 +59,19 @@ def __init__(self, hz):
         self.last_time = rospy.rostime.get_rostime()
         self.sleep_dur = rospy.rostime.Duration(0, int(1e9/hz))
 
+    def remaining(self):
+        """
+        Get the time ramaining for rate to sleep
```
dirk-thomas: Please fix the spelling (`remaining`) and finish the sentence with a period (`.`). Please also the return value and type.

dirk-thomas:
@ros-pull-request-builder

Adirio:
As this was the first time I pulled a request for ROS (and also the first time I used GitHub) I would apreciate some feedback.
I would also like to know if this will be included in the official build and if so, when will it be done.

dirk-thomas:
Thank you for contributing.

Any pull request is review by the maintainer(s) and (for this repo) also runs the unit tests associated to the packages in this repo. Usually you will receive feedback and after iterating the PR gets merged to the branch it is proposed for. In regular intervals new releases are being made - the time frame varies significantly from repo to repo.

Since `ros_comm` is at the very core releases are less frequent since it requires all ROS packages depending on it to be rebuild in the farm. (The next release will likely be early in April). Shortly after the release the binary packages are available in the `shadow-fixed` repository and move on to the `public` repo after a soak time (which is usually around 1-2 weeks).

dirk-thomas:
Thanks, you have already addressed the cosmetic comments I made. But it seems like you have not actually tried to run your changed code.

The build farm reports several failing test. I just looked at one of them (http://jenkins.ros.org/job/_pull_request-indigo-ros_comm/166/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/testReport/junit/__main__/TestRospyClientOnline/test_Rate/) and the changed code seems to be responsible for it. I will comment on the specific line.

clients/rospy/src/rospy/timer.py
```diff
@@ -70,14 +85,7 @@ def sleep(self):
         @raise ROSTimeMovedBackwardsException: if ROS time is set
         backwards
         """
-        curr_time = rospy.rostime.get_rostime()
```
dirk-thomas, line 73: With the removal of the local variable `curr_time` line 93 (https://github.com/ros/ros_comm/pull/588/files#diff-098242a094fe0fb6b870f1949712b77fR93) now fails.

Adirio:
It still finds the "global curr_time does not exist error" when now it is properly defined.

dirk-thomas:
The failing integration job was started long before you committed the fix. This (@ros-pull-request-builder) will trigger another integration build.

dirk-thomas:
@Adirio It passes all unit tests now and the patch looks good to me.

clients/rospy/src/rospy/timer.py
```diff
@@ -59,6 +59,21 @@ def __init__(self, hz):
         self.last_time = rospy.rostime.get_rostime()
         self.sleep_dur = rospy.rostime.Duration(0, int(1e9/hz))
 
+    def remaining(self):
```
esteve: This just a suggestion (feel free not to implement it), but this method could take an optional `curr_time` argument, so that you could reuse curr_time in line 73 and be slightly more accurate. E.g.:

``` python
def remaining(self, curr_time_time=None):
    if curr_time is None:
        curr_time = rospy.rostime.get_rostime()
    # detect time jumping backwards
    if self.last_time > curr_time:
        self.last_time = curr_time

    # calculate remaining time
    elapsed = curr_time - self.last_time
    return self.sleep_dur - elapsed
```

and then you would call this inside `sleep` like this:

``` python
    curr_time = rospy.rostime.get_rostime()
    sleep(self.remaining(curr_time=curr_time))
    self.last_time = self.last_time + self.sleep_dur
```

wjwwood:
Thanks for the contribution @Adirio, I just had one comment. With this change there is a new redundant call to `rospy.rostime.get_rostime()`. You could structure it such the the implementation of `remaining` was moved to a function named something like `def _remaining(self, curr_time)` and then restructure the code like so:

``` python
def remaining(self):
    curr_time = rospy.rostime.get_rostime()
    return self._remaining(curr_time)

def _remaining(self, curr_time):
    ...

def sleep(self):
    curr_time = rospy.rostime.get_rostime()
    sleep(self._remaining(curr_time))
    ...
```

This way there is never an extra call to `rospy.rostime.get_rostime()`. Not sure it makes a different one way or another, the performance impact is probably negligible, but it was just a thought I had.

Adirio:
Both of you suggestions would work, private method sounded like legitimate way to do it
Could you restart the test?

dirk-thomas:
@ros-pull-request-builder

esteve:
Changes look good to me, thanks @Adirio !

tfoote:
lgtm

dirk-thomas:
@Adirio Thank you for your first patch. I have squashed your changes in a single commit and cherry-picked them: 91f4a452c00d6fece236d3739517c958ccf2f14f It will be available in the next patch release (1.11.11) which should be released in about 1-2 weeks.

PR #590, opened 2015-03-30
Added a parameter to setPeriod to enable changing a Timer's period while taking into account the current cycle's elapsed time

sterlingm:
Currently, when you call setPeriod for a Timer, it starts the new cycle at the time setPeriod is called. So the next time the callback function is called will be current_elapsed_time + new_period_time. In my research project, it would be extremely useful for me to able to change the period for a Timer and have that change take effect on the immediate next cycle, rather than "starting over" each time I call setPeriod.

In terms of code changes, I just added a parameter to setPeriod to indicate whether or not you want the Timer to "reset". I put in a default value of true so any existing code using that function should not be affected.

### DIFF
--- a/clients/roscpp/include/ros/timer.h
+++ b/clients/roscpp/include/ros/timer.h
@@ -67,8 +67,9 @@ public:
 
   /**
    * \brief Set the period of this timer
+   * \param reset Whether to reset the timer. If true, timer ignores elapsed time and next cb occurs at now()+period
    */
-  void setPeriod(const Duration& period);
+  void setPeriod(const Duration& period, bool reset=true);
 
   bool isValid() { return impl_ && impl_->isValid(); }
   operator void*() { return isValid() ? (void*)1 : (void*)0; }
@@ -99,7 +100,7 @@ private:
 
     bool isValid();
     bool hasPending();
-    void setPeriod(const Duration& period);
+    void setPeriod(const Duration& period, bool reset=true);
 
     void start();
     void stop();
--- a/clients/roscpp/include/ros/timer_manager.h
+++ b/clients/roscpp/include/ros/timer_manager.h
@@ -93,7 +93,7 @@ public:
   void remove(int32_t handle);
 
   bool hasPending(int32_t handle);
-  void setPeriod(int32_t handle, const D& period);
+  void setPeriod(int32_t handle, const D& period, bool reset=true);
 
   static TimerManager& global()
   {
@@ -419,7 +419,7 @@ void TimerManager<T, D, E>::updateNext(const TimerInfoPtr& info, const T& curren
 }
 
 template<class T, class D, class E>
-void TimerManager<T, D, E>::setPeriod(int32_t handle, const D& period)
+void TimerManager<T, D, E>::setPeriod(int32_t handle, const D& period, bool reset)
 {
   boost::mutex::scoped_lock lock(timers_mutex_);
   TimerInfoPtr info = findTimer(handle);
@@ -431,9 +431,33 @@ void TimerManager<T, D, E>::setPeriod(int32_t handle, const D& period)
 
   {
     boost::mutex::scoped_lock lock(waiting_mutex_);
+  
+    if(reset)
+    {
+      info->next_expected = T::now() + period;
+    }
+    
+    // else if some time has elapsed since last cb (called outside of cb)
+    else if( (T::now() - info->last_real) < info->period)
+    {
+      // if elapsed time is greater than the new period
+      // do the callback now
+      if( (T::now() - info->last_real) > period)
+      {
+        info->next_expected = T::now();
+      }
+   
+      // else, account for elapsed time by using last_real+period
+      else
+      {
+        info->next_expected = info->last_real + period;
+      }
+    }
+    
+    // Else if called in a callback, last_real has not been updated yet => (now - last_real) > period
+    // In this case, let next_expected be updated only in updateNext
+    
     info->period = period;
-    info->next_expected = T::now() + period;
-
     waiting_.sort(boost::bind(&TimerManager::waitingCompare, this, _1, _2));
   }
 
--- a/test/test_roscpp/test/src/timer_callbacks.cpp
+++ b/test/test_roscpp/test/src/timer_callbacks.cpp
@@ -61,6 +61,8 @@ public:
   : expected_period_(period)
   , failed_(false)
   , total_calls_(0)
+  , testing_period_(false)
+  , calls_before_testing_period_(0)
   {
     NodeHandle n;
     timer_ = n.createWallTimer(expected_period_, &WallTimerHelper::callback, this, oneshot);
@@ -82,12 +84,72 @@ public:
       }
     }
 
-    expected_next_call_ = e.current_expected + expected_period_;
+    if(testing_period_)
+    {
+
+      // Inside callback, less than current period, reset=false
+      if(total_calls_ == calls_before_testing_period_)
+      {
+        WallDuration p(0.5);
+        pretendWork(0.15);
+        setPeriod(p);
+      }
+      
+      // Inside callback, greater than current period, reset=false
+      else if(total_calls_ == (calls_before_testing_period_+1))
+      {
+        WallDuration p(0.25);
+        pretendWork(0.15);
+        setPeriod(p);
+      }
+      
+      // Inside callback, less than current period, reset=true
+      else if(total_calls_ == (calls_before_testing_period_+2))
+      {
+        WallDuration p(0.5);
+        pretendWork(0.15);
+        setPeriod(p, true);
+      }
+      
+      // Inside callback, greater than current period, reset=true
+      else if(total_calls_ == (calls_before_testing_period_+3))
+      {
+        WallDuration p(0.25);
+        pretendWork(0.15);
+        setPeriod(p, true);
+      }
+    }
+    else
+    { 
+      expected_next_call_ = e.current_expected + expected_period_;
+    }
 
     WallTime end = WallTime::now();
     last_duration_ = end - start;
 
     ++total_calls_;
+  }
+
+  void setPeriod(const WallDuration p, bool reset=false)
+  {
+    if(reset)
+    {
+      expected_next_call_ = WallTime::now() + p;
+    }
+    else
+    {
+      expected_next_call_ = last_call_ + p;
+    }
+    
+    timer_.setPeriod(p, reset);
+    expected_period_ = p;
+  }
+
+
+  void pretendWork(const float t)
+  {
+    ros::Rate r(1. / t);
+    r.sleep();
   }
 
   WallTime last_call_;
@@ -99,6 +161,9 @@ public:
 
   WallTimer timer_;
   int32_t total_calls_;
+
+  bool testing_period_;
+  int  calls_before_testing_period_;
 };
 
 TEST(RoscppTimerCallbacks, singleWallTimeCallback)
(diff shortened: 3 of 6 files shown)


clients/roscpp/include/ros/timer.h
```diff
@@ -68,7 +68,7 @@ class ROSCPP_DECL Timer
   /**
    * \brief Set the period of this timer
    */
-  void setPeriod(const Duration& period);
+  void setPeriod(const Duration& period, const bool reset=true);
```
dirk-thomas: For a plain `bool` argument `const` does not make much sense. Please remove it from all signatures. Please also add documentation for the new parameter.

clients/roscpp/include/ros/timer_manager.h
```diff
@@ -431,13 +431,17 @@ void TimerManager<T, D, E>::setPeriod(int32_t handle, const D& period)
 
   {
     boost::mutex::scoped_lock lock(waiting_mutex_);
+    ros::Duration diff(info->period.toSec() - period.toSec());
```
dirk-thomas: The new variable `diff` is not used anywhere?

sterlingm:
Is there anywhere else I need to add documentation?

dirk-thomas:
If you commit changes to an existing PR please also comment on it otherwise nobody gets notified. I will look into it shortly.

dirk-thomas:
After looking closer to the patch I am not sure if the desired behavior is clear to me. I will describe an example to clarify your intentions with the change:

A timer is first created with a period of 10 seconds. After 5 seconds the period is changed to 20 seconds.

The old behavior (as well as when `restart` is `true`) will call the next callback 20 seconds after `setPeriod(20)` has been invoked.

With `restart` being `false` which of the two options do you intend to implement:
1. The "originally" intended callback will be called 5 seconds after `setPeriod(20)` has been invoked. Afterwards the callback is getting called in intervals of 20 seconds.
2. The next callback will be called 15 seconds after `setPeriod(20)` has been invoked (because 5 seconds have already passed). Afterwards the callback is getting called in intervals of 20 seconds.
   - If the already passed time is bigger then the newly set period the callback should be called immediately.

From reading the code I would assume that you have implemented option 1. I would think option 2 would be the more intuitive behavior?

sterlingm:
So my implementation is actually a little different than both of those. I want to be able to change the period in the callback and have the next callback occur according to the new period. Here is an example of the difference:

When `restart` is `true` (Current behavior):

> t=0: timer created for 2s
> ...
> t=2: callback occurs, timer period set to 1.5s
> ...
> t=4: next callback occurs
> ...
> t=5.5: next callback occurs
> ...
> t=7: next callback occurs

When `restart` is `false`:

> t=0: timer created for 2s
> ...
> t=2: callback occurs, timer period set to 1.5s
> ...
> t=3.5: next callback occurs
> ...
> t=5: next callback occurs
> ...
> t=6.5: next callback occurs 

If the period is set at any time between the callbacks (i.e. outside of the callback), my implementation does not change the current behavior. 

My use case for this is to be able to do some work in the callback and then decide when the next callback should occur based on what was done in the current callback. For example, if our callback is computing trajectories, the length of the trajectory (in time) may depend on environment information (how far we can see to plan a collision-free path). If the computed trajectory is T seconds long, we'll set the next callback to occur T seconds in the future. However, the current behavior would make the next callback not take place T seconds in the future - the change in period would take affect _after_ the next callback (so 2 callbacks in the future) rather than affecting the immediate next callback.

dirk-thomas:
I think in your use case where you change the period within the callback itself the only "reasonable" behavior is the second one where the next callback is being invoked according to the new period. For this case I would rather not add a new argument but "fix" the current behavior.

Regarding the two options I mentioned I think the code should use the second one (next callback happens in `new period - already passed duration). And the optional flag only chooses if the timer should be "reset" when a new period is set (which means setting the passed duration to 0).

sterlingm:
Okay, the code in my latest commit achieves what you listed. 

`setPeriod` will consider the elapsed time whenever it is called unless the parameter `reset` is set to `true`. In the case that the new period is less than the elapsed time, the timer will fire the callback immediately and then continue in intervals of the new period.

I tested the following cases:
1. `setPeriod` called outside of callback with `elapsed_time < new_period` and `reset=false`: next callback occurs at `new_period-elapsed_time` in the future (`new_period` since the last callback) and then continues in intervals of `new_period`
2. `setPeriod` called outside of callback with `elapsed_time < new_period` and `reset=true`: next callback happens at `new_period` time in the future
3. `setPeriod` called outside of callback with `elapsed_time > new_period` and `reset=false`: next callback happens immediately after `setPeriod` finishes and the timer continues at the rate of the `new_period`
4. `setPeriod` called outside of callback with `elapsed_time > new_period` and `reset=true`: next callback happens at `new_period` time in the future (instead of firing immediately) and continues at the rate of `new_period`
5. `setPeriod` called inside of callback with `new_period < current_period`, `reset=false`
6. `setPeriod` called inside of callback with `new_period < current_period`, `reset=true`
7. `setPeriod` called inside of callback with `new_period > current_period`, `reset=false`
8. `setPeriod` called inside of callback with `new_period > current_period`, `reset=true`

Cases 5-8 all have the same behavior: next callback happens at `new_period` time in the future after the current callback fired.

There may have been a better or more elegant way to achieve this behavior working with other methods in the code, but I tried to contain everything in `setPeriod`.

clients/roscpp/include/ros/timer_manager.h
```diff
@@ -428,11 +428,36 @@ void TimerManager<T, D, E>::setPeriod(int32_t handle, const D& period)
   {
     return;
   }
+  
+  if(reset)
+  {
+    info->next_expected = T::now() + period;
```
dirk-thomas: Changing `info->next_expected` should only happen inside the locked mutex scope. Therefore the whole logic from line 432 to 456 should go after line 459.

sterlingm: Oh whoops, of course, fixed in latest commit.

clients/roscpp/include/ros/timer_manager.h
```diff
@@ -428,11 +428,36 @@ void TimerManager<T, D, E>::setPeriod(int32_t handle, const D& period)
...
+  }
+  
+  // If called in a callback, last_real has not been updated yet so (now - last_real) > period
+  // In this case, let next_expected be updated only in updateNext
```
dirk-thomas: A problem here might be that it is not clear that the callback from which `setPeriod` has been called is triggered by the same timer or if they are unrelated. Only if they are the same the assumption that `last_real` is not up-to-date holds.

dirk-thomas: Actually I don't think this check is needed at all.
- If a callback calls `setPeriod` of an unrelated timer it can just updated the `next_expected` value(either with `now` or `last_real + period`).
- If a callbacks calls `setPeriod` on its own timer The only thing that matter is the new period since the `next_expected` will be updated immediately after the invocation of the callback returns in the `updateNext` function.

Therefore I would think the following should work:

```
{
  boost::mutex::scoped_lock lock(waiting_mutex_);
  info->period = period;

  if (reset)
  {
    // schedule the next invocation "period" from now
    info->next_expected = T::now() + period;
  }
  else
  {
    if (T::now() - info->last_real > period)
    {
      // the timer has already elapsed according to new period
      // schedule the next invocation immediately
      info->next_expected = T::now();
    }
    }
    else
    {
      // schedule the next invocation based on the last invocation + period
      info->next_expected = info->last_real + period;
    }
  }

  waiting_.sort(boost::bind(&TimerManager::waitingCompare, this, _1, _2));
}
```

sterlingm: The issue with this is that if the callback calls setPeriod on its own timer, `info->last_real` is not updated by that point - it still holds the value of the previous callback. That seems like unintended behavior though.

Using the logic you posted, the following happens:
If `setPeriod` is called on the very first callback, it works fine:

> t=0 create Timer of 2s
> t=2 callback occurs, call setPeriod(3)
> t=5 callback occurs
> t=8 callback occurs

If `setPeriod` is called on the second callback:

> t=0 create Timer of 2s
> t=2 callback occurs
> t=4 callback occurs, call setPeriod(3)
> t=5 callback occurs
> t=8 callback occurs

dirk-thomas: Is that actual tested behavior or only assuming the behavior? I thought `updateNext` (https://github.com/ros/ros_comm/blob/indigo-devel/clients/roscpp/include/ros/timer_manager.h#L403-L410) would move the `next_expected` forward in case the callback set it to `now` (so not calling at `t=5` but `t=7`).

dirk-thomas:
The complexity of this ticket indicates that this would really benefit from unit test coverage...

clients/roscpp/include/ros/timer_manager.h
sterlingm: Tested. I have a small node I am using to just run a Timer's callback, call setPeriod, and report the time that elapses between each callback.

dirk-thomas: You should convert your small node into a unit test if it is not yet one. Then you can add it to the `test_roscpp` package (https://github.com/ros/ros_comm/tree/indigo-devel/test/test_roscpp/test). 

You can add the test to this PR.

If you don't want to take the effort of making a unit test please consider sharing your small node via gist or even better in a github repo with all the other stuff to easily compile and run your node.

sterlingm:
The latest commit has a unit test for calling `setPeriod`. There are eight cases total listed above in a previous comment. I also committed a version of timer_manager.h that passes the unit test.

Let me know what changes you would like made. I see a couple issues:
1. I only edited the code for the `WallTimerHelper` class. I see that there is a `TimerHelper`, but its callback doesn't seem to check anything so I left it alone. Should I add in a unit test for `TimerHelper` as well?
2. I added all the cases into a single unit test. The cases are all essentially checking the same thing, but with different parameters so I wasn't sure if they all warranted individual unit tests. If you want it all in a single test, should I print any info about which case is failing?

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
It looks like the patched code does not even compile anymore: http://jenkins.ros.org/job/_pull_request-indigo-ros_comm/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/229/console

sterlingm:
Latest commit should fix build errors

sterlingm:
@dirk-thomas Can you shed some more light on what the failed test results are for? I'm not sure I understand why my previous commit passed, but my latest one did not. The latest one only removed two ROS_INFO lines and commented out another ROS_INFO line.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Some of the tests are flaky. The random rosbag play tests are some of them. I have retriggered the job - usually it should pass after some attempts (not a nice way to handle it but currently the best we can do).

clients/roscpp/include/ros/timer_manager.h
```diff
@@ -490,7 +514,7 @@ void TimerManager<T, D, E>::threadFunc()
         {
           current = T::now();
 
-          //ROS_DEBUG("Scheduling timer callback for timer [%d] of period [%f], [%f] off expected", info->handle, info->period.toSec(), (current - info->next_expected).toSec());
+          //ROS_INFO("Scheduling timer callback for timer [%d] of period [%f], [%f] off expected", info->handle, info->period.toSec(), (current - info->next_expected).toSec());
```
dirk-thomas: Please revert this change.

test/test_roscpp/test/src/timer_callbacks.cpp
```diff
@@ -80,16 +82,78 @@ class WallTimerHelper
         ROS_ERROR("Call came at wrong time (%f vs. %f)", expected_next_call_.toSec(), start.toSec());
         failed_ = true;
       }
+
```
dirk-thomas: Please revert whitespace only change.

test/test_roscpp/test/src/timer_callbacks.cpp
```diff
@@ -80,16 +82,78 @@ class WallTimerHelper
...
+        pretendWork(0.15);
+        setPeriod(p, true);
+      }
+    }
+    else
+    { 
+      expected_next_call_ = e.current_expected + expected_period_;
+    }
 
+
```
dirk-thomas: Please revert whitespace only change.

test/test_roscpp/test/src/timer_callbacks.cpp
```diff
@@ -453,6 +602,7 @@ class Tracked
   }
 };
 
+
```
dirk-thomas: Please revert whitespace only change.

dirk-thomas:
You don't have to add tests for `TimerHelper`. Also having all tests in a single test function is fine. If any of them fail the output will tell which one did.

I will try your patch locally and will see if I can simply the code in the `setPeriod` function as we have discussed before.

sterlingm:
@dirk-thomas I reverted those changes you requested. The test seems to have failed again, but jenkins lists no failed tests and the console output just says failure. Is this just flakiness again?

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

sterlingm:
@dirk-thomas Are there any updates on this?

dirk-thomas:
Since there hasn't been a ros_comm release for a while I haven't looked at this since the last comment. It looks to be in a good shape to be merge before the next patch release.

But one thing I have noticed while reading the patch again:
the previous behavior (https://github.com/ros/ros_comm/pull/590/files#diff-3decb5e6adcb333e676de1a4812e1378L435) was setting the next timer to `now() + period`. The new code does the same when `reset == true` (https://github.com/ros/ros_comm/pull/590/files#diff-3decb5e6adcb333e676de1a4812e1378R437) but the current default value is `false`. Shouldn't the default value of `reset` be `true` to maintain the same behavior as before?

sterlingm:
@dirk-thomas When I first submitted the patch, I was adding new behavior. After I explained what I was doing more thoroughly, you suggested we change the default behavior because it made more sense. Now the default behavior will consider the elapsed time between start of the last callback and calling setPeriod. Users will need to set `reset=true` if they don't want to consider the elapsed time, i.e. "resetting" the timer. See our comments on Apr 8th to read over this again.

dirk-thomas:
Thanks for mentioning that. I am not sure what the better approach is. Lets see what others think.

@esteve @tfoote @wjwwood Please review this PR and comment on what you think would be a "better" default for the new `reset` flag.

tfoote:
Since this is potentially breaking algorithms we should make the default keep the same behavior and the alternative behavior require the alternative. 

I agree that the non-reset makes more sense. But we should just document clearly the default behavior and the options.

wjwwood:
I'd say keep the behavior as is, others might be working around this issue with custom logic and changing the default behavior would break them. Though I do agree the new option would be a better default behavior if it were not already in use.

sterlingm:
@dirk-thomas @tfoote @wjwwood It seems the majority opinion is to leave the default behavior unchanged. The latest commit sets `reset=true` as the default value which will not change the current default behavior.

dirk-thomas:
The patch looks good. Thank you for iterating on it with us. Can you please squash the commits (if not I can squash and merge them manually).

sterlingm:
@dirk-thomas Okay, they should be squashed into one commit now.

sterlingm:
@dirk-thomas I missed changing `reset=true` in the Impl class on that last commit, will fix asap.

sterlingm:
@dirk-thomas Okay, should be ready now.

PR #592, opened 2015-03-31
Alternative versions of ros::param::param and ros::NodeHandle::param with return value

xperroni:
(Note: my editor automatically removed a number of line-ending whistespaces in "node_handle.h". If this is a problem, feel free to reject this pull request, and I'll resubmit it with these changes rolled back.)

Template methods ros::param::param() and ros::NodeHandle::param() set a
variable either to the value of a parameter or, if not available, a predefined
default value. Often the variable is passed straight to a constructor or
function call and then disregarded.

The alternative versions of the param() methods introduced here enable
retrieving the result value from a function return, rather than through an
output variable. This makes possible to write more streamlined code.

For example, the code below:

```
std::string path;
std::string format;
double fps;

param("~path", path, "video.mpg");
param("~format", format, "MPEG");
param("~fps", fps, 30.0);

VideoRecorder recorder(path, format, fps);
```

Could be more succinctly written as:

```
VideoRecorder recorder(
    param<std::string>("~path", "video.mpg"),
    param<std::string>("~format", "MPEG"),
    param<double>("~fps", 30.0)
);
```

Or:

```
NodeHandle nh("~");

VideoRecorder recorder(
    nh.param<std::string>("path", "video.mpg"),
    nh.param<std::string>("format", "MPEG"),
    nh.param<double>("fps", 30.0)
);
```

Signed-off-by: Helio Perroni Filho xperroni@gmail.com

### DIFF
--- a/test/test_roscpp/test/src/params.cpp
+++ b/test/test_roscpp/test/src/params.cpp
@@ -533,14 +533,14 @@ TEST(Params, paramTemplateFunction)
   EXPECT_EQ( param::param<std::string>( "string", "" ), "test" );
   EXPECT_EQ( param::param<std::string>( "gnirts", "test" ), "test" );
 
-  EXPECT_TRUE( param::param<int>( "int", 0 ) == 10 );
-  EXPECT_TRUE( param::param<int>( "tni", 10 ) == 10 );
+  EXPECT_EQ( param::param<int>( "int", 0 ), 10 );
+  EXPECT_EQ( param::param<int>( "tni", 10 ), 10 );
 
   EXPECT_DOUBLE_EQ( param::param<double>( "double", 0.0 ), 10.5 );
   EXPECT_DOUBLE_EQ( param::param<double>( "elbuod", 10.5 ), 10.5 );
 
-  EXPECT_FALSE( param::param<bool>( "bool", true ) );
-  EXPECT_TRUE( param::param<bool>( "loob", true ) );
+  EXPECT_EQ( param::param<bool>( "bool", true ), false );
+  EXPECT_EQ( param::param<bool>( "loob", true ), true );
 }
 
 TEST(Params, paramNodeHandleTemplateFunction)
@@ -550,14 +550,14 @@ TEST(Params, paramNodeHandleTemplateFunction)
   EXPECT_EQ( nh.param<std::string>( "string", "" ), "test" );
   EXPECT_EQ( nh.param<std::string>( "gnirts", "test" ), "test" );
 
-  EXPECT_TRUE( nh.param<int>( "int", 0 ) == 10 );
-  EXPECT_TRUE( nh.param<int>( "tni", 10 ) == 10 );
+  EXPECT_EQ( nh.param<int>( "int", 0 ), 10 );
+  EXPECT_EQ( nh.param<int>( "tni", 10 ), 10 );
 
   EXPECT_DOUBLE_EQ( nh.param<double>( "double", 0.0 ), 10.5 );
   EXPECT_DOUBLE_EQ( nh.param<double>( "elbuod", 10.5 ), 10.5 );
 
-  EXPECT_FALSE( nh.param<bool>( "bool", true ) );
-  EXPECT_TRUE( nh.param<bool>( "loob", true ) );
+  EXPECT_EQ( nh.param<bool>( "bool", true ), false );
+  EXPECT_EQ( nh.param<bool>( "loob", true ), true );
 }
 
 int
--- a/test/test_roscpp/test/src/params.cpp
+++ b/test/test_roscpp/test/src/params.cpp
@@ -530,8 +530,8 @@ TEST(Params, mapBoolParam)
 
 TEST(Params, paramTemplateFunction)
 {
-  EXPECT_TRUE( param::param<std::string>( "string", "" ) == "test" );
-  EXPECT_TRUE( param::param<std::string>( "gnirts", "test" ) == "test" );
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Thank you for the patch. As you also mentioned it please remove any whitespace changes since they make backporting changes between branches way more difficult.

xperroni:
OK. Should I close the pull request and submit it again? I couldn't find a
way to edit the request directly, or even restore the deleted whitespace
without first rolling back the original commit.

dirk-thomas:
Simply continue committing to the branch on your forked repo and the commits will "appear" in this PR.

xperroni:
Done. Sorry it took this long to sort it out.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:

esteve:
+1

test/test_roscpp/test/src/params.cpp
```diff
@@ -528,6 +528,38 @@ TEST(Params, mapBoolParam)
   ASSERT_TRUE(std::equal(map_b.begin(), map_b.end(), map_b2.begin()));
 }
 
+TEST(Params, paramTemplateFunction)
+{
+  EXPECT_TRUE( param::param<std::string>( "string", "" ) == "test" );
```
tfoote: This and the other EXPECT_TRUE on strings being equal should use EXPECT_EQ [docs on string comparisions](https://code.google.com/p/googletest/wiki/Primer#String_Comparison) It gives better visibility into the error reports.

wjwwood:
lgtm, +1

tfoote:
Thanks for putting this together. 

lgtm with the small fixup.

xperroni:
OK, changed the assertion used to check the string results as pointed.

tfoote:
lgtm

test/test_roscpp/test/src/params.cpp
```diff
@@ -528,6 +528,38 @@ TEST(Params, mapBoolParam)
...
+  EXPECT_TRUE( param::param<std::string>( "gnirts", "test" ) == "test" );
+
+  EXPECT_TRUE( param::param<int>( "int", 0 ) == 10 );
```
esteve: Please apply change suggested by @tfoote here too (EXPECT_EQ instead of EXPECT_TRUE)

xperroni: Argh, forgot about those other two. Changed as requested.

test/test_roscpp/test/src/params.cpp
```diff
@@ -528,6 +528,38 @@ TEST(Params, mapBoolParam)
   ASSERT_TRUE(std::equal(map_b.begin(), map_b.end(), map_b2.begin()));
 }
 
+TEST(Params, paramTemplateFunction)
+{
+  EXPECT_EQ( param::param<std::string>( "string", "" ), "test" );
+  EXPECT_EQ( param::param<std::string>( "gnirts", "test" ), "test" );
+
+  EXPECT_TRUE( param::param<int>( "int", 0 ) == 10 );
```
esteve: Could you make the EXPECT_EQ change here and for every integer test too? Thanks!

test/test_roscpp/test/src/params.cpp
```diff
@@ -528,6 +528,38 @@ TEST(Params, mapBoolParam)
...
+  EXPECT_EQ( nh.param<std::string>( "string", "" ), "test" );
+  EXPECT_EQ( nh.param<std::string>( "gnirts", "test" ), "test" );
+
+  EXPECT_TRUE( nh.param<int>( "int", 0 ) == 10 );
+  EXPECT_TRUE( nh.param<int>( "tni", 10 ) == 10 );
+
+  EXPECT_DOUBLE_EQ( nh.param<double>( "double", 0.0 ), 10.5 );
+  EXPECT_DOUBLE_EQ( nh.param<double>( "elbuod", 10.5 ), 10.5 );
+
+  EXPECT_FALSE( nh.param<bool>( "bool", true ) );
```
esteve: I'd replace this with EXPECT_EQ too. true/false are values returned by `nh.param<>()`, it's not like it returns a binary answer (as opposed to methods like `is_open()` or `is_valid()`), there's a slight semantic difference IMO. @dirk-thomas @tfoote @wjwwood thoughts?

wjwwood: In this case I think it doesn't matter. The most important thing is to use `EXPECT_EQ` on non-bool types so that when they do not match it can print out what the actual versus expected was. In this case there are only to states, so being explicit doesn't help.

xperroni:
Changed test assertions as requested.

esteve:
@xperroni thank you very much for addressing our comments!

xperroni:
@esteve My pleasure. :)

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Thank you! Squashed and merged: ba130b948707b3c6705e28f86ec07f2fb7d7628d

xperroni:
Yay! Thank you for your comments and support. :)

PR #595, opened 2015-04-02
Pause playback for topics

gerkey:
This PR is a replacement for #569, targeting Indigo instead of Hydro.

It's @tdenewiler's patch, with a couple of my tweaks.  The only substantial change I made was to start in topic-pausing mode if `--pause-topics` is given at the command line, because I found that to be more intuitive (as opposed to requiring the user to hit `t` to go into that mode).

I tested with a few different cases, found it to work as advertised, and not to have any obvious side-effects.

To @dirk-thomas to approve and merge.

### DIFF
--- a/tools/rosbag/include/rosbag/player.h
+++ b/tools/rosbag/include/rosbag/player.h
@@ -90,6 +90,7 @@ struct ROSBAG_DECL PlayerOptions
 
     std::vector<std::string> bags;
     std::vector<std::string> topics;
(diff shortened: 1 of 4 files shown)


dirk-thomas:
Thanks! Squashed and merged: b7796fea666aac7b330aad870f7e7b994eda8daf

dirk-thomas:
It looks like this caused a regression: #620

PR #597, opened 2015-04-02
Pass through the topic field of an incoming connection header, issue #543.

### LINKED ISSUE #543: roscpp: Empty connection header topic
I am trying to use a driver that reads the name of the incoming topic from the connection header.  When I publish data from a C++ node, it is always empty but when I use `rostopic pub` or python it works just fine.  Is there a way to ensure the connection header is populated?  

OS: Ubuntu 14.04

ROS:indigo 1.11.9

roscpp: 1.11.9-0trusty-20141029-0127-+0000

roscpp_core: 0.5.4-0trusty-20140724-0053-+0000

The subscriber is in C++ and is [here](https://github.com/smd-ros-devel/lynxmotion_ssc32/blob/master/src/ssc32_driver.cpp#L391)

Here is the code in [python](https://gist.github.com/tonybaltovski/e1470fc1175b423ca5f4) and the [C++ node](https://gist.github.com/tonybaltovski/ac75f7e3273e7086a433)

gerkey:
fixes #543

### DIFF
--- a/clients/roscpp/src/libros/transport_subscriber_link.cpp
+++ b/clients/roscpp/src/libros/transport_subscriber_link.cpp
@@ -111,6 +111,7 @@ bool TransportSubscriberLink::handleHeader(const Header& header)
   m["message_definition"] = pt->getMessageDefinition();
   m["callerid"] = this_node::getName();
   m["latching"] = pt->isLatching() ? "1" : "0";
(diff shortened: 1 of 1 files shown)


wjwwood:
@ros-pull-request-builder

PR #600, opened 2015-04-03
Fix roscpp crash in topic statistics

scpeters:
This includes @ericperko 's test from #556 with an alternative fix. While looking at the code for computing standard deviation with @dirk-thomas , we didn't like using `ros::Duration` as an intermediate type for this calculation. Instead, I use a `double` to compute variance then another `double` to compute the standard deviation. I then wrap the final conversion to `ros::Duration` in a `try`/`catch` block similar to the one in #556.

@ericperko how does this look?

### DIFF
--- a/clients/roscpp/src/libros/statistics.cpp
+++ b/clients/roscpp/src/libros/statistics.cpp
@@ -144,13 +144,26 @@ void StatisticsLogger::callback(const boost::shared_ptr<M_string>& connection_he
 
       msg.stamp_age_mean *= 1.0 / stats.age_list.size();
 
-      msg.stamp_age_stddev = ros::Duration(0);
+      double stamp_age_variance = 0.0;
       for(std::list<ros::Duration>::iterator it = stats.age_list.begin(); it != stats.age_list.end(); it++)
       {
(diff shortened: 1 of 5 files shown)


dirk-thomas:
@ros-pull-request-builder

ericperko:
Looks pretty good to me. I'm a little surprised that you didn't have to catch around `ros::Duration t = msg.stamp_age_mean - *it` since I mentioned in my original PR that I had seen the exception in that line as well as the stddev computation, but I can't remember what case tripped that line that wouldn't have been an issue in my test (ros::Time::now() - ros::Time(0) should be the worst case I could have seen).

The only difference between the real system and my test case is that the real system usually had correct timestamps in the message header with only a few stamps set to 0, while the test has no correct timesamps. If I can find the original bagfile that I used to verify the patch, I'll try running this change through to verify.

scpeters:
Right, I noticed that you said that, but the test passes even without wrapping `ros::Duration t = msg.stamp_age_mean - *it` in a try/catch. If we can reproduce a failure with the current code, then we should update the test accordingly.

PR #603, opened 2015-04-13
start thread only after successfully adding the connection

dirk-thomas:
Aims to address #544

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_pubsub.py
+++ b/clients/rospy/src/rospy/impl/tcpros_pubsub.py
@@ -264,8 +264,9 @@ class TCPROSHandler(rospy.impl.transport.ProtocolHandler):
             t.start()
             return 1, "Connected topic[%s]. Transport impl[%s]"%(resolved_name, conn.__class__.__name__), dest_port
         else:
+            # _SubscriberImpl already closed or duplicate subscriber created
             conn.close()
-            return 0, "ERROR: Race condition failure: duplicate topic subscriber [%s] was created"%(resolved_name), 0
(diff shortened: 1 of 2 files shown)


dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

tfoote:
This looks reasonable. If we merge this with the symptom workaround with a good error message pointing back to the ticket it should be safe.

dirk-thomas:
The first commit tries to fix the actual problem so that a connection should never be added to a closed subscriber in the first place.

The second commit will still catch the case and prevent a memory leak. If that still happens the error message points the user to the ticket to provide feedback.

tfoote:
+1

esteve:
LGTM

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

mpatalberta:
can you please direct me to where in indigo ros the fix has been made for 
@dirk-thomas dirk-thomas referenced this pull request on Apr 13, 2015
 Closed
rospy: Connection / Memory Leak, adding conn. to closed _SubscriberImpl #544

dirk-thomas:
@mpatalberta Please see the line above:

> dirk-thomas merged commit a88e77b into indigo-devel on Apr 16, 2015

The commit mentions version 1.11.11 as the first release the patch has been part of.

PR #606, opened 2015-04-20
rostopic: handling multiple array index specifications

rhaschke:
Please consider this patch that fixes an issue with multiple array index specifications for subfields in rostopic echo. E.g.
- `rostopic echo /tactile_mid_prox/sensors[1]/proximal[1]` failed with

> Topic name 'sensors[1]/proximal' contains '[' but does not end with ']'
- `rostopic echo /tactile_mid_prox/sensors[1]/proximal[1]` failed with

> Topic name 'sensors[1]/proximal[1]' contains invalid slice argument '1]/proximal[1': non-integer array index step '1]/proximal[1'

The reason were two issues:
- First, in function `msgevalgen`, it was assumed that there will be only a single index specification. I fixed, that by a) looking for the matching closing bracket (instead of using the whole rest of the string) and b) calling `msgevalgen` recursively on trailing sub fields.
- Second, in function `_rostopic_echo` when extracting the `type_information` for the to-be-displayed values, a second index specification was not expected too, resulting in an attempt to extract the type for `subtype[]` instead of `subtype`. This was easily fixed, dropping the brackets.

### DIFF
--- a/test/test_rostopic/CMakeLists.txt
+++ b/test/test_rostopic/CMakeLists.txt
@@ -8,10 +8,12 @@ if(CATKIN_ENABLE_TESTING)
   add_message_files(DIRECTORY msg
     FILES
     Arrays.msg
+    ArrayVal.msg
     Embed.msg
     Floats.msg
     Simple.msg
     TVals.msg
+    Val.msg
   )
   generate_messages(DEPENDENCIES std_msgs)
 endif()
--- a/test/test_rostopic/test/test_rostopic_unit.py
+++ b/test/test_rostopic/test/test_rostopic_unit.py
@@ -390,7 +390,7 @@ class TestRostopicUnit(unittest.TestCase):
         self.assertEquals(['seq', 'stamp'], list(f(m)))
 
     def test_slicing(self):
-        from test_rospy.msg import ArrayVal, Val
+        from test_rostopic.msg import ArrayVal, Val
         from rostopic import msgevalgen as f
 
         # prepare a sliceable msg
@@ -400,7 +400,7 @@ class TestRostopicUnit(unittest.TestCase):
 
         self.assertEqual(f(''), None)
         self.assertEqual(f('/'), None)
-        self.assertListEqual(f('/vals')(msg),    msg.vals)
+        self.assertListEqual(f('/vals')(msg), msg.vals)
         # first-level slicing
         self.assertListEqual(f('/vals[:]')(msg), msg.vals)
         self.assertListEqual(f('/vals[0:2]')(msg), msg.vals[0:2])
--- a/test/test_rostopic/test/test_rostopic_unit.py
+++ b/test/test_rostopic/test/test_rostopic_unit.py
@@ -388,3 +388,45 @@ class TestRostopicUnit(unittest.TestCase):
         self.assertEquals(['t', 'd'], list(f(m)))
         m = Header()
         self.assertEquals(['seq', 'stamp'], list(f(m)))
+
+    def test_slicing(self):
+        from test_rospy.msg import ArrayVal, Val
+        from rostopic import msgevalgen as f
+
+        # prepare a sliceable msg
+        msg = ArrayVal()
+        for v in ['ABCDEFG', 'abcdefg', '1234567', 'short']:
+            msg.vals.append(Val(val=v))
+
+        self.assertEqual(f(''), None)
+        self.assertEqual(f('/'), None)
+        self.assertListEqual(f('/vals')(msg),    msg.vals)
+        # first-level slicing
+        self.assertListEqual(f('/vals[:]')(msg), msg.vals)
+        self.assertListEqual(f('/vals[0:2]')(msg), msg.vals[0:2])
+        # element access
+        self.assertEqual(f('/vals[0]')(msg), msg.vals[0])
+        self.assertEqual(f('/vals[1]')(msg), msg.vals[1])
+        # element access continued
+        self.assertEqual(f('/vals[0]/val')(msg), msg.vals[0].val)
+        self.assertEqual(f('/vals[1]/val')(msg), msg.vals[1].val)
+        # second-level slicing
+        self.assertEqual(f('/vals[0]/val[:]')(msg), msg.vals[0].val)
+        self.assertEqual(f('/vals[0]/val[0:2]')(msg), msg.vals[0].val[0:2])
+        self.assertEqual(f('/vals[0]/val[:-3]')(msg), msg.vals[0].val[:-3])
+        self.assertEqual(f('/vals[0]/val[2]')(msg), msg.vals[0].val[2])
+        # first-level slicing + second-level access
+        self.assertListEqual(f('/vals[:3]/val[0]')(msg), ['A', 'a', '1'])
+        self.assertListEqual(f('/vals[:3]/val[0]')(msg), ['A', 'a', '1'])
+        self.assertListEqual(f('/vals[1:3]/val[0]')(msg), ['a', '1'])
+        self.assertListEqual(f('/vals[:]/val[-1]')(msg), ['G', 'g', '7', 't'])
+        # multiple slicing
+        self.assertListEqual(f('/vals[:3]/val[1:3]')(msg), ['BC', 'bc', '23'])
+        # out-of-range errors
+        self.assertEqual(f('/vals[5]/val')(msg), None)
+        self.assertListEqual(f('/vals[:]/val[6]')(msg), ['G', 'g', '7', None])
+        # invalid descriptions
+        self.assertEqual(f('/vals[:]/val[]'), None)
+        self.assertEqual(f('/unknown[:]/val[0]')(msg), None)
+        self.assertListEqual(f('/vals[:]/unknown[0]')(msg), [None, None, None, None])
+        self.assertEqual(f('/vals/unknown[0]')(msg), None)
--- a/test/test_rostopic/test/test_rostopic_unit.py
+++ b/test/test_rostopic/test/test_rostopic_unit.py
@@ -408,6 +408,7 @@ class TestRostopicUnit(unittest.TestCase):
         # element access
         self.assertEqual(f('/vals[0]')(msg), msg.vals[0])
         self.assertEqual(f('/vals[1]')(msg), msg.vals[1])
+        self.assertEqual(f('/vals[0'), None)
         # element access continued
         self.assertEqual(f('/vals[0]/val')(msg), msg.vals[0].val)
         self.assertEqual(f('/vals[1]/val')(msg), msg.vals[1].val)
--- a/test/test_rostopic/test/test_rostopic_unit.py
+++ b/test/test_rostopic/test/test_rostopic_unit.py
@@ -401,6 +401,7 @@ class TestRostopicUnit(unittest.TestCase):
         self.assertEqual(f(''), None)
         self.assertEqual(f('/'), None)
         self.assertListEqual(f('/vals')(msg), msg.vals)
+        self.assertListEqual(f('/vals/')(msg), msg.vals)
         # first-level slicing
         self.assertListEqual(f('/vals[:]')(msg), msg.vals)
         self.assertListEqual(f('/vals[0:2]')(msg), msg.vals[0:2])
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -311,7 +311,7 @@ def msgevalgen(pattern):
         else:
             evals.append((f, None))
 
-    def _msgeval(msg, evals):
+    def msgeval(msg, evals):
         for i, (field_name, slice_object) in enumerate(evals):
             try: # access field first
                 msg = getattr(msg, field_name)
@@ -331,13 +331,10 @@ def msgevalgen(pattern):
                 # in order to handle nested slices
                 if isinstance(msg, list):
                     rest = evals[i + 1:]
-                    return [_msgeval(m, rest) for m in msg]
+                    return [msgeval(m, rest) for m in msg]
         return msg
 
-    def msgeval(msg):
-        return _msgeval(msg, evals)
-
-    return msgeval if evals else None
+    return (lambda msg: msgeval(msg, evals)) if evals else None
 
 
 def _get_array_index_or_slice_object(index_string):
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -285,64 +285,59 @@ def _rostopic_bw(topic, window_size=-1):
         _sleep(1.0)
         rt.print_bw()
 
-# TODO: port to the version I wrote for rxplot instead as it should be more efficient
-
+# code adapted from rqt_plot
 def msgevalgen(pattern):
     """
-    Generates a function that returns the relevant field (aka 'subtopic') of a Message object
-    :param pattern: subtopic, e.g. /x. Must have a leading '/' if specified, ``str``
+    Generates a function that returns the relevant field(s) (aka 'subtopic(s)') of a Message object
+    :param pattern: subtopic, e.g. /x[2:]/y[:-1]/z, ``str``
     :returns: function that converts a message into the desired value, ``fn(Message) -> value``
     """
-    if not pattern or pattern == '/':
-        return None
-    assert pattern[0] == '/'
-    msg_attribute = pattern[1:]
+    evals = [] # list of (field_name, slice_object) pairs
+    fields = [f for f in pattern.split('/') if f]
+    for f in fields:
+        if '[' in f:
+            field_name, rest = f.split('[')
+            rest = rest[:rest.find(']')]
+            try:
+                array_index_or_slice_object = _get_array_index_or_slice_object(rest)
+            except AssertionError as e:
+                sys.stderr.write("field '%s' has invalid slice argument '%s': %s\n"
+                                 % (field_name, rest, str(e)))
+                return None
+            evals.append((field_name, array_index_or_slice_object))
+        else:
+            evals.append((f, None))
 
-    # use slice arguments if present
-    array_index_or_slice_object = None
-    rec_msgeval = None
-    index_start = msg_attribute.find('[')
-    if index_start != -1:
-        index_end = msg_attribute.find(']', index_start)
-        if index_end == -1:
-            sys.stderr.write("Topic name '%s' contains '[' but does not end with ']'\n" % msg_attribute)
-            return None
-        index_string = msg_attribute[index_start + 1:index_end]
-        try:
-            array_index_or_slice_object = _get_array_index_or_slice_object(index_string)
-        except AssertionError as e:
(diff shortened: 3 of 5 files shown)


tools/rostopic/src/rostopic/__init__.py
```diff
@@ -300,28 +300,37 @@ def msgevalgen(pattern):
 
     # use slice arguments if present
     array_index_or_slice_object = None
-    index = msg_attribute.find('[')
-    if index != -1:
-        if not msg_attribute.endswith(']'):
+    rec_msgeval = None
+    index_start = msg_attribute.find('[')
+    if index_start != -1:
+        index_end = msg_attribute.find(']')
```
dirk-thomas: Shouldn't the find search start from `index_start`? Or at least check the order of the start and end indices afterwards?

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -300,28 +300,37 @@ def msgevalgen(pattern):
...
+        rec_msgeval = msgevalgen(msg_attribute_tail)
+        msg_attribute = msg_attribute[:index_start]
 
     def msgeval(msg):
         # I will probably replace this with some less beautiful but more efficient
         try:
             value = _get_nested_attribute(msg, msg_attribute)
         except AttributeError as e:
-            sys.stdout.write("no field named [%s]"%pattern+"\n")
+            if len(msg) > 1:
```
dirk-thomas: Isn't this happening if you try to output e.g. `/tactile_mid_prox/sensors[1]/proximal`? Or if the message field is a string?

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -300,28 +300,37 @@ def msgevalgen(pattern):
...
+                sys.stdout.write("slicing only supported for last array index\n")
```
dirk-thomas: Should this go to `stderr` instead? Same two lines below.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -788,7 +797,9 @@ def _rostopic_echo(topic, callback_echo, bag_file=None, echo_all_topics=False):
                     index = submsg_class.__slots__.index(field)
                     type_information = submsg_class._slot_types[index]
                     if fields:
-                        submsg_class = roslib.message.get_message_class(type_information)
+                        submsg_class = roslib.message.get_message_class(type_information.split('[')[0])
+                        if not submsg_class:
```
dirk-thomas, line 804: Why is this necessary? Can you please provide an example where this is the case.

dirk-thomas:
Since the complexity of this message evaluation gets more and more complex to support all these cases can you please add unit tests for this new feature.

tools/rostopic/src/rostopic/__init__.py
rhaschke: You are right. One should better use `msg_attribute.find(']', index_start)'.

tools/rostopic/src/rostopic/__init__.py
rhaschke: This addition is related to multiple slicings like /topic/field[0:3]/field[0:2]. In this case it's neither clear how to output the product of array results from these multiply slicings, nor can _get_nested_attribute() handle lists of msgs.
Hence, the warning about multiple slicings. The standard warning would show `no field named [field[0:2]]`, which doesn't explain the issue for the user.

tools/rostopic/src/rostopic/__init__.py
rhaschke: I simply kept the stdout that was used before. If you want to change it, go ahead.

tools/rostopic/src/rostopic/__init__.py
rhaschke, line 804: `type_information` for an array element will contain the array subscript, e.g. `field[2]`. 
However, this type string doesn't have a valid message class. The type is named `field` in this case. Hence, we need to drop the brackets.

rhaschke:
I guess, in order to have unit tests for these case, I will have to launch a talker that publishes a nested array message. Do you suggest to add such a msg definition for testing only? Or can you suggest an existing message type using nested arrays?
Where should this talker code go to? Currently the talker node used for testing comes from package rospy.

rhaschke:
@dirk-thomas 
As I didn't get feedback on my question how and where to add nested arrayed messages for testing,
I have chosen another approach for unit tests: offline testing of rostopic.msgevalgen() using existing msg types `ArrayVal` and `Val` (i.e. a list of strings) from `test_rospy.msg`.

To this end, **I had to re-enable nosetests in cmake**. However, the **other tests** already existing in test_rostopic_unit.py **failed. Please decide how to proceed with these tests** (fix or disable).

Further, I reworked rostopic.msgevalgen() following a similar pattern as in rqt_plot (as was already suggested by a [TODO](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rostopic/src/rostopic/__init__.py#L288). However, I allow for multiple slicings now, resulting in nested arrays as output (see [unit tests](https://github.com/ubi-agni/ros_comm/blob/indigo-devel/tools/rostopic/test/test_rostopic_unit.py#L428) for examples).

Finally, I fixed an unrelated typo: Probably [valid_type should be named valid_type_s_.](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rostopic/src/rostopic/__init__.py#L263)

rhaschke:
@ros-pull-request-builder

dirk-thomas:
Please don't add unrelated changes into the same pull request but create separate ones. It makes it not only more difficult to review but also prevent some changes from being merged early while waiting for others to be iterated on.

I have committed the following changes separately:
- using the `queue_size` argument to avoid a warning: 05ad10f1750260a9f6f9e31a6c7842b7fbf511ac
- fix the string type check: 13c44d17f5725809686ae1457f072e850ee7a305
- reenable the offline test: 3c2a9d9b4e45451c301e03855e1fd84479dd5f4a
- moved the remaining disabled unit test to a separate test package where it can create the necessary test messages and was reenabled again: bd10f702dcf4c447cc421ed3fd4838f9173afe28

Please rebase the PR. The newly added test code should go into https://github.com/ros/ros_comm/blob/bd10f702dcf4c447cc421ed3fd4838f9173afe28/test/test_rostopic/test/test_rostopic_unit.py

rhaschke:
@dirk-thomas 
I rebased the PR as requested.

test/test_rostopic/test/test_rostopic_unit.py
```diff
@@ -388,3 +388,45 @@ def test_create_field_filter(self):
         self.assertEquals(['t', 'd'], list(f(m)))
         m = Header()
         self.assertEquals(['seq', 'stamp'], list(f(m)))
+
+    def test_slicing(self):
+        from test_rospy.msg import ArrayVal, Val
```
dirk-thomas: The `test_rostopic` package does not depend on `test_rospy`. Please add the necessary messages to `test_rostopic` and use them from there.

test/test_rostopic/test/test_rostopic_unit.py
```diff
@@ -388,3 +388,45 @@ def test_create_field_filter(self):
...
+        from rostopic import msgevalgen as f
+
+        # prepare a sliceable msg
+        msg = ArrayVal()
+        for v in ['ABCDEFG', 'abcdefg', '1234567', 'short']:
+            msg.vals.append(Val(val=v))
+
+        self.assertEqual(f(''), None)
+        self.assertEqual(f('/'), None)
+        self.assertListEqual(f('/vals')(msg),    msg.vals)
```
dirk-thomas: Please remove the extra white spaces following PEP8.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -285,45 +285,59 @@ def _rostopic_bw(topic, window_size=-1):
...
-            array_index_or_slice_object = _get_array_index_or_slice_object(index_string)
-        except AssertionError as e:
-            sys.stderr.write("Topic name '%s' contains invalid slice argument '%s': %s\n" % (msg_attribute, index_string, str(e)))
-            return None
-        msg_attribute = msg_attribute[:index]
+    evals = [] # list of (field_name, slice_object) pairs
+    fields = [f for f in pattern.split('/') if f]
+    for f in fields:
+        if '[' in f:
+            field_name, rest = f.split('[')
```
dirk-thomas: The `split` operation should be limited to return only 2 values using the `maxsplit` argument.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -285,45 +285,59 @@ def _rostopic_bw(topic, window_size=-1):
...
+            rest = rest[:rest.find(']')]
```
dirk-thomas: Since we don't allow `field[1][2]` here the closing bracket should always be the last character, no?

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -285,45 +285,59 @@ def _rostopic_bw(topic, window_size=-1):
...
+                sys.stderr.write("field '%s' has invalid slice argument '%s': %s\n"
+                                 % (field_name, rest, str(e)))
+                return None
+            evals.append((field_name, array_index_or_slice_object))
+        else:
+            evals.append((f, None))
+
+    def _msgeval(msg, evals):
+        if not evals:
+            return msg
```
dirk-thomas: Could this conditional be removed? It seems to be already covered by the conditional return line of the `msgevalgen` function.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -285,45 +285,59 @@ def _rostopic_bw(topic, window_size=-1):
...
+        for i, (field_name, slice_object) in enumerate(evals):
+            try: # access field first
+                msg = getattr(msg, field_name)
+            except AttributeError:
+                sys.stderr.write("no field named [%s]\n" % field_name)
```
dirk-thomas: May be more context information would be helpful for users. E.g. show the full "path" to the unresolvable field.

Same at the next except block.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -285,45 +285,59 @@ def _rostopic_bw(topic, window_size=-1):
...
+                    msg = msg.__getitem__(slice_object)
+                except IndexError as e:
+                    sys.stderr.write("%s: %s\n" % (str(e), pattern))
+                    return None
+
+                # if a list is returned here (i.e. not only a single element accessed),
+                # we need to recursively call msg_eval() with the rest of evals
+                # in order to handle nested slices
+                if isinstance(msg, list):
+                    rest = evals[i+1:]
```
dirk-thomas: Please add spaces around the `+` following PEP8.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -788,7 +802,9 @@ def _rostopic_echo(topic, callback_echo, bag_file=None, echo_all_topics=False):
                     index = submsg_class.__slots__.index(field)
                     type_information = submsg_class._slot_types[index]
                     if fields:
-                        submsg_class = roslib.message.get_message_class(type_information)
+                        submsg_class = roslib.message.get_message_class(type_information.split('[')[0])
+                        if not submsg_class:
+                            raise ROSTopicException("Cannot load message class for [%s]. Are your messages built?"%type_information)
```
dirk-thomas: Please add spaces around the`%` following PEP8.

dirk-thomas:
Besides the few small comments the algorithm looks solid. This is a great improvement. Thanks.

tools/rostopic/src/rostopic/__init__.py
rhaschke: Sure. Didn't know about this arg. Always learning new stuff ;-)

tools/rostopic/src/rostopic/__init__.py
rhaschke: Yes. Good point.

tools/rostopic/src/rostopic/__init__.py
rhaschke: Indeed. Cross-checked with additional unittest, checking for /field/ (slash in the end).

tools/rostopic/src/rostopic/__init__.py
rhaschke: added full pattern

rhaschke:
@dirk-thomas I fixed the small issues you listed. Thanks a lot for your comments.

test/test_rostopic/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ if(CATKIN_ENABLE_TESTING)
     Floats.msg
     Simple.msg
     TVals.msg
+    Val.msg
+    ArrayVal.msg
```
dirk-thomas: Please add the new messages in alphabetical order.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -280,50 +280,62 @@ def _rostopic_bw(topic, window_size=-1):
...
-        try:
-            value = _get_nested_attribute(msg, msg_attribute)
-        except AttributeError as e:
-            sys.stdout.write("no field named [%s]"%pattern+"\n")
-            return None
-        if array_index_or_slice_object is not None:
-            value = value[array_index_or_slice_object]
-        return value
-    return msgeval
+        return _msgeval(msg, evals)
```
dirk-thomas: Since `msgeval` is now basically empty you could move the implementation of `_msgeval` here and get rid of `_msgeval`.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

test/test_rostopic/CMakeLists.txt
rhaschke: OK. Fixed.

tools/rostopic/src/rostopic/__init__.py
dirk-thomas: Can you please address this comment too.

tools/rostopic/src/rostopic/__init__.py
dirk-thomas: Can you please address this comment too.

rhaschke: No, it cannot be removed easily. _msgeval() is called recursively with different arguments.
msgeval() defines the one-arg function to be returned, while _msgeval() is a two-arg function.
Probably I could use a lambda expression to mediate the arguments. But, IMHO, that wouldn't increase readability.

dirk-thomas: Oh, I missed that. It can stay as it is then. Sorry for the noise.

tools/rostopic/src/rostopic/__init__.py
rhaschke: Sorry, missed that one.

rhaschke:
I rebased the pull request to integrate the fixups into their corresponding commit.
Does this make the build check fail?

dirk-thomas:
One job failed since the hash was not available anymore when you rebased and force pushed. The newly triggered one passed.

@esteve @tfoote @wjwwood Please review (and if you can give it a try to ensure we don't break anything).

dirk-thomas:
@ros-pull-request-builder

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -280,50 +280,62 @@ def _rostopic_bw(topic, window_size=-1):
...
-        except AssertionError as e:
-            sys.stderr.write("Topic name '%s' contains invalid slice argument '%s': %s\n" % (msg_attribute, index_string, str(e)))
-            return None
-        msg_attribute = msg_attribute[:index]
+    evals = []  # list of (field_name, slice_object) pairs
+    fields = [f for f in pattern.split('/') if f]
+    for f in fields:
+        if '[' in f:
+            field_name, rest = f.split('[', 1)
+            rest = rest[:-1]  # slice content, removing closing bracket
```
wjwwood, line 303: Is there an assumption here about having a trailing `]`? Should there be an assertion to that end? If it's not there, should there be an error? It seems to me that an input of `foo[123` could be incorrectly parsed the same as `foo[12]`. Maybe this isn't an issue due to the overall context, but it isn't clear to me here. There seemed to be a similar check and error on the removed lines 305-307.

wjwwood, line 303: It looks like the test cases do not cover this type of "syntax" exception either.

wjwwood:
Other than the comments this looks good to me, @rhaschke thanks a lot for the contribution and iterating with us on it.

tools/rostopic/src/rostopic/__init__.py
rhaschke, line 303: Valid point. I added an assertion and an appropriate unittest.

dirk-thomas:
@ros-pull-request-builder

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -280,50 +280,60 @@ def _rostopic_bw(topic, window_size=-1):
...
-        except AssertionError as e:
-            sys.stderr.write("Topic name '%s' contains invalid slice argument '%s': %s\n" % (msg_attribute, index_string, str(e)))
-            return None
-        msg_attribute = msg_attribute[:index]
+    evals = []  # list of (field_name, slice_object) pairs
+    fields = [f for f in pattern.split('/') if f]
+    for f in fields:
+        if '[' in f:
+            field_name, rest = f.split('[', 1)
+            assert rest[-1] == ']', "missing closing ']' in slice spec '%s'" % f
```
dirk-thomas: Instead of using `assert` the function should report the problem in the same way as other similar errors (print error message to `stderr` and return `None`).

wjwwood: I agree, that would be preferable.

rhaschke: Sorry, it was too late yesterday. I simply took code from `_get_array_index_or_slice_object()` without noticing that the AssertionError was catched.
Replaced now assert by a simple if and an appropriate warning on stderr.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -788,7 +800,9 @@ def _rostopic_echo(topic, callback_echo, bag_file=None, echo_all_topics=False):
                     index = submsg_class.__slots__.index(field)
                     type_information = submsg_class._slot_types[index]
                     if fields:
-                        submsg_class = roslib.message.get_message_class(type_information)
+                        submsg_class = roslib.message.get_message_class(type_information.split('[',1)[0])
```
dirk-thomas, line 803: Please add a space between `,` and `1` following PEP8

dirk-thomas, line 803: Never mind, I will just apply the change when squashing all commits. Thanks.

dirk-thomas:
Squashed and merged in 7022c7f53dac06587f2c46e185aa3d8286f9e575

Thank you for contributing this enhancement!

rhaschke:
Thanks for patiently guiding me through your code style rules.

Issue #577, opened 2015-03-10
roscpp: crash in ros::TransportPublisherLink::onRetryTimer()


stwirth:
We witnessed a crash of `diagnostic_agregator/aggregator_node` on one of our robots, that is what i could find:

```
Reading symbols from /opt/ros/indigo/lib/diagnostic_aggregator/aggregator_node...(no debugging symbols found)...done.
[New LWP 2295]
[New LWP 2289]
[New LWP 6780]
[New LWP 2284]
[New LWP 2288]
[New LWP 2290]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/opt/ros/indigo/lib/diagnostic_aggregator/aggregator_node __name:=diagnostic_ag'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f59196794ef in ros::TransportPublisherLink::onRetryTimer(ros::WallTimerEvent const&) () from /opt/ros/indigo/lib/libroscpp.so
(gdb) bt
#0  0x00007f59196794ef in ros::TransportPublisherLink::onRetryTimer(ros::WallTimerEvent const&) () from /opt/ros/indigo/lib/libroscpp.so
#1  0x00007f591962cb00 in ros::TimerManager<ros::WallTime, ros::WallDuration, ros::WallTimerEvent>::TimerQueueCallback::call() ()
   from /opt/ros/indigo/lib/libroscpp.so
#2  0x00007f59196554f7 in ros::CallbackQueue::callOneCB(ros::CallbackQueue::TLS*) () from /opt/ros/indigo/lib/libroscpp.so
#3  0x00007f5919656303 in ros::CallbackQueue::callAvailable(ros::WallDuration) () from /opt/ros/indigo/lib/libroscpp.so
#4  0x00007f59196825c4 in ros::internalCallbackQueueThreadFunc() () from /opt/ros/indigo/lib/libroscpp.so
#5  0x00007f5917bcba4a in ?? () from /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.54.0
#6  0x00007f59179aa182 in start_thread (arg=0x7f590bfff700) at pthread_create.c:312
#7  0x00007f59184f947d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```

[...]

PR #615, opened 2015-05-08
fix using callback info reference instead of copy

dirk-thomas:
Addresses a problem mentioned in #577. It is not yet clear if that actually solves the problem completely though.


dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Closing since it is not solving the problem.

Issue #616, opened 2015-05-08
Deadlock in ros::ServiceManager


xqms:
Hi,

I just discovered a deadlock that is happening quite often (once or twice a day in continuous operation) in my application.

The issue is that `ros::ServiceManager` has an internal `shutting_down_mutex_`, which is usually held by all functions operating on the manager.

In my case, the call chains (read from top to bottom) are as follows:

```
First thread:
  ros::service::call()
  ...
  ros::ServiceManager::createServiceServerLink() (acquires the shutting_down_mutex_)
  ...
  TransportTCP::initialize()
  ...
  TransportTCP::enableWrite() (tries to acquire TransportTCP::close_mutex_)

Second thread:
  TransportTCP::socketUpdate() (acquires the TransportTCP::close_mutex_)
  ...
  TransportTCP::close()
  ...
  ServiceManager::removeServiceServerLink() (tries to acquire the shutting_down_mutex_)
```

It seems to me that the `ServiceManager::shutting_down_mutex_` should only be held while checking `ServiceManager::shutting_down_`. But it's hard to tell without a comment on the mutex whether it is also supposed to protect against running `ServiceManager::shutdown()` while another method is already running. Can someone shed light on that?

Here are the full backtraces from gdb (sadly without debug information):

```
First thread:

#0  __lll_lock_wait () at ../nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007f1321329672 in _L_lock_953 () from /lib/x86_64-linux-gnu/libpthread.so.0
#2  0x00007f13213294da in __GI___pthread_mutex_lock (mutex=0x7f12cb4d7610) at ../nptl/pthread_mutex_lock.c:114
#3  0x00007f12feb80d8d in ros::TransportTCP::enableWrite() () from /opt/ros/indigo/lib/libroscpp.so
#4  0x00007f12feb272d9 in ros::Connection::write(boost::shared_array<unsigned char> const&, unsigned int, boost::function<void (boost::shared_ptr<ros::Connection> const&)> const&, bool) () from /opt/ros/indigo/lib/libroscpp.so
#5  0x00007f12feb27588 in ros::Connection::writeHeader(std::map<std::string, std::string, std::less<std::string>, std::allocator<std::pair<std::string const, std::string> > > const&, boost::function<void (boost::shared_ptr<ros::Connection> const&)> const&) () from /opt/ros/indigo/lib/libroscpp.so
[...]

PR #617, opened 2015-05-13
Fix for #616 (Transport tcp deadlock)

xqms:
This is a quick fix for #616. I wrote this on-the-fly, so please check that I interpreted the function of the `close_mutex_` correctly.

It also makes a lot of the `TransportTCP` methods thread-safe. I don't know if that is necessary, since I don't understand all call chains in roscpp...

This has been running here for three days without the deadlocks occuring (which usually happen a few times per day).

### DIFF
--- a/clients/roscpp/include/ros/transport/transport_tcp.h
+++ b/clients/roscpp/include/ros/transport/transport_tcp.h
@@ -141,6 +141,16 @@ private:
 
   socket_fd_t sock_;
   bool closed_;
+
+  /**
+   * This mutex protects the closed_ property and is also held while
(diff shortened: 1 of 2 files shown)


dirk-thomas:
@ros-pull-request-builder

xqms:
This is not ready for merge yet, there is a further deadlock issue involving close(), which is sometimes called with the mutex held. I think we should replace the `recursive_mutex` with a straightforward `mutex`. This way, this type of issue _always_ results in a deadlock, which is much easier to debug than the sporadic deadlocks we currently see.

I'll push a new version for review as soon as it has run for a few days in my environment.

dirk-thomas:

Issue #620, opened 2015-06-03
rosbag play --topics is broken


mintar:
On Indigo, I couldn't find a way to get `rosbag play --topics` to work. IIRC, on Hydro, the following used to work, but produces an error on Indigo:

```
$ rosbag play mybag.bag --clock --topics /mytopic
[ERROR] [1433321952.976544128]: Error reading options: When using --topics or --pause-topics, --bags should be specified to list bags.
```

---

Ok, so I tried to humor rosbag:

```
$ rosbag play --clock --topics /mytopic --bags mybag.bag  
Usage: rosbag play BAGFILE1 [BAGFILE2 BAGFILE3 ...]

rosbag: error: You must specify at least 1 bag file to play back.
```

---

Last try:

```
$ rosbag play mybag.bag --clock --topics /mytopic --bags mybag.bag
[ INFO] [1433321989.060645972]: Opening mybag.bag
[ INFO] [1433321989.061010974]: Opening mybag.bag

Waiting 0.2 seconds after advertising topics... done.

Hit space to toggle paused, or 's' to step.
```

This doesn't produce an error, but of course now every message is replayed twice. Did I miss something obvious, or is this a regression from Hydro?

mintar:
Update: Just confirmed that the first variant (`rosbag play mybag.bag --clock --topics /mytopic`) works in Hydro.

Versions:
- hydro: rosbag 1.10.12 (works)
- indigo: rosbag 1.11.13 (broken)

mintar:
Update: This was introduced in b7796fea by @gerkey : in [line 238](https://github.com/ros/ros_comm/blob/b7796fea666aac7b330aad870f7e7b994eda8daf/tools/rosbag/src/rosbag/rosbag_main.py#L238), setting the --bags arg was removed.

mintar:
So, in summary the problem is this. This is how this worked in hydro:

```
rosbag play mybag.bag --clock --topics /mytopic
```
1. are positional args for bag files present? **yes** [(here)](https://github.com/ros/ros_comm/blob/d13d0263fa632e97500a81bf16250ab0d32cb169/tools/rosbag/src/rosbag/rosbag_main.py#L135-L136)
2. is `--topics` set? **yes** => add `--bags` to args [(here)](https://github.com/ros/ros_comm/blob/d13d0263fa632e97500a81bf16250ab0d32cb169/tools/rosbag/src/rosbag/rosbag_main.py#L224)
[...]

PR #624, opened 2015-06-08
fix handling of bag file arguments

dirk-thomas:
Addresses #620.

@mintar Can you please test the patch in this PR and confirm that it fixes the regression you reported?


mintar:
Yes, just tested the patch, and it fixes #620 ! Thanks!

dirk-thomas:
Thank you for testing it.

PR #625, opened 2015-06-09
rostopic: catch some more error conditions in array index specifications

rhaschke:
This is a follow up on #606. We have overlooked an error condition (namely array index spec with opening bracket but nothing more), which is now caught and covered in the unittest.

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -297,7 +297,7 @@ def msgevalgen(pattern):
     for f in fields:
         if '[' in f:
             field_name, rest = f.split('[', 1)
-            if not rest or rest[-1] != ']':
+            if not rest.endswith(']'):
                 print("missing closing ']' in slice spec '%s'" % f, file=sys.stderr)
                 return None
             rest = rest[:-1]  # slice content, removing closing bracket
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -297,7 +297,7 @@ def msgevalgen(pattern):
     for f in fields:
         if '[' in f:
             field_name, rest = f.split('[', 1)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
@ros-pull-request-builder

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -297,7 +297,7 @@ def msgevalgen(pattern):
     for f in fields:
         if '[' in f:
             field_name, rest = f.split('[', 1)
-            if rest[-1] != ']':
+            if not rest or rest[-1] != ']':
                 print("missing closing ']' in slice spec '%s'" % f, file=sys.stderr)
```
wjwwood, line 301: This error message should also include the possibility that the brackets are just empty, not just that a trailing `]` is missing.

Also, I missed this last time, it would be more idiomatic Python to do `not rest.endswith(']')` rather than `rest[-1] != ']'`.

One more thought is that it might be good to include a test for something like `/vals[/foo` and/or `/vals[]/foo`.

dirk-thomas, line 301: The case `/vals[]` is not caught here but with an assertion in line 306.

rhaschke, line 301: Exactly. That's why I didn't modified the error message. Should I add the suggested test cases?

wjwwood, line 301: I see, nvm then.

dirk-thomas, line 301: Just using `not rest.endswith(']')` seems like a good suggestion. Then the second check is also not necessary.

The additional tests would be nice but the important part is that they are already captured.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Thanks for catching this.

PR #628, opened 2015-06-16
avoid redefining ROS_ASSERT_ENABLED

bricerebsamen:
In order to compile in release mode with ROS_ASSERT enabled, one can force catkin to define ROS_ASSERT_ENABLED on the command line. That works but you end up with a lot of warning about ROS_ASSERT_ENABLED being redefined


dirk-thomas:
Looks straight forward to me. Thanks for the contribution.

PR #594, opened 2015-04-02
Add rosconsole command to change logger levels from command line

gerkey:
This PR replaces #576, adding a handful of fixes.  I'm happy with @cmansley's changes.  @dirk-thomas to review my changes and merge.

PR #631, opened 2015-06-23
Fix bug with missing current logger levels

cmansley:
Since get_loggers is not called, the names and statuses of the loggers is not known to the object. So, it fails in the following if statement because it cannot access current_levels. The effect is that both set and get do not work in the latest jade package.

This was apparently overlooked by @dirk-thomas and @gerkey when we cleaned up from #576 to #594. You can see the offending commit here 8d6b842. @dirk-thomas most likely didn't realize there were side effects of that function and the return value was not used. I should have caught it in review.


dirk-thomas:
Thanks.

PR #632, opened 2015-06-24
Added explicit bool cast for c++11

chadrockey:
I have a use case where I need to wrap a c++11 library with ROS.

While building ROS with the following command, libros fails to build.

./src/catkin/bin/catkin_make_isolated -DCMAKE_CXX_FLAGS="-std=c++11"

I don't believe this will cause a problem for backwards compatibility, so while c++11 isn't officially supported, this shouldn't stop anyone either.

gcc --version
gcc (Ubuntu 4.9.2-0ubuntu1~14.04) 4.9.2

### DIFF
--- a/clients/roscpp/src/libros/timer.cpp
+++ b/clients/roscpp/src/libros/timer.cpp
@@ -95,7 +95,7 @@ Timer::Timer(const TimerOptions& ops)
   impl_->callback_ = ops.callback;
   impl_->callback_queue_ = ops.callback_queue;
   impl_->tracked_object_ = ops.tracked_object;
-  impl_->has_tracked_object_ = ops.tracked_object;
+  impl_->has_tracked_object_ = (ops.tracked_object != NULL);
   impl_->oneshot_ = ops.oneshot;
 }
 
--- a/clients/roscpp/src/libros/wall_timer.cpp
+++ b/clients/roscpp/src/libros/wall_timer.cpp
@@ -95,7 +95,7 @@ WallTimer::WallTimer(const WallTimerOptions& ops)
   impl_->callback_ = ops.callback;
   impl_->callback_queue_ = ops.callback_queue;
   impl_->tracked_object_ = ops.tracked_object;
-  impl_->has_tracked_object_ = ops.tracked_object;
+  impl_->has_tracked_object_ = (ops.tracked_object != NULL);
   impl_->oneshot_ = ops.oneshot;
 }
 


clients/roscpp/src/libros/timer.cpp
```diff
@@ -95,7 +95,7 @@ Timer::Timer(const TimerOptions& ops)
   impl_->callback_ = ops.callback;
   impl_->callback_queue_ = ops.callback_queue;
   impl_->tracked_object_ = ops.tracked_object;
-  impl_->has_tracked_object_ = ops.tracked_object;
+  impl_->has_tracked_object_ = (bool)ops.tracked_object;
```
wjwwood: I think it might be more idiomatic of C++ and safer to use `static_cast<bool>(...)`.

dirk-thomas: Or `ops.tracked_object != NULL`.

wjwwood:
@chadrockey is it possible to get a copy-paste of the build failure for future Google'ers and my own edification. It isn't clear to me why this patch addresses anything.

chadrockey:
@wjwwood Here's the output

```
/home/chadrockey/wkspc/src/ros_comm/roscpp/src/libros/timer.cpp: In constructor ‘ros::Timer::Timer(const ros::TimerOptions&)’:
/home/chadrockey/wkspc/src/ros_comm/roscpp/src/libros/timer.cpp:98:30: error: cannot convert ‘const VoidConstPtr {aka const boost::shared_ptr<const void>}’ to ‘bool’ in assignment
   impl_->has_tracked_object_ = ops.tracked_object;
                              ^
/home/chadrockey/wkspc/src/ros_comm/roscpp/src/libros/wall_timer.cpp: In constructor ‘ros::WallTimer::WallTimer(const ros::WallTimerOptions&)’:
/home/chadrockey/wkspc/src/ros_comm/roscpp/src/libros/wall_timer.cpp:98:30: error: cannot convert ‘const VoidConstPtr {aka const boost::shared_ptr<const void>}’ to ‘bool’ in assignment
   impl_->has_tracked_object_ = ops.tracked_object;

```

From my research, c++11 removes some auto casting to boolean (the safe bool idiom).  The fix is to explicity call the type you wish to cast to.  I think this is so you don't do funny things like bitshifting an auto bool cast of a pointer and whatnot.

clients/roscpp/src/libros/timer.cpp
chadrockey: @wjwwood I think I agree, so I've updated to the static_cast.

wjwwood:
I see, I didn't notice that it was a boost shared pointer. I think @dirk-thomas's solution is clearer than a cast. The other appropriate solution would be:

``` c++
impl_->has_tracked_object_ = ops.tracked_object ? true : false;
```

Since that forces the evaluation of `ops.tracked_object` into a scenario where the safe bool idiom works. Like I said @dirk-thomas solution is probably the clearest and most correct.

chadrockey:
@wjwwood thanks for your advice.  I've updated the pull request accordingly.  I think it's still a little hard to grasp the details of this change in c++11, especially as the behavior before seems very prevalent.

PR #634, opened 2015-06-29
mplayer like speed controls

lucasw:
Press [ to slow playback speed 10%, or ] to increase speed 10%, or backspace to reset to rate 1.0  (though maybe if the user supplied a rate at the command line backspace should reset to that?).  

I'd like to add the ability to skip forward or back five seconds next, though my first attempt at using time_publisher_.setTime() hasn't worked.

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -88,13 +88,15 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
     elif os.path.isfile(logfile_dir):
         raise LoggingException("Cannot save log files: file [%s] is in the way"%logfile_dir)
 
-    if sys.platform not in ['win32']:
-        try:
-            success = renew_latest_logdir(logfile_dir)
-            if not success:
-                sys.stderr.write("INFO: cannot create a symlink to latest log directory\n")
-        except OSError as e:
-            sys.stderr.write("INFO: cannot create a symlink to latest log directory: %s\n" % e)
+    # the log dir itself should not be symlinked as latest
+    if logfile_dir != log_dir:
+        if sys.platform not in ['win32']:
+            try:
+                success = renew_latest_logdir(logfile_dir)
+                if not success:
+                    sys.stderr.write("INFO: cannot create a symlink to latest log directory\n")
+            except OSError as e:
+                sys.stderr.write("INFO: cannot create a symlink to latest log directory: %s\n" % e)
 
     if 'ROS_PYTHON_LOG_CONFIG_FILE' in os.environ:
         config_file = os.environ['ROS_PYTHON_LOG_CONFIG_FILE']


lucasw:
I figured out how to skip forward and backward any number of seconds, I'll put it in a future pull request.

wjwwood:
Thanks for the pr @lucasw.

I'll have a look at the pr, but it might take a few days for all of us to review it, as some people are traveling.

tools/rosbag/src/player.cpp
```diff
@@ -201,7 +202,7 @@ void Player::publish() {
     options_.advertise_sleep.sleep();
     std::cout << " done." << std::endl;
 
-    std::cout << std::endl << "Hit space to toggle paused, or 's' to step." << std::endl;
+    std::cout << std::endl << "Hit space to toggle paused, or 's' to step, [ or ] to decrease/increase playback speed, backspace to set speed to normal." << std::endl;
```
wjwwood: `Hit space to toggle paused, 's' to step, '['/']' to decrease/increase playback speed, or backspace to set speed to normal.`

tools/rosbag/src/player.cpp
```diff
@@ -209,17 +210,25 @@ void Player::publish() {
         // Set up our time_translator and publishers
 
         time_translator_.setTimeScale(options_.time_scale);
-
+
```
wjwwood: Please remove any trailing whitespace.

tools/rosbag/src/player.cpp
```diff
@@ -227,15 +236,20 @@ void Player::publish() {
             time_publisher_.setPublishFrequency(-1.0);
 
         paused_time_ = now_wt;
-
-        // Call do-publish for each message
```
wjwwood, line 237: Why remove this comment?

tools/rosbag/src/player.cpp
```diff
@@ -364,6 +382,33 @@ void Player::doPublish(MessageInstance const& m) {
...
+                time_translator_.setTimeScale(options_.time_scale);
+                break;
+            case ',':
+                skip_time_ = true;
+                new_time_ = time_publisher_.getTime() - ros::Duration(5.0);
+                break;
+            case '.':
+                skip_time_ = true;
+                new_time_ = time_publisher_.getTime() + ros::Duration(5.0);
+                break;
```
wjwwood: This feature isn't mentioned in the line above that describes the possible hotkeys.

tools/rosbag/src/player.cpp
```diff
@@ -329,7 +343,11 @@ void Player::doPublish(MessageInstance const& m) {
         bool charsleftorpaused = true;
         while (charsleftorpaused && node_handle_.ok())
         {
-            switch (readCharFromStdin()){
+            const int key = readCharFromStdin();
+            //if (key >= 0)
+            //    ROS_INFO_STREAM("key " << key);
```
wjwwood: You should change this back to just `switch(readCharFromStdin()) {`.

tools/rosbag/src/player.cpp
```diff
@@ -227,15 +236,20 @@ void Player::publish() {
...
-        // Call do-publish for each message
+        
         foreach(MessageInstance m, view) {
             if (!node_handle_.ok())
                 break;
 
             doPublish(m);
+
+            if (skip_time_)
+                break;
```
wjwwood: Why wouldn't this either wrap around the `foreach` as an `if` statement or at least `break` before the first `doPublish`? Put another way, why does this only do one `doPublish` when `skip_time_` is `true`?

wjwwood:
Ok, I've added some comments. I'll have to try out the pr later; the skipping seems tricky and seems like it might not cover all the corner cases, like skipping backwards at the beginning of a bag and skipping forwards at the end. It might handle these as is, but that's not clear to me from looking at the code.

tools/rosbag/src/player.cpp
lucasw: I meant those changes for a future pull request (and to debug them more), but as it is I'll address the other comments and add the skip forward/back to the hotkey description.

tools/rosbag/src/player.cpp
lucasw: `skip_time_` ought to always be false entering into the `foreach`, and is only set false within `doPublish` if a skip key is pressed.  (Though there are one or more unwanted doPublishes whenever I use the skip function, I see a glitch in rqt_plot when playing a bag with a steadily increasing published value and skipping forward or backward)

lucasw:
Cool, thanks for all the comments!  The skipping definitely deserves more attention, there is logic there to set the time to `start_time_` when it is set earlier than it.  When skipping too far forward you should see it does exit rosbag if looping is off, and if looping is true it does loop, but I haven't tested the interactions with the other options yet.

dirk-thomas:
Thanks for contributing new features like this.

As you mentioned before the `reset` operation should set the speed back to the initial rate value (which might be different from `1.0`).

Please also address the problem you mentioned regarding not exiting when fast forwarding and verify the interaction with other options.

lucasw:
I'd like to put this onto https://github.com/lucasw/ros_comm/tree/mplayer_like_speed_controls (so my indigo-devel can track the one here)- is it possible to edit this pr to reflect that?  If not and I get back to this should I start a new pr anyhow?

dirk-thomas:
You can't change the branch name used for a PR. In that case you need to open a new PR.

lucasw:
Okay, closing this since it no longer points at the right branch, but I'll refer back to this upon making a new pr.

PR #636, opened 2015-07-09
Allowed saving timestamp-less messages to Cache (using rospy.Time.now() as their timestamp).

peci1:
Also added a convenience `getLast` method for retrieving the last message in the cache.


tfoote:
I don't believe we should ever automatically stamp unstamped data. An alternative would be to provide a 2nd argument to which the user could add the timestamp explicitly.

peci1:
Okay, what about adding an `__init__` argument defaulting to False, which, set to True, would allow for automatic timestamping?

dirk-thomas:
That constructor argument sounds like a good option.

utilities/message_filters/src/message_filters/__init__.py
```diff
@@ -121,19 +122,21 @@ def add(self, msg):
     def getInterval(self, from_stamp, to_stamp):
         """Query the current cache content between from_stamp to to_stamp."""
         assert from_stamp <= to_stamp
-        return [m for m in self.cache_msgs
-                if m.header.stamp >= from_stamp and m.header.stamp <= to_stamp]
+        return [self.cache_msgs[i] for i in range(0, len(self.cache_msgs))
+                if self.cache_times[i] >= from_stamp and self.cache_times[i] <= to_stamp]
```
dirk-thomas: Please keep the list comprehension as simple as possible, e.g.:

```
[ m for i, m in enumerate(self.cache_msgs) if self.cache_times[i] >= from_stamp and self.cache_times[i] <= to_stamp]
```

Or even better you could even use `zip` to iterate over both lists at the same time.

peci1:
I've implemented your suggestion with the constructor argument. I've also added tests. Should I retarget to kinetic-devel, or is it okay this way?

tfoote:
lgtm, it would be better to target for kinetic and consider backporting only if there's strong demand

peci1:
Ok, see #806 .

Issue #137, opened 2013-01-31
Allow rostest to attach to existing master  (ros ticket #2351)


tfoote:
Sometimes it's much more convenient to attach rostest to an existing master on the default port (11311).  I can leave things like rxplot and rviz up and connected as I switch between running rostests and running other applications.  An option to rostest for attaching to my existing master would be helpful.

trac data:
- Owner: **kwc**
- Reporter: **sglaser**
- Reported at: **Fri Jan 15 16:25:02 2010**
- URL: https://code.ros.org/trac/ros/ticket/2351

tfoote:
[kwc] This is a fairly tough thing to allow, at least as described. Part of rostest's architecture is that for full launches it brings up and tears down a master between each run in order to keep tests independent and provide them with the same initial state (e.g. parameters).

There are other ways of looking at the problem that might provide more straightforward technical solutions and not confuse the contract of what rostest is doing. One is that a rostest file can be roslaunch'd, so I could just add an option to rostest to launch a named test within a test file, which would run against any existing launch/master. 

I'm assuming what you're describing implies that you want to focus on a specific test, right? Having rxplot bounce around as a bunch of different tests run seems less useful.

tfoote:
[sglaser] The use-case is that some test(s) is failing and I want to run the test locally, attaching visualization nodes, possibly some scaffolding nodes, and maybe changing parameters around.
1. The master should stay up (so that the scaffolding and viz nodes stay connected).
2. The master should be on the default port (all my environments assume 11311.  I think it's convenient to have the test connect to this one too).
3. Different tests should connect to the same master (so I can quickly switch between tests as I change functionality).

Understand that I don't expect rostest to have this behavior by default, I just want an option to enable this behavior.

Looking closely, I'm pretty sure you've described what I want:

> One is that a rostest file can be roslaunch'd, so I could just add an option to rostest to launch a named test within a test file, which would run against any existing launch/master.

trainman419:
+1. This would be useful to me as well.

dirk-thomas:
Closed by #637.

PR #614, opened 2015-05-04
Arg port number for rostest

130s:
Mentioned in http://answers.ros.org/question/208361/port-to-connect-to-rostest-rosmaster-after-indigo/?answer=208387#post-id-208387

With this PR, you can specify the port number of the `rostest`. E.g.,

```
term-1$ rostest leap_motion test_leap_motion.test --port 22423

term-2$ ROS_MASTER_URI=http://tork-kudu1:22423 rosparam get /sender_freq/freq
0.1
```

Drawback:
- Apparently this only works with a single test case. So I don't know if we want to merge this as it is. 
- I've tried to add test files for this too which I couldn't get working.

PR #637, opened 2015-07-14
Add --local option to rostest. Implements #137

trainman419:
Add an option to rostest which suppresses the behavior of spawning a ROS
master, and instead uses the default roslaunch behavior of connecting to
the ROS master specified in ROS_MASTER_URI, or creating a new ROS
master.

There is one final question here: with the --local option, the rostest environment is polluted by any parameters that already exist on the running ROS master.

I think it is reasonable to have an option for rostest to delete all of the existing parameters, so that the test can run with a clean parameter space. Should this option default to being on or off?

### DIFF
--- a/tools/rostest/package.xml
+++ b/tools/rostest/package.xml
@@ -16,9 +16,10 @@
   <build_depend>rosunit</build_depend>
 
   <run_depend>boost</run_depend>
-  <run_depend>rospy</run_depend>
+  <run_depend>rosgraph</run_depend>
   <run_depend>roslaunch</run_depend>
   <run_depend>rosmaster</run_depend>
+  <run_depend>rospy</run_depend>
   <run_depend>rosunit</run_depend>
 
   <export>
--- a/tools/rostest/src/rostest/rostest_main.py
+++ b/tools/rostest/src/rostest/rostest_main.py
@@ -98,7 +98,16 @@ def rostestmain():
     parser.add_option("--results-base-dir", metavar="RESULTS_BASE_DIR",
                       help="The base directory of the test results. The test result file is " +
                            "created in a subfolder name PKG_DIR.")
+    parser.add_option("-r", "--reuse-master", action="store_true",
+                      help="Connect to an existing ROS master instead of spawning a new ROS master on a custom port")
+    parser.add_option("-c", "--clear", action="store_true",
+                      help="Clear all parameters when connecting to an existing ROS master (only works with --reuse-master)")
     (options, args) = parser.parse_args()
+
+    if options.clear and not options.reuse_master:
+        print("The --clear option is only valid with --reuse-master", file=sys.stderr)
+        sys.exit(1)
+
     try:
         args = roslaunch.rlutil.resolve_launch_arguments(args)
     except roslaunch.core.RLException as e:
@@ -147,7 +156,7 @@ def rostestmain():
         parser.error("test file is invalid. Generated failure case result file in %s"%results_file)
         
     try:
-        testCase = rostest.runner.createUnitTest(pkg, test_file)
+        testCase = rostest.runner.createUnitTest(pkg, test_file, options.reuse_master, options.clear)
         suite = unittest.TestLoader().loadTestsFromTestCase(testCase)
 
         if options.text_mode:
--- a/tools/rostest/src/rostest/rostest_parent.py
+++ b/tools/rostest/src/rostest/rostest_parent.py
@@ -35,6 +35,7 @@
 import logging
 import sys
 
+import rosgraph
 import roslaunch.config
 from roslaunch.core import printlog_bold, RLException
 import roslaunch.launch
@@ -45,17 +46,46 @@ import roslaunch.xmlloader
 import roslaunch.parent
 
 from rosmaster.master import Master
+from rospy import logwarn
 
 class ROSTestLaunchParent(roslaunch.parent.ROSLaunchParent):
 
-    def __init__(self, config, roslaunch_files, port=0):
+    def __init__(self, config, roslaunch_files, port=0, reuse_master=False, clear=False):
         if config is None:
             raise Exception("config not initialized")
         # we generate a run_id for each test
-        run_id = roslaunch.core.generate_run_id()
+        if reuse_master:
+            param_server = rosgraph.Master('/roslaunch')
+            try:
+                run_id = param_server.getParam('/run_id')
+            except Exception as e:
+                # The user asked us to connect to an existing ROS master, and
+                # we can't. Throw an exception and die
+                raise Exception("Could not connect to existing ROS master. "
+                                + "Original exception was: %s" % str(e))
+            except:
+                # oh boy; we got something that wasn't an exception.
+                # Throw an exception and die
+                raise Exception("Could not connect to existing ROS master.")
+
+            if clear:
+                params = param_server.getParamNames()
+                # whitelist of parameters to keep
+                whitelist = ['/run_id', '/rosversion', '/rosdistro']
+                for i in reversed(range(len(params))):
+                    param = params[i]
+                    if param in whitelist:
+                        del params[i]
+                    elif param.startswith('/roslaunch/'):
+                        del params[i]
+                for param in params:
+                    param_server.deleteParam(param)
+        else:
+            run_id = roslaunch.core.generate_run_id()
         super(ROSTestLaunchParent, self).__init__(run_id, roslaunch_files, is_core=False, is_rostest=True)
         self.config = config
         self.port = port
+        self.reuse_master = reuse_master
         self.master = None
         
     def _load_config(self):
@@ -67,9 +97,10 @@ class ROSTestLaunchParent(roslaunch.parent.ROSLaunchParent):
         initializes self.config and xmlrpc infrastructure
         """
         self._start_infrastructure()
-        self.master = Master(port=self.port)
-        self.master.start()
-        self.config.master.uri = self.master.uri
+        if not self.reuse_master:
+            self.master = Master(port=self.port)
+            self.master.start()
+            self.config.master.uri = self.master.uri
         self._init_runner()
 
     def tearDown(self):
--- a/tools/rostest/src/rostest/runner.py
+++ b/tools/rostest/src/rostest/runner.py
@@ -190,7 +190,7 @@ def setUp(self):
     # new test_parent for each run. we are a bit inefficient as it would be possible to
     # reuse the roslaunch base infrastructure for each test, but the roslaunch code
     # is not abstracted well enough yet
-    self.test_parent = ROSTestLaunchParent(self.config, [self.test_file])
+    self.test_parent = ROSTestLaunchParent(self.config, [self.test_file], reuse_master=self.reuse_master, clear=self.clear)
     
     printlog("setup[%s] run_id[%s] starting", self.test_file, self.test_parent.run_id)
 
@@ -212,7 +212,7 @@ def tearDown(self):
         
     printlog("rostest teardown %s complete", self.test_file)
     
-def createUnitTest(pkg, test_file):
+def createUnitTest(pkg, test_file, reuse_master=False, clear=False):
     """
     Unit test factory. Constructs a unittest class based on the roslaunch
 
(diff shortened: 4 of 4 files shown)


bobsomers:
I think it makes sense for the "delete all existing parameters" option to default to off. Although deleting the parameters by default _would_ help ensure the integrity of the test environment, I think more importantly that behavior could be considered surprising. I wouldn't immediately expect that running rostest against my local master would delete all of my parameters without asking.

adolfo-rt:
What would be the gain of a) using an existing ROS master and deleting all parameters, over b) spawning a new ROS master with a different ROS_MASTER_URI?.

trainman419:
@adolfo-rt : The existing ROS master may have an instance of rqt_plot or rviz attached to it, so that the user doesn't have to include their visualization as part of the test.

If also means that the existing ROS master will persist after the test is over, allowing the user to introspect any parameters that were set on the master as a result of the test.

adolfo-rt:
@trainman419, I get your point. When I want to debug a rostest, or manually introspect what's going on (during the test, or post-mortem), what I usually do is a two-step approach:
- fire up the `.test` file with `roslaunch`, which starts everything except the actual rostest.
- Use `rosrun` to start the rostest.

The above is something I rarely do, typically only when coding a test or fixing a regression. Considering that there is already a human in the loop, I don't mind running two commands instead of just one. For automated testing, which accounts for most of the times the tests are ran, I've never had the need to spawn a rostest in an existing ROS master. Of course, your usecases and preferences might vary.

trainman419:
I've implemented the parameter clearing as discussed.

@adolfo-rt : our team is currently creating and debugging a large volume of automated tests, and the ability to easily run both the test and the test node against an existing ROS master will improve our productivity significantly.

This PR also addresses a longstanding need for better control over how tests are run, and the ability to easily attach additional nodes to a test session for debugging; see #137 and #614.

This change also implicitly provides the functionality requested in #614, by auto-starting a new master on the port specified in the ROS_MASTER_URI if one does not exist.

trainman419:
This passed the pull request builder. Can one of the package maintainers take a look at it please?

wjwwood:
Sorry for the delay, we're planning on discussing these open issues in our next ROS 1 meeting.

tools/rostest/src/rostest/rostest_parent.py
```diff
@@ -35,6 +35,7 @@
 import logging
 import sys
 
+import rosgraph
```
wjwwood, line 38: Should this translate to an explicit dependency on `rosgraph`? It doesn't currently depend directly on it:

https://github.com/ros/ros_comm/blob/indigo-devel/tools/rostest/package.xml

tools/rostest/src/rostest/rostest_parent.py
```diff
@@ -48,14 +49,37 @@
...
         # we generate a run_id for each test
-        run_id = roslaunch.core.generate_run_id()
+        if local:
+            param_server = rosgraph.Master('/roslaunch')
+            have_master = False
+            try:
+                run_id = param_server.getParam('/run_id')
+                have_master = True
+            except:
+                run_id = roslaunch.core.generate_run_id()
```
wjwwood: The catch all here is a bit sketchy, maybe something like this would be more appropriate?:

``` python
have_master = False
try:
    run_id = param_server.getParam('/run_id')
    have_master = True
except <Some expected exception> as e:
    rospy.logwarn(...)
<repeat as needed>
finally:
    if have_master is False:
        run_id = roslaunch.core.generate_run_id()
```

I'm worried the catch all might erroneously suppress some networking exception or some other unforeseen error.

tools/rostest/src/rostest/rostest_parent.py
```diff
@@ -48,14 +49,37 @@
...
+
+            if clear and have_master:
+                params = param_server.getParamNames()
+                # whitelist of parameters to keep
+                whitelist = ['/run_id', '/rosversion', '/rosdistro']
```
wjwwood, line 74: This seems fragile to me. The best thing would be to extend the Master API to provide this functionality and then the parameter server could clear itself. That being said, this is definitely much less intrusive and probably ok in practice.

tools/rostest/src/rostest/rostest_parent.py
trainman419: This is copied nearly verbatim from roslaunch, so I'm not really sure which exceptions to expect. Would a catch-all and print be more reasonable?

https://github.com/ros/ros_comm/blob/78aa1c4ec879668f686aba873f52968b1bd7aa73/tools/roslaunch/src/roslaunch/rlutil.py#L173-L181

wjwwood:
Makes sense, and code changes look ok to me (other than a few comments). +1

tools/rostest/src/rostest/rostest_parent.py
trainman419, line 38: Yes; I'll add this as a dependency.

tools/rostest/src/rostest/rostest_parent.py
wjwwood: Yeah, that code from `roslaunch` has the same deficiency. I'd be happy with a message.

One thing that brings up is that the roslaunch snippet you referenced busy waits on the master, why shouldn't this code do the same? Is it because that should never happen? And if that's the case why catch and silently pass at all? It's not clear to me what might happen and what this `--local` option should do in each of those cases.

trainman419: roslaunch only busy-waits if the `--wait` command line option in specified (which sets `options_wait_for_master`). rostest does not have a `--wait` option, so I've omitted that behavior here.

otherwise, both rostest and roslaunch will spawn a new, default master to use for the run.

I'll add a generic exception catch here and print a warning if the master could not be reached.

wjwwood: Sounds good to me.

trainman419:
I've implemented the final exception catch and warning as suggested.

trainman419:
It looks like the automated test failed because it aborted; is there a way to retry it?

dirk-thomas:
@ros-pull-request-builder

trainman419:
Poke. I think I've made all of the requested changes, and the pull request builder is passing. Is there anything else I can do to help this along?

wjwwood:
I think we're just trying to schedule a time to do a core packages release. This will definitely get a final review and an opportunity to get merged before the next release.

dirk-thomas:
I have created PR #655 with some changes as well as fixes to this patch. Therefore closing this.

dirk-thomas:
@trainman419 is going to update this PR instead.

trainman419:
I've renamed the `--local` option to `--reuse-master` and updated the short option from `-l` to `-r`.

tools/rostest/src/rostest/rostest_main.py
```diff
@@ -98,6 +98,10 @@ def rostestmain():
     parser.add_option("--results-base-dir", metavar="RESULTS_BASE_DIR",
                       help="The base directory of the test results. The test result file is " +
                            "created in a subfolder name PKG_DIR.")
+    parser.add_option("-r", "--reuse-master", action="store_true",
+                      help="Connect to an existing ROS master instead of spawning a new ROS master on a custom port")
+    parser.add_option("-c", "--clear", action="store_true",
+                      help="Clear all parameters when connecting to existing ROS master (useful with --reuse-master)")
```
dirk-thomas: Please clarify that this option is only effective when `--reuse-master` is specified. argparse could even error out when the option is used without `--reuse-master`.

tools/rostest/package.xml
```diff
@@ -20,6 +20,7 @@
   <run_depend>roslaunch</run_depend>
   <run_depend>rosmaster</run_depend>
   <run_depend>rosunit</run_depend>
+  <run_depend>rosgraph</run_depend>
```
dirk-thomas: Please insert in alphabetical order.

tools/rostest/src/rostest/rostest_parent.py
```diff
@@ -48,14 +49,46 @@
...
         # we generate a run_id for each test
-        run_id = roslaunch.core.generate_run_id()
+        if reuse_master:
+            param_server = rosgraph.Master('/roslaunch')
+            have_master = False
+            try:
+                run_id = param_server.getParam('/run_id')
+                have_master = True
+            except Exception as e:
+                rospy.logwarn("Could not connect to existing ROS master; will create"
```
dirk-thomas: `rospy.logwarn` has not been imported in this file and fails at runtime.

Same below.

trainman419:
I've made the requested changes to clarify --clear and clean up the dependencies and imports.

What do you think about @chpatton013's comment on #655 about returning an error in the event that an existing ROS master is not available, instead of spawning a new master?

dirk-thomas:
That sounds like a good change. The user asked for it so it is ok to exit if that can't be satisfied.

trainman419:
Ok; I updated the PR so that rostest throws an exception if it can't contact an existing master.

dirk-thomas:
@ros-pull-request-builder

trainman419:
@ros-pull-request-builder

dirk-thomas:
The patch looks good to me. The failing 3 unit tests are (sadly) expected. Can you please squash the 10 commits into one or a few semantic commits before the merge? Thanks!

trainman419:
Squashed and pushed.

--clear is dependent on --reuse-master, so I squashed those commits together.

trainman419:
@ros-pull-request-builder

dirk-thomas:
Thank you for iterating fast on this.

trainman419:
Do you have an estimated timeline for getting this released into Indigo?

dirk-thomas:
I am still waiting for others to look into the currently failing 3 tests: http://jenkins.ros.org/job/devel-indigo-ros_comm/

But I expect that a decision to release will be made next week - either with the tests being fixed before (highly preferred) or even if these tests can't be resolved by then. It has hold up the release for way to long already.

trainman419:
All of the tests pass on my Trusty/Indigo machine, but it has out-of-date system packages.

If I have time, I'll try upgrading the packages on it one at a time to see if I can isolate this to a particular upstream package change.

dirk-thomas:
Locally all tests pass for me too.

I also don't think that it is related to specific code change in `ros_comm` itself. The CI job had no failing tests for the last release but when being built. But when I run the job with the git hash from the last release it fails the three tests in question...

trainman419:
Which tests are failing?

From [this test report](http://jenkins.ros.org/job/_pull_request-indigo-ros_comm/367/testReport/), the top-level tags list `testrosservice_unit` , `testrostopic_unit` and `testcheck_master_false` as failing, but the standard output from all three tests shows the test file for `message_filters/test/test_subscriber.xml`

To me, it almost looks like the wrong test is running, or there's a stale test file from a previously failed test on the build machine.

dirk-thomas:
The three test you mentioned are exactly the ones which are failing.

I can't follow why you think the wrong is being executed. I can find `test_subscriber.xml` only in one area of the console output which is related to that specific test and which passes. Can you please describe how you came to that conclusion?

Since the workspace is cleaned before every build I doubt any result files can be from a previous build.

trainman419:
I see references to `test_subscriber.xml` when expanding the `Standard Output` of all three tests on this page: http://jenkins.ros.org/job/_pull_request-indigo-ros_comm/367/testReport/

Screenshot: http://namniart.com/Test_ros_comm.pdf

dirk-thomas:
I assume you refer to http://jenkins.ros.org/job/_pull_request-indigo-ros_comm/367/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/testReport/rostest.runner/RosTest/testrosservice_unit/ ?

I am not sure where Jenkins gets that information from. The test result file (which I think is the corresponding one) looks different: http://jenkins.ros.org/job/_pull_request-indigo-ros_comm/ws/ARCH_PARAM/amd64/UBUNTU_PARAM/trusty/label/devel/test_results/repos/test_rosservice/rostest-test_rosservice.xml/*view*/

trainman419:
I can't follow that link; I get a permission error: `ahendrix is missing the Job/Workspace permission`

PR #640, opened 2015-07-17
update error message from manifest.xml -> package.xml

k-okada:
update error message, manifest.xml is old name.


dirk-thomas:
Thanks.

PR #641, opened 2015-07-17
using rl_config.nodes() to get all node information

k-okada:
current roslaunch-check checks if all nodes exists, even if that tag is disabled by if statement, for example

```
<launch>
 <group if="true"> <node pkg="foo" name="bar_exists" type="valid" /> </group>
 <group if="false"> <node pkg="foo" name="bar_dummy" type="invalid" /> </group>
</launch>
```

and `valid` node exists and `invalid` node does not, current roslaunch-check implementation is try to check both nodes and returns failure because of missing `invalid` node. (real example is [1]  ,which naoqi_joint_states_cpp is not distributed ad deb due to some reason (may be copyright issue)

[1] https://github.com/ros-naoqi/naoqi_bridge/blob/0.4.8/naoqi_driver/launch/naoqi_driver.launch

### DIFF
--- a/tools/roslaunch/src/roslaunch/rlutil.py
+++ b/tools/roslaunch/src/roslaunch/rlutil.py
@@ -211,8 +211,8 @@ def check_roslaunch(f):
     
     # load all node defs
     nodes = []
-    for filename, rldeps in file_deps.items():
-        nodes.extend(rldeps.nodes)
+    for node in rl_config.nodes: # using rl_config instead of 'for filename, rldeps in file_deps.items():' for evaluating if/unless
+        nodes.extend([(node.package, node.type)])
 
     # check for missing packages
     rospack = rospkg.RosPack()


tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -211,8 +211,8 @@ def check_roslaunch(f):
     
     # load all node defs
     nodes = []
-    for filename, rldeps in file_deps.items():
-        nodes.extend(rldeps.nodes)
+    for node in rl_config.nodes():
```
dirk-thomas: This result in the following error for me:

```
TypeError: 'list' object is not callable
```

dirk-thomas: Even after removing the parenthesis `roslaunch-check` does not work for me.

k-okada:
sorry for confusion, I have updated the PR.

dirk-thomas:
I am not if this is a good idea. Currently you can run `roslaunch-check` and know if the launch file will find all resources. WIth these changes it could be that some parts are dynamically enabled / disabled and `roslaunch-check` only checks part of the launch file. So in order to know if the launch file would work (in all cases) you would need to invoke the script with any kind of input in order to ensure that all conditional parts are valid.

Maybe you can elaborate on your use case for this?

k-okada:
I need this feature when I try to check my launch file including,https://github.com/ros-naoqi/naoqi_bridge/blob/0.4.8/naoqi_driver/launch/naoqi_driver.launch#L21.
Usually we do not run naoqi_joint_state.py because it requires Python NaoQi SDK, which is not able to install via rosdep program nor we need additional licence to download them. 

Of course, there are way to avoid this use case, but in general the case we met is when launch file provide input option to select opensource environment OR environment including propriety software, and user only test in opensource environment.

dirk-thomas:
The problem with this patch is that there is no way anymore to check your complete roslaunch file. I would argue that in your use case the launch file should be split into:
- the main launch file which either includes
  - the launch file using the opensource environment or
  - the launch file using the proprietary software

Then `roslauch-check` can check the combination which is selected as a default.

k-okada:
> The problem with this patch is that there is no way anymore to check your complete roslaunch file. 

yes, that's true

PR #642, opened 2015-07-22
make support for lz4 in rosbag optional

dirk-thomas:
When `roslz4` can't be imported (e.g. when using the rosbag API with a different Python version) the patch makes rosbag gracefully fall back to work except without support for lz4.

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -62,7 +62,14 @@ import genpy.message
 
 import roslib.names # still needed for roslib.names.canonicalize_name()
 import rospy
-import roslz4
+try:
+    import roslz4
+    found_lz4 = True
+except ImportError:
+    rospy.logwarn(
+        'Failed to load Python extension for LZ4 support. '
+        'LZ4 compression will not be available.')
+    found_lz4 = False
 
 class ROSBagException(Exception):
     """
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@ros-pull-request-builder

tools/rosbag/src/rosbag/bag.py
```diff
@@ -62,7 +62,11 @@
 
 import roslib.names # still needed for roslib.names.canonicalize_name()
 import rospy
-import roslz4
+try:
+    import roslz4
+    found_lz4 = True
+except ImportError:
+    found_lz4 = False
```
wjwwood, line 72: I would at least do a rospy.logdebug message to mention that the roslz4 import failed. The only time it's ok for this to happen is in the very narrow use case you have. Otherwise it is an exception and one that has occurred before (when lz4 was built for a different version of Python than what's currently installed, e.g. 2.7.9 versus 2.7.10). I'd even argue for a warning or info message, but at the very least there should be some message about it. Examples of issues found because lz4 is not importable:

http://answers.ros.org/question/213728/issues-with-lz4/
http://answers.ros.org/question/197278/importerror-no-module-named-_roslz4-after-installing-from-source/

Granted they may have not been trying to use lz4, but they would have never known their systems were at fault unless this produced an error.

dirk-thomas:
I have added a warning message.

wjwwood:
+1

Issue #644, opened 2015-07-24
topic_tools transform doesn't get ros arguments


lucasw:
It isn't possible to use `transform` in a launch file, or run it in a command line with `__name` or `__ns` or similar ros arguments, it just results in `transform: error: unrecognized arguments` errors for them.

http://answers.ros.org/question/214269/topic_tools-transform-from-launch-file

dirk-thomas:
This has been addressed by 61d987de96b672291dc76592d3c5be28ce72ff5a.

PR #645, opened 2015-07-24
Allow transform to be used with ros arguments and in a launch file #644

lucasw:
Get ros args with rospy.argv before passing to argparse.


dirk-thomas:
Thanks. I have cherry-picked your patch (but renamed `arg0` to `argv`) in 61d987de96b672291dc76592d3c5be28ce72ff5a

PR #646, opened 2015-08-04
topic_tools/scripts/transform: add --wait-for-start

k-okada:
add wait-for-start option to avoid exitting this script in the first time

current implemnetation exit the scirpt if the input topic is not found, but if you start `transform` with other node in the launch file, usually it takes a few seconds to start publishing topics, so `transform` node exits in the first time, this option avoid such situation

### DIFF
--- a/tools/topic_tools/scripts/transform
+++ b/tools/topic_tools/scripts/transform
@@ -55,6 +55,10 @@ class TopicOp:
             '-i', '--import', dest='modules', nargs='+', default=['numpy'],
             help='List of Python modules to import.'
         )
+        parser.add_argument(
+            '--wait-for-start', dest='wait_for_start', action='store_true',
+            help='Wait for input messages.'
+        )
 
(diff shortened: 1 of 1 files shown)


wjwwood:
Why does this option take a time parameter? Based on the situation you described shouldn't it just be wait for ever? Even if you have an option that takes a timeout, I think it should also be possible to wait for ever.

k-okada:
I just thought that we'd like to know if the input message is not published for some reason (wrong remap ore something) or just wait for 1st message, but yes,  wait for ever  is much more simple.

note : old implementation -> https://github.com/k-okada/ros_comm/commit/f3b7b20a8afdb57319bb7fa09962e8ac650b15e1

dirk-thomas:
Thank you.

I have slightly modified the patch and cherry-picked it in 32edfbf86ddda1cb14c3638e4ebc5dd9850a6aa3

Issue #647, opened 2015-08-05
rosbag record: specifying the same topic multiple times causes duplicate messages 


dlaz:
For example

```
rosbag record /foo /foo
rostopic pub -1 /foo std_msgs/Empty
```

results in

```
types:       std_msgs/Empty [d41d8cd98f00b204e9800998ecf8427e]
topics:      /foo   2 msgs    : std_msgs/Empty
```

While this can probably be considered user error, I can't think of a situation when this is actually the desired behavior that rosbag should support. If this is indeed not the desired behavior, I'll submit a PR fixing it (eventually).

dirk-thomas:
This was addressed by #648.

PR #648, opened 2015-08-15
Remove duplicate topics from being recorded into one topic

spaghetti-:
Fixes https://github.com/ros/ros_comm/issues/647

### DIFF
--- a/tools/rosbag/src/record.cpp
+++ b/tools/rosbag/src/record.cpp
@@ -231,6 +231,7 @@ rosbag::RecorderOptions parseOptions(int argc, char** argv) {
     if (vm.count("topic"))
     {
       std::vector<std::string> bags = vm["topic"].as< std::vector<std::string> >();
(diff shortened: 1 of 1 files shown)


wjwwood:
Seems straight forward to me, and since it's in the cli it shouldn't break any code. However, it does change behavior, so we might want to consider it for release into k-turtle.

dirk-thomas:
The proposed code simply does not do what you describe. E.g. if the vector contain `a b a` the "filtered result is still `a b a` since the code currently only filters consecutive duplicates.

Please also remove the second commit from this PR. That patch was a separate PR which has been merged already.

spaghetti-:
@dirk-thomas good catch thanks, I forgot to sort it. Please call the ci again before you merge thanks!

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Thank you for the patch.

Squashed and merged in fac98b93e978974af34badd325eee306e8adfe58.

PR #649, opened 2015-08-17
made docstrings consistent for callback argument

CatherineH:
Why is callback currently defined as type str in the Subscriber class constructor? Would it be better defined as type function, and to have the same type where ever this callback is used?


wjwwood:
Looks like an oversight in the documentation. I'd say +1. Thanks for the pull request.

Issue #651, opened 2015-08-20
Docstring for rospy.on_shutdown() is wrong


opqdonut:
The docstring claims that the hook should take no arguments, but in the "already shutdown" case the hook gets called with one argument.

The call is here: https://github.com/ros/ros_comm/blob/ace5b9c11af766564441bad2cb08f5fca5b73784/clients/rospy/src/rospy/core.py#L316

spaghetti-:
I think we can just take out the argument. It is kind of difficult to work around this otherwise.

wjwwood:
@opqdonut I do not believe that is true. AFAIKT, those callbacks do take a string argument:
- https://github.com/ros/ros_comm/blob/ace5b9c11af766564441bad2cb08f5fca5b73784/clients/rospy/src/rospy/core.py#L360
- https://github.com/ros/ros_comm/blob/ace5b9c11af766564441bad2cb08f5fca5b73784/clients/rospy/src/rospy/core.py#L374

opqdonut:
the docstring for on_shutdown claims otherwise, as I said in my first comment :)

https://github.com/ros/ros_comm/blob/ace5b9c11af766564441bad2cb08f5fca5b73784/clients/rospy/src/rospy/client.py#L109

wjwwood:
Probably related to this:
- https://github.com/ros/ros_comm/commit/cdf44f3b103a951dbde25f8b6a4066a4afb40e82
- https://github.com/ros/ros_comm/issues/586

dirk-thomas:
The function `client.py:on_shutdown()` (https://github.com/ros/ros_comm/blob/ace5b9c11af766564441bad2cb08f5fca5b73784/clients/rospy/src/rospy/client.py#L109-L116) registers a **client** shutdown function (https://github.com/ros/ros_comm/blob/ace5b9c11af766564441bad2cb08f5fca5b73784/clients/rospy/src/rospy/core.py#L343-L352).

Those are supposed to be always called without arguments: https://github.com/ros/ros_comm/blob/ace5b9c11af766564441bad2cb08f5fca5b73784/clients/rospy/src/rospy/core.py#L396-L399

Therefore I would argue the current docstring is correct and does not need to be modified.

[...]

PR #652, opened 2015-08-24
removed argument to shutdown hook called after shutdown

spaghetti-:
Fixes https://github.com/ros/ros_comm/issues/651


dirk-thomas:
This changes behavior drastically. Therefore I propose #662 instead.

PR #653, opened 2015-08-25
Extend timer.Timer to include startup method

keithhamilton:
In working with the `rospy.timer.Timer` class, it feels to me like it could be improved with two subtle changes:
- A kwarg in its constructor that allows its invoker to either instantiate it in a running or shutdown state
- A companion method to `shutdown` that starts running the timer instead of turning it off.

This PR adds a kwarg to the `timer.Timer` constructor, `start_running` that defaults to False. In the `__init__` method, `self._shutdown` is set to `not start_running` so that if you wish to start the Timer in a running state, `start_running=True` and `self._shutdown = False`.

This PR also adds a `startup` method as a companion to `shutdown`, adding the ability to retain one instance of a `timer.Timer` object with the ability to toggle its running state. 

I have found this great for establishing a timer that controls a certain behavior with our robot, that may start and stop depending on repeating conditions.

Thanks for checking this out!
-- Keith

### DIFF
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -207,7 +207,7 @@ class Timer(threading.Thread):
         """
         self._shutdown = True
 
-    def start(self):
+    def startup(self):
         """
         Start run method by setting self._shutdown to False.
         """
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -194,7 +194,8 @@ class Timer(threading.Thread):
         @type  start: bool
         """
         threading.Thread.__init__(self)
-        self._period   = period
+        self.__period = period
+        self._period   = self.__period
         self._callback = callback
         self._oneshot  = oneshot
         self._shutdown = not start_running
@@ -206,6 +207,7 @@ class Timer(threading.Thread):
         Stop firing callbacks.
         """
         self._shutdown = True
+        self._period = self.__period
 
     def startup(self):
         """
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -190,8 +190,8 @@ class Timer(threading.Thread):
         @type  callback: function taking rospy.TimerEvent
         @param oneshot: if True, fire only once, otherwise fire continuously until shutdown is called [default: False]
         @type  oneshot: bool
-        @param start: if True, begin running the `run` method immediately, otherwise wait until self.start() is called [default: False]
-        @type  start: bool
+        @param start_running: if True, begin running the `run` method immediately, otherwise wait until self.start() is called [default: False]
+        @type  start_running: bool
         """
         threading.Thread.__init__(self)
         self.__period = period
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -48,7 +48,7 @@ class Rate(object):
     """
     Convenience class for sleeping in a loop at a specified rate
     """
-    
+
     def __init__(self, hz):
         """
         Constructor.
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/timer.py
```diff
@@ -48,7 +48,7 @@ class Rate(object):
     """
     Convenience class for sleeping in a loop at a specified rate
     """
-    
+
```
dirk-thomas: Please avoid any white space change.

clients/rospy/src/rospy/timer.py
```diff
@@ -190,12 +190,15 @@ def __init__(self, period, callback, oneshot=False):
         @type  callback: function taking rospy.TimerEvent
         @param oneshot: if True, fire only once, otherwise fire continuously until shutdown is called [default: False]
         @type  oneshot: bool
+        @param start: if True, begin running the `run` method immediately, otherwise wait until self.start() is called [default: False]
+        @type  start: bool
```
dirk-thomas: The name in the documentation does not match the new argument.

clients/rospy/src/rospy/timer.py
```diff
@@ -190,12 +190,15 @@ def __init__(self, period, callback, oneshot=False):
...
         """
         threading.Thread.__init__(self)
-        self._period   = period
+        self.__period = period
```
dirk-thomas, line 197: What is this needed for? `self._period` is never altered anywhere.

clients/rospy/src/rospy/timer.py
```diff
@@ -204,7 +207,14 @@ def shutdown(self):
         Stop firing callbacks.
         """
         self._shutdown = True
-        
+        self._period = self.__period
+
+    def startup(self):
```
dirk-thomas, line 212: It is not clear to me how this is supposed to be used by the user. If the timer has been shut down before and the thread has terminated already calling this function has no effect on its own. The user also has to call `start()` again.

If the thread is still running calling this method _undos_ a previous `shutdown` call which seems to be unintuitive.

keithhamilton, line 212: The point is that "shutting down" a threading.Thread object doesn't actually shut it down, right? It just flips a bit that causes the `run` method to ignore executing its main logic block. (line 222)

The thread is technically still running until its out of scope or is assigned a NoneType value. 

My use case was having a Timer object that I could start, shutdown, retain, and restart as a condition changed.

keithhamilton, line 212: Also, calling `start()` is not necessary. In this case you have a Timer object whose `_shutdown` property is True (`run` is not executing its main logic block), yet the thread still exists. Calling `startup()` flips the bit so that the `run` method begins paying attention to its main logic block again.

dirk-thomas, line 212: Have you tried to use the timer in this way? I don't think this is the behavior of a Python `Thread` instance. The `run()` is invoked **once** when the thread is being started. As soon as the run function returns the thread is done.

When the `_shutdown` flag is being set the while loop will be stopped and the run function returns and therefore the thread is finished.

keithhamilton, line 212: What's the protocol for suddenly realizing I had been misinterpreting something rather key to the functionality of a Thread object, and now feel like an idiot?

I was mistaking the run method for **not** exiting when `_shutdown` was flipped to `True`, hence the idea of making it possible to continue its existence after telling it to stop. I suppose it would be possible to have an implementation that did that, but as-is, the code in this PR does not make that functionally possible..

Sorry bout that! Feel free to close this with a vengeance.

dirk-thomas:
Since a different approach is necessary to achieve the goal of the feature request I will close this PR for now.

PR #655, opened 2015-08-26
add --local option to rostest, replaces #637

dirk-thomas:
This has been derived from PR #637 with the following changes:
- since the reused master might not be local the option has been renamed from `-local` to `--reuse-master`
- the `metavar` argument is not relevant for boolean flags in argparse and has been removed
- it has been clarified that `--clear` only works together with `--reuse-master`
- missing imports for `rospy.logwarn` added

### DIFF
--- a/tools/rostest/src/rostest/rostest_main.py
+++ b/tools/rostest/src/rostest/rostest_main.py
@@ -98,6 +98,8 @@ def rostestmain():
     parser.add_option("--results-base-dir", metavar="RESULTS_BASE_DIR",
                       help="The base directory of the test results. The test result file is " +
                            "created in a subfolder name PKG_DIR.")
+    parser.add_option("--reuse-master", action="store_true",
+                      help="Connect to an existing ROS master instead of spawning a new ROS master on a custom port")
     (options, args) = parser.parse_args()
     try:
         args = roslaunch.rlutil.resolve_launch_arguments(args)
@@ -147,7 +149,7 @@ def rostestmain():
         parser.error("test file is invalid. Generated failure case result file in %s"%results_file)
         
     try:
-        testCase = rostest.runner.createUnitTest(pkg, test_file)
+        testCase = rostest.runner.createUnitTest(pkg, test_file, options.reuse_master)
         suite = unittest.TestLoader().loadTestsFromTestCase(testCase)
 
         if options.text_mode:
--- a/tools/rostest/src/rostest/rostest_main.py
+++ b/tools/rostest/src/rostest/rostest_main.py
@@ -100,6 +100,8 @@ def rostestmain():
                            "created in a subfolder name PKG_DIR.")
     parser.add_option("--reuse-master", action="store_true",
                       help="Connect to an existing ROS master instead of spawning a new ROS master on a custom port")
+    parser.add_option("-c", "--clear", action="store_true",
+                      help="Clear all parameters when connecting to an existing ROS master (only works with --reuse-master)")
     (options, args) = parser.parse_args()
     try:
         args = roslaunch.rlutil.resolve_launch_arguments(args)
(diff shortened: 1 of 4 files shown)


dirk-thomas:
The anticipated 3 failing unit tests are not related to this change and have been failing before. I will deal with them separately. They should not block this PR to be merged.

@trainman419 Can you please check if this modified patch still works for your use case.

trainman419:
These changes look reasonable, but I won't have a chance to test this for a few days.

My team has been using the version from the original PR (#637) and has gotten used to the `--local` option. I'd prefer not to change things on them.

dirk-thomas:
Ok, if it passes tests I will then go ahead and merge this. If you later find a chance to test it please consider to comment to confirm that it works for you.

tools/rostest/src/rostest/rostest_main.py
```diff
@@ -98,6 +98,10 @@ def rostestmain():
     parser.add_option("--results-base-dir", metavar="RESULTS_BASE_DIR",
                       help="The base directory of the test results. The test result file is " +
                            "created in a subfolder name PKG_DIR.")
+    parser.add_option("--reuse-master", action="store_true",
```
chpatton013, line 101: There doesn't seem to be a short-form argument anymore.

tools/rostest/src/rostest/rostest_parent.py
```diff
@@ -45,17 +46,50 @@
...
 
 class ROSTestLaunchParent(roslaunch.parent.ROSLaunchParent):
 
-    def __init__(self, config, roslaunch_files, port=0):
+    def __init__(self, config, roslaunch_files, port=0, reuse_master=False, clear=False):
         if config is None:
             raise Exception("config not initialized")
         # we generate a run_id for each test
-        run_id = roslaunch.core.generate_run_id()
+        if not reuse_master:
```
chpatton013, line 58: Negating this conditional seems unnecessary. Wouldn't it read cleaner without an additional `not`?

tools/rostest/src/rostest/rostest_main.py
dirk-thomas, line 101: Most other options don't have a short option either. And `--reuse-master` is pretty uncommon. Since the original proposed `-l` does not make sense anymore I removed it. If you want to propose one I can easily update the PR and add a short option.

tools/rostest/src/rostest/rostest_parent.py
```diff
@@ -45,17 +46,50 @@
...
+            try:
+                run_id = param_server.getParam('/run_id')
+                have_master = True
+            except Exception as e:
+                logwarn("Could not connect to existing ROS master; will create a new master: " +
+                        str(e))
+            except:
+                # oh boy; we got something that wasn't an exception...
+                # ignore it, the same way that roslaunch does in this situation
+                logwarn("Could not connect to existing ROS master; will create a new master")
```
chpatton013, line 72: I don't know that this is even the right thing to do. When you explicitly ask to use an existing instance of ROS master, I would expect an error if one doesn't already exist. It just seems like unnecessary jimmy-proofing to me that covers up an error mode I might want to know about.

tools/rostest/src/rostest/rostest_parent.py
dirk-thomas, line 58: Usually I try to keep the standard case first. Special cases then follow later.

tools/rostest/src/rostest/rostest_parent.py
dirk-thomas, line 72: The user explicitly asked for reusing an existing master. If that is not possible it should not be silently done imo.

chpatton013:
There's really no easy way to tell what you changed on top of @trainman419's changes. To get an accurate diff I had to overwrite my local checkout of #637 with this code. Amending these commits instead of adding new commits subverts the entire principle of traceability.

dirk-thomas:
I will change the commits to be based on top of #637.

dirk-thomas:
I will close this since @trainman419 is going to update ##637 instead.

PR #656, opened 2015-08-26
disable failing unit tests

dirk-thomas:
These tests currently fail on the build farm only. Locally they pass and the same code passed on the build farm earlier. Since we could not figure out what has changed on the build farm regarding hostname lookups we have to disable these tests for now.


dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
With the problematic tests disabled it passes: http://jenkins.ros.org/job/_pull_request-indigo-ros_comm/ARCH_PARAM=amd64,UBUNTU_PARAM=trusty,label=devel/370/

PR #657, opened 2015-08-28
rosbag: better exception on empty bag

galou:
Bag.get_start_time and Bag.get_end_time thorw a better exception
if the bag contains no message.

Main modifications are on line approx. 496 and 512.
The rest is whitespace deletion.

### DIFF
--- a/test/test_rosbag/test/test_bag.py
+++ b/test/test_rosbag/test/test_bag.py
@@ -317,8 +317,12 @@ class TestRosbag(unittest.TestCase):
         fn = '/tmp/test_get_time_emtpy_bag.bag'
 
         with rosbag.Bag(fn, mode='w') as bag:
-            self.assertRaises(rosbag.ROSBagException, bag.get_start_time)
-            self.assertRaises(rosbag.ROSBagException, bag.get_end_time)
+            self.assertRaisesRegexp(rosbag.ROSBagException,
+                                    'Bag contains no message',
+                                    bag.get_start_time)
+            self.assertRaisesRegexp(rosbag.ROSBagException,
+                                    'Bag contains no message',
+                                    bag.get_end_time)
 
     def test_get_type_and_topic_info(self):
         fn = '/tmp/test_get_type_and_topic_info.bag'
--- a/test/test_rosbag/test/test_bag.py
+++ b/test/test_rosbag/test/test_bag.py
@@ -312,6 +312,14 @@ class TestRosbag(unittest.TestCase):
             self.assertEquals(start_stamp, 0.0)
             self.assertEquals(end_stamp, 99.0)
 
+    def test_get_time_empty_bag(self):
+        """Test for issue #657"""
+        fn = '/tmp/test_get_time_emtpy_bag.bag'
+
+        with rosbag.Bag(fn, mode='w') as bag:
+            self.assertRaises(rosbag.ROSBagException, bag.get_start_time)
+            self.assertRaises(rosbag.ROSBagException, bag.get_end_time)
+
     def test_get_type_and_topic_info(self):
         fn = '/tmp/test_get_type_and_topic_info.bag'
         topic_1 = "/test_bag"
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -492,6 +492,8 @@ class Bag(object):
         if self._chunks:
             start_stamp = self._chunks[0].start_time.to_sec()
         else:
+            if not self._connection_indexes:
+                raise ROSBagException('Bag contains no message')
             start_stamp = min([index[0].time.to_sec() for index in self._connection_indexes.values()])
         
         return start_stamp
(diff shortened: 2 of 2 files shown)


tfoote:
Please refactor this without the whitespace changes. 

Although our policy on new files is no trailing whitespace. For maintainability we don't want to have large sweeping changes for little patches like this. It makes any changes almost impossible to review and breaks future merging and our ability to backport. If you change a line clean up the whitespace, but if not please do not create larger diffs.

galou:
I thought this would be your reaction but I just hoped this would not be the case. After I tough fight with git I think I managed to remove whitespace changes.
Next time, I'll have to study how to disable automatic whitespace deletion with my editor (vim+python-mode).

galou:
Can you please reconsider merging?

wjwwood:
@galou can you post a snippet of the current behavior? In your OP you say that this is a "better" error message, which seems to indicate there is one already which is not so good.

It would also be good to have a test for this, but that may be more than you have time for. If you're interested in adding one, they're located here:

https://github.com/ros/ros_comm/tree/indigo-devel/test/test_rosbag/test

galou:
Sorry, I don't have any empty bag right now but the error was provoked by the giving an empty list to min: "`min() arg is an empty sequence`".
The error message now is "`'Bag contains no message`".
This can be tested with:

```
from rosbag import Bag
bag = Bag('/tmp/empty.bag', 'w')
bag.get_start_time()
```

I'll try to add the test.

galou:
Test added.

test/test_rosbag/test/test_bag.py
```diff
@@ -312,6 +312,14 @@ def test_get_time(self):
             self.assertEquals(start_stamp, 0.0)
             self.assertEquals(end_stamp, 99.0)
 
+    def test_get_time_empty_bag(self):
+        """Test for issue #657"""
+        fn = '/tmp/test_get_time_emtpy_bag.bag'
```
wjwwood, line 317: How does this file get put into `/tmp`?

Usually we download test data as part of the test procedure, for example:

https://github.com/ros/ros_comm/blob/indigo-devel/test/test_rosbag/CMakeLists.txt#L18

If you want I can place the test bag file on our servers.

test/test_rosbag/test/test_bag.py
```diff
@@ -312,6 +312,14 @@ def test_get_time(self):
...
+
+        with rosbag.Bag(fn, mode='w') as bag:
+            self.assertRaises(rosbag.ROSBagException, bag.get_start_time)
+            self.assertRaises(rosbag.ROSBagException, bag.get_end_time)
```
wjwwood: You can assert that it raises the correct `ROSBagException` as well by checking the error message with: https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp

wjwwood:
Thanks for taking the time to add the test. This error message is definitely an improvement.

I have a few comments about the test which I think can be pretty easily resolved and after that, I'd give it a +1.

Thanks again for iterating on this with us.

test/test_rosbag/test/test_bag.py
wjwwood, line 317: Never mind I see now that you're creating the bag in the `with ...` statement.

tfoote, line 317: It should use a unique path though. [tempfile](https://docs.python.org/2/library/tempfile.html) provides methods for creating temporary directories properly. 

You can either generate unique filenames in each test or setup a directory for all the tests.

wjwwood, line 317: That would be better, but all of the other tests in this file do the same as @galou is doing. So it's probably ok.

galou:
I used `assertRaisesRegexp` to test the content of the thrown message.

It's true that it was a long iterating process for 4 lines of codes!

I could moreover change bag file with tempfile but this would make again a lot of line changes. By the way, what is the preferred way to delete all these files on test end?

wjwwood:
@galou your latest commit adds back a lot of the whitespace changes. I'd do something like this to fix it:

```
$ git reset --soft HEAD~
$ git diff -w --no-color | git apply --cached --ignore-whitespace
$ git commit
$ git push -f <your origin> <your branch>
```

galou:
Dough, it looks that I forgot to deactivate automatic whitespace corrections before saving.

Thank you for the help with git. However, I had to add a line:

```
$ git reset --soft HEAD~
$ git reset HEAD test/test_bag.py
$ git diff -w --no-color | git apply --cached --ignore-whitespace
$ git commit
$ git push -f <your origin> <your branch>
```

wjwwood:
@ros-pull-request-builder

wjwwood:
@dirk-thomas can you kick off the pr builder?

tfoote:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
http://jenkins.ros.org/job/_pull_request-indigo-ros_comm/368/

dirk-thomas:
The three failing tests are unrelated to this change and are being worked on in #656.

wjwwood:
Ok, then I'm done with my review. +1

dirk-thomas:

Squashed and merged in b9c9192200a7f251c0cf359d792f97dbe27d7aeb.

PR #659, opened 2015-08-31
Create a symlink to the latest log directory.

yoneken:
This change makes a symbolic link named "latest" to the latest log directory.

Many times I would like to grep error messages in the latest log directory.
This change may make someone's life more easier.

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -45,6 +45,13 @@ from rospkg.environment import ROS_LOG_DIR
 
 class LoggingException: pass
 
+def renew_latest_logdir(logfile_dir):
+    log_dir = os.path.dirname(logfile_dir)
+    latest_dir = os.path.join(log_dir, 'latest')
+    if os.path.lexists(latest_dir):
+        os.remove(latest_dir)
+    os.symlink(logfile_dir, latest_dir)
+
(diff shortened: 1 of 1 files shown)


tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -45,6 +45,13 @@
 
 class LoggingException: pass
 
+def renew_latest_logdir(logfile_dir):
+    log_dir = os.path.dirname(logfile_dir)
+    latest_dir = os.path.join(log_dir, 'latest')
+    if os.path.exists(latest_dir):
```
dirk-thomas: I don't think this check will work if the symlink exists by points to a not existing folder.

yoneken: @dirk-thomas 
Thank you for the comment.
I fixed the issue to use os.path.lexists instead of os.path.exists.

dirk-thomas:
Thanks for the patch. I would like to improve it slightly before merging. Please see yoneken/ros_comm#1 for the suggested changed.

yoneken:
@dirk-thomas 
Thank you for the patch. I confirmed that the function becomes better!

dirk-thomas:
Thanks. I have squashed and merged the changes in 28f836135ce3e2ba3e05776cb9c7270411faa302.

PR #661, opened 2015-09-17
Prevent QueuedConnection memory/thread leak

mvollrath:
Added a timeout to `QueuedConnection`'s conditional wait() to prevent indefinite hanging threads.  These threads would eventually fill the process table when started by applications like rosbridge_server, which is expected to create and destroy many pub/sub connections.

There was a cleanup callback set on `TCPROSTransport` that was supposed to `notify()` the `Condition`, but it was being overwritten by the cleanup callback of `_PublisherImpl` and `_SubscriberImpl`.

Switching from the (broken) `notify()` callback to a timeout was less efficient than turning the transport cleanup callbacks into a list, so I did that too.

Fixes RobotWebTools/rosbridge_suite#165

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_pubsub.py
+++ b/clients/rospy/src/rospy/impl/tcpros_pubsub.py
@@ -394,7 +394,8 @@ class QueuedConnection(object):
         self._thread.start()
 
     def _closed_connection_callback(self, connection):
-        self._cond_data_available.notify()
+        with self._lock:
+            self._cond_data_available.notify()
 
     def __getattr__(self, name):
         if name.startswith('__'):
--- a/clients/rospy/src/rospy/topics.py
+++ b/clients/rospy/src/rospy/topics.py
@@ -426,7 +426,16 @@ class _TopicImpl(object):
             self.connections = new_connections
 
             # connections make a callback when closed
-            c.set_cleanup_callback(self.remove_connection)
+            # don't clobber an existing callback
+            if not c.cleanup_cb:
+                c.set_cleanup_callback(self.remove_connection)
+            else:
+                previous_callback = c.cleanup_cb
+                new_callback = self.remove_connection
+                def cleanup_cb_wrapper(s):
+                    new_callback(s)
+                    previous_callback(s)
+                c.set_cleanup_callback(cleanup_cb_wrapper)
             
             return True
 


dirk-thomas:
@ros-pull-request-builder

clients/rospy/src/rospy/impl/tcpros_pubsub.py
```diff
@@ -386,15 +386,17 @@ def __init__(self, connection, queue_size):
...
         self._lock = threading.Lock()
         self._cond_data_available = threading.Condition(self._lock)
-        self._connection.set_cleanup_callback(self._closed_connection_callback)
         self._queue = []
         self._error = None
 
         self._thread = threading.Thread(target=self._run)
         self._thread.start()
 
+        self._connection.add_cleanup_callback(self._closed_connection_callback)
```
dirk-thomas: Why did you move the callback registration below starting the thread?

clients/rospy/src/rospy/impl/tcpros_pubsub.py
```diff
@@ -386,15 +386,17 @@ def __init__(self, connection, queue_size):
...
+
     def _closed_connection_callback(self, connection):
-        self._cond_data_available.notify()
+        with self._lock:
```
dirk-thomas, line 397: Why did you add the lock here?

clients/rospy/src/rospy/impl/transport.py
```diff
@@ -93,15 +93,16 @@ def fileno(self):
     ## closed. Function will be passed this transport as an argument.
     ## @param self
     ## @param cleanup_callback fn(Transport): callback for when connection is closed
-    def set_cleanup_callback(self, cleanup_callback):
-        self.cleanup_cb = cleanup_callback
+    def add_cleanup_callback(self, cleanup_callback):
```
dirk-thomas: This is a breaking API change which can not be merged into already released distro. I am pretty sure this is going to break some of the tests.

Maybe the new function `add_cleanup_callback` should be added and the old function can call the new function to keep it kind of compatible?

clients/rospy/src/rospy/impl/transport.py
```diff
@@ -93,15 +93,16 @@ def fileno(self):
...
+        self.cleanup_callbacks.append(cleanup_callback)
 
     ## terminate i/o. Leaf subclasses should override and call this implementation
     ## @param self
     def close(self):
         self.done = True
-        if self.cleanup_cb:
-            self.cleanup_cb(self)
+        for cleanup_callback in self.cleanup_callbacks:
```
dirk-thomas: Would it make sense to call the callbacks in reverse order?

dirk-thomas:
An alternative could be that the QueuedConnection wraps the existing callback to ensure it is still being invoked. That would avoid any API changes but only works if the callbacks are registered _before_ the QueuedConnection wraps them.

clients/rospy/src/rospy/impl/tcpros_pubsub.py
mvollrath, line 397: It silently failed without it.  The Lock must be acquired before you can access the Condition.  See https://docs.python.org/2/library/threading.html#condition-objects

"If the calling thread has not acquired the lock when this method is called, a RuntimeError is raised."

I made a little script to test and illustrate this:

```
#!/usr/bin/env python

import threading

class ConditionTester(threading.Thread):
    def __init__(self):
        super(self.__class__, self).__init__()
        self.lock = threading.Lock()
        self.condition = threading.Condition(self.lock)

    def test_locked_notify(self):
        with self.lock:
            self.condition.notify()

    def test_unlocked_notify(self):
        self.condition.notify()

    def run(self):
        with self.lock:
            self.condition.wait()
            print 'got the condition'

if __name__ == '__main__':
    tester = ConditionTester()
    tester.start()
    tester.test_locked_notify()  # works
    #tester.test_unlocked_notify()  # RuntimeError and hang
    tester.join()
```

dirk-thomas, line 397: This would imply that this code path was either never triggered or didn't work before?

clients/rospy/src/rospy/impl/tcpros_pubsub.py
mvollrath: That's just where it ended up when I added it back, but it doesn't matter.  Here's how I understand it.

There is potentially a race here either way, but slightly less with something following up the thread start.  I'd love to find a cleaner way to make sure that `QueuedConnection._run()` acquires the `Lock` before the `_TopicImpl.close()` does, because if it doesn't then it will never be notified.  It is _probably_ never going to be an issue because of what follows this constructor.

Starting from `_PublisherImpl.add_connection()`, the `QueuedConnection` wrapper is created before the critical section of `_TopicImpl.add_connection()` which uses the same `Lock` as `_TopicImpl.close()` (which is what closes the  the cleanup callback).  So, there's no way the callback can happen before this constructor is off the stack and the critical section of `_TopicImpl.add_connection()` will _probably_ not yield for `close()` before `QueuedConnection._run()` can make it to the `Lock`.

clients/rospy/src/rospy/impl/transport.py
mvollrath: Makes sense to me.  I will reverse the list here.

clients/rospy/src/rospy/impl/transport.py
mvollrath: I'll make an alias, should not affect anybody's implementations unless they were banking on callback refs being overwritten.

Should I consider the cleanup_cb attribute to be public, or fix the tests to check the new list instead?

dirk-thomas: Since the attribute is not explicitly private we have to assume someone is using it directly.

mvollrath:
I feel like, between wrapping existing callbacks and waiting with a timeout, the timeout is more maintainable by far.  If the API change of single callback to a list is too disruptive, simply reverting https://github.com/ros/ros_comm/commit/d65247166d340eee9572fc5d9c9713fe47c522d8 will fix this leak without any hackery.

clients/rospy/src/rospy/impl/tcpros_pubsub.py
mvollrath, line 397: Was never triggered because topic impl was overwriting the single cleanup callback.

mvollrath:
See #663 for a more compatible and maintainable solution.

mvollrath:
I think the interface can be preserved completely but the behavior of `set_cleanup_callback()` will be slightly different.  At a glance the tests will still pass because it will only affect callbacks added after the first.

We can implicitly wrap an existing callback in `set_cleanup_callback()` completely transparent to everybody.  The wrapping closure will call an existing callback after the newly submitted one, keeping the FILO order.  This will increase the stack depth with each closure, but the rospy implementation shouldn't suffer.  Implicit is not great, but it's better than the memory leak.  Let's see if it works..

mvollrath:
Do you think a specific callback wrapper for `QueuedConnection` will be better than this implicit wrapping in `set_cleanup_callback()`?

mvollrath:
To be exact, it's `_TopicImpl.add_connection()` that would need to wrap an existing callback.  I can see the benefit.

mvollrath:
This change set fixes the rosbridge leak without any stacking performance regression or dirty implicit behavior under the hood.  Pretty cool.  Looks like the PR test broke when I stacked pushes, though.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
I just tried the current patch but I get the same increasing number of child processes with and without this patch. I also added a debug message when the wrapping occurs and that does appear on the console. Do you see a different result?

mvollrath:
I tested this against rosbridge with the script described in https://github.com/RobotWebTools/rosbridge_suite/issues/165#issuecomment-140908182 which makes two connections to the web socket and does some pubsub.

I checked two things:
- `ps -eLf | grep rosbridge_websocket | wc -l` (no increase in proc tree)
- adding `objgraph.show_growth()` to the `on_close()` handler ( https://github.com/RobotWebTools/rosbridge_suite/blob/0.7.13/rosbridge_server/src/rosbridge_server/websocket_handler.py#L93-L97 )

On indigo-devel, `objgraph.show_growth()` consistently shows growth in `TCPROSTransport`, `QueuedConnection`, `_PublisherImpl`, and other classes which are instance members of those.  `ps` also shows an additional pid for each cycle.

On fix_queued_connection_leak it stabilizes and there is no consistent growth in `objgraph` or `ps`.

I'll see about making a more isolated test.  What are you running?

dirk-thomas:
Nvm, I only modified the `topics.py` file and missed the lock of the condition. With that change applied too I do get a constant count in `ps` with this patch.

clients/rospy/src/rospy/topics.py
```diff
@@ -426,7 +426,16 @@ def add_connection(self, c):
             self.connections = new_connections
 
             # connections make a callback when closed
-            c.set_cleanup_callback(self.remove_connection)
+            # XXX: don't clobber an existing callback
```
dirk-thomas: Can you please remove the `XXX:` form this comment line since it usually indicates a todo.

dirk-thomas:
I made one minor comment about the comment line. After addressing it can you please squash the commits?

PR #663, opened 2015-09-17
Fix QueuedConnection memory leak with timeout

mvollrath:
This reverts commit d65247166d340eee9572fc5d9c9713fe47c522d8.

This fixes a memory/thread leak in QueuedConnection.  Its cleanup callback is never run because it is overwritten by _TopicImpl subclasses.


mvollrath:
This is an alternative to #661

mvollrath:
This is also a performance regression for nodes with many pubsubs.

PR #666, opened 2015-09-18
Test for publisher impl deregister leak

mvollrath:
This is a test for #661

PR #667, opened 2015-09-18
Test publisher deregister leak with patch

dirk-thomas:
#666 but with the patch applied to check that unit test passes now. @mvollrath FYI

This should not be merged!


dirk-thomas:
I confirmed in the console output that the test passed now. So both PRs are good to go.

PR #668, opened 2015-09-18
fix location of rosunit result files generated by rostests

dirk-thomas:
Based on the discussion from https://github.com/ros/ros_comm/pull/666#issuecomment-141535738

### DIFF
--- a/tools/rostest/src/rostest/runner.py
+++ b/tools/rostest/src/rostest/runner.py
@@ -39,6 +39,7 @@ import time
 import unittest
 
 import rospkg
+from rospkg.environment import ROS_TEST_RESULTS_DIR
 import roslaunch
 import roslib.packages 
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
@ros/ros_team Please take a look to review this patch.

wjwwood:
lgtm

tools/rostest/src/rostest/runner.py
```diff
@@ -121,7 +122,10 @@ def fn(self):
 
             #setup the test
             # - we pass in the output test_file name so we can scrape it
-            test_file = xmlResultsFile(test_pkg, test_name, False)
+            env = None
+            if results_base_dir:
+                env = {ROS_TEST_RESULTS_DIR: results_base_dir}
```
tfoote, line 127: Are there potentially other env values that this is going to override or suppress? It might be better to append to the env instead of overriding it.

dirk-thomas, line 127: Other code uses the same pattern: https://github.com/ros/ros_comm/blob/44a089aef8f90243084045db6ad8fe6b95f53052/tools/rostest/src/rostest/rostest_main.py#L148-L154

And at the end of the call stack only this value is checked: https://github.com/ros-infrastructure/rospkg/blob/d3c12bdfbf990b9910220642bafef16b4c830b63/src/rospkg/environment.py#L162-L166

esteve:
+1

tfoote:
+1

PR #669, opened 2015-09-21
Ft rospy validate name

### LINKED ISSUE #664: Inconsistent node name validation between roscpp and rospy
This works:

```
$ rosrun any_pkg any_rospy_node __name:=foo-bar
```

This does not:

```
$ rosrun any_pkg any_roscpp_node __name:=foo-bar
terminate called after throwing an instance of 'ros::InvalidNameException'
  what():  Character [-] at element [3] is not valid in Graph Resource Name [foo-bar].  Valid characters are a-z, A-Z, 0-9, / and _.
Aborted (core dumped)
```

minshallj:
This will validate names passed from the command line which were previously going un-validated.

Fix #664

### DIFF
--- a/clients/rospy/src/rospy/client.py
+++ b/clients/rospy/src/rospy/client.py
@@ -261,10 +261,7 @@ def init_node(name, argv=None, anonymous=False, log_level=None, disable_rostime=
     # this test can be eliminated once we change from warning to error in the next check
     if rosgraph.names.SEP in name:
         raise ValueError("namespaces are not allowed in node names")
-    if not rosgraph.names.is_legal_base_name(name):
-        import warnings
-        warnings.warn("'%s' is not a legal ROS base name. This may cause problems with other ROS tools"%name, stacklevel=2)
-    
+
     global _init_node_args
 
     # #972: allow duplicate init_node args if calls are identical
@@ -289,8 +286,7 @@ def init_node(name, argv=None, anonymous=False, log_level=None, disable_rostime=
     # check for name override
     mappings = rospy.names.get_mappings()
     if '__name' in mappings:
-        # use rosgraph version of resolve_name to avoid remapping
-        name = rosgraph.names.resolve_name(mappings['__name'], rospy.core.get_caller_id())
+        name = mappings['__name']
         if anonymous:
             logdebug("[%s] WARNING: due to __name setting, anonymous setting is being changed to false"%name)
             anonymous = False
@@ -300,7 +296,13 @@ def init_node(name, argv=None, anonymous=False, log_level=None, disable_rostime=
         # hostname as that is not guaranteed to be a legal ROS name
         name = "%s_%s_%s"%(name, os.getpid(), int(time.time()*1000))
 
-    resolved_node_name = rospy.names.resolve_name(name)
+    # check for legal base name once all changes have been made to the name
+    if not rosgraph.names.is_legal_base_name(name):
+        import warnings
+        warnings.warn("'%s' is not a legal ROS base name. This may cause problems with other ROS tools."%name, stacklevel=2)
+
+    # use rosgraph version of resolve_name to avoid remapping
+    resolved_node_name = rosgraph.names.resolve_name(name, rospy.core.get_caller_id())
     rospy.core.configure_logging(resolved_node_name)
     # #1810
     rospy.names.initialize_mappings(resolved_node_name)
--- a/tools/rosgraph/src/rosgraph/names.py
+++ b/tools/rosgraph/src/rosgraph/names.py
@@ -212,7 +212,7 @@ def load_mappings(argv):
 import re
 
 #~,/, or ascii char followed by (alphanumeric, _, /)
-NAME_LEGAL_CHARS_P = re.compile('^[\~\/A-Za-z][\w_\/]*$') 
+NAME_LEGAL_CHARS_P = re.compile('^[\~\/A-Za-z][\w\/]*$')
 def is_legal_name(name):
     """
     Check if name is a legal ROS name for graph resources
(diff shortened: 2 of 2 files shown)


dirk-thomas:
@ros-pull-request-builder

minshallj:
rewording the first commit message, but I'll wait until after the pull request builder has finished before pushing.

```
rospy: normalizing name validation and resolution

    Validation is moved to after the name has been possibly loaded by the
    command line. Supplying a node name with a '-' from the command line
    will now warn.
```

minshallj:
After this is merged, I would like to move this line:
https://github.com/ros/ros_comm/pull/669/files#diff-2727faacb27a9b39ef999b2efa6ce184R278
down below the `is_legal_base_name` check. This is because we should not set the `_init_node_params` in stone, if they do not have valid data.

Currently there is this behavior for a name that is invalid because of a separator in the name:

``` python
In [2]: rospy.init_node('foo/bar')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-65a1c17e9ba1> in <module>()
----> 1 rospy.init_node('foo/bar')

/home/jacob/dev/catkin_rospy/src/ros_comm/clients/rospy/src/rospy/client.pyc in init_node(name, argv, anonymous, log_level, disable_rostime, disable_rosout, disable_signals, xmlrpc_port, tcpros_port)
    261     # this test can be eliminated once we change from warning to error in the next check
    262     if rosgraph.names.SEP in name:
--> 263         raise ValueError("namespaces are not allowed in node names")
    264 
    265     global _init_node_args

ValueError: namespaces are not allowed in node names

In [3]: rospy.init_node('foo_bar') 

In [4]: 
```

This works because `_init_node_args` is set only after the first validation has passed. Once the second validation starts throwing errors, this behavior will no longer be possible. `_init_node_args` will be stuck with invalid data, and will throw an error if the arguments have been changed, and will throw an error if they have not. This is what it will look like once errors are thrown by the second validation:

``` python
In [2]: rospy.init_node('foo-bar')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-7473ffcfe157> in <module>()
----> 1 rospy.init_node('foo-bar')

/home/jacob/dev/catkin_rospy/src/ros_comm/clients/rospy/src/rospy/client.py in init_node(name, argv, anonymous, log_level, disable_rostime, disable_rosout, disable_signals, xmlrpc_port, tcpros_port)
    302         import warnings
    303         #warnings.warn("'%s' is not a legal ROS base name. This may cause problems with other ROS tools. This will be an error in future ROS releases."%name, stacklevel=2)
--> 304         raise ValueError("'%s' is not a legal ROS base name. This may cause problems with other ROS tools. This will be an error in future ROS releases."%name)
    305 
    306     resolved_node_name = rosgraph.names.resolve_name(name, rospy.core.get_caller_id())

ValueError: 'foo-bar' is not a legal ROS base name. This may cause problems with other ROS tools. This will be an error in future ROS releases.

In [3]: rospy.init_node('foo_bar')
---------------------------------------------------------------------------
ROSException                              Traceback (most recent call last)
<ipython-input-3-6908650eebfb> in <module>()
----> 1 rospy.init_node('foo_bar')

/home/jacob/dev/catkin_rospy/src/ros_comm/clients/rospy/src/rospy/client.py in init_node(name, argv, anonymous, log_level, disable_rostime, disable_rosout, disable_signals, xmlrpc_port, tcpros_port)
    269     if _init_node_args:
    270         if _init_node_args != (name, argv, anonymous, log_level, disable_rostime, disable_signals):
--> 271             raise rospy.exceptions.ROSException("rospy.init_node() has already been called with different arguments: "+str(_init_node_args))
    272         else:
    273             return #already initialized

ROSException: rospy.init_node() has already been called with different arguments: ('foo-bar', ['/home/jacob/.local/bin/ipython'], False, None, False, False)

In [4]: 
```

One note on moving the assignment of _init_node_args, the original `name` and original `disable_signals` value will need to be stored.

clients/rospy/src/rospy/client.py
```diff
@@ -289,8 +286,7 @@ def init_node(name, argv=None, anonymous=False, log_level=None, disable_rostime=
     # check for name override
     mappings = rospy.names.get_mappings()
     if '__name' in mappings:
-        # use rosgraph version of resolve_name to avoid remapping
-        name = rosgraph.names.resolve_name(mappings['__name'], rospy.core.get_caller_id())
```
dirk-thomas, line 293: Just for the review: the second argument (`caller_id=rospy.core.get_caller_id()`) is the default argument and therefore not necessary.

minshallj, line 293: So that is true for `rospy.names.resolve_name` https://github.com/ros/ros_comm/blob/indigo-devel/clients/rospy/src/rospy/names.py#L163-L178

But not for `rosgraph.names.resolve_name` https://github.com/ros/ros_comm/blob/indigo-devel/tools/rosgraph/src/rosgraph/names.py#L260-L275

clients/rospy/src/rospy/client.py
```diff
@@ -300,7 +296,13 @@ def init_node(name, argv=None, anonymous=False, log_level=None, disable_rostime=
         # hostname as that is not guaranteed to be a legal ROS name
         name = "%s_%s_%s"%(name, os.getpid(), int(time.time()*1000))
 
-    resolved_node_name = rospy.names.resolve_name(name)
+    # check for legal base name once all changes have been made to the name
+    if not rosgraph.names.is_legal_base_name(name):
+        import warnings
+        warnings.warn("'%s' is not a legal ROS base name. This may cause problems with other ROS tools. This will be an error in future ROS releases."%name, stacklevel=2)
```
dirk-thomas: After thinking about this a bit more I am not sure if I would escalate it to be an error in the future. If the user wants to ignore the recommendation and warning we don't have to block him from doing it.

Therefore I would propose to remove `This will be an error in future ROS releases.` for now.

clients/rospy/src/rospy/client.py
dirk-thomas, line 293: Very good point. I followed the wrong namespace.

clients/rospy/src/rospy/client.py
minshallj: Maybe it should be replaced with `This will be an error in a roscpp program.` or something similar?

dirk-thomas: I would rather not mention `roscpp` behavior in a `rospy` program. What about other languages then `roslisp`, `rosjava`?

minshallj: That makes sense. Done.

dirk-thomas:
The patch looks good to me and seems to achieve the goal when using a remapping for `__name` to warn about non-legal values.

@wjwwood Can you please take a look too in order to double check.

wjwwood:
+1

dirk-thomas:
Thanks.

Issue #602, opened 2015-04-09
rostopic could warn users once that it is using sim time 


guihomework:
Just a suggestion:

It would be nice to have a one time WARN/INFO message when calling rostopic hz telling if it is using simulated time. On rostopic echo this is quite obvious as the timestamps are not walltime (small values for seconds) but on rate aspects there is no such indication, one can easily be confused and the rate then seems perfect although the real rate is slower.

Here under is the context in which we found it could have been useful :
We have usually have simulation running at a realtime factor of 1.0, so we kind of forgot about simulation time and walltime differences and everything works fine when using a ROS only environment.

However, we also use another transport framework (RSB) in our environment that is not aware of simulation time. A few days ago, a new piece of code was slow and we (among others) used _rostopic hz_ and the equivalent function of RSB to debug which part was slow (not publishing correctly)

Unfortunately it took a while to understand the new process was using more CPU than expected, making the realtime factor 0.3 on the simulation side (which we did not see as there was no movement involved) or overall slowdown, and hence we got the walltime and simtime to be streched by the same factor regarding rates. Both topic monitoring tools indicated different rates of publication and we did not realize this could come from the sim time, we thought our second transport framework had issues.

we found that after adding our own rate calculation in our loop, first with ros::time (which again failed to display anything wrong due to awareness of simtime) and then with non-ros timing functions that finally uncovered the problem.

An indication in rostopic hz at the very beginning might help to recall the rate we see is according to simulated time.

spaghetti-:
I'll take a look into it but if I'm not wrong we've to check if the node publishing the topic has `/use_sim_time` set. @wjwwood is there another approach that could work?

wjwwood:
@spaghetti- Yes, that is the most reliable way that I know of to see if simulation time is being used. I agree that a one time warning would be useful.

You could conceivably have `rostopic hz` show a simulation rate and a real time rate, but there might be issues implementing that given the possibility of time travel, e.g. when a bag file loops.

dirk-thomas:
This has been implemented in #672.

PR #672, opened 2015-09-26
Warn user about simulated time when using rostopic hz

spaghetti-:
https://github.com/ros/ros_comm/issues/602

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -187,6 +187,11 @@ def _rostopic_hz(topic, window_size=-1, filter_expr=None):
     :param window_size: number of messages to average over, -1 for infinite, ``int``
     :param filter_expr: Python filter expression that is called with m, the message instance
     """
+
+    if rospy.get_param('use_sim_time', False):
+        print("WARNING: topic [%s] may be using simulated time" % (topic),
+              file=sys.stderr)
+
     msg_class, real_topic, _ = get_topic_class(topic, blocking=True) #pause hz until topic is published
     if rospy.is_shutdown():
         return


tools/rostopic/src/rostopic/__init__.py
```diff
@@ -187,6 +187,10 @@ def _rostopic_hz(topic, window_size=-1, filter_expr=None):
     :param window_size: number of messages to average over, -1 for infinite, ``int``
     :param filter_expr: Python filter expression that is called with m, the message instance
     """
+
+    if rospy.get_param('use_sim_time', False):
+        print("WARNING: topic [%s] may be using simulated time\n" % (topic))
```
dirk-thomas: This will be printed periodically. Is that really intended here?

spaghetti-: I tested it on a topic and it only printed once. Am I missing anything here?

``````
$python -c 'import rostopic; rostopic.rostopicmain()' hz /navigation/RPY                
WARNING: topic [/navigation/RPY] may be using simulated time

subscribed to [/navigation/RPY]
average rate: 91.676
    min: 0.000s max: 0.020s std dev: 0.00300s window: 28
average rate: 90.521
    min: 0.000s max: 0.023s std dev: 0.00270s window: 119
average rate: 89.307
    min: 0.000s max: 0.023s std dev: 0.00375s window: 207
average rate: 89.285
    min: 0.000s max: 0.024s std dev: 0.00366s window: 296
average rate: 89.144
    min: 0.000s max: 0.024s std dev: 0.00355s window: 385
average rate: 89.090
    min: 0.000s max: 0.024s std dev: 0.00365s window: 474```
``````

Comment (dirk-thomas) [inline tools/rostopic/src/rostopic/__init__.py]: You are right, it only gets printed once since the function loop internally.

Can you print the warning to `stderr` please.

Comment (spaghetti-) [inline tools/rostopic/src/rostopic/__init__.py]: Done.

File: tools/rostopic/src/rostopic/__init__.py [hunk 2]
```diff
@@ -187,6 +187,10 @@ def _rostopic_hz(topic, window_size=-1, filter_expr=None):
     :param window_size: number of messages to average over, -1 for infinite, ``int``
     :param filter_expr: Python filter expression that is called with m, the message instance
     """
+
+    if rospy.get_param('use_sim_time', False):
+        sys.stderr.write("WARNING: topic [%s] may be using simulated time\n" % (topic))
```
dirk-thomas: Can you please use the `print()` function for this (similar as in https://github.com/spaghetti-/ros_comm/blob/46857a2483a38648635740a16b63cc869d9e8177/tools/rostopic/src/rostopic/__init__.py#L305).

spaghetti-:
updated with print function instead

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -187,6 +187,11 @@ def _rostopic_hz(topic, window_size=-1, filter_expr=None):
     :param window_size: number of messages to average over, -1 for infinite, ``int``
     :param filter_expr: Python filter expression that is called with m, the message instance
     """
+
+    if rospy.get_param('use_sim_time', False):
+        print("WARNING: topic [%s] may be using simulated time\n" % (topic),
```
dirk-thomas: Please remove the extra newline from the string.

spaghetti-: Done

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -187,6 +187,11 @@ def _rostopic_hz(topic, window_size=-1, filter_expr=None):
     :param window_size: number of messages to average over, -1 for infinite, ``int``
     :param filter_expr: Python filter expression that is called with m, the message instance
     """
+
```
dirk-thomas, line 190: Please remove the empty after the docblock (D202 of PEP257).

dirk-thomas:
I have cherry-picked the change with my comment addressed in 5197c6ecb5174c0cd35187bd347d0178dbf6b1b6.

dirk-thomas:
Moved warning (and removed the reference to the topic since it's not relevant) to pass unit tests: bd838b69d5a042fc382943cd7c3e1abfbce4f633

Issue #564, opened 2015-02-06
TimeSequencer for PYTHON


nrkumar93:
I could not find TimeSequencer API for PYTHON under ROS message_filters. Could someone kindly work on it and build that?

dirk-thomas:
Please consider to provide a pull request for this feature.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #673, opened 2015-09-26
Python implementation of the message filter TimeSequencer

spaghetti-:
https://github.com/ros/ros_comm/issues/564

Tested with https://gist.github.com/spaghetti-/6e11c0789363a66456bc however I do agree that this isnt a very comprehensive test. 

I would love verification/testing from anyone.

### DIFF
--- a/utilities/message_filters/src/message_filters/__init__.py
+++ b/utilities/message_filters/src/message_filters/__init__.py
@@ -226,3 +226,46 @@ class ApproximateTimeSynchronizer(TimeSynchronizer):
                 for q,t in qt:
                     del q[t]
         self.lock.release()
+
+class TimeSequencer(SimpleFilter):
+    """
+    Guarantees that messages will be called in temporal order according to their
+    header's timestamp.
+
+    :class:`TimeSequencer` is constructed with a specific delay which specifies
+    how long to queue up messages before passing them through. A callback for a
+    message is never invoked until the messages' time stamp is out of date by at
+    least delay. However, for all messages which are out of date by at least the
+    delay, their callback are invoked and guaranteed to be in temporal order.
+    If a message arrives from a time prior to a message which has already had
+    its callback invoked, it is thrown away. 
+    """
+
+    def __init__(self, fs, delay, rate, queue_size):
+        SimpleFilter.__init__(self)
+        self.lock       = threading.Lock()
+        self.queue      = {}
+        self.queue_size = queue_size
+        self.last_stamp = rospy.Time(0)
+        self.rate       = rospy.Duration.from_sec(rate)
+        self.delay      = rospy.Duration.from_sec(delay)
+        self.timer      = rospy.Timer(self.rate, self.check)
+        fs.registerCallback(self.add)
+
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please make the new code follow PEP 8 / 257 and also include documentation for the constructor arguments.

utilities/message_filters/src/message_filters/__init__.py
```diff
@@ -226,3 +226,46 @@ def add(self, msg, my_queue):
...
+
+    def __init__(self, fs, delay, rate, queue_size):
+        SimpleFilter.__init__(self)
+        self.lock       = threading.Lock()
+        self.queue      = {}
+        self.queue_size = queue_size
+        self.last_stamp = rospy.Time(0)
+        self.rate       = rospy.Duration.from_sec(rate)
+        self.delay      = rospy.Duration.from_sec(delay)
+        self.timer      = rospy.Timer(self.rate, self.check)
```
dirk-thomas, line 252: Please don't align the equal signs.

utilities/message_filters/src/message_filters/__init__.py
```diff
@@ -226,3 +226,46 @@ def add(self, msg, my_queue):
...
+        fs.registerCallback(self.add)
+
+    def add(self, msg):
+        self.lock.acquire()
+        if msg.header.stamp > self.last_stamp:
+            self.queue[msg.header.stamp] = msg
```
dirk-thomas, line 258: This seems to overwrite other messages with the same timestamp?

utilities/message_filters/src/message_filters/__init__.py
```diff
@@ -226,3 +226,46 @@ def add(self, msg, my_queue):
...
+        while len(self.queue) > self.queue_size:
+            t = min(self.queue)
+            del self.queue[t]
+        self.lock.release()
```
dirk-thomas, line 262: Please use `with` instead to ensure unlock even when exceptions are raised.

Same below.

spaghetti-:
Thanks for your comments, will update the issues and test against the C++ TimeSequencer as well.

However when I ran pep8 against the file there are many other issues that are raised, should I address them in a different PR and just fix for my changes here?

dirk-thomas:
Just make sure that the newly added code block is "clean". You should not update existing code since it makes backporting patches way more difficult. Thanks.

utilities/message_filters/src/message_filters/__init__.py
```diff
@@ -226,3 +226,46 @@ def add(self, msg, my_queue):
...
+
+    def check(self, _):
+        self.lock.acquire()
+        to_call = sorted(filter(lambda t: t + self.delay <= rospy.Time.now(),
+                              self.queue.keys()))
```
mikepurvis, line 267: I suppose it depends on the length of the queue and frequency of this function being called, but sorting the dictionary keys every time seems not ideal.

I wonder what the performance difference would be using a deque which stays sorted by inserting new messages into the correct position.

dirk-thomas:

Issue #52, opened 2013-01-08
rostopic: hz should have --wall-time option


ablasdel:
This can allow clocking of topics when simulated time isn't being published.

migration of trac ticket 3429: 
  https://code.ros.org/trac/ros/ticket/3429

evenator:
Can this now be closed?

dirk-thomas:
Thanks for pointing it out.

PR #674, opened 2015-09-27
wall-time option for rostopic hz

spaghetti-:
```
This can allow clocking of topics when simulated time isn't being published.

migration of trac ticket 3429:
https://code.ros.org/trac/ros/ticket/3429
```

https://github.com/ros/ros_comm/issues/52

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -97,7 +97,7 @@ class ROSTopicHz(object):
     """
     ROSTopicHz receives messages for a topic and computes frequency stats
     """
-    def __init__(self, window_size, filter_expr=None):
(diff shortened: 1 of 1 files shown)


tools/rostopic/src/rostopic/__init__.py
```diff
@@ -120,7 +121,8 @@ def callback_hz(self, m):
         if self.filter_expr is not None and not self.filter_expr(m):
             return
         with self.lock:
-            curr_rostime = rospy.get_rostime()
+            curr_rostime = rospy.get_rostime() if not self.use_wtime else \
+                    rospy.Time.from_sec(time.time())
```
dirk-thomas, line 125: Please indent the wrapped line with only +4 (not with +8 according to PEP8).

Same a few lines below.

dirk-thomas:

PR #542, opened 2014-12-31
Enable caching of catkin_pkg.packages.find_packages when using roslaunch

garaemon:
patch for https://github.com/ros/ros_comm/issues/541

PR #676, opened 2015-09-29
[roslaunch] cache result of path finding for faster roslaunch behavior

### LINKED ISSUE #541: [roslaunch] cache result of catkin_pkg.packages.find_packages to speed up roslaunch
roslaunch_caller

originally discussed in [here](https://github.com/ros-infrastructure/catkin_pkg/pull/111)
## Problem
- roslaunch is slow if catkin workspace has a lot of files
- `catkin_pkg.packages.find_packages` always scan workspace and it's not fast enough
- When rosbuild was used, roslaunch was not so slow, as far as I remember
## Approach
- During running roslaunch, catkin workspace must not be changed, so we can cache `catkin_pkg.packages.find_packages` like old rospack
- In order to keep backward compatibility, add keyword arguments to functions
## Benchmark
- Use [test.rosinstall](https://gist.github.com/garaemon/7a39457795e7c61d7d8c) to build workspace
  
  ```
  $ find src | wc -l
  17239
  $ find src -name package.xml | wc -l
  286
  $ find src -name manifest.xml | wc -l
  146
  ```
- Measure time of `roslaunch openni2_launch openni2.launch`
- Result
  - without caching -- 17 sec
  - with caching -- 5 sec
  - Movie
    - [without caching](http://youtu.be/_Utax0R3u0g)
    - [with caching](http://youtu.be/bAFNlsNcihU)

wkentaro:
This is for faster roslaunch's node launching.

Closes #541
Closes #542 

This is firstly reported by @garaemon .

### DIFF
--- a/tools/roslaunch/src/roslaunch/node_args.py
+++ b/tools/roslaunch/src/roslaunch/node_args.py
@@ -52,6 +52,7 @@ from rosgraph.names import script_resolve_name
 
 import roslib.packages
 
+from . import resource_cache
 from . import substitution_args
 
 from roslaunch.core import setup_env, local_machine, RLException
@@ -239,7 +240,7 @@ def create_local_process_args(node, machine, env=None):
     try:
         #TODO:fuerte: pass through rospack and catkin cache
         rospack = rospkg.RosPack(rospkg.get_ros_paths(env=env))
-        matches = roslib.packages.find_node(node.package, node.type, rospack)
+        matches = resource_cache.find_node(node.package, node.type, rospack)
     except rospkg.ResourceNotFound as e:
         # multiple nodes, invalid package
         raise NodeParamsException(str(e))
--- a/tools/roslaunch/src/roslaunch/resource_cache.py
+++ b/tools/roslaunch/src/roslaunch/resource_cache.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import roslib.packages
+import catkin.find_in_workspaces
+
+from catkin.workspace import get_source_paths, get_workspaces
+from catkin_pkg.packages import find_packages
+
+
+def _get_source_path_to_packages(workspaces=None,
+                                 workspace_to_source_spaces=None):
+    """Return dictionay with mappings from source paths to packages.
+
+    :param workspaces: (optional, used for unit tests),
+        the list of workspaces to use.
+    :param workspace_to_source_spaces: the dictionary is populated
+        with mappings from workspaces to source paths,
+        pass in the same dictionary to avoid repeated reading
+        of the catkin marker file.
+    """
+    if workspaces is None:
+        workspaces = get_workspaces()
+    if workspace_to_source_spaces is None:
+        workspace_to_source_spaces = {}
+
+    source_path_to_packages = {}
+    for ws in workspaces:
+        if ws not in workspace_to_source_spaces:
+            workspace_to_source_spaces[ws] = get_source_paths(ws)
+        for src_path in workspace_to_source_spaces[ws]:
+            if src_path not in source_path_to_packages:
+                source_path_to_packages[src_path] = find_packages(src_path)
+    return source_path_to_packages
+
+
+# dictionary with mappings from source paths to packages
+_source_path_to_packages = None
+
+
+def _use_cached_resource(func):
+    def __use_cached_resource(*args, **kwargs):
+        global _source_path_to_packages
+        if _source_path_to_packages is None:
+            _source_path_to_packages = _get_source_path_to_packages()
+        return func(*args, **kwargs)
+    return __use_cached_resource
+
+
+@_use_cached_resource
+def find_node(*args, **kwargs):
+    kwargs['source_path_to_packages'] = _source_path_to_packages
+    return roslib.packages.find_node(*args, **kwargs)
+
+
+@_use_cached_resource
+def find_resource(*args, **kwargs):
+    kwargs['source_path_to_packages'] = _source_path_to_packages
+    return roslib.packages.find_resource(*args, **kwargs)
+
+
+@_use_cached_resource
+def find_in_workspaces(*args, **kwargs):
+    kwargs['source_path_to_packages'] = _source_path_to_packages
+    return catkin.find_in_workspaces.find_in_workspaces(*args, **kwargs)
--- a/tools/roslaunch/src/roslaunch/rlutil.py
+++ b/tools/roslaunch/src/roslaunch/rlutil.py
@@ -51,6 +51,7 @@ import rosgraph
 import roslaunch.core
 import roslaunch.config
 import roslaunch.depends
+import roslaunch.resource_cache
 from rosmaster import DEFAULT_MASTER_PORT
 
 def check_log_disk_usage():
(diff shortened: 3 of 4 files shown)


wkentaro:
Demo: https://drive.google.com/uc?id=0B9P1L--7Wd2veFZkSl82NzI4YVE

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -89,7 +90,7 @@ def resolve_launch_arguments(args):
     # try to resolve launch file in package first
     if len(args) >= 2:
         try:
-            resolved = roslib.packages.find_resource(args[0], args[1])
+            resolved = cached.find_resource(pkg=args[0], resource_name=args[1])
```
dirk-thomas: `pkg` and `resource_name` are not keyword arguments. Therefore please pass them as positional arguments.

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -47,6 +47,9 @@
 import rosgraph.names
 import rospkg
 
+from . import cached
```
dirk-thomas: I am not sure why the import was being done below but there might be a good reason for it.Therefore please keep the import in the same location as before.

tools/roslaunch/src/roslaunch/cached.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import roslib.packages
+import catkin.find_in_workspaces
+from catkin.find_in_workspaces import collect_path_data_in_workspaces
```
dirk-thomas: As mentioned in the catkin PR this function should be implemented here rather then in catkin itself. Also only a part of the return values is used so the function can be simplified to only return these information.

tools/roslaunch/src/roslaunch/rlutil.py
wkentaro: I think they are keyword arguments.
https://github.com/ros/ros/blob/jade-devel/core/roslib/src/roslib/packages.py#L463

and cached.py in this package supports both.
https://github.com/ros/ros_comm/pull/676/files#diff-66abf4062401099443c44fc29a06e40aR28

```
cached.find_resource(args[0], args[1])
```

```
cached.find_resource(pkg=args[0], resource_name=args[1])
```

tools/roslaunch/src/roslaunch/cached.py
wkentaro: Ok, I amended for this.

tools/roslaunch/src/roslaunch/substitution_args.py
wkentaro: Ok, I amended for this.

tools/roslaunch/src/roslaunch/rlutil.py
wkentaro: The change I made about the arguments are just a refactoring.
Using keyword arguments is good habit I think.

wkentaro:
I updated the commit, so please review again.

tools/roslaunch/src/roslaunch/rlutil.py
dirk-thomas: You are passing the two standard arguments as if they were keyword arguments. Please pass them without the keyword instead.

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -47,6 +47,7 @@
 import rosgraph.names
 import rospkg
 
+
```
dirk-thomas: Please avoid unrelated white space changes.

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -51,6 +51,7 @@
 import roslaunch.core
 import roslaunch.config
 import roslaunch.depends
+from . import cached
```
dirk-thomas: Please avoid relative imports but use the absolute package name instead (PEP 328).

Same below.

tools/roslaunch/src/roslaunch/cached.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
```
dirk-thomas: Please rename the file since `cached` does not clarify what the file is providing. Maybe `resource_cache`?

dirk-thomas:
Beside the comment the PRs look good. the performance improvement is definitely notable when the launch file has several resources to look up.

tools/roslaunch/src/roslaunch/cached.py
wkentaro: resource_cache would be better.

wkentaro:
I updated the commit following your comments.
Please review again.

tools/roslaunch/src/roslaunch/node_args.py
```diff
@@ -53,6 +53,7 @@
 import roslib.packages
 
 from . import substitution_args
+from . import resource_cache
```
dirk-thomas, line 55: Please move the import one line up to follow PEP8.

wkentaro:
I updated again.

tools/roslaunch/src/roslaunch/resource_cache.py
```diff
@@ -0,0 +1,64 @@
...
+import roslib.packages
+import catkin.find_in_workspaces
+
+from catkin.workspace import get_source_paths, get_workspaces
+from catkin_pkg.packages import find_packages
+
+
+def _get_source_path_to_packages(workspaces=None,
+                                 workspace_to_source_spaces=None):
+    """Returns dictionay with mappings from source paths to packages.
```
dirk-thomas: Sorry for the late tiny feedback but some of it only appears when I open the code in my editor which does all kind of style checks.

Please change first word from `Returns` to `Return` (PEP257 D401).
Please add an empty line between this and the next line (PEP257 D205).

tools/roslaunch/src/roslaunch/resource_cache.py
```diff
@@ -0,0 +1,64 @@
...
+            if src_path not in source_path_to_packages:
+                source_path_to_packages[src_path] = find_packages(src_path)
+    return source_path_to_packages
+
+
+# dictionary with mappings from source paths to packages
+_source_path_to_packages = None
+
+
+def _memoize(func):
```
dirk-thomas: Do you mean `memorize`? What about calling it `_use_cached_resources` instead?

Could this line:

```
kwargs['source_path_to_packages'] = _source_path_to_packages
```

be moved into the function instead of all three callers?

wkentaro: I mean memoize. 
https://en.m.wikipedia.org/wiki/Memoization

But maybe this is caching not memoization.

tools/roslaunch/src/roslaunch/resource_cache.py
wkentaro: I will.

tools/roslaunch/src/roslaunch/resource_cache.py
wkentaro: So I will change it.

dirk-thomas: Oh, great. I just learned a new word. That sound good then.

wkentaro:
I updated the commit. Please review again.

dirk-thomas:
Thanks for iterating on the patch.

wkentaro:
Thanks! :smile:

garaemon:
awesome feature!! :)

dirk-thomas:
I reviewed the change again and think an alternative approach will be even better. Currently the already passed in rospack instance is supposed to do some caching. But for separate nodes new instances are created. The following patches revert the previous changes and ensure that a single rospack instance is used which also caches the additional package mapping information:
- https://github.com/ros-infrastructure/rospkg/pull/89
- https://github.com/ros/ros/pull/95
- https://github.com/ros/ros_comm/pull/682

@wkentaro Can you please confirm that these changes also work in your scenario and that they improve performance even more?

dirk-thomas:
I have reverted the changes from this PR in favor of #682.

PR #677, opened 2015-10-07
Re #577: Signal slot & onRetryTimer callback fix

### LINKED ISSUE #577: roscpp: crash in ros::TransportPublisherLink::onRetryTimer()

ghost:
This issue manifested with random coredumps in the `onRetryTimer()`
callback in `transport_publisher_link.cpp`. These crashes were infrequent,
and extremely difficult to reproduce. They appeared to happen when
many Subscriptions were being destroyed, which made it likely to be
a destruction scheduling anomaly.

It turned there were two issues at play: the `onRetryTimer` function did
not having a valid reference to the `TransportPublusherLink` class, and
the `onConnectionDropped` function was not registered to the
`boost::signals2::signal` in such a way that tracked the existence of
the underlying ROS Subscriber connection. The first issue (which was
uncovered in #577) was resolved by calling `getInternalTimerManger()::add`
with `shared_from_this` instead of an empty shared pointer. Testing this
fix then revealed the second issue. Initially, the `onConnectionDropped`
seemed to be fixed by keeping track of the `TransportPublisherLink`
class with a `shared_from_this` pointer, though it did not pass all of
the `ros_comm` tests. However, changing anything this deep in ROS runs the
risk of unintended consequences, and upon further testing, we discovered
that `ros_controls/realtime_tools::RealTimePublishers` were not being
properly notified when connections were dropped, and entering deadlock
conditions.

Taking inspiration from the `boost::signals2` Automatic Connection Management
tutorial [[1](http://www.boost.org/doc/libs/1_54_0/doc/html/signals2/tutorial.html#signals2.tutorial.connection-management)], we then used `signals2::slot_type and track()` mechanisms in order
to maintain a reference to the `TransportPublisherLink` class during execution
of the `onConnectionDropped()` code. This was functionally similar to the
previous approach, with the added benefit that `boost::signals2` handles
all `shared_from_this()` reference counting under the hood, without preventing
destruction of the parent class (which we believe to be the source of the
unintended `RealTimePublisher` deadlocking). Upon the tracked object's
destruction, the `signal2` library automatically disconnects the signal.

Reproduction of this issue was incredibly difficult, since it required
the timing of the class destruction to perfectly coincide with the
execution of the callback function using that class. We found this could
only be reliably reproduced when running overnight simulations of hundreds
of Subscribers by Rethink Robotics Automated QA team. The occurrence of
this bug was producing roughly 1 to 5 coredumps per night in this testing
environment. Since this patch was applied, zero coredumps and
`RealTimePublisher` deadlocks have been seen over the course of a several
days of testing.

Major credit to Jeff Adams (@jeffadamsc) for pursuing the `shared_from_this`
callback connection strategy, and for QA allowing us to run these tests on
their nightly runs.

[1] http://www.boost.org/doc/libs/1_54_0/doc/html/signals2/tutorial.html#signals2.tutorial.connection-management

### DIFF
--- a/clients/roscpp/src/libros/transport_publisher_link.cpp
+++ b/clients/roscpp/src/libros/transport_publisher_link.cpp
@@ -78,7 +78,11 @@ TransportPublisherLink::~TransportPublisherLink()
 bool TransportPublisherLink::initialize(const ConnectionPtr& connection)
 {
   connection_ = connection;
-  connection_->addDropListener(boost::bind(&TransportPublisherLink::onConnectionDropped, this, _1, _2));
+  // slot_type is used to automatically track the TransporPublisherLink class' existence
+  // and disconnect when this class' reference count is decremented to 0. It increments
+  // then decrements the shared_from_this reference count around calls to the
+  // onConnectionDropped function, preventing a coredump in the middle of execution.
+  connection_->addDropListener(Connection::DropSignal::slot_type(&TransportPublisherLink::onConnectionDropped, this, _1, _2).track(shared_from_this()));
 
   if (connection_->getTransport()->requiresHeader())
   {
@@ -265,9 +269,11 @@ void TransportPublisherLink::onConnectionDropped(const ConnectionPtr& conn, Conn
     {
       retry_period_ = WallDuration(0.1);
       next_retry_ = WallTime::now() + retry_period_;
+      // shared_from_this() shared_ptr is used to ensure TransportPublisherLink is not
+      // destroyed in the middle of onRetryTimer execution
       retry_timer_handle_ = getInternalTimerManager()->add(WallDuration(retry_period_),
           boost::bind(&TransportPublisherLink::onRetryTimer, this, _1), getInternalCallbackQueue().get(),
-          VoidConstPtr(), false);
+          shared_from_this(), false);
     }
     else
     {


rethink-imcmahon:
@jeffadamsc could you take a quick look over the explanation and diff to see if I've missed anything?

rethink-imcmahon:
@dirk-thomas could I get a pull request builder please? Thanks!

dirk-thomas:
@ros-pull-request-builder

jeffadamsc:
Looks good to me!

The only thing that I would add is that we eventually ran into two callback issues. One was for onRetryTimer, which we fixed by just calling getInternalTimerManger()::add with shared_from_this instead of an empty shared pointer. This just tied into the tracked_object functionality that already existed in the timer manager callbacks, so the TransportPublisherLink object wouldn't go away while in the onRetryTimer method

The other problem was onDroppedConnection was being called and the TransportPublisherLink object went away. There wasn't already a tracked_object system, so we created one to parallel the code being used for the timer manager, and then when we connected the boost signals for the callback, we added the shared_from_this object and boost::signals2 did the rest, like Ian explained above.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros-pull-request-builder

rethink-imcmahon:
I'm trying something a bit less invasive based on this rather wonderful Boost Signals tutorial:
http://www.boost.org/doc/libs/1_54_0/doc/html/signals2/tutorial.html#signals2.tutorial.connection-management
Regardless of whether this passes the pull request builder, I don't think it's ready for merging yet. We need to do some more nightly runs / deployments on our robots.

dirk-thomas:
Thank you for your ongoing effort!

rethink-imcmahon:
Woot :) I'll put this into a patch for tomorrow's virtual test runs.

rethink-imcmahon:
Just rebased to squash everything together & rewrite the commit message (I'll update the initial comment as well). Everything ran smoothly over the weekend. No coredumps & no blatant unexpected / unintended consequences to other parts of ROS.

rethink-imcmahon:
That's, erm, confusing. This is the same patch that passed before, just rebased...

rethink-imcmahon:
@dirk-thomas, is there any way to build/test this separately from commit: 72a9a91fdacb5913991395733668d276b1246b3d ? I'm incredibly surprised this failed

dirk-thomas:
The PR job will always to a merge before. Until the `ros` repos is being released this check won't work. I don't think there is a workaround for it. I will retrigger this once I released the dependency.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
The current patch looks reasonable to me. Are you confident that it addresses the problem (based on your extensive testing) and is ready to be merged?

rethink-imcmahon:
Yes, I believe this patch finally solves the problem, based on our testing. We have 3 longterm simulations that have been running since Saturday with no issue. Roughly a dozen virtual "high stress" simulations each night during that time have completed without coredumps / publisher deadlocks. The QA team has also been deploying software (using this patch) on all of our physical robots each morning and testing extensively during the day.
Likely, it would be safe to merge this patch right now. Of course, each day we continue this testing, our confidence grows a bit higher. If you would prefer to wait another week before merging, I can report back on our status at that point. If we don't see anything in all this testing by then, it would show beyond much doubt that this patch has solved the issue without unintended side effects.

dirk-thomas:
Thank you very much for the effort you put into this fix :clap: I am happy to merge it with the already done testing. We then still have the soak time in shadow-fixed to gather more feedback.

Petrox:
Reading through this interesting story, I must say, it's impressive, that you've caught this, and also it's good to see the legions of simulated robots working hard to prove everything working.

So cheers to Ian and the QA team. :)

PR #682, opened 2015-10-12
Roslaunch performance improvement

dirk-thomas:
This reverts #676 and implements a different approach.


dirk-thomas:
@ros-pull-request-builder

wjwwood:
+1

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -90,7 +89,7 @@ def resolve_launch_arguments(args):
     # try to resolve launch file in package first
     if len(args) >= 2:
         try:
-            resolved = roslaunch.resource_cache.find_resource(args[0], args[1])
+            resolved = roslib.packages.find_resource(args[0], args[1])
```
wkentaro, line 92: Can we use cached resource for find_resource? (not only find_node at https://github.com/ros/ros_comm/pull/682/files#diff-67998749e41c63572aa0901b6715e9a0R251)

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -226,7 +225,7 @@ def check_roslaunch(f):
     # check for missing nodes
     for pkg, node_type in nodes:
         try:
-            if not roslaunch.resource_cache.find_node(pkg, node_type):
+            if not roslib.packages.find_node(pkg, node_type):
```
wkentaro: Can we use cached resource for find_node? (not only find_node at https://github.com/ros/ros_comm/pull/682/files#diff-67998749e41c63572aa0901b6715e9a0R251)

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -163,8 +163,7 @@ def _find_executable(resolved, a, args, _context):
     # we try to find the specific executable in libexec via catkin
     # which will search in install/devel space
     full_path = None
-    # don't know why not imported at the top
-    from roslaunch.resource_cache import find_in_workspaces
+    from catkin.find_in_workspaces import find_in_workspaces
     paths = find_in_workspaces(['libexec'], project=args[0], first_matching_workspace_only=True)  # implicitly first_match_only=True
```
wkentaro: Can we use cached resource for find_in_package? (not only find_node at https://github.com/ros/ros_comm/pull/682/files#diff-67998749e41c63572aa0901b6715e9a0R251)

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -190,8 +189,7 @@ def _find_resource(resolved, a, args, _context):
     path = _sanitize_path(args[1])
     # we try to find the specific path in share via catkin
     # which will search in install/devel space and the source folder of the package
-    # don't know why not imported at the top
-    from roslaunch.resource_cache import find_in_workspaces
+    from catkin.find_in_workspaces import find_in_workspaces
     paths = find_in_workspaces(['share'], project=args[0], path=path, first_matching_workspace_only=True, first_match_only=True)
```
wkentaro: Can we use cached resource for find_in_package? (not only find_node at https://github.com/ros/ros_comm/pull/682/files#diff-67998749e41c63572aa0901b6715e9a0R251)

tools/roslaunch/src/roslaunch/rlutil.py
dirk-thomas: This is easy since the calling scope already contains a rospack instance: 47f6c58931acda0f44cd5cb29395596500cc58c6

tools/roslaunch/src/roslaunch/rlutil.py
dirk-thomas, line 92: This function seems to be only called once. What is your use case where caching data here would improve anything?

wkentaro:
The finding paths is really fast. It's even faster than the implementation before in #676.

tools/roslaunch/src/roslaunch/substitution_args.py
dirk-thomas: I have added it in fa79ca8c23c9f0ebcdb197d6848e6c7d341a4883 but I don't have any launch file to exercise this and confirm the performance improvement. Can you please point me to a use case which benefits from this change?

tools/roslaunch/src/roslaunch/substitution_args.py
dirk-thomas: I have added it in fa79ca8c23c9f0ebcdb197d6848e6c7d341a4883 but I don't have any launch file to exercise this and confirm the performance improvement. Can you please point me to a use case which benefits from this change?

tools/roslaunch/src/roslaunch/rlutil.py
wkentaro, line 92: Actually I don't fully understand about rospack instance. I just wonder we can pass `rospack` instance to `find_resource` function which has cached `source _path_to_package` as previous implementation.

tools/roslaunch/src/roslaunch/substitution_args.py
wkentaro: the original issue has benchmark section.
https://github.com/ros/ros_comm/issues/541#issue-53175676

and It uses `openni.launch` for the testing.

dirk-thomas:
@ros-pull-request-builder

tools/roslaunch/src/roslaunch/substitution_args.py
dirk-thomas: Since I don't have a depth camera attached I can't use that example. Also based on the size I doubt it will show a different performance for the `find_in_workspaces` calls.

dirk-thomas:
@ros-pull-request-builder

guihomework:
@dirk-thomas , thanks a lot on solving this caching problem. 

However, we encountered a similar problem also in gazebo_ros_pkgs (https://github.com/ros-simulation/gazebo_ros_pkgs/pull/362) when resolving URDF resources (package://) 
Indeed, the `ros::package::getPath()` is used there and is not caching anything either.  We cached locally now which makes the resource resolution 6 times faster, but it might make sense to cache in the `ros::package` implementation. That caching at low level would probably also accelerate rviz loading models, and probably resolving URDF resources the same way.

PR #687, opened 2015-10-15
Add `-w` and `-t` options to roscore.

lucasb-eyer:
These define the number of worker-threads of the master and the socket timeout to use.

For large projects, the default of 3 workers with the OS' default timeout (of roughly 2-3mins) will cause the topic notification queue to stall as soon as a computer doesn't answer. So large projects want to set the timeout much lower and increase the thread-pool size a bit.

I'd like to add that having to define the option in three places and pass it through multiple objects almost made me go insane.

Ping @tlind as he debugged this together with me.

### DIFF
--- a/tools/roslaunch/scripts/roscore
+++ b/tools/roslaunch/scripts/roscore
@@ -33,6 +33,7 @@
 
 import sys
 from optparse import OptionParser
+from rosmaster.master_api import NUM_WORKERS
 
 NAME = 'roscore'
 
@@ -49,6 +50,12 @@ def _get_optparse():
     parser.add_option("-v", action="store_true",
                       dest="verbose", default=False,
                       help="verbose printing")
+    parser.add_option("-w", "--numworkers",
+                      dest="num_workers", default=NUM_WORKERS, type=int,
+                      help="override number of worker threads", metavar="NUM_WORKERS")
+    parser.add_option("-t", "--timeout",
+                      dest="timeout",
+                      help="override the socket connection timeout (in seconds).", metavar="TIMEOUT")
     return parser
 
 
--- a/tools/roslaunch/src/roslaunch/__init__.py
+++ b/tools/roslaunch/src/roslaunch/__init__.py
@@ -65,7 +65,9 @@ try:
     from rosmaster import DEFAULT_MASTER_PORT
 except:
     DEFAULT_MASTER_PORT = 11311
-    
+
+from rosmaster.master_api import NUM_WORKERS
+
 NAME = 'roslaunch'
 
 def configure_logging(uuid):
@@ -172,6 +174,12 @@ def _get_optparse():
     parser.add_option("--disable-title", default=False, action="store_true",
                       dest="disable_title",
                       help="Disable setting of terminal title")
+    parser.add_option("-w", "--numworkers",
+                      dest="num_workers", default=NUM_WORKERS, type=int,
+                      help="override number of worker threads. Only valid for core services.", metavar="NUM_WORKERS")
+    parser.add_option("-t", "--timeout",
+                      dest="timeout",
+                      help="override the socket connection timeout (in seconds). Only valid for core services.", metavar="TIMEOUT")
 
     return parser
     
@@ -294,7 +302,8 @@ def main(argv=sys.argv):
                     options.port = options.port or DEFAULT_MASTER_PORT
                 p = roslaunch_parent.ROSLaunchParent(uuid, args, roslaunch_strs=roslaunch_strs,
                         is_core=options.core, port=options.port, local_only=options.local_only,
-                        verbose=options.verbose, force_screen=options.force_screen)
+                        verbose=options.verbose, force_screen=options.force_screen,
+                        num_workers=options.num_workers, timeout=options.timeout)
                 p.start()
                 p.spin()
             finally:
--- a/tools/roslaunch/src/roslaunch/launch.py
+++ b/tools/roslaunch/src/roslaunch/launch.py
@@ -57,6 +57,8 @@ from roslaunch.pmon import start_process_monitor, ProcessListener
 
 from roslaunch.rlutil import update_terminal_name
 
+from rosmaster.master_api import NUM_WORKERS
+
 _TIMEOUT_MASTER_START = 10.0 #seconds
 _TIMEOUT_MASTER_STOP  = 10.0 #seconds
 
@@ -233,7 +235,7 @@ class ROSLaunchRunner(object):
     monitored.
     """
     
-    def __init__(self, run_id, config, server_uri=None, pmon=None, is_core=False, remote_runner=None, is_child=False, is_rostest=False):
+    def __init__(self, run_id, config, server_uri=None, pmon=None, is_core=False, remote_runner=None, is_child=False, is_rostest=False, num_workers=NUM_WORKERS, timeout=None):
         """
         @param run_id: /run_id for this launch. If the core is not
             running, this value will be used to initialize /run_id. If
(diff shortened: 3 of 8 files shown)


tools/rosmaster/src/rosmaster/master_api.py
```diff
@@ -73,8 +73,6 @@
 from rosmaster.registrations import RegistrationManager
 from rosmaster.validators import non_empty, non_empty_str, not_none, is_api, is_topic, is_service, valid_type_name, valid_name, empty_or_valid_name, ParameterInvalid
 
-NUM_WORKERS = 3 #number of threads we use to send publisher_update notifications
```
dirk-thomas, line 76: This is public API and can't be removed without potentially breaking existing code. Please keep it as-is.

tools/roslaunch/scripts/roscore
```diff
@@ -49,6 +49,12 @@ def _get_optparse():
     parser.add_option("-v", action="store_true",
                       dest="verbose", default=False,
                       help="verbose printing")
+    parser.add_option("-w", "--numworkers",
+                      dest="nwork", default=3, type="int",
+                      help="override number of worker threads", metavar="NUM_WORKERS")
+    parser.add_option("-t", "--timeout",
+                      dest="timeout", default=-1.0, type="float",
+                      help="override the socket connection timeout (in seconds). Negative values keep the default", metavar="TIMEOUT")
```
dirk-thomas: Please mention that these two arguments are for the ros master.

Same below in `roslaunch/__init__.py`.

tools/roslaunch/scripts/roscore
```diff
@@ -49,6 +49,12 @@ def _get_optparse():
     parser.add_option("-v", action="store_true",
                       dest="verbose", default=False,
                       help="verbose printing")
+    parser.add_option("-w", "--numworkers",
+                      dest="nwork", default=3, type="int",
```
dirk-thomas: `nwork` is an abbreviation which might not be obvious when reading the code in the future. Please use something more verbose like `num_workers` instead (throughout the patch).

tools/roslaunch/scripts/roscore
dirk-thomas: I don't think the remark about negative values is necessary. If someone really specifies negative values he will notice that no message with the desired timeout will appear.

tools/roslaunch/scripts/roscore
dirk-thomas: The default should come from the constant in the code instead of being hard coded here.

tools/roslaunch/scripts/roscore
```diff
@@ -49,6 +49,12 @@ def _get_optparse():
...
+                      help="override number of worker threads", metavar="NUM_WORKERS")
+    parser.add_option("-t", "--timeout",
+                      dest="timeout", default=-1.0, type="float",
```
dirk-thomas: I would suggest no default here which implies `None` and handle that accordingly.

tools/roslaunch/src/roslaunch/nodeprocess.py
```diff
@@ -61,7 +61,7 @@ def _next_counter():
     _counter += 1
     return _counter
 
-def create_master_process(run_id, type_, ros_root, port):
+def create_master_process(run_id, type_, ros_root, port, nworkers, timeout):
```
dirk-thomas: This is public API which can't be changed in a breaking way like this. Please make the new arguments optional.

tools/rosmaster/src/rosmaster/master_api.py
```diff
@@ -236,13 +234,13 @@ class ROSMasterHandler(object):
     this parameter (see ros::client::getMaster()).
     """
     
-    def __init__(self):
+    def __init__(self, nwork):
```
dirk-thomas: This is public API which can't be changed in a breaking way like this. Please make the new arguments optional.

lucasb-eyer:
I believe I have addressed all of your concerns. Note, though, that with this change `NUM_WORKERS` might not reflect the truth, as it's only used as default value. I can instead make it reflect the true value if you want.

tools/roslaunch/scripts/roscore
```diff
@@ -49,6 +50,12 @@ def _get_optparse():
     parser.add_option("-v", action="store_true",
                       dest="verbose", default=False,
                       help="verbose printing")
+    parser.add_option("-w", "--numworkers",
+                      dest="num_workers", default=NUM_WORKERS, type="int",
```
dirk-thomas: According to the `argparse` documentation the `type` argument should be `int` instead of `"int"`.

tools/roslaunch/scripts/roscore
```diff
@@ -49,6 +50,12 @@ def _get_optparse():
...
+                      help="override number of worker threads", metavar="NUM_WORKERS")
+    parser.add_option("-t", "--timeout",
```
dirk-thomas, line 56: Should this argument be of type float?

tools/roslaunch/src/roslaunch/__init__.py
```diff
@@ -295,6 +303,9 @@ def main(argv=sys.argv):
                 p = roslaunch_parent.ROSLaunchParent(uuid, args, roslaunch_strs=roslaunch_strs,
                         is_core=options.core, port=options.port, local_only=options.local_only,
                         verbose=options.verbose, force_screen=options.force_screen)
+                if options.core:
+                    p.num_workers = options.num_workers
+                    p.timeout = options.timeout
```
dirk-thomas: Please don't add additional members to the `ROSLaunchParent` instance but add them to the class.

tools/roslaunch/src/roslaunch/launch.py
```diff
@@ -386,7 +388,9 @@ def _launch_master(self):
             validate_master_launch(m, self.is_core, self.is_rostest)
 
             printlog("auto-starting new master")
-            p = create_master_process(self.run_id, m.type, get_ros_root(), m.get_port())
+            num_workers = getattr(self, "num_workers", NUM_WORKERS)
+            timeout = getattr(self, "timeout", None)
```
dirk-thomas: With these attributes being members of the class `getattr` is not necessary here.

dirk-thomas:
@ros-pull-request-builder

tools/roslaunch/scripts/roscore
lucasb-eyer: Good catch! I don't know how this possibly worked, but it did when I tested it. Will fix.

tools/roslaunch/scripts/roscore
lucasb-eyer, line 56: It previously was, and I used `-1` to mean "default". You asked me to use `None` for default instead (I agree), but that doesn't allow using `Float` as a type anymore, because it will fail converting `None` to `Float`.

tools/roslaunch/scripts/roscore
lucasb-eyer: While grepping, I found this same mistake in [rosmsg](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rosmsg/src/rosmsg/__init__.py#L262) and [rosnode](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rosnode/src/rosnode/__init__.py#L726).

lucasb-eyer:
I addressed all comments, except for the `Float` type one, see my inline comment for why.

lucasb-eyer:
The failure seems to be git-related: `FATAL: Could not checkout null with start point 8c7de6996601395f6cb28a424f548d3f232f6503`, not sure why though.

dirk-thomas:
Since `roslaunch` uses a new command line option of `rosmaster` it must add a minimum version to the dependency. Since it is not yet clear when this will be merged I will add it during the merge process.

dirk-thomas:
@ros-pull-request-builder

lucasb-eyer:
If I understand correctly, there's nothing more to do from my side, right?

dirk-thomas:
Correct.

wjwwood:
lgtm, but you should update the wiki in these places (at least) mentioning the new options:
- http://wiki.ros.org/roscore
- http://wiki.ros.org/roslaunch/Commandline%20Tools

Make sure to note what versions of ROS will contain this feature.

dirk-thomas:
Thanks, cherry-picked (including a version dependency) in: 08f6bf06fe4e68efbf6b2c13b558a4d5298d1a41

PR #690, opened 2015-10-26
Shutdown rospy timers cleanly

stonier:
Timers go off and start periodically running in the background, and will throw exceptions when ros shutdown occurs because of the sleep function.

These are impossible to catch as they're off in a background thread and can be cleanly handled as in the code change.

Others with the same problem:
- [baxter examples](https://github.com/RethinkRobotics/baxter_examples/issues/48)

### DIFF
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -210,7 +210,11 @@ class Timer(threading.Thread):
         current_expected = rospy.rostime.get_rostime() + self._period
         last_expected, last_real, last_duration = None, None, None
         while not rospy.core.is_shutdown() and not self._shutdown:
-            r.sleep()
+            try:
+                r.sleep()
+            except rospy.exceptions.ROSInterruptException as e:
+                if rospy.core.is_shutdown():
+                    break
             if self._shutdown:
                 break
             current_real = rospy.rostime.get_rostime()


stonier:
Note - to reproduce, these exceptions are only getting thrown in the simulation branch of that sleep function. I am getting them when simulating with gazebo.

clients/rospy/src/rospy/timer.py
```diff
@@ -210,7 +210,13 @@ def run(self):
         current_expected = rospy.rostime.get_rostime() + self._period
         last_expected, last_real, last_duration = None, None, None
         while not rospy.core.is_shutdown() and not self._shutdown:
-            r.sleep()
+            try:
+                r.sleep()
+            except rospy.exceptions.ROSInterruptException as e:
+                if rospy.core.is_shutdown():
+                    break
+                else:
```
dirk-thomas: Please remove the `else` line since it is not necessary.

stonier:
Failed because of the rebase?

clients/rospy/src/rospy/timer.py
```diff
@@ -210,7 +210,11 @@ def run(self):
         current_expected = rospy.rostime.get_rostime() + self._period
         last_expected, last_real, last_duration = None, None, None
         while not rospy.core.is_shutdown() and not self._shutdown:
-            r.sleep()
+            try:
+                r.sleep()
+            except rospy.exceptions.ROSInterruptException as e:
+                if rospy.core.is_shutdown():
+                    break
```
dirk-thomas, line 217: Before the update there was a `raise` line in here. My previous comment only referred to the `else` line. I think the `raise` still needs to stay otherwise this function changes behavior significantly.

dirk-thomas:
Thanks for the patch.

Cherry-picked with my previous comment: 2e84880a6f494e32bc016d414990e6d3f17b8de6

stonier:
Faster than I was....thanks :neckbeard:

PR #694, opened 2015-11-03
filterROSArgs function in init.cpp

Karsten1987:
Allows a pre-filtering of ROS keywords before calling `ros::init()`.

This is an alternative to `ros::this_node::getName()` [[0]](http://docs.ros.org/api/roscpp/html/namespaceros_1_1this__node.html#a2a991f67172b59ccd7e5b4aeca9b76b7), since this requires to call `ros::init(argc, argv, "node_name")` beforehand and forcing to specify a node name. 
ros::filterROSArgs can be called wo/ init for filter various keywords such as `__name` or `__ns`. 

```
$ rosrun hello_world hello_world __name:=foo
```

```
// hello_world.cpp
int main( int argc, char** argv )
{ 
  std::string node_name = ros::filterROSArgs(argc, argv, "__name");
  if (!node_name.empty())
    ros::init(argc, argv, node_name);
  else
    ros::init(argc, argv, "bar");

  return 0;
}
```

In case the desired argument is not found, an empty string is returned.

### DIFF
--- a/clients/roscpp/include/ros/init.h
+++ b/clients/roscpp/include/ros/init.h
@@ -196,6 +196,16 @@ ROSCPP_DECL bool isStarted();
 ROSCPP_DECL CallbackQueue* getGlobalCallbackQueue();
 
 /**
+ * \brief searches the command line arguments for the given arg parameter. In case this argument is not found
+ * an empty string is returned.
+ *
+ * \param argc the number of command-line arguments
+ * \param argv the command-line arguments
+ * \param arg argument to search for
+ */
+ROSCPP_DECL std::string getROSArg(int argc, const char* const* argv, const std::string& arg);
+
+/**
  * \brief returns a vector of program arguments that do not include any ROS remapping arguments.  Useful if you need
  * to parse your arguments to determine your node name
  *
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -509,6 +509,20 @@ void init(const VP_string& remappings, const std::string& name, uint32_t options
   init(remappings_map, name, options);
 }
 
+std::string getROSArg(int argc, const char* const* argv, const std::string& arg)
+{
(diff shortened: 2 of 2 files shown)


wjwwood:
I guess you'll be using this for some other ROS parameter or in some other way? The reason I ask is that this code is equivalent to your snippet above:

``` c++
// hello_world.cpp
int main( int argc, char** argv )
{ 
  ros::init(argc, argv, "bar");

  return 0;
}
```

If you run it without a `__name:=` argument it will be called `bar`, but if you call it with `__name:=foo` it will be called `foo`.

dirk-thomas:
Can you please describe your use case?

Beside that I am not sure if `filterROSArgs` is a suitable name for the function. Since it doesn't really filter anything and only checks for one specific argument I would suggest calling it `getROSArg` instead.

Karsten1987:
@wjwwood I might be wrong here, but `ros::this_node::getName()` resolves the complete namespace. So the example you gave would lead to `/foo`. Launching the executable inside a namespace `bar_ns` would return `/bar_ns/foo` as a name. Can you confirm that?

Initial motivation comes from this PR https://github.com/ros-naoqi/naoqi_driver/pull/44. 
Other use cases are launching the same executable for different devices (e.g. IP cameras) and adjusting their unique topic/rosparam names on runtime via a respective launch file. 

In general I was wondering how you'd access the `__name` variable anyways which is passed by a launch file when inside your `ros::init` you overwrite this by a hard-coded name?

@dirk-thomas Sure, there is no problem in renaming this. You guys might find better names according to the context.

dirk-thomas:
@Karsten1987 Can you please update the function according to my previous comment.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Thanks for iterating on this. Next time please also comment on the ticket since there is no notification for pushing to a PR.

PR #697, opened 2015-11-08
Show size unit for --size of rosbag record

wkentaro:
Expressly declare the size unit on help display for rosbag record --size option


dirk-thomas:
Thanks.

PR #702, opened 2015-11-19
rostopic pub updates keyword time

chris-smith:
This adds a new flag (-u, --update_time) to rostopic pub that will update timestamps when using the 'now' or 'auto' keywords and the rate flag.

Without this flag, rostopic pub works as before

rostopic pub /my_topic geometry_msgs/PointStamped "header: auto
point:
  x: 0.0
  y: 0.0
  z: 0.0" -r 10

(echoing /my_topic in another window will show this message coming in at 10Hz with the same timestamp)

With this flag, the timestamp will update for each new message

rostopic pub /my_topic geometry_msgs/PointStamped "header: auto
point:
  x: 0.0
  y: 0.0
  z: 0.0" -r 10 -u

This flag is only used when publishing messages with the rate flag from the command line.

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -1306,7 +1306,7 @@ def create_publisher(topic_name, topic_type, latch):
     pub = rospy.Publisher(topic_name, msg_class, latch=latch, queue_size=100)
     return pub, msg_class
 
-def _publish_at_rate(pub, msg, rate, verbose=False):
+def _publish_at_rate(pub, msg, rate, verbose=False, substitute_keywords=False, pub_args=None):
     """
     Publish message at specified rate. Subroutine of L{publish_message()}.
     
@@ -1320,6 +1320,8 @@ def _publish_at_rate(pub, msg, rate, verbose=False):
     except ValueError:
         raise ROSTopicException("Rate must be a number")
     while not rospy.is_shutdown():
+        if substitute_keywords:
+            _fillMessageArgs(msg, pub_args)
         if verbose:
             print("publishing %s"%msg)
         pub.publish(msg)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Since the substitution changes more things then just a _time_ the name `update_time` doesn't seem to be fitting very well. Beside this naming the patch lgtm. Can you please suggest alternative names.

chris-smith:
How about `update_keywords`?

dirk-thomas:
Since its more of a replace / substitute operation than an update: `replace_keywords` / `substitute_keywords`?

chris-smith:
I would say `substitute_keywords` if only because `rostopic pub` could then support `-s, --substitute_keywords` flags whereas if we went with `replace_keywords` people might get confused about which argument `-r` is supposed to be.

chris-smith:
I liked update though because it implies that it is being reevaluated, the keyword substitution will happen whether the flag is supplied or not. Maybe `update_keyword_substitution`?

dirk-thomas:
The option will replace / substitute keyword / placeholders in the message values with some autogenerated values. `update_keyword_substitution` sounds weird to me too. The _keyword substitution_ is not being _updated_ in any way.

chris-smith:
Definitely fair to say the keyword substitution is not being updated. It is being performed at constant intervals, as opposed to only being done once at the beginning though. I can't think of a better term than `substitute_keywords` without being unnecessarily verbose - combined with the 'help' information I think `substitute_keywords` purpose should be clear enough.

dirk-thomas:
Sounds good to me. And the help text of the command line option should resolve any remaining uncertainty :wink:

chris-smith:
Updated pull request with flag renamed to `substitute_keywords` instead of `update_time`

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1417,6 +1423,8 @@ def _rostopic_cmd_pub(argv):
                       help="read args from YAML file (Bagy)")
     parser.add_option("-l", '--latch', dest="latch", default=False, action="store_true",
                       help="enable latching for -f, -r and piped input.  This latches the first message.")
+    parser.add_option("-s", '--substitute_keywords', dest="substitute_keywords", default=False, action="store_true",
```
dirk-thomas: Please use `--substitute-keywords`. Underscore separators are rather uncommon for command line arguments.

chris-smith:
Ah! Whoops. Done.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Thank you for the patch.

PR #708, opened 2015-11-30
Enable more tests

gerkey:
Remove exe bit on test files and remove setup.cfg, to avoid confusing test results.

Without this change, depending on how you run `nosetests`, you might not be running all of the tests.  For example, as `catkin_make` runs `nosetests`, in a different directory and with the `--with=/path/to/src/test/directory` argument, any test files with their executable bit set will be ignored.

There seem to be other potentially confusing situations possible here.  The sanest course of action appears to be dropping executable bits on all Python test files and also removing `setup.cfg` files.


gerkey:
+1

dirk-thomas:
I have created a separate ticket for the `rosmsg` tests: #709.

PR #710, opened 2015-12-01
Add pass_all_args attr to include tag

gerkey:
Add a new attribute, `pass_all_args`, to the `include` tag.  It's a bool, with default of false.  E.g.:

``` xml
<arg name="foo" value="bar"/>
<arg name="bat" value="bang"/>

<!-- Pass down foo=bar and bat=bang -->
<include file="foo.launch" pass_all_args="true"/>

<!-- Pass down foo=override and bat=bang -->
<include file="foo.launch" pass_all_args="true">
  <arg name="foo" value="override"/>
</include>

<!-- The existing syntax can still be used to explicitly pass down foo=bar and bat=bang -->
<include file="foo.launch" pass_all_args="false">
  <arg name="foo" value="bar"/>
  <arg name="bat" value="bang"/>
</include>
```

If false, then nothing changes from existing behavior.

If true, then all args set in the current context are added to the child context that is created for processing the included file.  You can do this instead of explicitly listing each argument that you want to pass down.

Notes on behavior when `pass_all_args="true"`:
- In the case that the parent passes down a value for an arg that is defined as constant in the child, then the child's value is used.
- If an `<arg>` tag explicitly passes down an argument that is also set in the parent, then that explicit value overrides.

### DIFF
--- a/tools/roslaunch/test/unit/test_xmlloader.py
+++ b/tools/roslaunch/test/unit/test_xmlloader.py
@@ -938,4 +938,115 @@ class TestXmlLoader(unittest.TestCase):
         self.assertEquals(param_d['/include3/include_test/p3_test'], 'set')
         self.assertEquals(param_d['/include3/include_test/p4_test'], 'new3')
         
+    # Test the new attribute <include pass_all_args={"true"|"false"}>
+    def test_arg_all(self):
+        loader = roslaunch.xmlloader.XmlLoader()
+        filename = os.path.join(self.xml_dir, 'test-arg-all.xml')
 
+        # Test suite A: load without an optional arg set externally
+        mock = RosLaunchMock()
+        loader.load(filename, mock, argv=["required:=test_arg"])
+
+        param_d = {}
+        for p in mock.params:
+            param_d[p.key] = p.value
+
+        # Sanity check: Parent namespace
+        self.assertEquals(param_d['/p1_test'], 'test_arg')
+        self.assertEquals(param_d['/p2_test'], 'not_set')
+        self.assertEquals(param_d['/p3_test'], 'parent')
+
+        # Test case 1: include without pass_all_args
+        self.assertEquals(param_d['/notall/include_test/p1_test'], 'test_arg')
+        self.assertEquals(param_d['/notall/include_test/p2_test'], 'not_set')
+        self.assertEquals(param_d['/notall/include_test/p3_test'], 'set')
+
+        # Test case 2: include without pass_all_args attribute, and pass optional arg
+        # internally
+        self.assertEquals(param_d['/notall_optional/include_test/p1_test'], 'test_arg')
+        self.assertEquals(param_d['/notall_optional/include_test/p2_test'], 'not_set')
+        self.assertEquals(param_d['/notall_optional/include_test/p3_test'], 'set')
+
+        # Test case 3: include with pass_all_args attribute, instead of passing individual
+        # args
+        self.assertEquals(param_d['/all/include_test/p1_test'], 'test_arg')
+        self.assertEquals(param_d['/all/include_test/p2_test'], 'not_set')
+        self.assertEquals(param_d['/all/include_test/p3_test'], 'set')
+
+        # Test case 4: include with pass_all_args attribute, and override one
+        # arg inside the include tag
+        self.assertEquals(param_d['/all_override/include_test/p1_test'], 'override')
+        self.assertEquals(param_d['/all_override/include_test/p2_test'], 'not_set')
+        self.assertEquals(param_d['/all_override/include_test/p3_test'], 'set')
+
+        # Test suite B: load with an optional arg set externally
+        mock = RosLaunchMock()
+        loader.load(filename, mock, argv=["required:=test_arg", "optional:=test_arg2"])
+
+        param_d = {}
+        for p in mock.params:
+            param_d[p.key] = p.value
+
+        # Sanity check: Parent namespace
+        self.assertEquals(param_d['/p1_test'], 'test_arg')
+        self.assertEquals(param_d['/p2_test'], 'test_arg2')
+        self.assertEquals(param_d['/p3_test'], 'parent')
+
+        # Test case 1: include without pass_all_args attribute
+        self.assertEquals(param_d['/notall/include_test/p1_test'], 'test_arg')
+        self.assertEquals(param_d['/notall/include_test/p2_test'], 'not_set')
+        self.assertEquals(param_d['/notall/include_test/p3_test'], 'set')
+
+        # Test case 2: include without pass_all_args attribute, and pass optional arg
+        # internally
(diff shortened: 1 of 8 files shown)


dirk-thomas:
@ros-pull-request-builder

scpeters:
Wow, this sounds amazing. Lots of `gazebo_ros` launch files are missing parameters because they all have to be manually defined

tools/roslaunch/test/unit/test_xmlloader.py
```diff
@@ -938,4 +938,80 @@ def test_arg(self):
...
+
+        # Sanity check: Parent namespace
+        self.assertEquals(param_d['/p1_test'], 'test_arg')
+        self.assertEquals(param_d['/p2_test'], 'test_arg2')
+        self.assertEquals(param_d['/p3_test'], 'parent')
+
+        # Test case 1: include without pass_all_args attribute
+        self.assertEquals(param_d['/notall/include_test/p1_test'], 'test_arg')
+        self.assertEquals(param_d['/notall/include_test/p2_test'], 'not_set')
+        self.assertEquals(param_d['/notall/include_test/p3_test'], 'set')
```
scpeters, line 998: I don't see expectations for `include_arg` or `if_test` in the test file, even though they are set below.

gerkey, line 998: I'm only setting those (https://github.com/ros/ros_comm/blob/add_pass_all_args/tools/roslaunch/test/xml/test-arg-all.xml#L6-L7) because they're required in the included file, which I'm reusing from another test (https://github.com/ros/ros_comm/blob/add_pass_all_args/tools/roslaunch/test/xml/test-arg-include.xml#L4-L5), where they're used for a different testing purpose. I could test their values here, but I feel that that case (passing an arg that is required in an included file) is already covered by testing the value of the `required` arg.

dirk-thomas:
Does the patch result in no exception being thrown in the following case anymore:
- `first.xml` has a set of arguments (but not `grounded`)
- `first.xml`includes `second.xml` with `pass_all_args="True"` and `grounded="foo"`
- `second.xml` defines an argument `grounded` with a fixed value

If that is the case I think it would be good to preserve the exception.

dirk-thomas:
Beside my previous comment the patch looks good to me (merge pending the decision into which ROS distro this and followup changes should be released).

tools/roslaunch/test/xml/test-arg-all.xml
```diff
@@ -0,0 +1,48 @@
+<launch>
+
+  <arg name="required" />
+  <arg name="optional" default="not_set" />
```
scpeters, line 4: this value `"not_set"` is the same as the [upstream test-arg-include.xml](https://github.com/ros/ros_comm/blob/add_pass_all_args/tools/roslaunch/test/xml/test-arg-include.xml#L7). Should we use a different value here?

scpeters, line 4: I believe this would affect the test for `/notall_optional/include_test/p2_test`

scpeters, line 4: I'm rethinking this...

scpeters:
~~One feature that just occurred to me and may be related to the exception that @dirk-thomas mentioned:~~

~~Base launch file with optional parameters:~~

```
<launch>
  <arg name="param1" default="this_is_optional" />
...
  <arg name="paramN" default="this_is_optional" />
</launch>
```

~~If I want to create a derived launch file without redefining all the optional parameters and duplicating their default values, can I do:~~

```
<launch>
  <include ns="all" file="base.launch" pass_all_args="true">
    <param20>gazebo_world_name.world</param20>
  </include>
</launch>
```

~~and then specify `param10:="--verbose"` when I invoke it from the command-line? That would be really helpful, but maybe that's the exception that @dirk-thomas wants to keep?~~

I think this should work fine.

gerkey:
@scpeters, your test case should work fine. Did you try it?

scpeters:
No, I didn't test it. Sorry for the noise.

scpeters:
Sorry I didn't understand @dirk-thomas 's comment, but now I do.

gerkey:
@dirk-thomas, I understand the motivation behind your desire to have that exception behavior, and I would go further to say that it would be nice to always throw an exception in the case that the parent launch file explicitly sets an arg while including a file that sets the same arg constant, whether or not `pass_all_args` was enabled for that include, and whether or not that arg was set in the parent.

That is, ideally, it would only be legal to attempt to pass an arg to a child that sets the same arg constant in the case that the passing is implicit, via `pass_all_args`.

However, I don't see a way to implement that behavior without making intrusive changes into how roslaunch passes arguments from parent to child. We'd need to carry with each argument _how_ it was set (implicitly via `pass_all_args` or explicitly in the include). Given that what we're talking about is stricter error-checking in a pretty narrow use case, I don't think that it's worth the risk of making those changes to get that exception behavior. (I also don't think that the lack of that exception behavior is sufficient to justify rejecting this PR.)

For future reference, I added a test that checks for that exception behavior, commented out for now: https://github.com/ros/ros_comm/blob/add_pass_all_args/tools/roslaunch/test/unit/test_xmlloader.py#L1040-L1053.

It is very good that you brought up this point, as I found a nasty bug while investigating the question, and added more tests as a result: https://github.com/ros/ros_comm/blob/add_pass_all_args/tools/roslaunch/test/unit/test_xmlloader.py#L1020-L1037.

dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
@ros/ros_team Please review and state your opinion to merge this into Indigo and Jade.

gerkey:
Having written it, I can't review it, but I vote to merge into both distros. It's a new feature that has no effect unless explicitly enabled, and I feel confident in both the characterization of the behavior and the coverage of the test suite.

However, I can appreciate the general stability argument for Indigo, in which case we can just put it in Jade.

gerkey:
Result from today: merge to indigo and jade.

scpeters:
Some tests failed. Are those flaky tests? Can the builds be re-triggered?

dirk-thomas:
The two failing roswtf tests are due to ros/catkin#777 and will pass once that fix is available in a released catkin package.

If there are more failing tests that is likely due to the patch.

dirk-thomas:

dirk-thomas:

dirk-thomas:

dirk-thomas:
And squashed.

dirk-thomas:
@gerkey Please update the ROS wiki in the near future to document this new argument and comment here with a link to the diff.

gerkey:
Docs updated!

http://wiki.ros.org/roslaunch/XML/include

dirk-thomas:
Thank you. I added the version number to it in case anyone wants to check: http://wiki.ros.org/action/diff/roslaunch/XML/include?action=diff&rev1=6&rev2=7

gerkey:
Awesome, thanks. I wanted to do that, but didn't know which version it would appear in.

PR #711, opened 2015-12-03
waiting for param server to be available before trying to get Param

asmodehn:
This fixes a problem when launching roscore and using scriptapi very quickly ( scripts for tests for exemple ). Now scriptapi wait for the param server to be available ( the same way than the main roslaunch method does )


dirk-thomas:
@ros-pull-request-builder

dirk-thomas:
Thanks!

PR #712, opened 2015-12-04
[rostopic] Get multiple topics with rostopic hz

wkentaro:


wkentaro:
it looks like 
image

wkentaro:
This is designed to find bottleneck of topic node flow. So it supports `--search-parent` option which automatically detect parent topics of specified one.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,8 +175,20 @@ def print_hz(self):
...
+
+    def print_hz(self):
+        """
+        print the average publishing rate to screen
+        """
+        ret = self.get_hz()
+        if ret is None and self.times:
+            print("no new messages")
+            return
+        rate, min_delta, max_delta, std_dev, window = ret
```
dirk-thomas: This line will fail if `ret is None` but `not self.times`.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -142,14 +142,16 @@ def callback_hz(self, m):
...
                 self.times.pop(0)
 
-    def print_hz(self):
+    def get_hz(self):
         """
-        print the average publishing rate to screen
+        calculate the average publising rate
+
+        @returns: tuple of stat results
+            rate, min_delta, max_delta, standard deviation, window number
```
dirk-thomas: The doc needs to mention the cases in which the method returns None.

tools/topic_tools/package.xml
```diff
@@ -28,6 +28,8 @@
   <build_depend>xmlrpcpp</build_depend>
 
   <run_depend>message_runtime</run_depend>
+  <run_depend>python-numpy</run_depend>
+  <run_depend>python-texttable</run_depend>
```
dirk-thomas: I am not sure that it is a good idea to pull in these extra dependencies in a patch release of already released stable distro (Indigo, Jade). If the API changes in `rostopic/__init__.py` get merged the command line tool itself could also be released as a standalone package? That also gives you more flexibility for future updated and maintenance releases (since ros_comm us usually a slow turnover time between patch releases due to its impact).

tools/rostopic/src/rostopic/__init__.py
wkentaro: Thanks. fixed.

tools/rostopic/src/rostopic/__init__.py
wkentaro: Thanks fixed.

tools/topic_tools/package.xml
wkentaro: I have removed the dependency on these packages.

I don't mind slow releasing of ros_comm, and putting tools that can be used in general cases on upstream package is better, I think. Tools in lowstream package will soon obsolete when the maintainer's status changes.
(We have some packages as shared package in my lab, but some of them are not used after the author student graduates.)

wkentaro:
Maybe this can be supported by `rostopic hz` by `rostopic hz [topic1] [topic2] ...`.
What do you think about this?

wkentaro:
Now it looks like
image

tools/topic_tools/package.xml
dirk-thomas: Sure, but adding more and more features into ros_comm also means offloading the maintenance work to me :wink:

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,40 +176,122 @@ def print_hz(self):
...
-        # have to subscribe with topic_type
-        sub = rospy.Subscriber(real_topic, msg_class, rt.callback_hz)
+    hz_checkers = []
+    for topic in topics:
+        rt = ROSTopicHz(window_size, filter_expr=filter_expr)
+        # we use a large buffer size as we don't know what sort of messages we're dealing with.
+        # may parameterize this in the future
+        if filter_expr is not None:
+            # have to subscribe with topic_type
+            sub = rospy.Subscriber(real_topic, msg_class, rt.callback_hz)
```
dirk-thomas: `msg_class` is only set in the previous loop (line 230) which doesn't seem to be what you want?

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,40 +176,122 @@ def print_hz(self):
...
+        else:
+            sub = rospy.Subscriber(real_topic, rospy.AnyMsg, rt.callback_hz)
+        hz_checkers.append(rt)
+    if len(topics) == 1:
+        print("subscribed to [%s]"%real_topic)
     else:
-        sub = rospy.Subscriber(real_topic, rospy.AnyMsg, rt.callback_hz)        
-    print("subscribed to [%s]"%real_topic)
+        print("subscribed %d topics" % len(topics))
```
dirk-thomas: This should still print which actual topics have been subscribed to.

dirk-thomas:
Can you please describe what a parent topic is?

Also the two features (multiple topics for hz and parent topics) seem to be independent from each other (except that the later one needs the former one). Please use separate PRs or at least commits for this in the future since it simplifies the review significantly.

tools/topic_tools/package.xml
wkentaro: Ok, so I will send pr only when I think it's really necessary.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -100,7 +100,7 @@ class ROSTopicHz(object):
     def __init__(self, window_size, filter_expr=None):
         import threading
         self.lock = threading.Lock()
-        self.last_printed_tn = 0
+        self.last_msg_tn = 0
```
dirk-thomas: Please do not change the name of this variable. It is publically visible and it would break the API to rename it.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,40 +176,95 @@ def print_hz(self):
...
+            stats.append(stat)
+            for i, s in enumerate(stat):
+                col_width = len(s)
+                if col_width > max_col_width[i]:
+                    max_col_width[i] = col_width
+        if not stats:
+            print('no new messages')
+            continue
+        stats = sorted(stats, key=lambda x: x[1], reverse=True)
+        # compose table with left alignment
```
dirk-thomas: I think left alignment is the wrong term here. The number are all right aligned if I understand the code correctly.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,40 +176,95 @@ def print_hz(self):
...
+                continue
+            rate, min_delta, max_delta, std_dev, window = hz_stat
+            stat = [topic, '{:.4}'.format(rate), '{:.4}'.format(min_delta),
+                    '{:.4}'.format(max_delta), '{:.4}'.format(std_dev),
+                    str(window)]
+            stats.append(stat)
+            for i, s in enumerate(stat):
+                col_width = len(s)
+                if col_width > max_col_width[i]:
+                    max_col_width[i] = col_width
```
dirk-thomas: These computation are all overly complicated. Since you have to manually iterate over all cells in a single column to determine the maximum width (including the header) I think it would make much more sense to store all cells of a column in a list (rather then the cells of a row). That way the majority of this code becomes obsolete.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,40 +176,95 @@ def print_hz(self):
...
+        if not stats:
+            print('no new messages')
+            continue
+        stats = sorted(stats, key=lambda x: x[1], reverse=True)
```
dirk-thomas: Does this sort the list of topics by their current rate? If yes, I don't think this is a good idea. This would result in a constantly changing order if the rates fluctuate. instead the order of the topics as passed in should be preserved.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1184,7 +1242,7 @@ def _rostopic_cmd_type(argv):
 def _rostopic_cmd_hz(argv):
     args = argv[2:]
     from optparse import OptionParser
-    parser = OptionParser(usage="usage: %prog hz /topic", prog=NAME)
+    parser = OptionParser(usage="usage: %prog hz /topic_0 [/topic_1] [topic_2] ..", prog=NAME)
```
dirk-thomas: The usage should be:

```
usage: %prog hz /topic_0 [/topic_1 [/topic_2 [..]]]
```

tools/rostopic/src/rostopic/__init__.py
wkentaro: I think it is left aligned. Am I missing something?
screen shot 2016-03-02 at 1 20 25 pm

wkentaro:
I updated the commit aside from above comment.

dirk-thomas:

dirk-thomas:
Jenkins fails to build this branch. Please make sure it is rebased to the latest devel branch to avoid potential problems.

dirk-thomas:
The test still times out.

wkentaro:
8db1799 should fix the problem.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -177,39 +180,93 @@ def print_hz(self):
...
-        print("average rate: %.3f\n\tmin: %.3fs max: %.3fs std dev: %.5fs window: %s"%(rate, min_delta, max_delta, std_dev, n+1))
+
+        return rate, min_delta, max_delta, std_dev, n+1
+
+    def print_hz(self):
+        """
+        print the average publishing rate to screen
+        """
+        if not self.times:
+            return
```
dirk-thomas, line 199: If `get_hz` would distinguish this case with a different return value this extra check would not be necessary.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -177,39 +180,93 @@ def print_hz(self):
...
+            col_widths.append(col_width)
+            header_aligned.append(h.center(col_width))
+            for i, el in enumerate(stats[h]):
+                stats[h][i] = stats[h][i].ljust(col_width)
+        # sum of col and each 3 spaces width
+        table_width = sum(col_widths) + 3 * (len(header) - 1)
+        body = '\n'.join('   '.join(stats[h][i] for h in header) for i in xrange(len(stats['topic'])))
+        table = '{header}\n{hline}\n{body}\n'.format(
+            header='   '.join(header_aligned), hline='=' * table_width, body=body)
+        print(table)
```
dirk-thomas: Should this be done by `print_hz` too? the function could have a verbose mode (used for printing a single topic) and a "single line" mode for this case.

Currently this printing logic is not accessible outside of this function and therefore also not easily testable.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1287,7 +1344,7 @@ def _rostopic_cmd_type(argv):
 def _rostopic_cmd_hz(argv):
     args = argv[2:]
     from optparse import OptionParser
-    parser = OptionParser(usage="usage: %prog hz /topic", prog=NAME)
+    parser = OptionParser(usage="usage: %prog hz /topic_0 [/topic_1 [[topic_2 [..]]]] ..", prog=NAME)
```
dirk-thomas: What are the trailing two dots for?

dirk-thomas:
Thanks for fixing the build and the tests. I added a few more comments inline but they should be fairly easy to address.

wkentaro:
Thanks, I applied changes according to your comments. Could you please check them again?

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1287,7 +1347,7 @@ def _rostopic_cmd_type(argv):
 def _rostopic_cmd_hz(argv):
     args = argv[2:]
     from optparse import OptionParser
-    parser = OptionParser(usage="usage: %prog hz /topic", prog=NAME)
+    parser = OptionParser(usage="usage: %prog hz [optioncs] /topic_0 [/topic_1 [[topic_2 [..]]]]", prog=NAME)
```
dirk-thomas: Spelling.

The previous question remains:

> What are the trailing two dots for?

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -112,10 +113,11 @@ def __init__(self, window_size, filter_expr=None, use_wtime=False):
             window_size = 50000
         self.window_size = window_size
                 
-    def callback_hz(self, m):
+    def callback_hz(self, m, topic):
```
dirk-thomas: This changes the public API of the class. Existing code using it would break.

tools/rostopic/src/rostopic/__init__.py
wkentaro: I mean `topic_0 [/topic_1 [[topic_2 [topic_3 [topic_4 [topic_5 [..]]]]]]]`

wkentaro:
Fixed spelling, and fixed the breaking public api with default value for argument.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1287,7 +1347,7 @@ def _rostopic_cmd_type(argv):
 def _rostopic_cmd_hz(argv):
     args = argv[2:]
     from optparse import OptionParser
-    parser = OptionParser(usage="usage: %prog hz /topic", prog=NAME)
+    parser = OptionParser(usage="usage: %prog hz [options] /topic_0 [/topic_1 [[topic_2 [..]]]]", prog=NAME)
```
dirk-thomas, line 1350: I will replace the double `[[` with a single `[` when cherry-picking the changes.

dirk-thomas:
Thank you for providing this enhancement. I have cherry-picked the changes to the _kinetic-devel_ branch: a6a4e57fbf6bdc8c7c05956282980cb53430b57a

wkentaro:
Thank you!

dirk-thomas:
I just run into a problem with this new feature when testing #874. If you pass two topic names but only one of them has any activity the command only prints "no new messages" and doesn't show the information about the topic which does have data.

@wkentaro Can you please look into this problem and make a PR for it (the tests should probably also cover that case).

wkentaro:
I think #886 fixes the problem, but not sure how to test it.
Currently the testing is disabled, and enabling it leads test fails.
https://github.com/ros/ros_comm/blob/kinetic-devel/tools/rostopic/test/check_rostopic_command_line_online.py#L116-L127

dirk-thomas:
Since this PR changes the public API of `ROSTopicHz` it breaks downstream code which uses the class. E.g. it breaks [rqt_topic](https://github.com/ros-visualization/rqt_common_plugins/blob/e9cbc6fce0b11037ad9db40a3d4346f45c250380/rqt_topic/src/rqt_topic/topic_info.py#L43). While this specific code can be updated to work with the old API (`list`) as well as the new API (`dict`) I am not sure if it is a good idea risking to break more downstream code. Maybe this PR (as well as all other ones building on top) should be reverted? Or this can be updated somehow to maintain backward compatibility?

wkentaro:
I agree that the code should be changed for the backward compatibility, but currently I don't have time to work on that, so it cannot be done soon.
So could you please revert this at this time, and open an issue for this? I will work for that later.
The better implementation seems using multiple `ROSTopicHz` class in `rostopic_hz` command and print hz with `get_hz` and `_get_ascii_table`, without changes on member variables of `ROSTopicHz`.

dirk-thomas:
Since we have already merged quite a bit of stuff on-top of this patch I rather spent time on actually fixing the API than trying to revert all the changes (knowing that I will have to reapply them in the near future again). Therefore I created #896 to restore API compatibility..

Issue #692, opened 2015-11-01
[roscpp] NodeHandle constructor with parent parameter does not work with a namespace with a leading slash


jim-rothrock:
In the examples below, model_nh.getNamespace() returns "/cerberus".

According to the NodeHandle constructor documentation,

```
ros::NodeHandle nh(model_nh, "/gazebo_ros_control/pid_gains/" + "arm_base_joint");
```

should be equivalent to

```
ros::NodeHandle nh(model_nh.getNamespace() + "/" + "/gazebo_ros_control/pid_gains/" + "arm_base_joint");
```

They are not. When using the second statement, nh.getNamespace() returns "/cerberus/gazebo_ros_control/pid_gains/arm_base_joint". When using the first statement, nh.getNamespace() returns "/gazebo_ros_control/pid_gains/arm_base_joint". Is the slash before "gazebo_ros_control" prohibited? The gazebo_ros_control package is using code similar to the first example, and no longer finds the PID parameters for my robot's joints. I think that this code worked as of June, 2015. I'll probably change "/gazebo_ros_control" to "gazebo_ros_control" in the gazebo_ros_control source code to fix the problem, but I wanted to point out this issue in case the NodeHandle constructor's current behavior is considered a bug. If it isn't, then the claim of constructor equivalence in the documentation is wrong.

dirk-thomas:
If the passed `namespace` argument is absolute the namespace of the parent node handle is _not_ prepended. I have created #715 to update the docblock to clarify that.

PR #715, opened 2015-12-09
improve NodeHandle constructor documentation

dirk-thomas:
Addresses #692.

### DIFF
--- a/clients/roscpp/include/ros/node_handle.h
+++ b/clients/roscpp/include/ros/node_handle.h
@@ -112,9 +112,15 @@ namespace ros
     /**
      * \brief Parent constructor
      *
-     * This version of the constructor takes a "parent" NodeHandle, and is equivalent to:
+     * This version of the constructor takes a "parent" NodeHandle.
+     * If the passed "ns" is relative (does not start with a slash), it is equivalent to calling:
      \verbatim
      NodeHandle child(parent.getNamespace() + "/" + ns);
+     \endverbatim
+     *
+     * If the passed "ns" is absolute (does start with a slash), it is equivalent to calling:
+     \verbatim
+     NodeHandle child(ns);
      \endverbatim
      *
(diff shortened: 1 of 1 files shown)


clients/roscpp/include/ros/node_handle.h
```diff
@@ -112,11 +112,17 @@ namespace ros
     /**
      * \brief Parent constructor
      *
-     * This version of the constructor takes a "parent" NodeHandle, and is equivalent to:
+     * This version of the constructor takes a "parent" NodeHandle.
+     * If the passes ns is relative (does not start with a slash), it is equivalent to calling:
```
wjwwood: "passed" not "passes" and should "ns" be expanded to "namespace" since "ns" can also mean "nanoseconds"?

clients/roscpp/include/ros/node_handle.h
```diff
@@ -112,11 +112,17 @@ namespace ros
...
      \verbatim
      NodeHandle child(parent.getNamespace() + "/" + ns);
      \endverbatim
      *
+     * If the passes ns is absolute (does start with a slash), it is equivalent to calling:
```
wjwwood: Same, "passed" not "passes".

wjwwood:
Other than small comments it makes sense to me.

clients/roscpp/include/ros/node_handle.h
dirk-thomas: Fixed the four locations of `passes`.

I would rather keep `ns` since it refers to the passed argument `ns`. `Namespace` could either mean `ns` or the namespace of the parent node handle and I think is therefore ambiguous.

dirk-thomas: I also quoted `ns` - that way it is similar to how the `parent` argument is referred to earlier.

PR #717, opened 2015-12-15
fix encoding of header fields (fix #704)

### LINKED ISSUE #704: Unicode Caller ID prevent receiving topic messages
I am currently investigating this, while diving into rospy code, so I have only a part of the information...

It seems that, whenever `rospy.names.get_name()` returns a `caller_id` that is a unicode string, it then goes into `rospy.impl.tcpros_pubsub.get_header_fields()` which forwards this to `tcpros_base.write_header()` and to `rosgraph.network.write_ros_handshake_header()`.
There `encode_ros_handshake_header` throw `'ascii' codec can't decode byte 0xc5 in position 0: ordinal not in range(128)` which is caught in `tcpros_base._tcp_server_callback` and the following message is displayed:

```
[WARN] [WallTime: 1448097800.943432] Inbound TCP/IP connection failed: 'ascii' codec can't decode byte 0xc5 in position 0: ordinal not in range(128)
```

And since the connection doesn't happen some communication breaks somewhere.
It took me a very long time to track this down because : 
- There is no information in the error message hinting at where things have gone wrong
- This doesn't break in the main node process, but in the topic subscriber thread, which means that to be able to catch it with a debugger, one needs to actually debug inside rospy itself.

Now here are my two questions :
- what can I do right now to make sure none of my node happen to set the caller_id with a unicode string ? I am currently clueless as to why that caller_id is unicode...
- what can we do to fix this / prevent the error in the long term ?

This happens for me now with https://github.com/asmodehn/rostful but the setup is a bit complicated, identifying the core issue with a simpler setup would be helpful.
I ll keep looking for more clues in the next few days...

dirk-thomas:
Fixes #704.

For Python 2 the field is explicitly converted to a string which usually means ASCII.

For Python 3 it actually fixes another bug: `len(f)` does not need to be equal to `len(f.encode('utf-8'))`. And in that case the length of the encoded data must be serialized.

### DIFF
--- a/tools/rosgraph/src/rosgraph/network.py
+++ b/tools/rosgraph/src/rosgraph/network.py
@@ -400,11 +400,13 @@ def encode_ros_handshake_header(header):
     # in the usual configuration, the error 'TypeError: can't concat bytes to str' appears:
     if python3 == 0:
         #python 2
+        fields = [str(f) for f in fields]
         s = ''.join(["%s%s"%(struct.pack('<I', len(f)), f) for f in fields])
         return struct.pack('<I', len(s)) + s
     else:
         #python 3 
-        s = b''.join([(struct.pack('<I', len(f)) + f.encode("utf-8")) for f in fields])
+        fields = [f.encode('utf-8') for f in fields]
+        s = b''.join([(struct.pack('<I', f) + f) for f in fields])
         return struct.pack('<I', len(s)) + s
                                         
 def write_ros_handshake_header(sock, header):


wjwwood:
+1

tools/rosgraph/src/rosgraph/network.py
```diff
@@ -400,11 +400,13 @@ def encode_ros_handshake_header(header):
     # in the usual configuration, the error 'TypeError: can't concat bytes to str' appears:
     if python3 == 0:
         #python 2
+        fields = [str(f) for f in fields]
```
eric-wieser, line 403: This is a no-op, because `f` is already a `str`, as it is the result of `"{}".format`

eric-wieser, line 403: ^ That was a lie

PR #718, opened 2015-12-15
[rostopic] Get parent topics in rostopic hz

wkentaro:
Parent topics means that topics to which the specified topic subscribes to
Depends on #712


dirk-thomas:
This pull request mixes several independent features which makes it unnecessary cumbersome to review. Please consider splitting it into separate PR.

dirk-thomas:
Never mind I just looked at #712. I will comment on that one first.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,40 +176,123 @@ def print_hz(self):
...
+        rate, min_delta, max_delta, std_dev, window = ret
+        print("average rate: %.3f\n\tmin: %.3fs max: %.3fs std dev: %.5fs window: %s"%(rate, min_delta, max_delta, std_dev, window))
 
 def _sleep(duration):
     rospy.rostime.wallsleep(duration)
 
-def _rostopic_hz(topic, window_size=-1, filter_expr=None):
+
+def get_parent_topics(topic, parent_topics=None, impl=False):
+    """Get topics to which the specified topic subscribes to.
```
dirk-thomas, line 245: Conceptionally a topic does not subscribe to another topic. Therefore I think the current naming is problematic. I don't have a good name for it though. But the docblock need to clarify what it means.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,40 +176,123 @@ def print_hz(self):
...
+
+    @type topic: ``str``
+    @param topic: topic name
+    @type parent_topics: ``NoneType`` or ``list``
+    @param parent_topics: list of parent topics which already got.
+    @type impl: ``bool``
+    @param impl: search parent topics recursively or not
```
dirk-thomas: The meaning of `impl` is unclear without consulting the docblock. The argument should be named `recursive` instead.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,40 +176,123 @@ def print_hz(self):
...
+    master = rosgraph.Master('/rostopic')
+    publications, subscriptions, _ = master.getSystemState()
+    if parent_topics is None:
+        parent_topics = []
+    pub_nodes = [l for t, l in publications if t == topic]
+    if not pub_nodes:
+        return parent_topics
+    for node in pub_nodes[0]:
+        subs = [t for t, l in subscriptions if node in l]
+        subs = filter(lambda x: x not in parent_topics, subs)  # get unique
```
dirk-thomas, line 263: The filter could be directly expressed in the list comprehension a line above.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -173,40 +176,123 @@ def print_hz(self):
...
+        parent_topics.extend(subs)
+        if not impl:
+            return parent_topics
+        for sub in subs:
+            parent_topics = get_parent_topics(sub, parent_topics, impl=impl)
```
dirk-thomas, line 268: Before calling this function recursively it should be checked that `sub` is not alread in `parent_topcis`. While that condition was true a few lines above after other recursive calls it might not anymore.

wkentaro:
Moved to #874

PR #722, opened 2015-12-28
rostest can load tests from a dotted name

### LINKED ISSUE #423: rostest: Allow test suites to be called in rostest.rosrun
See [this question](http://answers.ros.org/question/122786/any-way-to-use-a-testsuite-instead-a-testcase-with-rostest/).

The idea is to extend the functionality of `rostest.rosrun` so it can accept both `unittest.TestCase` and `unittest.TestSuite` in its parameters.

heuristicus:
This is a simple solution to the issue, using [`unittest.TestLoader().loadTestsFromName`](https://docs.python.org/2/library/unittest.html#unittest.TestLoader.loadTestsFromName). The dual-use (`str` or `unittest.TestCase`) of the `test` parameter could be changed to use a kwarg instead. That would necessitate `test` being an optional parameter which specifies the `unittest.TestCase` class, and some other parameter (`test_from_name` or something?) specifying a string from which tests should be loaded. That would probably require additional checks to ensure that at least one of the optional parameters existed.

You can check that the name-based loading works by changing [line 56](https://github.com/ros/ros_comm/blob/a0a45fddeb5ad94f2a657710d1db532d148bfe30/test/test_rosbag/test/latched_pub.py#L56) in `test_rosbag` to

```
rostest.rosrun('test_rosbag', 'latched_pub', 'latched_pub.LatchedPub', sys.argv)
```

Here's a test you can run to check that suite loading works correctly. You can run it with `rosrun test_rosbag tmp.py`. Don't forget to `chmod +x`.

`ros_comm/test/test_rosbag/test/tmp.py`:

``` python
#!/usr/bin/env python

import unittest
import rostest
import rospy
import sys

class CaseA(unittest.TestCase):

    def runTest(self):
        self.assertTrue(True)

class CaseB(unittest.TestCase):

    def runTest(self):
        self.assertTrue(True)

class SuiteTest(unittest.TestSuite):

    def __init__(self):
        super(SuiteTest, self).__init__()
        self.addTest(CaseA())
        self.addTest(CaseB())

if __name__ == '__main__':
      rostest.rosrun('test_rosbag', 'temp_tests', 'tmp.SuiteTest', sys.argv)

```

Fixes #423.

### DIFF
--- a/tools/rostest/src/rostest/__init__.py
+++ b/tools/rostest/src/rostest/__init__.py
@@ -109,8 +109,8 @@ def rosrun(package, test_name, test, sysargs=None):
     @type  package: str
     @param test_name: name of test that is being run
     @type  test_name: str
-    @param test: test class 
-    @type  test: unittest.TestCase
+    @param test: test class or name resolving to something returning TestCase or TestSuite
+    @type  test: unittest.TestCase, or string
     @param sysargs: command-line args. If not specified, this defaults to sys.argv. rostest
       will look for the --text and --gtest_output parameters
     @type  sysargs: list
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Can you please add a unit test for the new feature to this PR.

heuristicus:
I've added some tests, but couldn't really come up with a way of programmatically ensuring that the expected tests are actually loaded. I'm just assuming that `loadTestsFromName` will crash if you give it incorrect values for the dotted name. I'm not sure if there is a way to test expected failures, like giving a non-string, non-`unittest.TestCase` object to `rosrun` in the `test` parameter.

stonier:
Had a chat about this - test should probably be testing the function itself (nosetest?) rather than running a rostest and crashing rostest itself.

heuristicus:
I've moved the tests to using the python `unittest` framework, and added some test cases for expected failures.

Also squashed to a single commit.

dirk-thomas:

dirk-thomas:

dirk-thomas:

dirk-thomas:

dirk-thomas:

Jenkins is actually fine: http://build.ros.org/view/Jpr/job/Jpr__ros_comm__ubuntu_trusty_amd64/50/

I have cherry-picked the patch with the same modifications as ros/ros#104: 4c837a7f9cd72dc1c170754bf5231ad2628d4817

dirk-thomas:
Several callers pass types instead of instances of `unittest.TestCase`. Therefore I had to relax the check: e34122564c254252f201a778dd01466509bb325a

PR #724, opened 2015-12-30
[rostopic] Show stat when rostopic echo --noarr/nostr

wkentaro:
Should this be another option like `--statarr`?
Because this kind of function is not needed for string for me but for only array field.

screen shot 2015-12-31 at 3 17 50 am

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -619,7 +619,7 @@ def _sub_str_plot_fields(val, f, field_filter):
     return None
 
 
-def _str_plot(val, time_offset=None, current_time=None, field_filter=None, type_information=None, fixed_numeric_width=None):
+def _str_plot(val, time_offset=None, current_time=None, field_filter=None, type_information=None, fixed_numeric_width=None, value_transform_fn=None):
     """
     Convert value to matlab/octave-friendly CSV string representation.
 
@@ -627,6 +627,7 @@ def _str_plot(val, time_offset=None, current_time=None, field_filter=None, type_
     :param current_time: current :class:`genpy.Time` to use if message does not contain its own timestamp.
     :param time_offset: (optional) for time printed for message, print as offset against this :class:`genpy.Time`
     :param field_filter: filter the fields that are stringified for Messages, ``fn(Message)->iter(str)``
+    :param value_transform_fn: Not used but for same API as CallbackEcho.custom_strify_message
     :returns: comma-separated list of field values in val, ``str``
     """
         
@@ -1263,11 +1264,18 @@ def _rostopic_cmd_echo(argv):
     except ValueError:
         parser.error("NUM_WIDTH must be an integer")
 
-    value_transform_fn = create_value_transform(options.nostr, options.noarr)
+    if options.plot:
+        field_filter_fn = create_field_filter(options.nostr, options.noarr)
+        value_transform_fn = None
+    else:
+        field_filter_fn = None
+        value_transform_fn = create_value_transform(options.nostr, options.noarr)
+
     callback_echo = CallbackEcho(topic, None, plot=options.plot,
                                  filter_fn=filter_fn,
                                  echo_clear=options.clear, echo_all_topics=options.all_topics,
                                  offset_time=options.offset_time, count=msg_count,
+                                 field_filter_fn=field_filter_fn,
                                  value_transform_fn=value_transform_fn,
                                  fixed_numeric_width=fixed_numeric_width)
     try:
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -769,14 +769,9 @@ class CallbackEcho(object):
         # ensure to print uint8[] as array of numbers instead of string
         if type_information and type_information.startswith('uint8['):
             val = [ord(x) for x in val]
-        value_untransform_fn = None
         if value_transform is not None:
-            val, value_untransform_fn = value_transform(val)
+            val = value_transform(val)
         text = genpy.message.strify_message(val, indent=indent, time_offset=time_offset, current_time=current_time, field_filter=field_filter, fixed_numeric_width=fixed_numeric_width)
-        if value_untransform_fn is not None:
-            # this operation is necessary because value_transform does change the type of message
-            # and that causes not stringified values next time
-            val = value_untransform_fn(val)
         return text
 
     def callback(self, data, callback_args, current_time=None):
@@ -1285,35 +1280,35 @@ def _rostopic_cmd_echo(argv):
 
 def create_value_transform(echo_nostr, echo_noarr):
     def value_transform(val):
+
+        class TransformedMessage(genpy.Message):
+            # These should be copy because changing these variables
+            # in transforming is problematic without its untransforming.
+            __slots__ = val.__slots__[:]
+            _slot_types = val._slot_types[:]
+
+        val_trans = TransformedMessage()
+
         fields = val.__slots__
         field_types = val._slot_types
-        transformed = []
         for index, (f, t) in enumerate(zip(fields, field_types)):
-            f_val = val.__getattribute__(f)
+            f_val = getattr(val, f)
             if echo_noarr and '[' in t:
-                transformed.append((index, t, f, f_val))
-                val.__setattr__(f, '<array type: %s, length: %s>' %
+                setattr(val_trans, f, '<array type: %s, length: %s>' %
                                 (t.rstrip('[]'), len(f_val)))
-                val._slot_types[index] = 'string'
+                val_trans._slot_types[index] = 'string'
             elif echo_nostr and 'string' in t:
-                transformed.append((index, t, f, f_val))
-                val.__setattr__(f, '<string length: %s>' % len(f_val))
+                setattr(val_trans, f, '<string length: %s>' % len(f_val))
             else:
                 try:
                     msg_class = genpy.message.get_message_class(t)
                     if msg_class is None:
                         continue
-                    nested_transformed, _ = value_transform(f_val)
-                    val.__setattr__(f, nested_transformed)
-                    transformed.append((index, t, f, f_val))
+                    nested_transformed = value_transform(f_val)
+                    setattr(val_trans, f, nested_transformed)
                 except ValueError:
-                    pass
-        def untransform_fn(val):
-            for index, type_, f, f_val in transformed:
-                val._slot_types[index] = type_
-                val.__setattr__(f, f_val)
-            return val
-        return val, untransform_fn
+                    setattr(val_trans, f, f_val)
+        return val_trans
     return value_transform
 
 def create_field_filter(echo_nostr, echo_noarr):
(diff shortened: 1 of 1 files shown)


tools/rostopic/src/rostopic/__init__.py
```diff
@@ -657,11 +660,19 @@ def __init__(self, topic, msg_eval, plot=False, filter_fn=None,
...
         if type_information and type_information.startswith('uint8['):
             val = [ord(x) for x in val]
-        return genpy.message.strify_message(val, indent=indent, time_offset=time_offset, current_time=current_time, field_filter=field_filter, fixed_numeric_width=fixed_numeric_width)
+        if value_transform:
+            val, value_untransform_fn = value_transform(val)
+        text = genpy.message.strify_message(val, indent=indent, time_offset=time_offset, current_time=current_time, field_filter=field_filter, fixed_numeric_width=fixed_numeric_width)
+        if value_transform:
+            # this operation is necessary because value_transform does change the type of message
+            # and that causes not strified values next time
+            val = value_untransform_fn(val)
```
wjwwood: This might be easier to read like this:

``` python
value_untransform_fn = None
if value_transform:
    val, value_untransform_fn = value_transform(val)
text = genpy.message.strify_message(val, indent=indent, time_offset=time_offset, current_time=current_time, field_filter=field_filter, fixed_numeric_width=fixed_numeric_width)
if value_untransform_fn is not None:
    # this operation is necessary because value_transform does change the type of message
    # and that causes not strified values next time
    val = value_untransform_fn(val)
```

wjwwood:
Looks reasonable to me.

tools/rostopic/src/rostopic/__init__.py
wkentaro: Thanks, I updated the commit.

wkentaro:
Thanks for the feedback, I updated the commit.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -608,7 +608,8 @@ class CallbackEcho(object):
     def __init__(self, topic, msg_eval, plot=False, filter_fn=None,
                  echo_clear=False, echo_all_topics=False,
                  offset_time=False, count=None,
-                 field_filter_fn=None, fixed_numeric_width=None):
+                 field_filter_fn=None, value_transform_fn=None,
+                 fixed_numeric_width=None):
```
dirk-thomas: Please do not change the order of existing arguments but add new arguments at the end.

Same below.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -657,11 +660,20 @@ def __init__(self, topic, msg_eval, plot=False, filter_fn=None,
...
         if type_information and type_information.startswith('uint8['):
             val = [ord(x) for x in val]
-        return genpy.message.strify_message(val, indent=indent, time_offset=time_offset, current_time=current_time, field_filter=field_filter, fixed_numeric_width=fixed_numeric_width)
+        value_untransform_fn = None
+        if value_transform is not None:
+            val, value_untransform_fn = value_transform(val)
+        text = genpy.message.strify_message(val, indent=indent, time_offset=time_offset, current_time=current_time, field_filter=field_filter, fixed_numeric_width=fixed_numeric_width)
+        if value_untransform_fn is not None:
+            # this operation is necessary because value_transform does change the type of message
+            # and that causes not strified values next time
```
dirk-thomas: Spelling: `strified`

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1145,25 +1159,44 @@ def eval_fn(m):
...
+                val.__setattr__(f, '<array type: %s, length: %s>' %
+                                (t.rstrip('[]'), len(val.__getattribute__(f))))
+                index = fields.index(f)
+                val._slot_types[index] = 'string'
+                transformed_arrays.append((index, t))
             elif echo_nostr and 'string' in t:
-                continue
+                val.__setattr__(f, '<string length: %s>' % len(val.__getattribute__(f)))
+        def untransform_fn(val):
+            for index, type in transformed_arrays:
```
dirk-thomas: Please don't use reserved name like `type` for variable names.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1145,25 +1159,44 @@ def eval_fn(m):
...
+                val._slot_types[index] = type
+            return val
+        return val, untransform_fn
+    return value_transform
+
+def create_field_filter(echo_nostr, echo_noarr):
+    def field_filter(val):
+        fields = val.__slots__
+        field_types = val._slot_types
+        for f, t in zip(fields, field_types):
```
dirk-thomas: Why `zip` but then only use the first element?

dirk-thomas:
Please add tests for this new command.

dirk-thomas:
Please ignore the last comment about tests.

tools/rostopic/src/rostopic/__init__.py
wkentaro: `strified` is misspelling?

wjwwood: He probably thought you meant `stringified`, but you meant `str`-ified

wkentaro: I chose `strified` inspired by `custom_strify_message`. Is `stringified` better?

tfoote: It's still very much slang but `stringified` is much better. Without that it's hard to parse the base word from the made up ending, in my eyes I first see `strife` as the base and then try to make it a verb and past tense.

wkentaro:
I have updated `strified` -> `stringified`.

dirk-thomas:
Some of the unit tests are currently failing. Please see the referenced Jenkins builds.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -608,15 +608,18 @@ class CallbackEcho(object):
...
+                 value_transform_fn=None):
         """
         :param plot: if ``True``, echo in plotting-friendly format, ``bool``
         :param filter_fn: function that evaluates to ``True`` if message is to be echo'd, ``fn(topic, msg)``
         :param echo_all_topics: (optional) if ``True``, echo all messages in bag, ``bool``
         :param offset_time: (optional) if ``True``, display time as offset from current time, ``bool``
         :param count: number of messages to echo, ``None`` for infinite, ``int``
-        :param field_filter_fn: filter the fields that are strified for Messages, ``fn(Message)->iter(str)``
+        :param field_filter_fn: filter the fields that are stringified for Messages, ``fn(Message)->iter(str)``
+        :param value_transform_fn: transform the values of Messages, ``fn(Message)->Message``
```
dirk-thomas: Duplicate of line 622.

wkentaro: thanks, I have fixed.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1145,25 +1161,46 @@ def eval_fn(m):
...
+                val.__setattr__(f, '<string length: %s>' % len(f_val))
+        def untransform_fn(val):
+            for index, type_, f, f_val in transformed:
+                val._slot_types[index] = type_
+                val.__setattr__(f, f_val)
+            return val
+        return val, untransform_fn
+    return value_transform
+
+def create_field_filter(echo_nostr, echo_noarr):
```
dirk-thomas, line 1314: I am not sure if this function is used anywhere outside by now it doesn't filter anything anymore.

wkentaro:
Removed the field_filter function.

dirk-thomas:
Since you have removed all unit tests for `create_field_filter` are you planning to readd them for the new function?

wkentaro:
I'm not planning to readd them, because filtering field it not supported after this PR.
Aside from actually removing(filtering) the fields (string/array) this PR shows the stats for them.

wkentaro:
Is this solution OK? I removed the `create_field_filter` function which will be no longer used.

dirk-thomas:
`create_field_filter` is a public function of this Python package. While it might not be used within this file anymore other packages might still be using it and would break if it gets removed. So I don't think it should be removed since such a breakage is not necessary.

wkentaro:
@dirk-thomas 
OK, I restored the `create_field_filter` function.

And also, I found a bug of this code with nested messages as below.
It is fixed now with the last commit.
**BEFORE**

```
% rostopic echo /raw_image_bgr/image_color --nostr --noarr
header: 
  seq: 48
  stamp: 
    secs: 1463637710
    nsecs: 575849056
  frame_id: camera
height: 1080
width: 1920
encoding: <string length: 4>
is_bigendian: 0
step: 5760
data: <array type: uint8, length: 6220800>
```

**AFTER**

```
% rostopic echo /raw_image_bgr/image_color --nostr --noarr
header: 
  seq: 46
  stamp: 
    secs: 1463637665
    nsecs: 575885057
  frame_id: <string length: 6>
height: 1080
width: 1920
encoding: <string length: 4>
is_bigendian: 0
step: 5760
data: <array type: uint8, length: 6220800>
```

wkentaro:
@ros-pull-request-builder rostest this please

wkentaro:
Test passed. Please review again when you have time.

dirk-thomas:
The patch looks good. But I was thinking if the effort for transforming and then untransforming the message is necessary. Wouldn't it be easier to just copy the message and update the copied message? For efficiency you could also implement a custom copy operation which already avoids to copy the string / array fields in the first place?

wkentaro:
> The patch looks good. But I was thinking if the effort for transforming and then untransforming the message is necessary. Wouldn't it be easier to just copy the message and update the copied message? For efficiency you could also implement a custom copy operation which already avoids to copy the string / array fields in the first place?

Thanks for the suggestion. I worked around for a while, and the problem when the transformed message was untransformed will be fixed by correct `_field_type` of the message class. So I created `TransformedMessage` as the dummy class for printing.

wkentaro:
ping

wkentaro:
ping

dirk-thomas:
I am sorry for the late response but I currently have to focus on a different project and don't have much time to review ROS tickets.

The patch looks good now. Thank you for iterating on it. I cherry-picked the changed to the _kinetic-devel_ branch with two small cosmetic changes: e0f63978fe0dd239ba8002a7e79756cdf1a335d0

```
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -831,8 +831,7 @@ class CallbackEcho(object):
             val = [ord(x) for x in val]
         if value_transform is not None:
             val = value_transform(val)
-        text = genpy.message.strify_message(val, indent=indent, time_offset=time_offset, current_time=current_time, field_filter=field_filter, fixed_numeric_width=fixed_numeric_width)
-        return text
+        return genpy.message.strify_message(val, indent=indent, time_offset=time_offset, current_time=current_time, field_filter=field_filter, fixed_numeric_width=fixed_numeric_width)

     def callback(self, data, callback_args, current_time=None):
         """
@@ -1355,7 +1354,7 @@ def create_value_transform(echo_nostr, echo_noarr):
             f_val = getattr(val, f)
             if echo_noarr and '[' in t:
                 setattr(val_trans, f, '<array type: %s, length: %s>' %
-                                (t.rstrip('[]'), len(f_val)))
+                                      (t.rstrip('[]'), len(f_val)))
                 val_trans._slot_types[index] = 'string'
             elif echo_nostr and 'string' in t:
                 setattr(val_trans, f, '<string length: %s>' % len(f_val))
```

wkentaro:
Thanks!

PR #727, opened 2016-01-14
Include GTEST_INCLUDE_DIRS

v-lopez:
This makes sure that the proper gtest headers are found, instead of the ones in the standard header path.

Required for the fixes proposed in https://github.com/ros/catkin/issues/699


dirk-thomas:

dirk-thomas:
Thank you for the fix.

PR #730, opened 2016-01-19
do not fail if other package.xml provides package required

k-okada:
roslaunch-check fails for heqtor_quadrotor_gazebo

```
$ rosrun roslaunch roslaunch-check  /opt/ros/indigo/share/hector_quadrotor_gazebo/launch/quadrotor_empty_world.launch 
checking /opt/ros/indigo/share/hector_quadrotor_gazebo/launch/quadrotor_empty_world.launch
...writing test results to /home/k-okada/.ros/test_results/hector_quadrotor_gazebo/rosunit-roslaunch_check_quadrotor_empty_world_launch.xml
FAILURE:
[/opt/ros/indigo/share/hector_quadrotor_gazebo/launch/quadrotor_empty_world.launch]:
    Missing package dependencies: hector_quadrotor_gazebo/package.xml: gazebo_ros, hector_quadrotor_controller
Missing package dependencies: hector_quadrotor_controller/package.xml: controller_manager
```

but it has dependency like

```
$ rospack depends-why --target gazebo_ros hector_quadrotor_gazebo 
Dependency chains from hector_quadrotor_gazebo to gazebo_ros:
* hector_quadrotor_gazebo -> gazebo_plugins -> gazebo_ros 
* hector_quadrotor_gazebo -> hector_gazebo_plugins -> gazebo_ros 
* hector_quadrotor_gazebo -> hector_sensors_gazebo -> gazebo_plugins -> gazebo_ros 
* hector_quadrotor_gazebo -> hector_sensors_gazebo -> hector_gazebo_plugins -> gazebo_ros 
* hector_quadrotor_gazebo -> hector_quadrotor_gazebo_plugins -> gazebo_ros 
* hector_quadrotor_gazebo -> hector_quadrotor_gazebo_plugins -> hector_gazebo_plugins -> gazebo_ros 
* hector_quadrotor_gazebo -> hector_quadrotor_controller_gazebo -> gazebo_ros_control -> gazebo_ros
```

### DIFF
--- a/tools/roslaunch/src/roslaunch/rlutil.py
+++ b/tools/roslaunch/src/roslaunch/rlutil.py
@@ -193,6 +193,7 @@ def check_roslaunch(f):
     except roslaunch.core.RLException as e:
         return str(e)
     
+    rospack = rospkg.RosPack()
     errors = []
     # check for missing deps
     try:
@@ -206,8 +207,21 @@ def check_roslaunch(f):
         missing = {}
         file_deps = {}
     for pkg, miss in missing.items():
-        if miss:
+        # even if the pkgs is not found in packges.xml, if other package.xml provdes that pkgs, then it will be ok
+        all_pkgs = []
+        try:
+            for file_dep in file_deps.keys():
+                file_pkg = rospkg.get_package_name(file_dep)
+                all_pkgs.extend(rospack.get_depends(file_pkg,implicit=False))
+                miss_all = list(set(miss) - set(all_pkgs))
+        except Exception as e:
+            print(e, file=sys.stderr)
+            miss_all = True
+        if miss_all:
+            print("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)), file=sys.stderr)
             errors.append("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)))
+        elif miss:
+            print("Missing package dependencies: %s/package.xml: %s (notify upstream maintainer)"%(pkg, ', '.join(miss)), file=sys.stderr)
     
     # load all node defs
     nodes = []
@@ -215,7 +229,6 @@ def check_roslaunch(f):
         nodes.extend(rldeps.nodes)
 
     # check for missing packages
-    rospack = rospkg.RosPack()
     for pkg, node_type in nodes:
         try:
             rospack.get_path(pkg)


tfoote:
Can you clarify which cases this is covering? It sounds like your're allowing launch files with transitive dependencies. If the files are referenced locally I would consider that correct behavior, since if the transitive dependency gets removed then the launch file will break. If you're transitively using the launch files through other launch files it should be ok.

k-okada:
to reproduce this
1.  create package

```
catkin_create_pkg hoge gazebo_ros hector_quadrotor_gazebo gazebo_ros hector_quadrotor_controller
```
1. put launch file

```
$ cat hoge/hoge.launch 
<launch>
  <include file='$(find hector_quadrotor_gazebo)/launch/quadrotor_empty_world.launch' />
</launch>
```
1. run roslaunch-check

```
k-okada@kokada-t440s:/tmp/aaa/src$ rosrun roslaunch roslaunch-check hoge/hoge.launch
checking hoge/hoge.launch
^P...writing test results to /home/k-okada/.ros/test_results/hoge/rosunit-roslaunch_check_hoge_launch.xml
FAILURE:
[hoge/hoge.launch]:
    Missing package dependencies: hector_quadrotor_gazebo/package.xml: gazebo_ros, hector_quadrotor_controller
Missing package dependencies: hector_quadrotor_controller/package.xml: controller_manager
wrote test file to [/home/k-okada/.ros/test_results/hoge/rosunit-roslaunch_check_hoge_launch.xml]
```

it is true that `hector_quadrotor_gazebo` does not have correct dependency for `gazebo_ros`, but  if you look this from `hoge` package, it has dependency to the `gazebo_ros` so `hoge` package itself works. It is very painful to track all upstream package to correct dependency.

```
k-okada@kokada-t440s:/tmp/aaa/src$ rospack depends hoge | grep gazebo
gazebo_msgs
gazebo_ros
gazebo_plugins
hector_gazebo_plugins
hector_sensors_gazebo
hector_quadrotor_gazebo_plugins
gazebo_ros_control
hector_quadrotor_controller_gazebo
hector_quadrotor_gazebo
```

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -207,6 +207,10 @@ def check_roslaunch(f):
         file_deps = {}
     for pkg, miss in missing.items():
         if miss:
+            print("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)))
```
dirk-thomas: Why should this message be still printed? I thought the point was that this is not actually a problem?

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -207,6 +207,10 @@ def check_roslaunch(f):
...
+        # even if the pkgs is not found in packges.xml, if other package.xml provdes that pkgs, then it will be ok                                                                                           
+        miss_all = list(set(miss) - set(reduce(lambda x, y: x+y, [file_deps[pkg].pkgs for pkg in file_deps.keys()])))
```
dirk-thomas: The logic for the lambda looks overly complicated. Please try to simplify the expression.

dirk-thomas:
The CI build for this PR reports numerous failing tests. Please look into them and resolve them.

tools/roslaunch/src/roslaunch/rlutil.py
k-okada: I think missing package dependencies should be fixed, the point is upstream package dependencies can not be fixed by downstream  maintainers, so test should not be fail. But printing this is useful for downstream maintainers to notify upstream maintainers.

k-okada:
```
            print("Missing package dependencies: %s/package.xml: %s (notify upstream maintainer)"%(pkg, ', '.join(miss)))
```

added message.

tools/roslaunch/src/roslaunch/rlutil.py
k-okada: ok, fixed the logic; https://github.com/ros/ros_comm/pull/730/files#diff-b2e93710e71874c9ea655b6b0644f4c5R210

k-okada:
OK fixed the patch and now test passed, please review.

dirk-thomas:
I don't think this patch achieve the goal at all. From my point of view that check for one launch file should ensure that the containing package contains the necessary dependencies to everything which is being directly using within this launch file. Recursively  included launch files should not be checked since they should be covered by the dependencies of the package containing that launch file.

Currently `file_deps` returned by `roslaunch.depends.roslaunch_deps` already contains recursive information. The patched code has no way to distinguish between direct file dependencies and transitive file dependencies. Hence it can't warn about the first case while being quiet about the second.

k-okada:
Current implementation fails if upstream launch file did not contains the
necessary dependencies, even if my package depends on that package, for
example current fetch_navigation fails to include necessary dependencies
(see  https://github.com/fetchrobotics/fetch_ros/pull/35)
and if you have package that depends on fetch_navigation it fails as

```
$ rosrun roslaunch roslaunch-check  fetch_bringup.launch
checking fetch_bringup.launch
...writing test results to
/home/k-okada/.ros/test_results/jsk_fetch_startup/rosunit-roslaunch_check_fetch_bringup_launch.xml
FAILURE:
[fetch_bringup.launch]:
Missing package dependencies: fetch_navigation/package.xml: move_base,
map_server, amcl
wrote test file to
[/home/k-okada/.ros/test_results/jsk_fetch_startup/rosunit-roslaunch_check_fetch_bringup_launch.xml]
```

This mean if we want to add roslaunch_add_file_check to our CMakeLists.txt,
we have to wait upstream maintainer for fixing this problem.

with this PR,

```
k-okada@kokada-t440s:~/catkin_ws/ws_fetch/src/jsk_robot/jsk_fetch_robot/jsk_fetch_startup/launch$
rosrun roslaunch roslaunch-check  fetch_bringup.launch
checking fetch_bringup.launch
Missing package dependencies: fetch_navigation/package.xml: move_base,
map_server, amcl (notify upstream maintainer)
...writing test results to
/home/k-okada/.ros/test_results/jsk_fetch_startup/rosunit-roslaunch_check_fetch_bringup_launch.xml
passed
k-okada@kokada-t440s:~/catkin_ws/ws_fetch/src/jsk_robot/jsk_fetch_robot/jsk_fetch_startup/launch$
cat ../package.xml
<package>
  <name>jsk_fetch_startup</name>
  <version>1.0.3</version>
  <description>
     jsk_fetch_startup

     data/boost_sound.wav is downloaded from
https://www.youtube.com/watch?v=HM-xG0qXaeA
  </description>
  <maintainer email="k-okada@jsk.t.u-tokyo.ac.jp">Kei Okada</maintainer>
  <license>BSD</license>
  <url>http://ros.org/wiki/jsk_fetch_startup</url>
  <buildtool_depend>catkin</buildtool_depend>

  <run_depend>jsk_robot_startup</run_depend>
  <run_depend>jsk_maps</run_depend>
  <run_depend>jsk_pr2_startup</run_depend>
  <run_depend>fetch_navigation</run_depend>
  <run_depend>rviz</run_depend>
  <run_depend>tf2_ros</run_depend>

  <!-- for fetch_naviation -->
  <run_depend>amcl</run_depend>
  <run_depend>map_server</run_depend>
  <run_depend>move_base</run_depend>

  <test_depend>rostest</test_depend>

  <export>
  </export>
</package>

```

## 

◉ Kei Okada

dirk-thomas:
So your test would pass but it wouldn't actually work? I am not sure if that is a useful test.

k-okada:
No

1) both an upstream and my package has correct dependency
2) an upstream package does not have correct dependency and my package does not include missing dependency
3) an upstream package does not have correct dependency , but my package covers missing dependency

In current implementation, 1 passes the test, but 2 and 3 fails
this PR, both.1,3 pass the test, and only 2 fails

## 

◉ Kei Okada

2016/03/09 10:04、Dirk Thomas notifications@github.com のメッセージ:

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -206,8 +207,21 @@ def check_roslaunch(f):
...
-        if miss:
+        # even if the pkgs is not found in packges.xml, if other package.xml provdes that pkgs, then it will be ok
+        all_pkgs = []
+        try:
+            for file_dep in file_deps.keys():
+                file_pkg = rospkg.get_package_name(file_dep)
+                all_pkgs.extend(rospack.get_depends(file_pkg,implicit=False))
+                miss_all = list(set(miss) - set(all_pkgs))
+        except Exception as e:
+            print(e)
```
dirk-thomas: Since this is an error message please print it to `stderr` using:

```
print(e, file=sys.stderr)
```

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -206,8 +207,21 @@ def check_roslaunch(f):
...
+            miss_all = True
+        if miss_all:
+            print("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)))
```
dirk-thomas: Same here.

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -206,8 +207,21 @@ def check_roslaunch(f):
...
             errors.append("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)))
+        elif miss:
+            print("Missing package dependencies: %s/package.xml: %s (notify upstream maintainer)"%(pkg, ', '.join(miss)))
```
dirk-thomas: While this message isn't failing the test it is still an error message. I would think it should go to `stderr` (but could be convinced otherwise).

dirk-thomas:
Without a clear example to reproduce and follow what you are running it is really difficult to understand current and new expected behavior. I just had a very hard time following this.

Anyway I have tried the cases you mentioned and the behavior with this patch is for sure better then without. I have only small comment left regarding `print. If you could address that one and squash I am happy to merge this contribution.

k-okada:
ok fixed the patch and squashed, 

and added example https://github.com/k-okada/roslaunch_730, which is basically follow the instructions in  -> https://github.com/ros/ros_comm/pull/730#issuecomment-173428731
if you run `catkin run_test` and `catkin_test_results` you'll see both package fails without this PR

dirk-thomas:

Issue #593, opened 2015-04-02
roslaunch: Add additional conditional arguments ifempty/unlessempty


piyushk:
I've often found that I want a different behavior if an argument is empty or not. For instance, 

```
<arg name="tf_prefix" default="" />
<arg unlessempty="$(arg tf_prefix)" name="rgb_frame_id" default="$(arg tf_prefix)/$(arg camera)_rgb_optical_frame" />
<arg ifempty="$(arg tf_prefix)" name="rgb_frame_id" default="$(arg camera)_rgb_optical_frame" />
```
1. Is this a good idea? Is there a way I can achieve this with the current roslaunch xml syntax?
2. If I submit a PR with this feature, is it too late to be merged into Indigo?

gavanderhoorn:
Is this similar to #540?

piyushk:
@gavanderhoorn Looks like it. I'll follow up on that issue report.

PR #731, opened 2016-01-21
Added six comparison substitution arguments

bponsler:
Added six new substitution arguments:

```
- eq: compare an arg value to a given value for equality, e.g., $(eq arg value)
- neq: compare an arg value to a given value for inequality, e.g., $(neq arg value)
- eqarg: compare two arg values for equality, e.g., $(eqarg arg1 arg2)
- neqarg: compare two arg values for inequality, e.g., $(neqarg arg1 arg2)
- empty: check if the value of an arg is an empty string, e.g., $(empty arg)
- notempty: check if the value of an arg is not an empty string, e.g., $(notempty arg)
```

Each of the above substitution arguments evaluates to 'true' or 'false' making it possible to perform more advanced if/unless checking, e.g.:

``` xml
    <arg name="robot_type" default="mobile" />
    <include if="$(eq robot_type mobile)" file="$(find my_package)/launch/my_mobile_robot.launch" />
```

The "my_mobile_robot.launch" file would only get included when the "robot_type" argument equals "mobile" otherwise it will not get included.

This includes the changes in PR #540 (to add empty) which hasn't been updated in almost a year. Given that all of the related PRs (#296, #593, and #540) haven't been merged I imagine the same will happen with this PR. I'd be willing to make changes, and or add other operators -- just let me know.

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -89,7 +89,7 @@ def _optenv(resolved, a, args, context):
         return resolved.replace("$(%s)"%a, ' '.join(args[1:]))
     else:
         return resolved.replace("$(%s)"%a, '')
-    
+
 def _anon(resolved, a, args, context):
     """
     process $(anon) arg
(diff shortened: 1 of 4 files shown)


dirk-thomas:
These six special operations are all covered by the generic execution of logic covered by #707. Therefore I will close this.

Issue #709, opened 2015-12-01
reenable rosmsg tests


dirk-thomas:
Current the tests in the `rosmsg` are not being used: https://github.com/ros/ros_comm/tree/indigo-devel/tools/rosmsg/test

gerkey:
Mostly done in ea7cf134b8c429ffacf8cac84e374207ec8e5220.

There are a few remaining failures, which might be actual problems. I'll keep looking into it.

gerkey:
FYI, here are the remaining failures:

```
======================================================================
FAIL: test_get_msg_text (test_rosmsg.TestRosmsg)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gerkey/code/rosmsg-tests/src/ros_comm/tools/rosmsg/test/test_rosmsg.py", line 74, in test_get_msg_text
    self.assertEquals(text, rosmsg.get_msg_text(type_, raw=True))
AssertionError: 'test_rosmaster/Empty empty\n' != 'Empty empty\n'

======================================================================
FAIL: test_get_srv_text (test_rosmsg.TestRosmsg)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gerkey/code/rosmsg-tests/src/ros_comm/tools/rosmsg/test/test_rosmsg.py", line 136, in test_get_srv_text
    self.assertEquals(text, rosmsg.get_srv_text('test_rosmaster/RossrvB', raw=True))
AssertionError: 'test_rosmaster/Empty empty\n---\ntest_rosmaster/Empty empty\n' != 'Empty empty\n---\nEmpty empty\n'

======================================================================
FAIL: test_cmd_show (test_rosmsg_command_line.TestRosmsg)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/gerkey/code/rosmsg-tests/src/ros_comm/tools/rosmsg/test/test_rosmsg_command_line.py", line 156, in test_cmd_show
    self.assertEquals(text, output)
AssertionError: 'Empty empty\n\n' != 'test_rosmaster/Empty empty\n\n'

----------------------------------------------------------------------
```

dhood:
I've been looking into this to see if I could help - 

have opened https://github.com/ros/ros_comm/pull/734 with one way of fixing remaining failures

dirk-thomas:
Addressed by PR #751. Thank you.

PR #734, opened 2016-01-29
Fix remaining rosmsg tests

dhood:
Building on work related to issue https://github.com/ros/ros_comm/issues/709

```
FAIL: test_get_msg_text (test_rosmsg.TestRosmsg)
  File "/home/gerkey/code/rosmsg-tests/src/ros_comm/tools/rosmsg/test/test_rosmsg.py", line 74, in test_get_msg_text
    self.assertEquals(text, rosmsg.get_msg_text(type_, raw=True))
AssertionError: 'test_rosmaster/Empty empty\n' != 'Empty empty\n'
```

From what I can see, rosmsg.get_msg_text() used to have a 'fulltext' option which would refer to the msg types with their package name included ('test_rosmaster/Empty empty\n' vs 'Empty empty\n', as in the error output). This was removed in https://github.com/ros/ros_comm/commit/ef2beb715ed502d65e8a0c4e2149d9437e4d2483, but now having the parameter raw=False returns msgs in fulltext, [while raw=True doesn't](https://github.com/ros/ros_comm/commit/ef2beb715ed502d65e8a0c4e2149d9437e4d2483#diff-88c107efa9221d2911e1ca87e7246869R252). 

The tests are failing because they don't expect this discrepancy.

The current behaviour is:
With raw=False, messages are: recursive, whitespace removed, fulltext.
with raw=True, messages are not recursive, with whitespace not removed, and not fulltext.

I updated the tests to match the current behaviour of get_msg_text/get_srv_text, but it's possible that the appropriate fix is actually the other way around, as, judging from [this comment](https://github.com/ros/ros_comm/blob/indigo-devel/tools/rosmsg/test/test_rosmsg_command_line.py#L149) in the previously failing test, the behaviour when raw=False vs True used to be different (with respect to fulltext).

### DIFF
--- a/tools/rosmsg/test/test_rosmsg.py
+++ b/tools/rosmsg/test/test_rosmsg.py
@@ -133,16 +133,19 @@ test_rosmaster/String s2
     def test_get_srv_text(self):
         d = get_test_path()
         srv_d = os.path.join(d, 'srv')
-        with open(os.path.join(srv_d, 'RossrvA.srv'), 'r') as f:
-            text = f.read()
-        self.assertEquals(text, rosmsg.get_srv_text('test_rosmaster/RossrvA', raw=False))
-        self.assertEquals(text, rosmsg.get_srv_text('test_rosmaster/RossrvA', raw=True))
-
-        # std_msgs/empty / std_msgs/empty
-        with open(os.path.join(srv_d, 'RossrvB.srv'), 'r') as f:
-            text = f.read()
-        self.assertEquals(text, rosmsg.get_srv_text('test_rosmaster/RossrvB', raw=False))
-        self.assertEquals(text, rosmsg.get_srv_text('test_rosmaster/RossrvB', raw=True))
+        
+        test_srv_package = 'test_rosmaster'
+        rospack = rospkg.RosPack()
+        srv_raw_d = os.path.join(rospack.get_path(test_srv_package), 'srv')
+        for t in ['RossrvA', 'RossrvB']:
+            with open(os.path.join(srv_d, '%s.srv'%t), 'r') as f:
+                text = f.read()
+            with open(os.path.join(srv_raw_d, '%s.srv'%t), 'r') as f:
+                text_raw = f.read()
+                
+            type_ = test_srv_package+'/'+t
+            self.assertEquals(text, rosmsg.get_srv_text(type_, raw=False))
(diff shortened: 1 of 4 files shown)


tools/rosmsg/test/test_rosmsg.py
```diff
@@ -66,22 +68,29 @@ def test_fullusage(self):
...
         t = 'RosmsgC'
-        with open(os.path.join(msg_d, '%s.msg'%t), 'r') as f:
-            text = f.read()
+        with open(os.path.join(msg_raw_d, '%s.msg'%t), 'r') as f:
+            text_raw = f.read()
         type_ = 'test_rosmaster/'+t
-        self.assertEquals(text, rosmsg.get_msg_text(type_, raw=True))
-        self.assertEquals("""std_msgs/String s1
+        self.assertEquals(text_raw, rosmsg.get_msg_text(type_, raw=True))
+        self.assertEquals("""test_rosmaster/String s1
```
jackieokay: Consider using the `test_message_package` variable to format this string instead of the raw string here and elsewhere.

jackieokay:
Ran tests manually and the rosmsg tests pass.

This was the output of `catkin_test_results`:

```
roswtf/nosetests-test.xml: 3 tests, 0 errors, 1 failures
roswtf/rosunit-roswtf_command_line_online.xml: 2 tests, 0 errors, 1 failures
test_rosbag/rosunit-r02random_play.xml: 1 tests, 0 errors, 1 failures
test_rosbag/rosunit-random_play05.xml: 1 tests, 0 errors, 1 failures
test_rospy/rosunit-sub_to_multiple_pubs.xml: 1 tests, 0 errors, 1 failures
```

On origin/enable_tests, rosmsg has 3 failures.

Changing from `std_msgs/String` to `test_rosmaster/String` also looks good to me--why doesn't `RosmsgC.msg` use a raw `string` type? Is it important that we test a nested message here for some reason?

I'm not a maintainer on this package, will defer to @gerkey and @dirk-thomas.

dhood:
Thanks for running the tests @jacquelinekay 
Yeah, RosmsgC.msg is being used to test the recursive printing of message types.

gerkey:
Thanks for the fixes!

Given how long ago the behavior change was made, I'd prefer to stick with the current behavior, however imperfect it might be. It's more dangerous to "fix" the behavior to match the old tests, which could break somebody's system.

PR #735, opened 2016-02-01
Fixed a typo in unknown host error message

bponsler:
This fixes a typo when generating the error message when a machine's address was not found in the known_hosts file. The port string and the user string were reversed so if a username and port were provided it would generate an error like this:

```
Please manually:
   ssh username@-p 26 127.0.0.1
```

Rather than the correct version:

```
Please manually:
   ssh -p 26 username@127.0.0.1
```

The first is not a valid ssh command, while the second is.

### DIFF
--- a/tools/roslaunch/src/roslaunch/remoteprocess.py
+++ b/tools/roslaunch/src/roslaunch/remoteprocess.py
@@ -120,8 +120,8 @@ Please manually:
 then try roslaunching again.
 
 If you wish to configure roslaunch to automatically recognize unknown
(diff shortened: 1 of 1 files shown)


wjwwood:
@bponsler thanks for the pull request! Can you please update the commit to avoid touching whitespace? I know it's ugly, but it's better for forward/back porting if your change does not contain unnecessary diffs.

bponsler:
No problem, I reverted the whitespace changes for you.

dirk-thomas:
Thank you for the fix. I have squashed your commits into a single one: 8e24f232b0ee2e456c1bb71c46a7b19b7daf919c

Issue #706, opened 2015-11-25
Latched Topics in Split Bags


DLu:
I'm trying to record TF data across split bags. (`rosbag record --split --duration=5s /tf /tf_static`) The /tf_static topic is only recorded in the first bag. 

Is there anyway to force rosbag to record latched topics in each bag?

dirk-thomas:
There is currently no such option.

wjwwood:
I outlined some possible solutions in https://github.com/ros/ros_comm/issues/805#issue-155150413

JanezCim:
Just an FYI for anyone struggling, using the command

`rosbag play 0.bag 1.bag 2.bag`

will keep all the static transforms correctly for the whole duration of all the split bag files combined. The problem becomes when you try to play them back separately. But afaik this is being solved in Noetic.

chaoflow:
This has been implemented with #1850 included in noetic. I think the issue can be closed.

DLu:
Thanks @tappan-at-git !

Issue #608, opened 2015-04-21
ROS service interface for Rosbag


muhrix:
Hi all,

About a year ago I modified the `rosbag` package to act as a ROS service provider for local/remote logging control. That way I could start/stop/start-over the recording of messages being published via simple ROS service calls.

I believe this is a welcome feature to `rosbag`, and it is related to issue #428.

I have noticed there were changes to the `rosbag` source code since then. I had to make significant changes to the code in order to handle the queue of incoming messages and the fact that, once I stopped recording, the node would not shut down (service provider) and hence cause memory leaks when a new recording started.

Before I go into greater detail, I would like to ask what would be best in preparation for a release:
1. Release a separate package, with a different name
2. Release within `rosbag` an extra node which implements the service functionality
3. Release within `rosbag` and extend the current node with an optional parameter to start rosbag as a service

I will be tidying up the code and bringing it up to date with the latest changes, then make it publicly available soon-ish, but I would like to discuss the best approach so that a package with this functionality can be released.

@tfoote @dirk-thomas and everybody else, could you please share your points of view?

Many thanks!

tfoote:
I think that providing a separate node makes sense. For initial development it makes sense to use a separate package. But assuming it can be done without adding dependencies or conflicting names, promoting it into the rosbag package seams a reasonable goal.

muhrix:
I'll initially work on a separate package then. Thanks!

tfoote:
After thinking about this a bit, it might be even more powerful/flexible to design this to work as a nodelet. It could be launched in the same process to take advantage of the optimized intraprocess communications and possibly avoid network communications of the data entirely.

muhrix:
Do you mean attaching to existing publishers?
Wouldn't this assume publishers are implemented as nodelets too? (or am I missing something?)

In any case, both node and nodelet versions could be provided.

tfoote:
Yes it would only work if the publisher is a nodelet. But if implemented as a nodelet, it can be run as a standalone nodelet to provide the equivalent of a node without needing a parallel version.

DLu:
+1 
I would also like the functionality described in the original issue. Nodelet vs new node doesn't matter.

dirk-thomas:
Recently a rosbag nodelet has been developed which offers an action-based interface: https://github.com/osrf/nodelet_rosbag

Does this cover the requested feature and this ticket can be closed?

DLu:
Is there any documentation for this the nodelet_rosbag?

[...]

PR #737, opened 2016-02-04
Rosbag Recorder Private-->Protected

DLu:
To enable subclassing of rosbag record behavior.

### DIFF
--- a/tools/rosbag/include/rosbag/recorder.h
+++ b/tools/rosbag/include/rosbag/recorder.h
@@ -125,7 +125,7 @@ public:
 
     int run();
 
-private:
+protected:
     void printUsage();
 
     void updateFilenames();
@@ -150,7 +150,6 @@ private:
     template<class T>
     static std::string timeToStr(T ros_t);
 
-private:
     RecorderOptions               options_;
 
     Bag                           bag_;


dirk-thomas:

dirk-thomas:
Currently all this API is internal which means it can be easily changed in future releases. By changing the visibility it is implied that all the API is stable and stays exactly the same.

@ros/ros_team Please review this.

gerkey:
@DLu Can you give more information on your use case? Is there a chance that it could be satisfied via composition instead of inheritance?

jackieokay:
I believe changing functions from private to protected is an ABI change:

https://techbase.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B

I can't speak to the design decision of making Recorder subclassable; in principle it seems fine, but Brian makes a good point that, depending on your design, it might not be the right thing to do.

DLu:
I was exploring solutions to #608 so that I could control (i.e. start and stop) bags from a ROS interface. I ended up just writing a python wrapper, which I consider inelegant in this scenario.

wjwwood:
If there is a use case that needs it, then I think changing them to be protected is ok (potentially some of the members might be left private, but protected should ok too). However, since it is an ABI change I'd prefer it go into Kinetic and not Jade/Indigo.

@DLu do you have a prototype (even a partial one) you could share with us. I'd like to get a sense of what is actually needed for the tool you were discussing. It's clear this class wasn't designed to be inherited from, but maybe we change it in a more subtle way to enable your use case.

DLu:
#706 is another thing I was looking at this functionality for. Unfortunately, I no longer have the prototype I had hacked up.

dirk-thomas:
Should this ticket be closed then?

Issue #152, opened 2013-01-31
roscpp: Add service calls an optional timeout value  (ros ticket #2913)


tfoote:
I believe an optional timeout is mandatory in service calls.
If the service provider node is malfunctioning and is not answering a
service call, my node should not be stuck on the call forever.

Sometimes when I restart a node that provides a service, any call to that service will be locked until I kill the process.

Attached files to reproduce it:
 They are nearly the same used in
 http://www.ros.org/wiki/ROS/Tutorials/WritingServiceClient%28c%2B%2B%29

 The server has had the NoSigintHandler option added to recreate a hard
 crash.
 The client is on an infinite loop.

 To recreate it, run the server, run the client and kill the server using
 Ctrl+C.
 The client should be stuck calling the server.

trac data:
- Owner: **tfoote**
- Reporter: **vlopez**
- Reported at: **Sat Jul 24 02:08:37 2010**
- URL: https://code.ros.org/trac/ros/ticket/2913

tfoote:
[asomerville] I've recently run into this very issue. Is this still on the table? It seems to be pretty important not just for crashes on the server side, but also for clients to be able to make time guarantees when dealing with services that don't have them.

It seems deep down to depend on a boost::condition_variable::wait() which seems like it might be able to be switched to a timed_wait.

http://ros.org/doc/unstable/api/roscpp/html/service__server__link_8cpp_source.html#l00360

tfoote:
[lorenz] I'm reassigning the ticket to Tully since Ken is not at Willow anymore and I don't know who is maintaining roscpp at the moment. I personally think that this ticket is describing a severe problem that should be addressed quickly.

xqms:
Just a short note because I came across this ticket: Setting the `TCP_USER_TIMEOUT` socket option should be enough (on linux) to make this work. Of course, you cannot be sure that the call _actually_ failed, or if you just failed to get the response. I still believe it is better behavior to notify the caller that something is taking too long instead of waiting indefinitely.

jbohren:
@xqms Funny we should meet here. Have you had any luck setting `TCP_USER_TIMEOUT` for TCPROS topics?

mrdanbrooks:
I recently ran into this problem. I had to work around the issue by running the service call in a separate thread so that I could decide whether or not the call was taking to long to return. I'm guessing that this issue has taken a back seat in favor of encouraging people to use actionlib, but in this case I was working with a product and the service call was part of their API.

[...]

PR #738, opened 2016-02-06
Timeout can be specified for roscpp service calls

dkargin:
Added function ros::ServiceClient::setTimeout to limit the time roscpp service client waits for response. By defaut it is set to -1.0 seconds,  and the service call blocks until it gets a response or connection is dropped.

Straightforward solution for #152

### DIFF
--- a/clients/roscpp/include/ros/service_server_link.h
+++ b/clients/roscpp/include/ros/service_server_link.h
@@ -105,7 +105,8 @@ public:
    * If there is already a call happening in another thread, this will queue up the call and still block until
    * it has finished.
    */
-  bool call(const SerializedMessage& req, SerializedMessage& resp, double timeout = -1);
+  bool call(const SerializedMessage& req, SerializedMessage& resp);
+  bool call(const SerializedMessage& req, SerializedMessage& resp, double timeout);
 
 private:
   void onConnectionDropped(const ConnectionPtr& conn);
--- a/clients/roscpp/include/ros/service_server_link.h
+++ b/clients/roscpp/include/ros/service_server_link.h
@@ -105,7 +105,7 @@ public:
    * If there is already a call happening in another thread, this will queue up the call and still block until
    * it has finished.
    */
-  bool call(const SerializedMessage& req, SerializedMessage& resp);
+  bool call(const SerializedMessage& req, SerializedMessage& resp, double timeout = -1);
 
 private:
   void onConnectionDropped(const ConnectionPtr& conn);
--- a/clients/roscpp/src/libros/service_client.cpp
+++ b/clients/roscpp/src/libros/service_client.cpp
@@ -174,10 +174,10 @@ bool ServiceClient::isPersistent() const
 
 void ServiceClient::setTimeout(double timeout)
 {
-	if(impl_)
-	{
-		impl_->timeout = timeout;
-	}
+  if(impl_)
+  {
+    impl_->timeout = timeout;
+  }
 }
 
 void ServiceClient::shutdown()
--- a/clients/roscpp/src/libros/service_client.cpp
+++ b/clients/roscpp/src/libros/service_client.cpp
@@ -35,7 +35,7 @@ namespace ros
 {
 
 ServiceClient::Impl::Impl() 
-  : is_shutdown_(false)
+  : is_shutdown_(false), timeout (-1.0)
 { }
 
 ServiceClient::Impl::~Impl()
@@ -140,7 +140,7 @@ bool ServiceClient::call(const SerializedMessage& req, SerializedMessage& resp, 
     }
   }
 
-  bool ret = link->call(req, resp);
+  bool ret = link->call(req, resp, impl_->timeout);
   link.reset();
 
   // If we're shutting down but the node haven't finished yet, wait until we do
@@ -170,6 +170,14 @@ bool ServiceClient::isPersistent() const
   }
 
   return false;
+}
+
+void ServiceClient::setTimeout(double timeout)
+{
+	if(impl_)
+	{
+		impl_->timeout = timeout;
+	}
 }
 
 void ServiceClient::shutdown()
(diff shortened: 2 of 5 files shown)


evenator:
I'm not an expert on C++ linkers, but I think this would cause an ABI change, which would make it ineligible for released ROS distros, especially indigo. I hope someone with more knowledge on C++ ABIs than me could come up with a way to implement this that doesn't change the ABI, because it would be a useful feature. Barring that, I hope this PR will be accepted for KTurtle.

dkargin:
It 'should'  be fine for ABI. I've added a new variable into Impl class. And new class method should not break C++ ABI. There could be a problem with internal ABI, because I've added some new arguments. Does anybody uses that internal stuff directly? I mean ServiceServerLink::call method.

But I've found some other pull requests, i.e for solving 'reconnect' behaviour for service client, like #574 . The final solution, that supports both reconnect and timeout functionality, should be a bit more complex.

clients/roscpp/src/libros/service_client.cpp
```diff
@@ -172,6 +172,14 @@ bool ServiceClient::isPersistent() const
   return false;
 }
 
+void ServiceClient::setTimeout(double timeout)
+{
+	if(impl_)
+	{
+		impl_->timeout = timeout;
+	}
```
dirk-thomas: Please follow the style in the file ans use two spaces for indentation.

Same below.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -357,15 +358,30 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
+    			break;
+    		}
+    	}
+    	else
+    		info->finished_condition_.wait(lock);
     }
   }
 
+  if(interrupted)
+	  this->clearCalls();
```
dirk-thomas: This function call seems to require the same lock as used above: `info->finished_mutex_`

dkargin: clearCalls is mutex-protected by itself. Protecting it by info->finished_mutex_ will cause a deadlock at ServiceServerLink::cancelCall

prince-ph0en1x (approved):

clients/roscpp/include/ros/service_server_link.h
```diff
@@ -105,7 +105,7 @@ class ROSCPP_DECL ServiceServerLink : public boost::enable_shared_from_this<Serv
    * If there is already a call happening in another thread, this will queue up the call and still block until
    * it has finished.
    */
-  bool call(const SerializedMessage& req, SerializedMessage& resp);
+  bool call(const SerializedMessage& req, SerializedMessage& resp, double timeout = -1);
```
efernandez: This will break ABI compatibility (see http://www.ros.org/reps/rep-0009.html#what-s-not-allowed), but it's very easy to keep ABI compatibility in this case, simply create a new method with the new argument, so you have something like this:

``` c++
bool call(const SerializedMessage& req, SerializedMessage& resp);
bool call(const SerializedMessage& req, SerializedMessage& resp, double timeout);
```

I'm hesitant of providing the new implementation for the 2-argument version here because IIRC if the compiler makes it inline it'd also break ABI compatibility. Therefore, I'd add the implementation on the .cpp.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -321,7 +322,7 @@ void ServiceServerLink::processNextCall()
   }
 }
 
-bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
+bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp, double timeout)
```
efernandez, line 325: Following my previous comment above, here you'll have to do this:

``` c++
bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
{
  return call(req, resp, -1);
}

bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp, double timeout)
```

efernandez (commented):
Consider my coments if you want to keep ABI compatibility.

efernandez:
@dkargin I've made my ABI comments just a review "Comment" rather than a request for changes because you're right that those headers are sort of internal, but also note they're exported here https://github.com/sybotech/ros_comm/blob/e5e8d8ac3ffbc1b36ebd9f8bb3a1bf6067aed3a3/clients/roscpp/CMakeLists.txt#L133

clients/roscpp/src/libros/service_server_link.cpp
dkargin: Thanks for notice. Fixed

efernandez:
@dkargin Regarding the merging conflicts, I think all you have to do is:

1. Remove the "Trying to ease merging process" commit
2. From your branch, run `git rebase indigo-devel`
3. Then `git push -f sybotech ServiceCallTimeout` assuming `sybotech` is how you call your fork locally

Maybe you should target `lunar-devel` too, BTW.

efernandez (approved):
Thanks for applying the changes after my comments. LGTM.

efernandez:
Although this change should be ABI compatible now, I'd rather target `lunar-devel` instead of `indigo-devel`.

@dirk-thomas Thoughts?

It's also unclear to me what was the end of the story for #574

mikepurvis:
Rebased for `lunar-devel` in #1165. Let's take the discussion there.

peci1:
But we'd also like to have it in Indigo =)

efernandez:
IMHO this should go to the latest devel branch (`lunar-devel`) @peci1 and once it gets merged there we can discuss about cherry-picking it into previous versions. I don't have anything against that, but remember that indigo is 3 versions old.

prateek470:
@dkargin Do we have SetTimout in any ROS version?
I am stuck on this one for a long time now. Please merge these changes.

dkargin:
@prateek470 No, timeouts are not in the mainline now. You can fork  ros_comm, and adapt this changes and rebuild everything from scratch (like in meta-ros), or do not use ROS services at all, if you need sort of reliability and robustness.

PR #739, opened 2016-02-12
getParamNames

progtologist:
Added method for getting all the parameters from the parameter server as implemented in the rospy client. Implemented in NodeHandle and added simple unit test.

### DIFF
--- a/clients/roscpp/include/ros/param.h
+++ b/clients/roscpp/include/ros/param.h
@@ -543,6 +543,7 @@ ROSCPP_DECL bool getCached(const std::string& key, std::map<std::string, bool>& 
  * \throws InvalidNameException if the key is not a valid graph resource name
  */
 ROSCPP_DECL bool has(const std::string& key);
+
 /** \brief Delete a parameter from the parameter server.
  *
  * \param key The key to delete.
@@ -583,6 +584,13 @@ ROSCPP_DECL bool search(const std::string& ns, const std::string& key, std::stri
  * \throws InvalidNameException if the key is not a valid graph resource name
  */
 ROSCPP_DECL bool search(const std::string& key, std::string& result);
+
+/**
+ * \brief Get the list of all the parameters in the server
+ * \param keys The vector of all the keys
+ * \return false if the process fails
+ */
+ROSCPP_DECL bool getParamNames(std::vector<std::string>& keys);
 
 /** \brief Assign value from parameter server, with default.
  *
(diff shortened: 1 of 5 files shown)


clients/roscpp/include/ros/param.h
```diff
@@ -543,6 +543,7 @@ ROSCPP_DECL bool getCached(const std::string& key, std::map<std::string, bool>&
  * \throws InvalidNameException if the key is not a valid graph resource name
  */
 ROSCPP_DECL bool has(const std::string& key);
+
```
dirk-thomas, line 546: Please avoid any unrelated whitespace changes.

dirk-thomas:
Thank you for providing the patch.

On issue with the proposed API is that it might be very slow. But that is not a reason to not provide it in the first place. Python has the same function so C++ should provide it too.

clients/roscpp/src/libros/param.cpp
```diff
@@ -720,6 +720,31 @@ bool getCached(const std::string& key, std::map<std::string, bool>& map)
...
+  // Make sure it returned 3 elements
+  if (result.size() != 3)
+    return false;
+  // Get the actual parameter keys
+  XmlRpc::XmlRpcValue parameters = result[2];
+  // Resize the output
+  keys.resize(parameters.size());
+
+  // Fill the output vector with the answer
+  for (int i = 0; i < parameters.size(); i++) {
```
dirk-thomas, line 741: Please use `size_t i` here as well as `++i`.

dirk-thomas:
@ros/ros_team Please review this.

wjwwood:
Sounds good to me. +1

dirk-thomas:
Thank you for the patch. I have squashed the commit and applied my comments in cd10ec9cc333e33e12ec753872d6c55b2e6d22be

clients/roscpp/src/libros/param.cpp
progtologist, line 741: I would also prefer size_t, however, the XmlRpcValue::size() returns an int and not a long unsigned int. I tried to follow the same code style of the param.cpp file (as shown, for example in the getImpl function which uses i++). Anyway, I don't want to slow down this commit, so feel free to proceed anyway you see fit!

dirk-thomas, line 741: Thanks for pointing it out. I obviously didn't check the size but expected it to be `size_t`. I will change this back as part of PR #760.

PR #743, opened 2016-02-20
Preserve identity of `numpy_msg(T)`

eric-wieser:
This makes `numpy_msg(T) is numpy_msg(T)` always return true.
Otherwise, it is impossible to check whether a message is of a numpy type or not, since `isinstance(numpy_msg(T)(), numpy_msg(T))` would always return false


dirk-thomas:
Thank you for the patch. I have created #758 which modifies the logic a bit to reduce the diff size to make it easier to review the change.

eric-wieser:
Great, thanks!

PR #744, opened 2016-02-21
Update travis to use new install process

eric-wieser:
Install as recommended in http://wiki.ros.org/indigo/Installation/Source

This might make the build servers pass?


dirk-thomas:
See #747.

PR #750, opened 2016-02-23
fix max elements param for statistics window

dirk-thomas:
Fix `max_element` parameter for statistics windows. The wrong initialization has been the case since the feature was added in #398.


dirk-thomas:
retest this please

PR #751, opened 2016-02-23
Enable and fix rosmsg tests

gerkey:
As noted in #709, the test suite for `rosmsg` has long been disabled. This PR enables the test suite and updates the tests to make them pass. In some cases, it could be argued that instead the code should be updated instead of the tests. But I feel that's it's less risky to update the tests to match the long-standing behavior of the code.

### DIFF
--- a/clients/roscpp/CMakeLists.txt
+++ b/clients/roscpp/CMakeLists.txt
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(roscpp)
 
+if(NOT WIN32)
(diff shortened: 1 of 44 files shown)


dirk-thomas:
The failing test from roswtf (`test_offline`) is a known problem (see #752). But the other failing tests are related to this PR.

dirk-thomas:

gerkey:
The test failures were caused by:

```
Unknown message class: geometry_msgs/Polygon
```

I added a `test_depend` on `geometry_msgs` in efff5a4.

gerkey:
Replaced by #764.

PR #754, opened 2016-02-24
add EdgeList.__contains__

eric-wieser:
Python dictionaries support indexing by tuples

### DIFF
--- a/tools/rosgraph/src/rosgraph/impl/graph.py
+++ b/tools/rosgraph/src/rosgraph/impl/graph.py
@@ -104,6 +104,10 @@ class EdgeList(object):
         return itertools.chain(*[v for v in self.edges_by_start.values()])
     
     def has(self, edge):
+        """ deprecated """
+        return edge in self
+
+    def __contains__(self, edge):
         """
         @return: True if edge is in edge list
         @rtype: bool
@@ -239,9 +243,9 @@ class Graph(object):
         self.topic_ns = topic_ns or '/'
 
         # ROS nodes
-        self.nn_nodes = set([])
+        self.nn_nodes = set()
         # ROS topic nodes
-        self.nt_nodes = set([])
+        self.nt_nodes = set()
 
         # ROS nodes that aren't responding quickly
         self.bad_nodes = {}
@@ -249,7 +253,7 @@ class Graph(object):
         self.bad_nodes_lock = threading.Lock()
         
         # ROS services
-        self.srvs = set([])
+        self.srvs = set()
         # ROS node->node transport connections
         self.nn_edges = EdgeList()
         # ROS node->topic connections        


eric-wieser:
Obviously `has` is public api, so can't be removed entirely

tools/rosgraph/src/rosgraph/impl/graph.py
```diff
@@ -200,9 +203,9 @@ def __init__(self, start, end, label=''):
         self.start = start
         self.end = end
         self.label = label
-        self.key = "%s|%s"%(self.start, self.label)
+        self.key = (self.start, self.label)
```
dirk-thomas: These slots are public API of an edge and therefore can't change their value like this.

Same  for `rkey` below.

eric-wieser: So the value of `rkey` itself is considered public api, not just the value of the expression `edges_by_start[rkey]`? Darn.

dirk-thomas: An `Edge` is basically like a struct. All member are public and likely being used by user land code..

eric-wieser: Darn, I can't really argue with the fact that people _might_ be relying on this. Is `|` a valid character in node or edge names? (ie, can I frame this as a bug, in which case breaking reliance of a bug is perhaps more ok)

Or should I just rebase this PR to just be the `__contains__` above?

dirk-thomas: No, `|` is not a valid character in node or edge names. So it is safe to be used as a separator.

Yes, please update the PR - and squash the commits on this branch. The `__contains__` is certainly the significant of this PR. The slightly easier implementation of `__iter__` as well as the set initializations below are more cosmetic but you can keep those.

dirk-thomas:

dirk-thomas:

dirk-thomas:
Please rebase the patch against current `indigo-devel` since it seems to contain an old patch which results in a syntax error.

eric-wieser:
Looks like you did a rewrite of published history there! Corrected now

dirk-thomas:
Thanks for the patch. I have applied a modified version to the _kinetic-devel_ branch: 5de058ad1bbf3a525ae3f5288a76e59ec0dd62d0 Since the `has` method is not actually deprecated I removed the comment. I also skipped the cleanup changes.

eric-wieser:
> Since the has method is not actually deprecated

Part of this pull request was me requesting that it should be, in the same way that [`dict.has_key`](https://docs.python.org/2/library/stdtypes.html#dict.has_key) is. not assuming that it was.

If you think it shouldn't be, then that's fine - it's not my call to make.

Thanks for the merge!

dirk-thomas:
Personally I do agree with you that cleaning up the API is nice. But based on the amount of code using this API it is not worth "forcing" the people to update it. Sorry.

eric-wieser:
> But based on the amount of code using this API it is not worth "forcing" the people to update it. Sorry.

Deprecation doesn't ever need to lead to forcing people to update it - my goal was to stop anyone ever using `.has` in new code.

Also, this strikes me as a rather internal API, so I'd be surprised if that much code actually uses it anyway.

At any rate, I half-assed this, and if I was going for deprecation should probably have included a `DeprecationWarning`, so rejecting the rest of the patch is the right thing to do

PR #759, opened 2016-03-01
Actually send the length byte in python 3

eric-wieser:
Fixes regresssion of #717

### DIFF
--- a/tools/rosgraph/src/rosgraph/network.py
+++ b/tools/rosgraph/src/rosgraph/network.py
@@ -394,8 +394,8 @@ def encode_ros_handshake_header(header):
 
     :param header: header field keys/values, ``dict``
     :returns: header encoded as byte string, ``str``
-    """    
-    fields = ["%s=%s" % (k, header[k]) for k in sorted(header.keys())]
+    """
+    fields = ["%s=%s" % (k, v) for k, v in sorted(header.items())]
     
     # in the usual configuration, the error 'TypeError: can't concat bytes to str' appears:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch. The first commit is a crucial fix (0752719283c6f2d8b33d2d7c7eaf7b4e6b69499f). Since the remaining three commits only refactor the code I squashed them into one (7a3d14d89d89a528ff39816a11d477433d59282d).

eric-wieser:
I'm slightly worried by how this failed the CI...

dirk-thomas:
I am not :wink: The Jade job passed and the Indigo job just had a single failing test which is sadly flaky.

eric-wieser:
I'm guessing there are no python 3 tests on travis right now. Does ROS even official support Python 3 before ROS 2.0?

dirk-thomas:
We try to support users which build ROS 1 from source. Commonly it is only the case on other platforms like Gentoo that they want to use Python 3. I don't expect that any ROS 1 distribution in the near future will switch to Python 3 simply because the community does not want to deal with the effort necessary to update existing code.

Btw. the pull request testing doesn't use Travis at all but the ROS Jenkins server from build.ros.org. And yes, it only tests with Python 2.

dirk-thomas:
FYI this patch needed fixing as proposed in #759.

eric-wieser:
Did you mean to link to the same page you posted on?

dirk-thomas:
Of course - not :wink: #789

PR #761, opened 2016-03-04
Fix minor python issues

eric-wieser:
- Fix a crash when crashing
- Make `isinstance(obj, SimpleFilter)` work
- Future-proof decorator
- Use super where possible


dirk-thomas:
Thank you!

PR #762, opened 2016-03-09
evaluation of expressions in roslaunch

rhaschke:
This replaces PR #707, providing a first proof-of-concept implementation of a generic $(eval ...) substitution arg as discussed in #707.

Unfortunately, the new resolve method cannot be seamlessly integrated into the existing `resolve_args` mechanism (as suggested by @gerkey), because the existing code uses the simple parser in `_collect_args`, which cannot handle parentheses within `$(xxx )` expressions. Hence, I handle `$(eval ...)` expressions in a specific fashion before resorting to the other resolve args.
For a similar reason, $(eval ) substitution args can only be handled if they span the whole string.

The branch https://github.com/ubi-agni/ros_comm/tree/roslaunch-eval provides a basic implementation.
The extension branch https://github.com/ubi-agni/ros_comm/tree/implicit-arg-access comprises a simplification to allow implicit access to arg values. Hence, instead of writing
`$(eval arg('my_arg') == 42)`
one can directly write:
`$(eval my_arg == 42)`

What do you think about that implementation?

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -268,6 +268,18 @@ def _arg(resolved, a, args, context):
     else:
         raise ArgException(arg_name)
 
+# Create a dictionary of global symbols that will be available in the eval
+# context.  We disable all the builtins, then add back True and False, and also
+# add true and false for convenience (because we accept those lower-case strings
+# as boolean values in XML).
+_eval_dict=dict(true=True, false=False, __builtins__={'True': True, 'False': False})
+
+def _eval(s, context):
+    # ignore values containing double underscores (for safety)
+    # http://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html
+    if s.find('__') >= 0:
+        raise SubstitutionException("$(eval ...) may not contain double underscore expressions")
+    return str(eval(s, _eval_dict, None))
 
 def resolve_args(arg_str, context=None, resolve_anon=True):
     """
@@ -294,9 +306,11 @@ def resolve_args(arg_str, context=None, resolve_anon=True):
     """
     if context is None:
         context = {}
-    #parse found substitution args
     if not arg_str:
         return arg_str
+    # special handling of $(eval ...)
+    if arg_str.startswith('$(eval ') and arg_str.endswith(')'):
+        return _eval(arg_str[7:-1], context)
     # first resolve variables like 'env' and 'arg'
     commands = {
         'env': _env,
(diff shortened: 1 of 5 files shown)


dirk-thomas:
@gerkey Can you please review this.

Due to the advanced time I think we should not merge this into Jade but into Kinetic only. If it works there it can always be backported.

gerkey:
@rhaschke This looks good, thanks. I'm torn about the shortcut on args. It's certainly less typing, but should we be worried about name collisions with other symbols in Python? If there's any risk of the automatic substitution doing the wrong thing, I'd rather just force people to write `arg('foo')` in full.

Other than that issue, +1 from me. The implementation is clear and the test suite is good.

@dirk-thomas Agreed on targeting kinetic now, and considering a backport later once we have more experience with it. There's one failure on Jenkins with this branch, but it looks very unrelated: http://build.ros.org/job/Kpr__ros_comm__ubuntu_xenial_amd64/14/testReport/test_bag/TestRosbag/test_get_compression_info/. Is that a known flaky test?

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -284,18 +283,35 @@ def _arg(resolved, a, args, context):
...
 _eval_dict=dict(true=True, false=False, __builtins__={'True': True, 'False': False})
 
+class _DictWrapper(object):
+    def __init__(self, args, functions):
+        self._args = args
+        self._functions = functions
+
+    def __getitem__(self, key):
+        try:
+            return self._functions[key]
```
rhaschke, line 294: substitution functions take precedence over implicit argument usage

rhaschke:
@gerkey Indeed the automatic substitution might overwrite existing symbols. However, evaluation of substitions functions (arg, anon, find, etc) already takes precedence over implicit argument usage (as pointed out above and verified with an additional unittest).
There might be interference with the four builtin python symbols (True, False, true, false) that can be redefined by the user. Currently, there are no other exposed python symbols that could be redefined.
To resolve that issue, we could:
- warn the user about name collisions
- or silently give precedence to builtin symbols again

I would prefer the latter option, which I already implemented in 1c9d7a0.

gerkey:
lgtm

Can you please re-target for kinetic and then we'll consider it for merging?

rhaschke:
Retargeting for kinetic

arthurnunesq:
Is there any chance of this feature being backport to indigo?

dirk-thomas:
Larger feature like this is commonly not backported since the chance of regressions is outweighing the desire for stability for a ROS distro which has been released for over 2.5 years.

cpagravel:
We're using very ugly work-arounds for this missing feature at the robotics startup I'm working at. If it means anything, I'd really like to have this feature on our system without having to move to Kinetic. String comparisons would make really be a big help to make this project more maintainable.

PR #764, opened 2016-03-10
Enabled tests and then fixed them

gerkey:
Replacement for #751, where my branch got mangled to a point that I couldn't squash the commits. This branch contains the same changes as `enable_tests`, where on the last build all tests passed on Indigo, and there was one apparently spurious timing failure in `rostest`:
- http://build.ros.org/job/Ipr__ros_comm__ubuntu_trusty_amd64/109/
- http://build.ros.org/job/Jpr__ros_comm__ubuntu_trusty_amd64/105/

Thanks much to @dhood for the help!


tools/rosmsg/package.xml
```diff
@@ -25,6 +25,8 @@
   <run_depend>rosbag</run_depend>
   <run_depend>roslib</run_depend>
 
+  <test_depend>geometry_msgs</test_depend>
```
dirk-thomas: Currently nothing in this repository depends on `common_msgs`. Therefore I think this should be avoided and local custom messages should be used for these tests.

gerkey: In 1437c60 I deleted `test_rosmsgproto_geometry_py` and added to `test_rosmsgproto.py` tests of similar complexity to the deleted tests, but only using messages from `std_msgs`. The resulting test suite is not identical to the previous suite, but I don't see any significant changes in coverage.

dirk-thomas:
Thanks. Recreated against _kinetic-devel_: #826

In the future please comment on the pull request when committing changes in order to notify people.

Issue #733, opened 2016-01-26
rosbag::createAdvertiseOptions prototype missing parameter


amrahimi:
In branch indigo-devel, include/rosbag/player.h, the createAdvertiseOptions() prototypes are missing the third 'const string &prefix' parameter.

dirk-thomas:
Please see #774 for the proposed change in Kinetic.

dirk-thomas:
@ros-pull-request-builder retest this please

PR #626, opened 2015-06-11
add prefix option for prefixing output topics

Karsten1987:
The prefix option allows to publish all topics with the given string prefix. 

Use cases are e.g. playing a rosbag and the original source at the same time comparing the two outcomes simultaneously without topic conflicts.

PR #771, opened 2016-03-17
add missing parameter to AdvertiseOptions::createAdvertiseOptions (#733)

dirk-thomas:
Addresses #733.

The signature of the function was only updated in the .cpp file in #626.


dirk-thomas:
This would break ABI an therefore needs to go into the upcoming kinetic-devel branch only.

dirk-thomas:
Closing in favor of #774.

Issue #769, opened 2016-03-16
rosbag python API deserializes to non-identity-equal types


eric-wieser:
All the types that are created on the fly do not match the types that can be imported, ie:

```
bag = rosbag.Bag(...)
msg = next(iter(bag))
isinstance(msg, PoseStamped)
```

Will always return `False` on a bag containing stamped poses

dirk-thomas:
Currently the message type is being generated on-the-fly (https://github.com/ros/ros_comm/blob/98aa98e565ea623aae61c02f54ecb06c7ee190dd/tools/rosbag/src/rosbag/bag.py#L1562-L1570) in order to handle arbitrary message types (which might not even be available on the system.

It would be possible to look at `info.datatype` which can be parsed by `genmsg.package_resource_name` and try to import the message definition. If that is available (and has the same md5sum) it could be used instead of the dynamically generated type.

I will likely not have the bandwidth to work on this but would be happy to review a merge a pull request.

eric-wieser:
Is making the type reuse happen upstream in `genpy.dynamic.generate_dynamic` sensible?

dirk-thomas:
I would say no since `generate_dynamic` should still do what its name implies: generate a dynamic type.

eric-wieser:
I'd argue that you never want to have two identical but not identity-equal types floating around the system. Can you give me an example where this behavour is desirable?

eric-wieser:
Actually, I'd say this is required to give sane rosbag behaviour. If I have a custom message type `MyWrappedImage` containing an `sensor_msgs/Image`, and `MyWrappedImage` isn't installed on my system, I'd expect it to dynamically generate a class that uses the static `sensor_msgs/Image`. Right now, that gets dynamically generated too.

dirk-thomas:
The name of the function in `genpy` implies a very specific semantic behavior. Using existing message definitions when available should happen outside of that function.

This conditional logic can be put anywhere (maybe even in `genpy`) and then be used wherever useful.

eric-wieser:
Sounds like `generate_dynamic` needs to acquire a `use_existing = False` default argument then

dirk-thomas:
As argued above I don't think it is in the scope of a function called `generate_dynamic` to return non-dynamic message types.

A new function `get_message_type` should be introduced which either returns an existing message class or the result of `generate_dynamic` otherwise.

eric-wieser:
Seems reasonable

mikepurvis:
[...]

PR #772, opened 2016-03-17
Add a very simple testcase

eric-wieser:
This exposes #769, but currently marks it as an expected failure. It also contains a passing test that verifies some very simple roundtrip properties.


tools/rosbag/test/test_roundtrip.py
```diff
@@ -0,0 +1,61 @@
+import unittest
+import rosbag
+import rospy
+
+class TestRoundTrip(unittest.TestCase):
+    def _write_simple_bag(self):
+        from std_msgs.msg import Int32, String
+
+        with rosbag.Bag('/tmp/test.bag', 'w') as bag:
```
dirk-thomas: Please use a bag name which is kind of unique to this test. Otherwise it might collide with other tests being run in parallel.

eric-wieser: Does this need to also be unique to each `test_...` function? Or are those tests guaranteed to run sequentially?

dirk-thomas: By default each nosetest invocation runs its tests sequentially. Separate CTests (therefore separate nosetests) will run in parallel.

But since the user can override this with e.g. `NOSE_PROCESSES` I think it is good to avoid any cross talk if possible by design.

eric-wieser: This is fixed in the rewritten commits below

dirk-thomas:
It would be great if you could provide a PR to implement the missing functionality so that these test actually pass.

eric-wieser:
Yeah, I'm working on that PR, but I wanted to have some committed tests first in case I never get around to it

tools/rosbag/CMakeLists.txt
```diff
@@ -58,3 +58,7 @@ catkin_install_python(PROGRAMS
   scripts/savemsg.py
   scripts/topic_renamer.py
   DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
+
+if(CATKIN_ENABLE_TESTING)
+  catkin_add_nosetests(test)
+endif()
```
mikepurvis: Newline please.

mikepurvis:
Please retarget this to **kinetic-devel**.

dirk-thomas:

dirk-thomas:
Sorry, I meant #981.

dirk-thomas:

mikepurvis:
Looks good, thanks @eric-wieser!

PR #777, opened 2016-03-17
Promote the result of read_messages to a namedtuple

eric-wieser:
Which is backwards compatible

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -103,6 +103,8 @@ class Compression:
     BZ2  = 'bz2'
     LZ4  = 'lz4'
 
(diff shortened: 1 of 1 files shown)


eric-wieser:
This strikes me as an unrelated test failure

dirk-thomas:

eric-wieser:
Does that only work if you say it?

dirk-thomas:
Yes, you need to be admin on the orgunit for Jenkins to "accept" the magic phrases of the GitHubPullRequestBuilder plugin.

dirk-thomas:
Thank you for the patch.

dirk-thomas:
Cherry-picked to _kinetic-devel_ in: 1ee3fde8f873b45fcf53a3cec54723a137a0560f

Issue #736, opened 2016-02-03
rosbag filter adds two bytes per topic


chaoflow:
The bug occurs at least with ROS indigo, we did not test other versions.
cc @durko @dejanpan

```
$ rosbag info original.bag 
path:        original.bag
version:     2.0
duration:    11.0s
start:       Mar 06 2015 14:32:21.02 (1425648741.02)
end:         Mar 06 2015 14:32:32.04 (1425648752.04)
size:        60.8 KB
messages:    308
compression: none [1/1 chunks]
types:       rosgraph_msgs/Log [acffd30cd6b6de30f120938c17c593fb]
             std_msgs/String   [992ce8a1687cec8c8bd883ec73ca41d1]
topics:      /chatter       98 msgs    : std_msgs/String  
             /rosout       106 msgs    : rosgraph_msgs/Log (2 connections)
             /rosout_agg   104 msgs    : rosgraph_msgs/Log
$ rosbag filter original.bag x.bag True                  
$ rosbag filter x.bag y.bag True
$ rosbag filter y.bag z.bag True
$ du -bs *
62209   original.bag
59370   x.bag
59376   y.bag
59382   z.bag
```

The change of filesize from original.bag to x.bag is to be expected. After that the filesize (and md5sum) should remain the same. However, `rosbag filter` adds 2 bytes per topic.

At least one byte per topic we found: it's a `\n` (hex `0a`) appended to pytype._full_text:

```
$ xxd x.bag  |grep -A1 ^00014b0
00014b0: 7374 7269 6e67 2066 7261 6d65 5f69 640a  string frame_id.
00014c0: 0a27 0000 006d 6435 7375 6d3d 6163 6666  .'...md5sum=acff

$ xxd y.bag  |grep -A1 ^00014b0
00014b0: 7374 7269 6e67 2066 7261 6d65 5f69 640a  string frame_id.
00014c0: 0a0a 2700 0000 6d64 3573 756d 3d61 6366  ..'...md5sum=acf

$ xxd z.bag  |grep -A1 ^00014b0
00014b0: 7374 7269 6e67 2066 7261 6d65 5f69 640a  string frame_id.
00014c0: 0a0a 0a27 0000 006d 6435 7375 6d3d 6163  ...'...md5sum=ac
```

chaoflow:
Link to tarball with bags: https://ternaris.com/rosbag-filter-bags-issue736.tar.gz

dirk-thomas:
[...]

PR #780, opened 2016-03-25
writing rosbags adds a newline

dhood:
Potential fix for https://github.com/ros/ros_comm/issues/736 where it was noticed that newlines are added each time rosbag filter is called.

From what I can see a newline is added to the message text [here](https://github.com/ros/genpy/blob/6577cc3d3731801913b3fdf62951609f1bf41b72/src/genpy/generator.py#L773) when writing a message, but it's never accounted for when reading the message. So it should either not be written, or it should be removed when reading.

Have removed the newline in the rosbag code but it may be appropriate to remove it [here](https://github.com/ros/genpy/blob/06cc06b7914174ecdddca641a0490651236a0336/src/genpy/dynamic.py#L127) instead

### DIFF
--- a/test/test_rosbag/test/test_bag.py
+++ b/test/test_rosbag/test/test_bag.py
@@ -33,10 +33,12 @@
 #
 # test_bag.py
 
+import hashlib
 import heapq
 import os
 import shutil
 import sys
+import tempfile
 import time
 import unittest
 
@@ -187,6 +189,33 @@ class TestRosbag(unittest.TestCase):
         msgs = list(rosbag.Bag(outbag_filename).read_messages())
 
         self.assertEquals(len(msgs), 5)
+
+    # Regression test for issue #736
+    def test_trivial_rosbag_filter(self):
+        tempdir = tempfile.mkdtemp()
+        try: 
+            inbag_filename = os.path.join(tempdir, 'test_rosbag_filter__1.bag') 
+            outbag1_filename = os.path.join(tempdir, 'test_rosbag_filter__2.bag') 
+            outbag2_filename = os.path.join(tempdir, 'test_rosbag_filter__3.bag') 
+
+            with rosbag.Bag(inbag_filename, 'w') as b:
+                for i in range(30):
+                    msg = ColorRGBA()
+                    t = genpy.Time.from_sec(i)
+                    b.write('/ints' + str(i), msg, t)
+
+            # filtering multiple times should not affect the filtered rosbag
+            cmd = 'rosbag filter %s %s "True"'
+            os.system(cmd % (inbag_filename, outbag1_filename))
+            os.system(cmd % (outbag1_filename, outbag2_filename))
+            
+            with open(outbag1_filename, 'r') as h:
+                outbag1_md5 = hashlib.md5(h.read()).hexdigest()
+            with open(outbag2_filename, 'r') as h:
+                outbag2_md5 = hashlib.md5(h.read()).hexdigest()
+            self.assertEquals(outbag1_md5, outbag2_md5)
+        finally:
+            shutil.rmtree(tempdir)
 
     def test_reindex_works(self):
         fn = '/tmp/test_reindex_works.bag'


dirk-thomas:
Thank you for creating a regression test for this. The test itself looks good. I will comment on some minor style things inline.

Regarding the proposed fix: as you mentioned it would probably be cleaner to fix `genpy` instead. The function `compute_full_text_escaped` should escape a trailing double quote if it has not already been escaped by https://github.com/ros/genpy/blob/9c74c588955bc4398176bbc4b9d09224a7804307/src/genpy/generator.py#L316 Then the extra newline in https://github.com/ros/genpy/blob/6577cc3d3731801913b3fdf62951609f1bf41b72/src/genpy/generator.py#L773 would be obsolete and rosbag doesn't need to cut it off. Can you please try that alternative approach?

test/test_rosbag/test/test_bag.py
```diff
@@ -188,6 +190,29 @@ def test_rosbag_filter(self):
 
         self.assertEquals(len(msgs), 5)
 
+    def test_trivial_rosbag_filter(self):
+        # regression test for issue #736
+
+        inbag_filename  = '/tmp/test_rosbag_filter__1.bag'
+        outbag1_filename = '/tmp/test_rosbag_filter__2.bag'
+        outbag2_filename = '/tmp/test_rosbag_filter__3.bag'
```
dirk-thomas: Please avoid hard coding temporary paths. Use `tempfile` instead. That will ensure that the names don't collide with other tests and potentially work on other platforms.

test/test_rosbag/test/test_bag.py
```diff
@@ -188,6 +190,29 @@ def test_rosbag_filter(self):
...
+            for i in range(30):
+                msg = ColorRGBA()
+                t = genpy.Time.from_sec(i)
+                b.write('/ints' + str(i), msg, t)
+
+        expression = "True"
+        
+        # filtering multiple times should not affect the filtered rosbag
+        os.system('rosbag filter %s %s "%s"' % (inbag_filename, outbag1_filename, expression))
+        os.system('rosbag filter %s %s "%s"' % (outbag1_filename, outbag2_filename, expression))
```
dirk-thomas: I would suggest to extract the command string into a variable:

```
cmd = 'rosbag filter %s %s "True"'
```

That would avoid the duplication of the string as well as the expression variable.

test/test_rosbag/test/test_bag.py
```diff
@@ -188,6 +190,29 @@ def test_rosbag_filter(self):
...
+
+        outbag1_md5 = hashlib.md5(open(outbag1_filename,'r').read()).hexdigest()
+        outbag2_md5 = hashlib.md5(open(outbag2_filename,'r').read()).hexdigest()
```
dirk-thomas: Please use `with` to scope each file read:

```
with open(outbag1_filename, 'r') as h:
    outbag1_md5 = hashlib.md5(h.read()).hexdigest()
```

Also add a space after the comma before the `'r'` to comply with PEP8 (even if the existing code does a poor job with code style).

dhood:
Thanks for the review @dirk-thomas - I'm understanding better now where the problem originated.

I think the newline is trying to solve the same problem compute_full_text_escaped is. But, there was a bug in compute_full_text_escaped.

What do you think about this instead? https://github.com/ros/genpy/compare/indigo-devel...dhood:indigo-devel

I haven't been able to find a message type that this will actually affect in order to test it properly though. Do you know of any?

dirk-thomas:
While the missing assignment certainly needs fixing a single quote at the end of the string also needs to be escaped. For the following input `something ... "` the trailing quote would be a problem the the input is wrapped in `"""`.

I don't know any message out of my head which has a trailing quote. Obviously none has triple quotes since escaping that correctly is broken. :wink:

dhood:
How does https://github.com/ros/genpy/compare/indigo-devel...dhood:indigo-devel look now? I didn't bother with the beginning given the current msg format

dhood:
I'll open a PR on genpy so the tests can run

dhood:
Commits squashed. I think this can be closed now with https://github.com/ros/genpy/pull/47 right?

test/test_rosbag/test/test_bag.py
```diff
@@ -39,6 +39,7 @@
 import sys
 import time
 import unittest
+import tempfile
```
dirk-thomas: Please move this line two up (for alphabetical order of the core modules [PEP8]).

test/test_rosbag/test/test_bag.py
```diff
@@ -49,6 +50,8 @@
 from std_msgs.msg import ColorRGBA
 from std_msgs.msg import String
 
+import hashlib
```
dirk-thomas: Same here - should be before `import heapq`.

test/test_rosbag/test/test_bag.py
```diff
@@ -188,6 +191,35 @@ def test_rosbag_filter(self):
 
         self.assertEquals(len(msgs), 5)
 
+    def test_trivial_rosbag_filter(self):
+        # regression test for issue #736
+        
+        try: 
+            tempdir = tempfile.mkdtemp()
```
dirk-thomas: This line should be outside of the `try/except` block because if it raises there is nothing to cleanup yet.

dirk-thomas:
I think it still makes sense to merge this new test since it fill prevent regressions in this area in the future.

dhood:
thanks again - hopefully ok to merge now

dirk-thomas:
I will wait with the merge until `genpy` has been released and the test passes. But the current patch looks good. Thanks.

dirk-thomas:

dirk-thomas:

PR #782, opened 2016-03-30
Use defusedxml in rosmaster

codebot:
I'm not sure if this is the best way to do it (particularly the "del" statement), but anyway defusedxml.monkey_patch() shuts down a bunch of different ways to blow up XML-RPC servers.

### DIFF
--- a/tools/rosmaster/package.xml
+++ b/tools/rosmaster/package.xml
@@ -13,6 +13,7 @@
   <buildtool_depend version_gte="0.5.68">catkin</buildtool_depend>
 
   <run_depend>rosgraph</run_depend>
+  <run_depend>python-defusedxml</run_depend>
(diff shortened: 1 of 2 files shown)


wjwwood:
Is this the only place an XMLRPC server is created in Python? Do nodes in `rospy` not create one?

wjwwood:
That's fine, so long as you're aware that this isn't the only place.

dirk-thomas:

dirk-thomas:
#780 is only proposing a new test so it can't affect the results of other tests. The tests which failed in the last CI run are timing dependent and sadly flaky. I triggered another run already.

dirk-thomas:

dirk-thomas:

PR #783, opened 2016-03-31
Remove duplicated function, use decorator

eric-wieser:
Requires ros/genpy#52 to be merged/released first, so do not expect tests to pass


eric-wieser:
ros/genpy#52 is now merged, so this might now pass tests, depending on whether tests use the released version

dirk-thomas:
Yeah, this will require `genpy` to be released. I will retrigger the PR test once this has happened.

dirk-thomas:

dirk-thomas:
Thanks.

PR #784, opened 2016-04-05
python evaluation of expressions in roslaunch

rhaschke:
Retargeted #762 to kinetic.

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -46,6 +46,7 @@ except ImportError:
 
 import rosgraph.names
 import rospkg
+from roslaunch.loader import convert_value
 
 _rospack = None
 
@@ -60,6 +61,12 @@ class ArgException(SubstitutionException):
     """
     pass
 
+def _eval_env(name):
+    try:
+        return os.environ[name]
+    except KeyError as e:
+        raise SubstitutionException("environment variable %s is not set" % str(e))
+
 def _env(resolved, a, args, context):
     """
     process $(env) arg
@@ -69,10 +76,12 @@ def _env(resolved, a, args, context):
     """
     if len(args) != 1:
         raise SubstitutionException("$(env var) command only accepts one argument [%s]"%a)
-    try:
-        return resolved.replace("$(%s)"%a, os.environ[args[0]])
-    except KeyError as e:
-        raise SubstitutionException("environment variable %s is not set"%str(e))
+    return resolved.replace("$(%s)" % a, _eval_env(args[0]))
+
+def _eval_optenv(name, default=''):
+    if name in os.environ:
+        return os.environ[name]
+    return default
 
 def _optenv(resolved, a, args, context):
     """
@@ -83,13 +92,15 @@ def _optenv(resolved, a, args, context):
     """
     if len(args) == 0:
         raise SubstitutionException("$(optenv var) must specify an environment variable [%s]"%a)
-    if args[0] in os.environ:
-        return resolved.replace("$(%s)"%a, os.environ[args[0]])
-    elif len(args) > 1:
-        return resolved.replace("$(%s)"%a, ' '.join(args[1:]))
-    else:
-        return resolved.replace("$(%s)"%a, '')
-    
+    return resolved.replace("$(%s)" % a, _eval_optenv(args[0], default=' '.join(args[1:])))
+
+def _eval_anon(id, anons):
+    if id in anons:
+        return anons[id]
+    resolve_to = rosgraph.names.anonymous_name(id)
+    anons[id] = resolve_to
+    return resolve_to
+
 def _anon(resolved, a, args, context):
     """
     process $(anon) arg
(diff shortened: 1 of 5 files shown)


gerkey:
+1

The test failures in `test_rosbag` look to me like they're unrelated. The underlying failure is a path/permissions issue, e.g.: `Error opening file: /tmp/test_rosbag_random_record_54321.bag`. @dirk-thomas, is that due to some issue on the build machines?

gerkey:
Btw, the test failures are indeed unrelated to this change: #787.

gerkey:

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -46,6 +46,7 @@
 
 import rosgraph.names
 import rospkg
+import loader
```
dirk-thomas: Is this referring to the relative module `roslaunch.loader`? Then please use absolute import names (optionally with `from` and updating the usage) and insert them in alphabetical order. E.g.:

```
import rosgraph.names
from roslaunch.loader import convert_value
import rospkg
```

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -69,10 +76,13 @@ def _env(resolved, a, args, context):
...
-    try:
-        return resolved.replace("$(%s)"%a, os.environ[args[0]])
-    except KeyError as e:
-        raise SubstitutionException("environment variable %s is not set"%str(e))
+    return resolved.replace("$(%s)" % a, _eval_env(args[0]))
+
+def _eval_optenv(name, default=''):
+    if name in os.environ:
+        return os.environ[name]
+    else:
```
dirk-thomas: No `else` needed here. Just unindent the following `return`.

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -83,13 +93,16 @@ def _optenv(resolved, a, args, context):
...
-    if args[0] in os.environ:
-        return resolved.replace("$(%s)"%a, os.environ[args[0]])
-    elif len(args) > 1:
-        return resolved.replace("$(%s)"%a, ' '.join(args[1:]))
+    return resolved.replace("$(%s)" % a, _eval_optenv(args[0], default=' '.join(args[1:])))
+
+def _eval_anon(id, anons):
+    if id in anons:
+        return anons[id]
     else:
```
dirk-thomas: No `else` needed here.

tools/roslaunch/src/roslaunch/substitution_args.py
```diff
@@ -253,21 +269,50 @@ def _arg(resolved, a, args, context):
...
+    if 'anon' not in context:
+        context['anon'] = {}
+    if 'arg' not in context:
+        context['arg'] = {}
+
+    # inject correct anon context
+    def _eval_anon_context(id): return _eval_anon(id, anons=context['anon'])
+    # inject arg context
+    def _eval_arg_context(name): return loader.convert_value(_eval_arg(name, args=context['arg']), 'auto')
+    functions = dict(anon = _eval_anon_context, arg = _eval_arg_context)
```
dirk-thomas: Please remove the spaces around the equal sign for keyword args (PEP8).

dirk-thomas:
Beside the small comments this looks good. Can you please squash the commits after updating.

tools/roslaunch/src/roslaunch/substitution_args.py
eric-wieser: Aren't relative imports of the folloing form better, since they don't rely on `roslaunch` referring to the parent module (ie, if the pythonpath has two modules with that name)

```
from .loader import convert_value
```

dirk-thomas: PEP 8 recommends absolute imports (see https://www.python.org/dev/peps/pep-0008/#imports). If you use the `from` syntax it needs to be absolute.

(Also I don't think any code within this repo uses relative imports currently?)

rhaschke:
@dirk-thomas Fixed the PEP8 issues. Shall I squash? It's getting harder to follow the changes then... I guess, it's better when you do during merging.

dirk-thomas:
If I squash and merge using the GitHub interface the single commit won't have any reference to this ticket. Therefore I think it would be better that you squash the commits on this branch and we do a normal merge commit.

rhaschke:
@dirk-thomas Squashed into a single commit as requested.

tools/roslaunch/src/roslaunch/substitution_args.py
eric-wieser: >  If you use the from syntax it needs to be absolute

Not sure what you mean by that, but thanks for pointing out to me that PEP8 says _"Standard library code should avoid complex package layouts and always use absolute imports"_

dirk-thomas:
Can you please make sure to document this new functionality in the wiki for the _Kinetic_ version of roslaunch (and post the link here). Even though it doesn't require any migration step I think it would also be great to mentioned on this page: http://wiki.ros.org/kinetic/Migration

eric-wieser:
Would it be desirable to expose `math` in the `_eval_dict` (or more questionably everything within it), so that `$(eval math.pi)` can be used in parameters?

rhaschke:
> Would it be desirable to expose |math| in the |_eval_dict|, so that 
> |$(eval math.pi)| can be used in parameters?
> 
> If that is desired, it's easy to do. We have exposed math symbols as 
> well as ['list', 'dict', 'map', 'str', 'float', 'int'] in xacro.

eric-wieser:
I'd say there's a strong argument for making the environment be identical in `.launch` files and `.xacro` for the sake of consistency

rhaschke:
> Can you please make sure to document this new functionality in the 
> wiki for the /Kinetic/ version of roslaunch. Even though it doesn't 
> require any migration step I think it would also be great to mentioned 
> on this page: http://wiki.ros.org/kinetic/Migration
> 
> Done.

PR #786, opened 2016-04-11
Fix str conversion in encode_ros_handshake_header

benureau:
The previous code was not properly filtering unicode/str out of the header. Added the corresponding unit test.


benureau:
I realised too late that I was removing a lot of trailing whitespaces. Tell me if you want a cleaner commit.

dirk-thomas:
Thanks for the pull request. Yes, please remove any unrelated white space changes from the patch.

PR #788, opened 2016-04-11
Add missing inter-test dependency

### LINKED ISSUE #787: [test_rosbag] Tests rely on inter-test ordering
This kind of failure in `test_random_play`: http://build.ros.org/job/Kpr__ros_comm__ubuntu_xenial_amd64/46/testReport/__main__/RandomPlay/test_random_play/ can be traced to this console output:

```
08:04:33 [31m[FATAL] [1459868673.515029494]: Error opening file: /tmp/test_rosbag_random_record_54321.bag[0m
```

It turns out that the random_play test is assuming that the random_record test, which creates that bag file, has already run. Indeed, the record test is declared first in the CMakeLists.txt: https://github.com/ros/ros_comm/blob/kinetic-devel/test/test_rosbag/bag_migration_tests/CMakeLists.txt#L57-L60, but that's surely just wishful thinking on the part of some untrustworthy contributor from a long time ago: https://github.com/ros/ros_comm/commit/678a997ba34ce1fc3652632c4a50b663ae2cabe3.

I'm working on a fix now.

gerkey:
Fixes #787.

I'm not sure what caused this problem to suddenly pop up, but I would guess that it's a side effect of using a newer version of cmake/ctest that no longer runs tests in the order that they're declared in the CMakeLists.txt. In any case we never should have assumed that that was happening.

A more thorough cleanup step would be to restructure these three tests to not depend on each others' side effects in the file system. But adding the missing dependency in this manner will at least restore the previous behavior.

### DIFF
--- a/test/test_rosbag/bag_migration_tests/CMakeLists.txt
+++ b/test/test_rosbag/bag_migration_tests/CMakeLists.txt
@@ -55,9 +55,15 @@ catkin_add_nosetests(${PROJECT_BINARY_DIR}/test/migrate_test.py)
 configure_file(test/random_record.xml.in
   ${PROJECT_BINARY_DIR}/test/random_record.xml)
 add_rostest(${PROJECT_BINARY_DIR}/test/random_record.xml)
+# Make sure that the random_record test runs before either of the random_play
+# tests, both of which require a .bag file that the random_record test will
+# write into /tmp. We're making an assumption about the target names generated
+# by add_rostest(), but that naming scheme seems to be pretty stable.
 configure_file(test/random_play.xml.in
   ${PROJECT_BINARY_DIR}/test/random_play.xml)
-add_rostest(${PROJECT_BINARY_DIR}/test/random_play.xml)
+add_rostest(${PROJECT_BINARY_DIR}/test/random_play.xml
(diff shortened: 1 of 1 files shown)


wjwwood:
Seems like a reasonable to workaround to me. However, it would probably be better if the bag generation was done in each test if it doesn't already exist. For example, I imagine it makes it impossible to just run the test that was failing before (without first running the test that generates the bag).

gerkey:
Indeed, with this change, if you `make run_tests_test_rosbag_rostest_test_random_play.xml`, you'll first get a(nother) run of the random_record test. That seems to me a minor inconvenience and a practical trade-off compared to restructuring the tests.

wjwwood:
Sure, I'm in favor of merging as-is.

gerkey:
First try didn't work due to more-complex-than-expected target structure around the tests. f705e7f will hopefully do it, using the `DEPENDENCIES` keyword arg to `add_rostest()`

gerkey:
Hooray! http://build.ros.org/job/Kpr__ros_comm__ubuntu_xenial_amd64/49/

dirk-thomas:
Cherry-picked to _indigo-devel_: 535f16017054a6bf27515c62c626d3c0809e3feb

PR #789, opened 2016-04-12
Fix str conversion in encode_ros_handshake_header

benureau:
The previous code was not properly filtering unicode/str
out of the header. Added the corresponding unit test.

### DIFF
--- a/tools/rosgraph/src/rosgraph/network.py
+++ b/tools/rosgraph/src/rosgraph/network.py
@@ -398,10 +398,14 @@ def encode_ros_handshake_header(header):
     str_cls = str if python3 else unicode
     
     # encode all unicode keys in the header. Ideally, the type of these would be specified by the api
-    for k, v in sorted(header.items()):
-        if isinstance(k, str_cls): k = k.encode('utf-8')
-        if isinstance(v, str_cls): v = v.encode('utf-8')
-    
+    for k, v in header.items():
+        if isinstance(k, str_cls):
+            header.pop(k)
+            k = k.encode('utf-8')
+            header[k] = v
+        if isinstance(v, str_cls):
+            header[k] = v.encode('utf-8')
+
     fields = [k + b"=" + v for k, v in sorted(header.items())]
     s = b''.join([struct.pack('<I', len(f)) + f for f in fields])
     
--- a/tools/rosgraph/test/test_network.py
+++ b/tools/rosgraph/test/test_network.py
@@ -71,6 +71,21 @@ class NetworkTest(unittest.TestCase):
         encoded = encoded+struct.pack('<I', len(s))+s        
         assert struct.pack('<I', len(encoded))+encoded == \
             encode_ros_handshake_header(d)
+        if sys.version_info > (3, 0):
+            assert encode_ros_handshake_header({'a': 'b',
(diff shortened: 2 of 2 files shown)


humm:
Was PR #789, before I messed up branches trying to remove whitespace changes. So I opened a new one.

benureau:
Was PR #786, before I messed up with `git push` trying to remove whitespace changes. I opened this new one in its own branch.

eric-wieser:
Darn, that was careless of me

dirk-thomas:
Thank you for the patch. I have slightly updated it in #792 (while keep the commit in your name) to make it a bit simpler and easier to read.

eric-wieser:
It would make more sense here to define the API (at least for py3) as taking either `bytes` or `str` but not both. Which choice makes more sense? Should headers ever contain non-ascii characters?

dirk-thomas:
I would rather not spend time on revising the specified API. The current code should handle both input well.

eric-wieser:
I'd argue that the API is currently underspecified, especially for python 3. Python 3 makes a big point about enforcing the deliberate use of either `str` or `bytes`, and them not being interchangeable. Therefore, a python3 api should be strict about this distinction also.

Having said that, I get the impression that python3 support is not a big item on the roadmap, and so by association nor is being picky about the `str`/`bytes` distinction.

dirk-thomas:
I think it is "good" API even in Python 3 to allow passing in either `bytes` or `str` which will then be `UTF-8` encoded for the user.

Yes, there is currently no roadmap about Python 3 in ROS 1. We have made a significant push on that topic a year and a half ago. But then it was decided to not spend further resources on it. But any pull request improving Python 3 support (especially since some platforms use it by default) are highly appreciated.

benureau:
Having trudged through the installation of MORSE/ROS with python 3 (and on OS X, to make matters worse), I can definitely say that python 3 support is not a big item on the roadmap. But with packages such as the MORSE simulator that only support python 3, it should be. In the meantime, I'll keep posting patches for bugs I encounter and understand.

Thanks for all the attention and reactivity you brought to this PR.

eric-wieser:
ROS 2 is jumping ship to python 3 only, right?

dirk-thomas:
Yes, ROS 2 is Python 3 only.

eric-wieser:
> I think it is "good" API even in Python 3 to allow passing in either bytes or str which will then be UTF-8 encoded for the user.

I don't agree. Python 3 [makes a big distinction between these two types](https://docs.python.org/3.0/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit)

> Python 3.0 uses the concepts of _text_ and (binary) _data_ instead of Unicode strings and 8-bit strings. All text is Unicode; however _encoded_ Unicode is represented as binary data. The type used to hold text is `str`, the type used to hold data is `bytes`.

Are headers text or data? Should I be able to put a binary string in the header?  
- If they're data, then we have to commit to the api using `bytes`, otherwise `decode_ros_handshake_header` will raise a `UnicodeDecodeError` (this happens currently if you pass binary data in python 3).  
- If they're text, then we should commit to using `str`, to prevent the user from being able to try to send binary data.

PR #792, opened 2016-04-12
fix header encoding in encode_ros_handshake_header

dirk-thomas:
Replaces #789.


dirk-thomas:
Cherry-picked to _indigo-devel_: e7704b2dc083503a318c42a5a9ee0798d003a5f9

PR #795, opened 2016-04-20
the log dir itself should not be symlinked as latest

dirk-thomas:
See https://github.com/ros/ros_comm/commit/28f836135ce3e2ba3e05776cb9c7270411faa302#commitcomment-17183164


dirk-thomas:
Backported to _indigo-devel_: 4d215cea906d5e20f18e33a171f780eccfe9280d

Issue #748, opened 2016-02-22
ROS_LOG_THROTTLE not working when ros::time goes backward


hfarazi:
Sometimes I need to play a bag file in a loop (with -l) in those times any ROS_LOG_THROTTLE is not working.

It seems that the problem is related to the definition of last_hit in ROS_LOG_THROTTLE:
static double last_hit = 0.0;

My suggestion is reset the last_hits when an empty message arrived in /reset_time topic.

Another possibility is using wall time.

hfarazi:
it can be easily fixed by using:

#define ROS_LOG_THROTTLE2(rate, level, name, ...) \
  do \
  { \
    ROSCONSOLE_DEFINE_LOCATION(true, level, name); \
    static double last_hit = 0.0; \
    ::ros::WallTime now = ::ros::WallTime::now(); \
    if (ROS_UNLIKELY(**rosconsole_define_location__enabled) && ROS_UNLIKELY(last_hit + rate <= now.toSec())) \
    { \
      last_hit = now.toSec(); \
      ROSCONSOLE_PRINT_AT_LOCATION(__VA_ARGS**); \
    } \
  } while(0)

dirk-thomas:
Please consider to create a pull request for a proposed code change. That way the patch can be compiled and tested automatically.

dirk-thomas:
This issue was moved to ros/rosconsole#6

PR #796, opened 2016-04-25
Change ROS_LOG_THROTTLE in a way that it can work even if ros::time i…

hfarazi:
Change ROS_LOG_THROTTLE in a way that it can work even if ros::time is affected by /use_sim_time
More detail: https://github.com/ros/ros_comm/issues/748

PR #797, opened 2016-04-25
Change ROS_LOG_THROTTLE in a way that it can work even if ros::time

hfarazi:
Change ROS_LOG_THROTTLE in a way that it can work even if ros::time is affected by /use_sim_time
More detail: https://github.com/ros/ros_comm/issues/748


dirk-thomas:
Closing as a duplicate of #796.

If the patch is successfully released into Kinetic it can be considered to be backported.

PR #799, opened 2016-05-04
explicitly declare rosout message dependencies

codebot:
rosgraph_msgs was used in rosout.cpp but not included in the find_package() call, and was not declared as a run_depend. I'm not quite sure how it was working before. :confused: :poultry_leg:

### DIFF
--- a/tools/rosout/CMakeLists.txt
+++ b/tools/rosout/CMakeLists.txt
@@ -5,7 +5,7 @@ if(NOT WIN32)
   set_directory_properties(PROPERTIES COMPILE_OPTIONS "-Wall;-Wextra")
 endif()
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thanks for the patch. `roscpp` which is a build and run dependency and was also find_packag()-ed before depends on `rosgraph_msgs` and so this package got the dependency implicitly. Anyway it is good to make it explicit.

Just out of curiousity: did you had any issue with it which made you notice this?

dirk-thomas:
Cherry-picked to _indigo-devel_: 66a5f46b2d21980c8dda06f6e65798192d9a8f3c

dirk-thomas:
While I merged this with one failing test (which was unrelated and already happening on master) I followed up on this and was able to find the reason. See #803 for more information.

PR #803, opened 2016-05-05
update test to match new genpy behavior

dirk-thomas:
The bag embeds the message definition and until recently `genpy` was always append a newline at the end. To address a problem related to that `genpy` has been changed to not add any extra newline (https://github.com/ros/genpy/commit/4f7755b3863755db8ed6f8e8549ae26e9f650526). So the resulting header in the bagfile is now one byte shorter.


dirk-thomas:
Cherry-picked to _indigo-devel_: 2b2355a064e7f42901deac86aed382c1a5edf7c1

PR #806, opened 2016-05-17
Allow saving timestamp-less messages to Cache (using rospy.Time.now() as their timestamp).

peci1:
Retargeting #636 to kinetic-devel.

### DIFF
--- a/utilities/message_filters/src/message_filters/__init__.py
+++ b/utilities/message_filters/src/message_filters/__init__.py
@@ -104,13 +104,14 @@ class Cache(SimpleFilter):
         self.incoming_connection = f.registerCallback(self.add)
 
     def add(self, msg):
-        # Cannot use message filters with non-stamped messages
+        stamp = rospy.Time.now()
         if not hasattr(msg, 'header') or not hasattr(msg.header, 'stamp'):
-            rospy.logwarn("Cannot use message filters with non-stamped messages")
-            return
+            rospy.logdebug("Message saved to Cache has no header.stamp field. It will be assigned a timestamp\
+                based on the time it was saved to the Cache.")
+        else:
+            stamp = msg.header.stamp
 
         # Insert sorted
-        stamp = msg.header.stamp
         self.cache_times.append(stamp)
         self.cache_msgs.append(msg)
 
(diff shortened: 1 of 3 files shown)


utilities/message_filters/src/message_filters/__init__.py
```diff
@@ -125,19 +136,22 @@ def add(self, msg):
     def getInterval(self, from_stamp, to_stamp):
         """Query the current cache content between from_stamp to to_stamp."""
         assert from_stamp <= to_stamp
-        return [m for m in self.cache_msgs
-                if m.header.stamp >= from_stamp and m.header.stamp <= to_stamp]
+
+        return [msg for (msg, time) in zip(self.cache_msgs, self.cache_times)
+                if from_stamp <= time <= to_stamp]
```
dirk-thomas, line 140: This is not valid logic. The first operator results in a boolean which is then used in the second comparison.

utilities/message_filters/src/message_filters/__init__.py
```diff
@@ -99,18 +102,26 @@ def __init__(self, f, cache_size = 1):
...
         if not hasattr(msg, 'header') or not hasattr(msg.header, 'stamp'):
-            rospy.logwarn("Cannot use message filters with non-stamped messages")
-            return
+            if self._allow_headerless:
+                stamp = rospy.Time.now()
+            else:
+                rospy.logwarn("Cannot use message filters with non-stamped messages. "
+                              "Use the 'allow_headerless' constructor option to "
+                              "auto-assign ROS time to headerless messages.")
+                return
```
dirk-thomas, line 118: I would suggest to turn around the condition to error / return early:

```
if not self._allow_headerless:
    rospy.logwarn("Cannot use message filters with non-stamped messages. "
                  "Use the 'allow_headerless' constructor option to "
                  "auto-assign ROS time to headerless messages.")
    return
stamp = rospy.Time.now()
```

Same for the other class.

utilities/message_filters/src/message_filters/__init__.py
peci1, line 140: If I understand you objection correctly, I think it is incorrect :) Check http://stackoverflow.com/questions/26502775/pycharm-simplify-chained-comparison

utilities/message_filters/src/message_filters/__init__.py
```diff
@@ -87,10 +87,13 @@ class Cache(SimpleFilter):
...
     are cached in a ring buffer, from which time intervals of the cache
-    can then be retrieved by the client.
+    can then be retrieved by the client. The ``allow_headerless``
+    option specifies whether to allow storing headerless messages with
+    current ROS time instead of timestamp. You should avoid this as
+    much as you can, since the delays are unpredictable.
     """
 
-    def __init__(self, f, cache_size = 1):
+    def __init__(self, f, cache_size = 1, allow_headerless = False):
```
dirk-thomas: Can you please remove the spaces around the `=` sign (following PEP 8) (also for the other keyword).

utilities/message_filters/src/message_filters/__init__.py
```diff
@@ -99,18 +102,26 @@ def __init__(self, f, cache_size = 1):
         # Array to store msgs times, auxiliary structure to facilitate
         # sorted insertion
         self.cache_times = []
+        # Whether to allow storing headerless messages with current ROS
+        # time instead of timestamp.
+        self._allow_headerless = allow_headerless
```
dirk-thomas: I think it would be ok to make the property "public" so that users can check the property later. So I would propose to remove the leading underscore.

Same for the other class.

utilities/message_filters/src/message_filters/__init__.py
dirk-thomas, line 140: Omg scary syntax, I guess I just learned something new. Thanks for the reference. Please forget my comment.

dirk-thomas:
Thank you for iterating on this. I have squashed and cherry-picked your changes in 005191b76ae4e0e6d2dbe1aa544c42848f0c11de.

gavanderhoorn:
Just came across this and it seems like a useful feature.

@peci1: will you also update the wiki on [message_filters](http://wiki.ros.org/message_filters)? I do not see this mentioned anywhere right now.

peci1:
@gavanderhoorn Thanks for the hint, I've edited the page accordingly.

PR #807, opened 2016-05-19
Check for pointer return from createOutgoing and fail requestTopic if it is NULL.

jasonimercer:
It is possible for createOutgoing to return a null smart pointer if the connect fails. Connect can fail if it cannot resolve a hostname (possible in a mesh network if a client disconnects), if a file descriptor limit has been hit or if !isHostAllowed.

`
#6  0x00007fc4b3db5509 in ros::TopicManager::requestTopic (this=this@entry=0x2239060, topic="/move_base/PathTracker/plan_spline", protos=..., ret=...) at ../../../../../roscpp-release-indigo-roscpp-1-11-16-0/src/libros/topic_manager.cpp:669

669 ../../../../../roscpp-release-indigo-roscpp-1-11-16-0/src/libros/topic_manager.cpp: No such file or directory.
(gdb) print(transport)
$39 = {px = 0x0, pn = {pi_ = 0x0}}
`

Note, line 669 mentioned above is

`
      connection_manager_->udprosIncomingConnection(transport, h);
`

in indigo after createOutgoing()

The code in topic_manager assumes that transport is valid but we have encountered a case where it is not and a crash results. The crash happens when the transport sets the read callback in connection.cpp.

@dirk-thomas 

FYI @mikepurvis

### DIFF
--- a/clients/roscpp/src/libros/topic_manager.cpp
+++ b/clients/roscpp/src/libros/topic_manager.cpp
@@ -656,6 +656,11 @@ bool TopicManager::requestTopic(const string &topic,
       int max_datagram_size = proto[4];
       int conn_id = connection_manager_->getNewConnectionID();
       TransportUDPPtr transport = connection_manager_->getUDPServerTransport()->createOutgoing(host, port, conn_id, max_datagram_size);
+      if (!transport)
+      {
+        ROSCPP_LOG_DEBUG("Error creating outgoing transport for [%s:%d]", host.c_str(), port);
+        return false;
+      }
       connection_manager_->udprosIncomingConnection(transport, h);
 
       XmlRpcValue udpros_params;


clients/roscpp/src/libros/topic_manager.cpp
```diff
@@ -656,6 +656,11 @@ bool TopicManager::requestTopic(const string &topic,
       int max_datagram_size = proto[4];
       int conn_id = connection_manager_->getNewConnectionID();
       TransportUDPPtr transport = connection_manager_->getUDPServerTransport()->createOutgoing(host, port, conn_id, max_datagram_size);
+      if (!transport)
+      {
+        ROSCPP_LOG_DEBUG("Error creating outgoing transport for [%s:%d]", host.c_str(), port);
+        return false;
+      }
```
mikepurvis, line 663: What are the consequences of returning `false` here? Should the log message be a warn/error?

jasonimercer, line 663: The DEBUG/return false seems to be convention in this method and can happen 4 other times in this else block. 

My expectation is that the code calling this will decide if a false return needs to be elevated to an error or if there are alternative strategies.

mikepurvis:
+1

efernandez:
ping @dirk-thomas @tfoote Could we also have this cherry-picked into `indigo-devel`?

dirk-thomas:
Thanks for the patch.

For the record: `createOutgoing` might return an invalid pointer (https://github.com/ros/ros_comm/blob/82024ac2149b47884e59fe2165085c545d579aef/clients/roscpp/src/libros/transport/transport_udp.cpp#L693) which isn't checked before and passed on to `ConnectionManager::udprosIncomingConnection` and then to `Connection::initialize` where is will assert (https://github.com/ros/ros_comm/blob/82024ac2149b47884e59fe2165085c545d579aef/clients/roscpp/src/libros/connection.cpp#L71). This changes checks the returned pointer and signals the failure to the caller.

I agree that this should be backported since it is clearly fix and I don't think there is a chance for a regressions here.

dirk-thomas:
Cherry-picked to _indigo-devel_: 25dca9e9153c9a3e173ea6314a26e81b8e6cdcb3

Issue #804, opened 2016-05-11
rosnode info non verbose option flag


ghost:
Suggesting a not so verbose option for rosnode info - one that just prints publications, subscriptions and service and leaves out contacting the node to print other information like topic direction, transport and destination. 

Most times, we're only looking at the publications/subscriptions or services.

dirk-thomas:
That sounds like a useful option. Can you please create a pull request for this feature.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #809, opened 2016-05-22
Add a non verbose (quiet) mode for rosnode info

spaghetti-:
Addresses https://github.com/ros/ros_comm/issues/804

Default behavior not changed to avoid regression.

### DIFF
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -554,7 +554,7 @@ def get_node_connection_info_description(node_api, master):
         raise ROSNodeIOException("Communication with node[%s] failed!"%(node_api))
     return buff
 
-def rosnode_info(node_name):
+def rosnode_info(node_name, quiet=False):
     """
     Print information about node, including subscriptions and other debugging information. This will query the node over the network.
     
@@ -578,10 +578,10 @@ def rosnode_info(node_name):
     if not node_api:
         print("cannot contact [%s]: unknown node"%node_name, file=sys.stderr)
         return
-    
-    print("\ncontacting node %s ..."%node_api)
 
-    print(get_node_connection_info_description(node_api, master))
+    if not quiet:
+        print("\ncontacting node %s ..." % node_api)
+        print(get_node_connection_info_description(node_api, master))
 
 # backwards compatibility (deprecated)
 rosnode_debugnode = rosnode_info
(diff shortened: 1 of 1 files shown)


spaghetti-:
@dirk-thomas This file contains a lot of trailing whitespace which my editor removed by default, but I figured I'd open another pull request for that it makes reading this particular diff pretty annoying. Do let me know if you're interested.

https://github.com/spaghetti-/ros_comm/commit/0d51e75032a3e6ea824018d928119f1a759ac36b

tools/rosnode/src/rosnode/__init__.py
```diff
@@ -554,7 +554,7 @@ def get_node_connection_info_description(node_api, master):
         raise ROSNodeIOException("Communication with node[%s] failed!"%(node_api))
     return buff
 
-def rosnode_info(node_name):
+def rosnode_info(node_name, non_verbose):
```
dirk-thomas: Please avoid inverted logic. Either call the argument `quiet` or `verbose`.

Adding a positional argument breaks existing code. Please add the new argument as an optional keyword argument. And the default value should not change the current behavior.

tools/rosnode/src/rosnode/__init__.py
```diff
@@ -613,12 +613,17 @@ def _rosnode_cmd_info(argv):
     Implements rosnode 'info' command.
     """
     args = argv[2:]
-    parser = OptionParser(usage="usage: %prog info node1 [node2...]", prog=NAME)
+    parser = OptionParser(usage="usage: %prog info [options] node1 [node2...]",
+                          prog=NAME)
+    parser.add_option("-q", "--quiet",
+                      dest="non_verbose", default=False,
```
dirk-thomas: Please avoid `dest`. The option name should already be clear enough to map to a variable. Also the `default` argument is not necessary for a boolean action.

tools/rosnode/src/rosnode/__init__.py
```diff
@@ -613,12 +613,17 @@ def _rosnode_cmd_info(argv):
...
+                      action="store_true",
+                      help="non-verbose mode")
```
dirk-thomas: Please describe the difference of the output in the help text.

dirk-thomas:
Thanks for avoiding any whitespace changes. The are usually not being considered for merge anyway since it makes backporting changes between branches unnecessary cumbersome.

spaghetti-:
thanks for the comments - fixed.

dirk-thomas:
Thanks for the patch and update.

PR #811, opened 2016-05-27
Changed calling_in_this_thread restore mechanism in callOneCB from try-catch to RAII.

jasonimercer:
This PR prevents the try-catch on callOneCB from masking the source of the original exception in a callback. We have encountering this situations here at Clearpath and the community has reported similar problems:

http://answers.ros.org/question/167386/node-crashes-on-spinonce/

@dirk-thomas @mikepurvis

### DIFF
--- a/clients/roscpp/src/libros/callback_queue.cpp
+++ b/clients/roscpp/src/libros/callback_queue.cpp
@@ -34,6 +34,7 @@
 
 #include "ros/callback_queue.h"
 #include "ros/assert.h"
+#include <boost/scope_exit.hpp>
 
 namespace ros
 {
@@ -371,7 +372,17 @@ CallbackQueue::CallOneResult CallbackQueue::callOneCB(TLS* tls)
     tls->calling_in_this_thread = id_info->id;
 
     CallbackInterface::CallResult result = CallbackInterface::Invalid;
-    try {
+
+    {
+      // Ensure that thread id gets restored, even if callback throws.
+      // This is done with RAII rather than try-catch so that the source
+      // of the original exception is not masked in a crash report.
+      BOOST_SCOPE_EXIT(&tls, &last_calling)
+      {
+        tls->calling_in_this_thread = last_calling;
+      }
+      BOOST_SCOPE_EXIT_END
+
       if (info.marked_for_removal)
       {
         tls->cb_it = tls->callbacks.erase(tls->cb_it);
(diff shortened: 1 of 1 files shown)


mikepurvis:
LGTM.

clients/roscpp/src/libros/callback_queue.cpp
```diff
@@ -382,14 +393,6 @@ CallbackQueue::CallOneResult CallbackQueue::callOneCB(TLS* tls)
         result = cb->call();
       }
     }
-    catch (std::exception&)
-    {
-      // ensure that thread id gets restored, even in case of an exception
-      tls->calling_in_this_thread = last_calling;
-      throw;
```
dirk-thomas, line 389: `throw` without an exception argument re-throws the current exception and doesn't create a new one. Therefore I don't see why this change is necessary. Can you please provide a minimal example which demonstrates why the current code is not sufficient and "masks" the source of the original exception.

jasonimercer, line 389: Sure, the community link provided shows an exception that was thrown by user code but the strack trace in gdb leads to callOneCB. The exception is the same but the source of the exception is lost. 

The last remark of the user in that post was "The stack trace I posted was in Debug mode. It's strange that it didn't lead me to my callback function."

I'll put together a minimal example.

dirk-thomas, line 389: Nvm, I didn't think about it before. While it passes along the same exception it will change the reported line in e.g. gdb which is clearly undesired.

dirk-thomas:
Do you want this to be backported to Indigo / Jade as well?

dirk-thomas:
Thanks for the patch!

Do you want this to be backported to Indigo / Jade as well?

jasonimercer:
A backport isn't required for our needs. I'll let you decide if it is of value for the community and timeline accordingly. 

Thanks for merging this into kinetic-devel!

dirk-thomas:
@ros/ros_team I would like your feedback if you think this should be backported. While the patch isn't necessary it would improve the stack trace in the case of problems for users. If we agree that this change doesn't pose a risk for regressions I would move forward with the backport.

wjwwood:
I think should be safe to backport. I can't think of anyway this would affect existing code.

mikepurvis:
We've been running it already in our overnight testing— it's great getting full stack traces in our crash reporter!

PR #812, opened 2016-05-29
[rospy] Implement rospy.logXXX_throttle

wkentaro:
Periodicall logging feature for python client.

### DIFF
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -37,6 +37,8 @@
 
 
 import atexit
+import cPickle as pickle
+import inspect
 import logging
 import os
 import signal
@@ -149,6 +151,57 @@ logerr = logging.getLogger('rosout').error
 logerror = logerr # alias logerr
 
 logfatal = logging.getLogger('rosout').critical
+
+
+class LoggingThrottle(object):
+
+    last_logging_time_table = {}
+
+    def __call__(self, id, logging_func, period, msg):
+        """Do logging specified message periodically.
+
+        - id (str): Id to identify the caller
+        - logging_func (function): Function to do logging.
+        - period (float): Period to do logging in second unit.
+        - msg (object): Message to do logging.
+        """
+        now = rospy.Time.now()
+
+        last_logging_time = self.last_logging_time_table.get(id)
+
+        if (last_logging_time is None or
+              (now - last_logging_time) > rospy.Duration(period)):
+            logging_func(msg)
+            self.last_logging_time_table[id] = now
+
+
+_logging_throttle = LoggingThrottle()
+
+
+def logdebug_throttle(period, msg):
+    id = pickle.dumps(inspect.stack()[1][1:])
+    _logging_throttle(id, logdebug, period, msg)
+
+
+def loginfo_throttle(period, msg):
+    id = pickle.dumps(inspect.stack()[1][1:])
+    _logging_throttle(id, loginfo, period, msg)
+
+
+def logwarn_throttle(period, msg):
+    id = pickle.dumps(inspect.stack()[1][1:])
+    _logging_throttle(id, logwarn, period, msg)
+
+
+def logerr_throttle(period, msg):
+    id = pickle.dumps(inspect.stack()[1][1:])
+    _logging_throttle(id, logerr, period, msg)
+
+
+def logfatal_throttle(period, msg):
+    id = pickle.dumps(inspect.stack()[1][1:])
+    _logging_throttle(id, logfatal, period, msg)
+
 
 #########################################################
 # CONSTANTS
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -182,28 +182,28 @@ _logging_throttle = LoggingThrottle()
 
 
 def logdebug_throttle(period, msg):
-    id = pickle.dumps(inspect.stack()[1][1:])
-    _logging_throttle(id, logdebug, period, msg)
+    id_ = pickle.dumps(inspect.stack()[1][1:])
+    _logging_throttle(id_, logdebug, period, msg)
 
 
 def loginfo_throttle(period, msg):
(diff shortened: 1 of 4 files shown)


clients/rospy/src/rospy/core.py
```diff
@@ -149,6 +151,57 @@ def rospywarn(msg, *args):
...
+        """Do logging specified message periodically.
+
+        - id (str): Id to identify the caller
+        - logging_func (function): Function to do logging.
+        - period (float): Period to do logging in second unit.
+        - msg (object): Message to do logging.
+        """
+        now = rospy.Time.now()
+
+        last_logging_time = self.last_logging_time_table.get(id)
```
dirk-thomas: Is the id sufficient for this? From the following code I would expect both messages to be printed every three seconds (which I would assume is not the case currently?):

```
while True:
    loginfo_throttle(3, 'foo') or loginfo_throttle(3, 'foo')
```

clients/rospy/src/rospy/core.py
```diff
@@ -37,6 +37,8 @@
 
 
 import atexit
+import cPickle as pickle
```
dirk-thomas: Doesn't work with Python 3.

clients/rospy/src/rospy/core.py
```diff
@@ -149,6 +151,57 @@ def rospywarn(msg, *args):
...
+              (now - last_logging_time) > rospy.Duration(period)):
+            logging_func(msg)
+            self.last_logging_time_table[id] = now
+
+
+_logging_throttle = LoggingThrottle()
+
+
+def logdebug_throttle(period, msg):
+    id = pickle.dumps(inspect.stack()[1][1:])
```
dirk-thomas: Please avoid using reserved words for variable names.

test/test_rospy/test/rostest/test_rospy_client_online.py
```diff
@@ -100,6 +100,55 @@ def test_log(self):
             rospy.logfatal("test 4")            
             self.assert_("[FATAL]" in sys.stderr.getvalue())            
             self.assert_("test 4" in sys.stderr.getvalue())            
+
+            # logXXX_throttle
+            for i in xrange(3):
```
dirk-thomas: Doesn't work with Python 3.

clients/rospy/src/rospy/core.py
wkentaro: Maybe I need to get more detailed position where the function is called.
Currently, the `inspect.stack()[1][1:]` returns only line where the function is called:
Because there is the case where the logging functions, period and msg are all same.

```
% cat spam.py
import inspect


def loginfo_throttle():
    print(inspect.stack()[1])


def main():
    loginfo_throttle(); loginfo_throttle()


if __name__ == '__main__':
    main()

% python spam.py 
(<frame object at 0x7f3d6b2ed3a0>, 'spam.py', 9, 'main', ['    loginfo_throttle(); loginfo_throttle()\n'], 0)
(<frame object at 0x7f3d6b2ed3a0>, 'spam.py', 9, 'main', ['    loginfo_throttle(); loginfo_throttle()\n'], 0)
```

wkentaro: But currently I don't know how to get the called location with information about lines and columns.
`inspect.stack()[1][1:]` only returns the line and line number.

wkentaro: Posted on stackoverflow to asking about this: http://stackoverflow.com/questions/37606094/is-there-way-to-get-function-called-column-location-not-only-file-and-line-infor

gavanderhoorn:
Just a data point: I've looked into this some time in the past as well, and could not arrive at a solution that was both portable (`inspect` and / or `sys._getframe()` are not supported 'everywhere' and may not return the same information on all platforms / runtimes) and relatively straightforward to implement.

See [rospy: throttling logmessage rate?](http://answers.ros.org/question/94016) for the question on ROS Answers and (my own) answer. I'm not really a Python expert, so it is very well possible that I missed something or am just ignorant.

It would be really nice if this could be fixed though, so +1 for working on it @wkentaro.

wkentaro:
@gavanderhoorn In what cases, the `inspect` and `sys._getframe()` do not work? On windows?

gavanderhoorn:
I didn't say it doesn't work, I said it doesn't necessarily return the same information, and in the same format. I don't have my full notes anymore (it was 3 years ago), but iirc, some Python interpreters either don't return the required info, or don't return everything, or in a slightly different format. That makes it difficult to implement something that always works.

Now how much of a problem that is, I don't know. If we are happy with it working on the regular / official Python interpreters, fine by me. I just didn't want to implement something with if/else trees and / or special cases.

clients/rospy/src/rospy/core.py
wkentaro: I found the way to get column number where the function is called.
http://stackoverflow.com/questions/37606094/is-there-way-to-get-function-called-column-location-not-only-file-and-line-infor/37614347#37614347

wkentaro:
@dirk-thomas I have done the implementation. Could you please review again?

dirk-thomas:
As long as the `caller_id` is unique for every invocation of these new log functions this should be working. It tried several different way to "bypass" it but it worked flawless.

The `caller_id` doesn't have to be "the same" for different Python interpreters as long as the result is unique this should work.

Thanks for iterating on this. I have squashed and cherry-picked this in 5daef5631dc96825300769e437b6cf24a1c653a1. Can you please update the [wiki page](http://wiki.ros.org/rospy/Overview/Logging) to add some documentation about these new functions mentioning that they are new in Kinetic and comment with a link to the diff here.

wkentaro:
Thanks for the review.
I updated the wiki: http://wiki.ros.org/action/diff/rospy/Overview/Logging?action=diff&rev1=18&rev2=19

BTW, Is that possible to add this feature also to Indigo and Jade?

PR #816, opened 2016-06-05
More detailed help string for rostopic echo -p

spaghetti-:
I think that it is very obvious for people to overlook this very useful functionality as its in the niche spot between being easy to write your own script to do it vs googling for implementations.


dirk-thomas:
Thanks for the patch.

PR #818, opened 2016-06-10
Move queue operations to after debug message so that tenses of message match reality.

jasonimercer:
Before this change a queue with a limit of 1 element would report the following debug message:

Incoming queue full for topic "baz".  Discarding oldest message (current queue size [0])

which implies that the queue is full when it holds 0 elements and that it will discard a message even without elements. 

After this change you'll get the message:

Incoming queue full for topic "baz".  Discarding oldest message (current queue size [1])

(low-pri PR, quality of life only)

@dirk-thomas

### DIFF
--- a/clients/roscpp/src/libros/subscription_queue.cpp
+++ b/clients/roscpp/src/libros/subscription_queue.cpp
@@ -59,9 +59,6 @@ void SubscriptionQueue::push(const SubscriptionCallbackHelperPtr& helper, const 
 
   if(fullNoLock())
   {
-    queue_.pop_front();
-    --queue_size_;
-
     if (!full_)
     {
       ROS_DEBUG("Incoming queue full for topic \"%s\".  Discarding oldest message (current queue size [%d])", topic_.c_str(), (int)queue_.size());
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I am not sure if changing the order makes the message much clearer. I would read it as events in temporal order and with that the current number makes sense:
1. the incoming queue is full
2. the oldest message is discarded
3. the queue size is now N

jasonimercer:
If you are happy with the existing message then I'm OK with that. I found it difficult to interpret without looking at the code. I'm OK with this PR getting declined.

dirk-thomas:
Maybe the message can be improved to make it clear to everyone what it means. Can you suggest a rephrased message which is clear to you?

jasonimercer:
The following message could be used in place of moving the queue logic around. 

Incoming queue was full for topic "baz". Discarded oldest message (current queue size [0])

dirk-thomas:
Thanks. I have updated the message with your suggestions: c710fbba6d62c4fd7d442efcdc17f57eb48558e3

PR #828, opened 2016-07-08
Use logging Formatter, enabling printing exception info with exc_info=1

v-lopez:
This recovers the functionality provided by the python logging library,
which includes exception information if the kwarg exc_info is provided
with a True value.

I have seen no side effects


dirk-thomas:
Thank you for the patch. I have cherry-picked it to the _kinetic-devel_ branch: ee6a0a8e6a5ba85c2b77e1c7d7a5f2e6bdbcce67

PR #830, opened 2016-07-09
 add_rostest_gtest does now add the created gtest-target as a depende…

cwecht:
…ncy to the created rostest. This insures that on running catkin_make run_tests_PACKAGE_rostests_TEST underlaying binary is checked for changes and is rebuild if needed. This simplifies developing, because you just have to run the the test and everything will be up-to-date. Form now on you do not need to build and run separately.


dirk-thomas:

PR #831, opened 2016-07-13
Fix/107

### LINKED ISSUE #107: rospy: a single rospy latching publication continues to track dead nodes  (ros ticket #3038)
If you create a node which publishes once on a latching topic, and then bring a subscriber up and down, the list of connections in the publisher seems to just keep growing.

Terminal 1:
{{{
$ rostopic pub chatter std_msgs/String foo
}}}

Terminal 2:
{{{
$ rosrun rospy_tutorials listener.py ^C
$ rosrun rospy_tutorials listener.py ^C
$ rosrun rospy_tutorials listener.py ^C
$ rosnode info /rostopic_27665_1285637405612

...

Connections:
- topic: /chatter
  - to: /listener_27682_1285637409689
  - direction: outbound
  - transport: TCPROS
- topic: /chatter
  - to: /listener_27706_1285637411687
  - direction: outbound
  - transport: TCPROS
- topic: /chatter
  - to: /listener_27729_1285637413320
  - direction: outbound
  - transport: TCPROS
    }}}

The connection list seems to get reset on publication.

trac data:
- Owner: **kwc**
- Reporter: **leibs**
- Reported at: **Mon Sep 27 18:35:07 2010**
- URL: https://code.ros.org/trac/ros/ticket/3038

mgrrx:
Fixes #107 

I replaced the poll() behavior in rospy by epoll() which allows to check for EPOLLRDHUP events. Extract from the man page epoll_ctl:

> Stream socket peer closed connection, or shut down writing half of connection.  (This flag is especially useful for writing simple code to detect peer shutdown when using Edge Triggered monitoring.)

Using that event we can easily check for closed connections that were hung up by the remote node. I added that check to the regmanager to gracefully close those connections in its update loop. Especially latched publishers suffered from the old behavior as the messages are normally sent only once.

The fix works also on kinetic


clients/rospy/src/rospy/topics.py
```diff
@@ -189,7 +193,7 @@ class Poller(object):
     """
     def __init__(self):
         try:
-            self.poller = select.poll()
+            self.poller = select.epoll()
```
dirk-thomas, line 196: How does this affect cross platform compatibility? The [documentation](https://docs.python.org/2/library/select.html#select.epoll) only mentions Linux, what about OS X and Windows?

clients/rospy/src/rospy/impl/registration.py
```diff
@@ -318,6 +320,7 @@ def run(self):
     def _connect_topic_thread(self, topic, uri):
         try:
             code, msg, _ = self.handler._connect_topic(topic, uri)
+
```
dirk-thomas: Please avoid unrelated white space changes.

clients/rospy/src/rospy/topics.py
mgrrx, line 196: The Poller class in topics.py provides a unified interface for polling on all three platforms. I cleaned up the code a bit (48f4f0a) and added a fallback to select.poll for Linux <  2.5.44. I did not touch the code for OS X and Windows. So regarding the compatibility I'd say that this PR does not affect the compatibility.

clients/rospy/src/rospy/topics.py
```diff
@@ -439,6 +458,21 @@ def cleanup_cb_wrapper(s):
...
+        for fd in self.connection_poll.error_iter():
+            to_remove = [x for x in new_connections if x.fileno() == fd]
+
+            for x in to_remove:
+                removed_connection = True
+                rospydebug("removing connection to %s, connection error detected"%(x.endpoint_id))
+                self._remove_connection(new_connections, x)
+
+        if removed_connection:
+            self.connections = new_connections
```
dirk-thomas: Doesn't this require to lock `c_lock` since it is potentially modifying the `connections`?

test/test_rospy/nodes/listener_once.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2008, Willow Garage, Inc.
```
dirk-thomas: I guess this should be different since the code is not coming from Willow Garage.

test/test_rospy/test/rostest/test_latch_unsubscribe.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2008, Willow Garage, Inc.
```
dirk-thomas: I guess this should be different since the code is not coming from Willow Garage.

clients/rospy/src/rospy/impl/registration.py
```diff
@@ -306,6 +306,8 @@ def run(self):
                 if cond is not None:
                     cond.release()
 
+            get_topic_manager().check_all()
```
dirk-thomas, line 309: How "expensive" will this invocation be? It seems to require frequent locks.

mgrrx, line 309: Now that I rewrote the check function I'd say it is not expensive. The lock is now only acquired when one or more connections need to be removed. The only overhead that we add now is the check for connections that need to be closed which is O(1) with epoll/kqueue (with poll it would be O(n)!).

clients/rospy/src/rospy/impl/registration.py
dirk-thomas: Please avoid unrelated white space changes.

clients/rospy/src/rospy/topics.py
dirk-thomas: Doesn't this require to lock c_lock since it is potentially modifying the connections?

test/test_rospy/test/rostest/test_latch_unsubscribe.py
dirk-thomas: You could add a copyright line with your name / organization instead.

jspricke:
@dirk-thomas anything that keeps this from merging (to kinetic)?

dirk-thomas:
It is mostly due to not having the time to thoroughly test this patch. Hopefully I can get back to this in the near future.

alain-m:
We can reproduce the escalation of open file descriptors - `lsof -n | wc -l` - and sockets left in CLOSE_WAIT state - `sudo netstat -np | grep CLOSE_WAIT | wc -l` - on our indigo robots at Savioke during regular navigation tasks. The sockets in question are all associated with python nodes, as expected from reading https://github.com/ros/ros_comm/issues/107. 
We tested this PR on indigo and it effectively fixes the problem for us, thanks @mgrrx . We will test it on kinetic.

alain-m:
Any plans for releasing this fix soon @dirk-thomas? We've been using it on both indigo and kinetic robots since November and haven't found any regression. Let us know if we can help with specific tests.

mgrrx:
@alain-m thanks for testing this.
@mikepurvis I added the hitlist label ;) Can you integrate this PR into your kinetic-edge branch and test it at Clearpath? We've tested this on indigo here at Magazino for quiet some time now without regressions. I can create a PR for kinetic if this helps

dirk-thomas:
Since it has been tested by two parties for a while already I would be fine with merging a PR against kinetic-devel now. The kinetic-edge branch can be rebased against the latest master afterwards to also cover this until the next release.

jspricke:
@dirk-thomas I've moved this to kinetic-devel. No idea why Jenkins has problems with it. Can you have a look and merge it if acceptable?

dirk-thomas:

test/test_rospy/nodes/listener_once.py
```diff
@@ -0,0 +1,55 @@
...
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# Revision $Id: gossipbot.py 1013 2008-05-21 01:08:56Z sfkwc $
```
dirk-thomas: This line should probably be removed.

dirk-thomas:
I can't say why it didn't trigger the kinetic job after rebasing and "asking Jenkins nicely". I triggered the job manually (http://build.ros.org/view/Kpr/job/Kpr__ros_comm__ubuntu_xenial_amd64/272/) but it won't show up as a build status here when its finished.

dirk-thomas:
Your latest push to the branch triggered a new build anyway. Waiting for the result on that...

test/test_rospy/test/rostest/test_latch_unsubscribe.py
```diff
@@ -0,0 +1,83 @@
...
+PKG = 'test_rospy'
+NAME = 'test_latch_unsubscribe'
+
+import os
+import sys
+import unittest
+
+import psutil
+
+from std_msgs.msg import String
```
dirk-thomas, line 40: I was just about to merge this. But on a last view through the diff I saw that `std_msgs` is currently neither a runtime nor test dependency yet. Can you please add it as a `<test_depend>std_msgs</test_depend>`.

jspricke: Done

mikepurvis: ... but now it's upset because it was already listed as a build dependency. Maybe just switch this package to format 2?

dirk-thomas: I am sorry. I am only using format 2 nowadays. Can you please just remove the added test dependency again. No need to update to the new format in this PR.

jspricke:
No problem, done.

clients/rospy/src/rospy/topics.py
```diff
@@ -1136,6 +1166,15 @@ def close_all(self):
...
             self.subs.clear()        
+
+            
+    def check_all(self):
+        """
+        Check all registered publication and subscriptions.
+        """
+        with self.lock:
+            for t in chain(iter(self.pubs.values()), iter(self.subs.values())):
+                t.check()
```
dirk-thomas, line 1177: The `check()` might alter `self.pubs` / `self.subs` those this might result in a `RuntimeError: dictionary changed size during iteration`. See #1894 for the proposed fix.

Issue #610, opened 2015-04-27
rosmaster leaves sockets in CLOSE_WAIT state


lrasinen:
Somehow related to #495, but not reproducible as easily.

In some scenarios, likely involving a node (or nodes) restarting, rosmaster can leave sockets in a CLOSE_WAIT state, eventually exhausting the limit on open file descriptors and becoming unresponsive.

Our system has about 30 nodes and 150 topics, and when the system is run without restarting, there is a steady creep upwards in the amount of CLOSE_WAIT sockets in systemwide monitoring. Detailed examination assigns most of these to the rosmaster process. See picture.

![image](https://cloud.githubusercontent.com/assets/2937610/7347178/58e25b0c-ecee-11e4-951e-3acd63d433e6.png)

The sudden jump in the image has not been fully explained but it's likely associated with a hardware problem that caused parts of the system to restart repeatedly. Still, the trend is obvious.

After diagnosing the problem we've been trying to reproduce it with a simpler setup and/or collect logs, and the results are inconclusive. After comparing the rosmaster logs with lsof output, the leaks do appear to be related to ServerProxy objects.

It does not seem to matter whether the node uses roscpp or rospy, or whether it's a publisher-only or a subscriber-only. There are even instances of CLOSE_WAIT sockets being associated with nodes that do not publish or subscribe anything (they just read some parameters on startup).

We've run a belt-and-suspenders approach since hitting the limit, which involves a) nightly restarts and b) keeping the ServerProxy cache trimmed (patch below).

The nightly restarts have kept us from hitting the limit, but we're planning to run without them for a while to see if the ServerProxy patch helps. The default daily cycle already looks different:

![image](https://cloud.githubusercontent.com/assets/2937610/7347198/a951fc28-ecee-11e4-88d1-8c331c261dbe.png)

The spikes are due to rosbag restarting every hour so we get bags to manageable size.

When there is a leak, it seems to be contained:

![image](https://cloud.githubusercontent.com/assets/2937610/7347211/c7f68310-ecee-11e4-9d95-61bcdffe117c.png)

(These are system-wide numbers, rosmaster has 90 CLOSE_WAIT sockets)

```
--- util.py.old 2015-04-22 12:23:11.054085143 +0300
+++ util.py.new 2015-04-22 12:23:18.418073585 +0300
@@ -45,7 +45,12 @@
 except ImportError:
     from xmlrpclib import ServerProxy

-_proxies = {} #cache ServerProxys
+import collections
+import threading
[...]

PR #833, opened 2016-07-14
Using poll() in favor of select() in the XmlRPCDispatcher

mgrrx:
The XmlRPCDispatcher uses select() for checking the open file descriptors, but this function can't check more than FD_SETSIZE file descriptors in Linux. FD_SETSIZE is set to 1024 by default and increasing that value would require to rebuild the current kernel. I observed this bug during my work on the CLOSE_WAIT problem (#610, #831): Every now and then the rosout process died. I found out that it reached this limit of 1024 open connections (most of them were in the CLOSE_WAIT state) when relaunching our robot's bringup launch file several times. The easiest way to fix this bug is to use select's 'successor' poll() which isn't limited by 1024. From a performance point of view, both implementations are similar.

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -147,9 +147,8 @@ XmlRpcDispatch::work(double timeout)
         _sources.erase(thisIt);  // Stop monitoring this one
         if ( ! src->getKeepOpen())
           src->close();
-      } else if (newMask != (unsigned) -1) {
+      } else if (newMask != (unsigned) -1)
         thisIt->getMask() = newMask;
-      }
     }
 
     // Check whether to clear all sources
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -99,9 +99,9 @@ XmlRpcDispatch::work(double timeout)
     for (it=_sources.begin(); it!=_sources.end(); ++it, ++i) {
       int fd = it->getSource()->getfd();
       fds[i].fd = fd;
+      fds[i].events = 0;
       if (it->getMask() & ReadableEvent) fds[i].events |= POLLIN;
       if (it->getMask() & WritableEvent) fds[i].events |= POLLOUT;
-      if (it->getMask() & Exception)     fds[i].events |= POLLERR;
     }
 
     // Check for events
@@ -126,7 +126,7 @@ XmlRpcDispatch::work(double timeout)
         newMask &= src->handleEvent(ReadableEvent);
       if (fds[i].revents & POLLOUT)
         newMask &= src->handleEvent(WritableEvent);
-      if (fds[i].revents & POLLERR)
+      if (fds[i].revents & (POLLERR|POLLHUP|POLLNVAL))
         newMask &= src->handleEvent(Exception);
 
       // Find the source again.  It may have moved as a result of the way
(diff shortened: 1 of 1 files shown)


mgrrx:

dirk-thomas:
Can you please add a description to this pull request describing which exact use case this patch fixes. If possible also with a reproducible example.

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
```diff
@@ -81,83 +86,69 @@ XmlRpcDispatch::work(double timeout)
...
+        XmlRpcUtil::error("Error in XmlRpcDispatch::work: couldn't find source iterator");
+        continue;
+      }
+
+      if ( ! newMask) {
+        _sources.erase(thisIt);  // Stop monitoring this one
+        if ( ! src->getKeepOpen())
+          src->close();
+      } else if (newMask != (unsigned) -1) {
+        thisIt->getMask() = newMask;
```
mikepurvis, line 151: Since we're mutating these lines anyway, perhaps switch to ROS style for the braces? http://wiki.ros.org/CppStyleGuide#Formatting

mikepurvis:
Haven't tested yet, but this change seems reasonable to me, to get out from under the 1024 socket limit of Linux's `select` implementation. cf. https://www.ulduzsoft.com/2014/01/select-poll-epoll-practical-difference-for-system-architects/

Of course, fixing the CLOSE_WAIT issue will also have a big mediating effect, so maybe it will turn out not to be as necessary— but I could still see it as valid, especially for nodes which have to make a ton of subscribers or publishers, like rosbag record/play.

dirk-thomas:
Should we consider this for the "hitlist"?

dirk-thomas:
Should we consider this for the "hitlist"? I will add the label tentatively. Please feel free to remove it.

jspricke:
I've ported this to kinetic-devel, +1 for the hitlist.

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
```diff
@@ -81,84 +86,69 @@ XmlRpcDispatch::work(double timeout)
...
-    fd_set inFd, outFd, excFd;
-	  FD_ZERO(&inFd);
-	  FD_ZERO(&outFd);
-	  FD_ZERO(&excFd);
+    struct pollfd fds[_sources.size()];
 
-    int maxFd = -1;     // Not used on windows
     SourceList::iterator it;
-    for (it=_sources.begin(); it!=_sources.end(); ++it) {
+    std::size_t i = 0;
```
dirk-thomas, line 98: I would prefer the scope of this variable to be limited to the for loop since it isn't used afterwards.

The loop in line 119 can then declare its own local index.

jspricke: I don't agree. `i` is used the same way as `it`. Moving `i` into it's own scope would mean to add an extra pair of brackets and indention and breaking the overall flow. Or did I miss something?

dirk-thomas:
I have reverted this change in #981. Please see that PR for more details about the rational. Please create a new PR to apply these changes again. It looks like it will need some modifications to pass CI cleanly.

mikepurvis:
It's weird that there's a checkmark on c7c2087d0c121bbf6e77071b73eb8bec930b3de7, so it seems like it ran clean at least once. Maybe some tests need extra time buffers or something?

dirk-thomas:
It can also be that since that CI ran something else had changed. While each PR on its own passes it doesn't mean both being merged still passes. Just as one example why this is not  a 100% guarantee.

PR #835, opened 2016-07-14
rosbag_storage: make Bag constructor explicit

xqms:
This is a very small improvement to the rosbag API (adds "explicit" to the bag constructor)

Previously, it was possible to do something like this:

```
std::string bagFile = "...";
rosbag::Bag bag(bagFile);

rosbag::View view;
view.addQuery(bagFile, ...);
```

Note that bagFile is passed to View::addQuery, which does not cause a compiler error, since implicit conversion is allowed by the constructor. Using the view will cause a segfault, since the temporary bag will no longer be around.

I think that there are no good use cases for implicit std::string -> rosbag::Bag conversions, so we should disallow them.

### DIFF
--- a/tools/rosbag_storage/include/rosbag/bag.h
+++ b/tools/rosbag_storage/include/rosbag/bag.h
@@ -95,7 +95,7 @@ public:
      *
      * Can throw BagException
      */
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch.

Issue #836, opened 2016-07-14
XmlRpcValue is TypeInvalid returned from master::execute("registerSubscriber", ...)


mikepurvis:
We're seeing a bunch of crashes in our robots which take this form:

```
../../../../../ros-comm-d54e9be3421af71b70fc6b60a3bf916e779b43dc/clients/roscpp/src/libros/node_handle.cpp in ros::NodeHandle::subscribe at line 331

../../../../../ros-comm-d54e9be3421af71b70fc6b60a3bf916e779b43dc/clients/roscpp/src/libros/topic_manager.cpp in ros::TopicManager::subscribe at line 277

../../../../../ros-comm-d54e9be3421af71b70fc6b60a3bf916e779b43dc/clients/roscpp/src/libros/topic_manager.cpp in ros::TopicManager::registerSubscriber at line 485

../../../../../ros-comm-d54e9be3421af71b70fc6b60a3bf916e779b43dc/utilities/xmlrpcpp/src/XmlRpcValue.cpp in XmlRpc::XmlRpcValue::size at line 205

/usr/lib/x86_64-linux-gnu/libstdc++.so.6 in __cxa_throw ()

/usr/lib/x86_64-linux-gnu/libstdc++.so.6 in std::terminate() ()

/usr/lib/x86_64-linux-gnu/libstdc++.so.6 in ?? ()

/usr/lib/x86_64-linux-gnu/libstdc++.so.6 in __gnu_cxx::__verbose_terminate_handler() ()

abort.c in __GI_abort at line 89

../nptl/sysdeps/unix/sysv/linux/raise.c in __GI_raise at line 56
```

The line in question is here: https://github.com/ros/ros_comm/blob/d54e9be3421af71b70fc6b60a3bf916e779b43dc/utilities/xmlrpcpp/src/XmlRpcValue.cpp#L205

The cause is this logic which calls the master RPC without checking for the possibility of a `TypeInvalid` for the payload object: https://github.com/ros/ros_comm/blob/d54e9be3421af71b70fc6b60a3bf916e779b43dc/clients/roscpp/src/libros/topic_manager.cpp#L479-491

The simple fix is to just check for invalid and return false, similar to when `master::execute` itself returns false. However, my concern about this is that I don't really understand the underlying conditions which cause the RPC to fail. Should it be retried? Is there logic somewhere which already does this? AFAIK, the roscpp API makes no expectation of users to check for and retry subscriptions at the application layer.

FYI @jasonimercer

[...]

PR #837, opened 2016-07-15
Check xmlrpc response is valid before returning.

mikepurvis:
Proposed resolution to #836.

### DIFF
--- a/clients/roscpp/src/libros/master.cpp
+++ b/clients/roscpp/src/libros/master.cpp
@@ -199,7 +199,7 @@ bool execute(const std::string& method, const XmlRpc::XmlRpcValue& request, XmlR
 
     ok = !ros::isShuttingDown() && !XMLRPCManager::instance()->isShuttingDown();
 
-    if (!b && ok)
+    if (!(b && response.valid()) && ok)
     {
       if (!printed && wait_for_master)
       {


mikepurvis:

mikepurvis:
@dirk-thomas Thoughts on this?

dirk-thomas:
I am sorry for not replying earlier but I have to focus on a different project which doesn't allow for much time for ROS tickets.

In the case that the `XmlRpcClient::execute` returns `true` I don't see how the `response` could be invalid. The only case where this function returns `true` (https://github.com/ros/ros_comm/blob/54398ba8bedf254abead9dfffd5bd8e6bb8d72a1/utilities/xmlrpcpp/src/XmlRpcClient.cpp#L111) implies that the previous internal call to `parseResponse` also returned `true` (https://github.com/ros/ros_comm/blob/54398ba8bedf254abead9dfffd5bd8e6bb8d72a1/utilities/xmlrpcpp/src/XmlRpcClient.cpp#L106). And that function only returns `true` if the `result` (aka `response`) is valid (https://github.com/ros/ros_comm/blob/54398ba8bedf254abead9dfffd5bd8e6bb8d72a1/utilities/xmlrpcpp/src/XmlRpcClient.cpp#L470).

Without having run any code I would expect the added condition to be logically redundant. It looks like that it `b` can't be `true` at the same time as `response.valid()` is `false`.

That being said the current code only ensures that the `response` is valid. But the reason this was failing in the first place is that `payload` is invalid. So at the end of `master::execute` the following check might make sense:

```
if (!payload.isValid()) return false;
```

I am not sure though why that would be the case. (I am also not sure if there a valid cases where payload is intentionally invalid.) Eventually debugging into `XMLRPCManager::validateXmlrpcResponse` (https://github.com/ros/ros_comm/blob/d54e9be3421af71b70fc6b60a3bf916e779b43dc/clients/roscpp/src/libros/xmlrpc_manager.cpp#L235-L241) will shed some light into when that happens.

mikepurvis:
Oh you're right, of course, on `payload` vs. `response`— sorry for the distraction on that score.

If it would be of interest, I can try to instrument our bots to find out more about the circumstances of the TypeInvalid payload. We do a lot of rviz-over-wifi here, so it's possible this is down to dropped TCP sessions or something like that.

dirk-thomas:
That would be great. Any information will help to get this resolved. Thanks.

mikepurvis:
Closing this, will resume discussion in #836.

PR #839, opened 2016-07-20
Drop strict-aliasing warnings from roslz4 file.

mikepurvis:
Silences this build warning:

```
00:07:06.610 /tmp/binarydeb/ros-kinetic-roslz4-1.12.2/src/_roslz4module.c: In function ‘init_roslz4’:
00:07:06.610 /tmp/binarydeb/ros-kinetic-roslz4-1.12.2/src/_roslz4module.c:444:3: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
00:07:06.610    Py_INCREF(&LZ4Compressor_Type);
00:07:06.610    ^
00:07:06.610 /tmp/binarydeb/ros-kinetic-roslz4-1.12.2/src/_roslz4module.c:446:3: warning: dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]
00:07:06.610    Py_INCREF(&LZ4Decompressor_Type);
00:07:06.610    ^
```

Source: http://build.ros.org/job/Kbin_uX64__roslz4__ubuntu_xenial_amd64__binary/4/console

### DIFF
--- a/utilities/roslz4/CMakeLists.txt
+++ b/utilities/roslz4/CMakeLists.txt
@@ -41,7 +41,7 @@ include_directories(${PYTHON_INCLUDE_PATH})
 add_library(roslz4_py src/_roslz4module.c)
 set_source_files_properties(
   src/_roslz4module.c
-PROPERTIES COMPILE_FLAGS "-Wno-missing-field-initializers -Wno-unused-variable -Wstrict-aliasing")
+PROPERTIES COMPILE_FLAGS "-Wno-missing-field-initializers -Wno-unused-variable -Wno-strict-aliasing")
 set_target_properties(
   roslz4_py PROPERTIES OUTPUT_NAME roslz4 PREFIX "_" SUFFIX ".so"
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Are you sure that the warning is gone with this patch? I would think it needs either `-Wno-strict-aliasing` or `-fno-strict-aliasing` (http://cellperformance.beyond3d.com/articles/2006/06/understanding-strict-aliasing.html).

mikepurvis:
Oh, good catch. I was thrown off by the PR builder here not showing the warning.

dirk-thomas:
The devel job (and therefore the pr job) doesn't show the warning. I think the release job only shows it due to it optimization level.

The patch looks good to me (just suppressing the warning which has been there for a while. I am just not sure if the warning posses any problem. If yes, using `-fno-strict-aliasing` might be safer?

dirk-thomas:
I guess the Python API is designed like that on purpose (http://osdir.com/ml/python.cython.devel/2008-04/msg00080.html). Will merge once it's "green". Thanks.

PR #840, opened 2016-07-22
roslaunch: Fix <param command="..." /> for python3.

de-vri-es:
In python3, the standard output of the subprocess is of type bytes. That causes the data to be stored as a binary xmlrpc node which ends up base64 encoded when retrieving it later. Additionally, it crashed the `rosparam` command line utility which could not deal with that.

This PR checks if the read command output is not of type `str` and decodes it as UTF-8 in that case. This is a NOP in python2. We can't check if the read output is of type bytes, because in python2 `bytes` is just an alias for `str`.

### DIFF
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -479,7 +479,7 @@ class Loader(object):
         elif command is not None:
             try:
                 if type(command) == unicode:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch.

de-vri-es:
Thanks for the merge :)

PR #842, opened 2016-07-22
fix compiler warnings in release mode

### LINKED ISSUE #841: Warnings (unused-parameter, strict-aliasing) warnings from roscpp.
See: http://build.ros.org/view/Kbin_uX64/job/Kbin_uX64__roscpp__ubuntu_xenial_amd64__binary/11/consoleFull

15 instances of `unused parameter`, 10 instances of `strict aliasing`

These make a mess in our bundle builds, as catkin_tools dumps all stderr output to the main build log.

@dirk-thomas I can fix the causes or suppress the warnings; which would you prefer?

dirk-thomas:
Fixes #841.


mikepurvis:
Clean build for us now, thanks! 👍

PR #844, opened 2016-07-23
Fix confusing copyright messages/dates

jspricke:
Thanks to Wookey


dirk-thomas:
Thanks for the patch. Looks like a long time ago someone was doing a search-and-replace...

PR #846, opened 2016-07-23
Add 'Darwin' to unix-like platforms (resolve #845)

### LINKED ISSUE #845: rosgraph - roscore crash on OSX, depending on network setup
Crashes for me with a fresh Yosemite install. For a similar report, see http://answers.ros.org/question/220631/roscore-bug-errno8/

The issue is caused by OSX being categorised as non-unix-like in network.py, meaning the less sophisticated local address lookup is used. PR inbound.

luke-gumbley:
Prevents issue calling socket.getaddrinfo with the machine name where this is not a known local hostname (e.g. where the machine name hasn't been added to the hosts file).

### DIFF
--- a/tools/rosgraph/src/rosgraph/network.py
+++ b/tools/rosgraph/src/rosgraph/network.py
@@ -102,7 +102,7 @@ def _is_unix_like_platform():
     @rtype: bool
     """
     #return platform.system() in ['Linux', 'Mac OS X', 'Darwin']
-    return platform.system() in ['Linux', 'FreeBSD']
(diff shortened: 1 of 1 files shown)


luke-gumbley:
Proposed resolution for #845.

wjwwood:
The check is only used in one place that I can see, and that code looks ok for use on OS X. +1

dirk-thomas:
Thank you for the patch. I have cherry-picked it in 87412601f0eaec844c6d995d3b58595d2255c50a (and removed the commented line above).

PR #847, opened 2016-07-25
Fix `rosservice call` for python3.

de-vri-es:
`rosservice` was mistakenly passing a StringIO as buffer to `rosgraph.network.read_ros_handshake_header` even for python3. For python3 this needs to be a BytesIO object. This PR corrects that.

When running with `python2`, the old `StringIO` type is still used.

### DIFF
--- a/tools/rosservice/src/rosservice/__init__.py
+++ b/tools/rosservice/src/rosservice/__init__.py
@@ -50,9 +50,9 @@ import sys
 import socket
 
 try:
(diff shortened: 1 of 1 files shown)


wjwwood:
looks ok to me

dirk-thomas:
Thank you for the patch.

de-vri-es:
Thanks for the quick merge :)

Issue #849, opened 2016-07-27
Problem with many simultaneous connects to rosmaster on OSX 10.11 


jonfink:
As in [http://answers.ros.org/question/240430/problems-launching-many-nodes-on-osx](http://answers.ros.org/question/240430/problems-launching-many-nodes-on-osx)

I started experiencing intermittent problems with ROS (jade) on OSX 10.11 when running many nodes all trying to start up at once and access the rosmaster for subscriptions, publishes, and param server access.

Wrote a very simple test node:

``` c++
#include <ros/ros.h>

int main(int argc, char *argv[])
{
  ros::init(argc, argv, "test_node");
  ros::NodeHandle pnh("~");

  ros::Rate r(100.0);

  while(ros::ok()) {

    if(!ros::master::check()) {
      ROS_ERROR("[%s] Unable to see master", ros::this_node::getName().c_str());
    }

    ros::spinOnce();

    r.sleep();
  }

  return 0;
}
```

And ran many (up to 100, but ran into problems by 30 or 40) instances of it.

**Nodes periodically report that they are unable to contact the master.** This is in keeping with what I noticed in my "real-world" setting, nodes would intermittently fail to contact master and be unable to retrieve parameter settings.

dirk-thomas:
Fixed by 1e546976053ea948dfb22cba46e11ce2af53e541.

PR #850, opened 2016-07-27
Increase request_queue_size for xmlrpc server

jonfink:
Addresses issue #849

### DIFF
--- a/tools/rosgraph/src/rosgraph/xmlrpc.py
+++ b/tools/rosgraph/src/rosgraph/xmlrpc.py
@@ -93,6 +93,9 @@ class ThreadingXMLRPCServer(socketserver.ThreadingMixIn, SimpleXMLRPCServer):
         # to True to allow quick restart on the same port.  This is equivalent 
         # to calling setsockopt(SOL_SOCKET,SO_REUSEADDR,1)
         self.allow_reuse_address = True
+        # Increase request_queue_size to handle issues with many simultaneous
+        # connections in OSX 10.11
+        self.request_queue_size = 128
         if rosgraph.network.use_ipv6():
             logger = logging.getLogger('xmlrpc')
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please avoid any unrelated white space changes in the PR.

wjwwood:
@jonfink thanks for digging into the issue. I'll try to reproduce the problem and test your fix on my mac soon. The proposed change seems reasonable to me though.

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -93,6 +93,9 @@ def __init__(self, addr, log_requests=1):
         # to True to allow quick restart on the same port.  This is equivalent 
         # to calling setsockopt(SOL_SOCKET,SO_REUSEADDR,1)
         self.allow_reuse_address = True
+        # Increase request_queue_size to handle issues with many simultaneous
+        # connections in OSX 10.11
+        self.request_queue_size = 128
```
dirk-thomas, line 98: Can you please elaborate why you chose `128`.

jonfink, line 98: In short, I picked 128 because it is the max on OSX and Ubuntu 14.04
(socket.SOMAXCONN gives OS limit).

I suppose it is hard to say if a smaller queue size would eliminate the
problem since I can't say exactly what is causing the problem on my system
vs older OSX builds or Ubuntu.

dirk-thomas, line 98: Instead of hard coding a value (which might not be supported) this could set the queue size to e.g. `min(socket.SOMAXCONN, 128)`.

dirk-thomas:
I have modified the patch based on my comment and cherry-picked it in 1e546976053ea948dfb22cba46e11ce2af53e541.

PR #851, opened 2016-07-29
Indigo devel fix rospy reconnect

atiderko:
1. while roscpp reconnects on timeout or temporary 'No route to host'
   errors the rospy does it not. So on connection problems longer than 3
   min the connection between subscriber and publisher goes.
   I added some exceptions if these occurs rospy does not close the socket
   and reconnect now.
2. since rospy reconnects on timeout now, the reconnects goes very fast in
   times where it easier to restart the ros node then wait for a reconnect.
   So I added a maximal backoff time.
3. on connection problems while node start the subscription to a publisher
   will be stopped after 3 tries. Now the reconnection is stopped on
   shutdown of the node.

### DIFF
--- a/clients/rospy/src/rospy/impl/masterslave.py
+++ b/clients/rospy/src/rospy/impl/masterslave.py
@@ -437,20 +437,22 @@ class ROSHandler(XmlRpcHandler):
         #    individual socket timeouts, but this could potentially
         #    affect user code.
         socket.setdefaulttimeout(60.)
-        tries = 0
-        max_num_tries = 3
         success = False
-        while not success:
-            tries += 1
+        interval = 0.5 # seconds
+        # try to get the topic information until succes or the ROS node is shutdown.
+        # -> on connections problems while ROS node start
+        while not success and not is_shutdown():
             try:
                 code, msg, result = \
                       xmlrpcapi(pub_uri).requestTopic(caller_id, topic, protocols)
                 success = True
             except Exception as e:
-                if tries >= max_num_tries:
-                    return 0, "unable to requestTopic: %s"%str(e), 0
-                else:
+                if not is_shutdown():
                     _logger.debug("Retrying for %s" % topic)
+                    if interval < 30.0:
+                        # exponential backoff (maximum 32 seconds)
+                        interval = interval * 2
+                    time.sleep(interval)
 
         #Create the connection (if possible)
         if code <= 0:
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -562,13 +562,23 @@ class TCPROSTransport(Transport):
             raise
         except Exception as e:
             #logerr("Unknown error initiating TCP/IP socket to %s:%s (%s): %s"%(dest_addr, dest_port, endpoint_id, str(e)))
-            rospywarn("Unknown error initiating TCP/IP socket to %s:%s (%s): %s"%(dest_addr, dest_port, endpoint_id, traceback.format_exc()))            
-            if isinstance(e, socket.error):
-                if not isinstance(e, socket.timeout) and e.errno not in [100, 101, 102, 103, 110, 112, 113]:
-                    # 100: 'ENETDOWN', 101: 'ENETUNREACH', 102: 'ENETRESET', 103: 'ECONNABORTED'
-                    # 110: 'ETIMEDOUT', 112: 'EHOSTDOWN', 113: 'EHOSTUNREACH'
-                    # FATAL: no reconnection as error is unknown
-                    self.close()
+            rospywarn("Unknown error initiating TCP/IP socket to %s:%s (%s): %s"%(dest_addr, dest_port, endpoint_id, traceback.format_exc()))
+            # check for error type and reason. On unknown errors the socket will be closed
+            # to avoid reconnection and error reproduction
+            if not isinstance(e, socket.error):
+                # FATAL: no reconnection as error is unknown
+                self.close()
+            elif not isinstance(e, socket.timeout) and e.errno not in [100, 101, 102, 103, 110, 112, 113]:
+                # reconnect in follow cases, otherwise close the socket:
+                # 1. socket.timeout: on timeouts caused by delays on wireless links
+                # 2. ENETDOWN (100), ENETUNREACH (101), ENETRESET (102), ECONNABORTED (103):
+                #     while using ROS_HOSTNAME ros binds to a specific interface. Theses errors
+                #     are thrown on interface shutdown e.g. on reconnection in LTE networks
+                # 3. ETIMEDOUT (110): same like 1. (for completeness)
+                # 4. EHOSTDOWN (112), EHOSTUNREACH (113): while network and/or DNS-server is not reachable
+                #
+                # no reconnection as error is not 1.-4.
+                self.close()
             raise TransportInitError(str(e)) #re-raise i/o error
                 
     def _validate_header(self, header):
@@ -741,7 +751,7 @@ class TCPROSTransport(Transport):
                 self.socket = None
                 
             if self.socket is None and interval < 30.:
-                # exponential backoff
+                # exponential backoff (maximum 32 seconds)
                 interval = interval * 2
                 
             time.sleep(interval)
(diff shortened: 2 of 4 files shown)


dirk-thomas:
Some comments for the different changes:

First commit: Can you please elaborate what exceptions you expect and in which cases you want to continue retrying? Currently it looks like if the exception is not a `socket.error` it retries?

Second commit: An upper limit for the exponential backoff is a good point. Why does the patch use two different limits instead of the same?

Third commit: the current retry logic (3 times) is implemented in `_connect_topic`. This patch adds another level of retry on top of that in `_connect_topic_thread`. What is the rational to implement the new behavior in a different location and not modify the existing retry logic instead?

atiderko:
First commit:
1. socket.timeout: on timeouts caused by delays on wireless links 
2. ENETDOWN (100), ENETUNREACH (101), ENETRESET (102), ECONNABORTED (103): while using `ROS_HOSTNAME` ros binds to a specific interface. Theses errors are thrown on interface shutdown e.g. on reconnection in LTE networks
6. ETIMEDOUT (110): same like 1 (for completeness)
7. EHOSTDOWN (112), EHOSTUNREACH (113): while network and/or DNS-server is not reachable
Perhaps it is easier to reconnect on each `socket.error`, but there are also some errors on which the reconnection does make no sense.
Since there are no error message in user space, if the connection was aborted it would be useful to add a `logerror` output if the connection was closed.

Second commit:
I used the same values as they are used for timeout in `connect`-call (some lines above).

Third commit:
The implemented `_connect_topic` blocks in worst case 3 min. The new reconnect blocks in worst case until the rosnode is stopped. Therefore I added the new reconnect in `_connect_topic_thread` because this method is called in a thread as suggested by method name. Since it is nowhere documented that  `_connect_topic` should only called in a thread, I did not want to add a non ending block. I should prefer to remove the 3 retries from `_connect_topic`, but it can also be done in the additional commit.

dirk-thomas:
Thank you for the explanation. Let me try to rephrase / clarify my questions / concerns:

First commit:
- Please add the rational for the error codes in a comment in the code so that future reader understand why they were added.
- In the case of a timeout wouldn't it make sense to try to reconnect again?
- I don't see a reason why it should retry if a different exception than `socket.error` is raised. I would expect the opposite:
  
  ```
  if not isinstance(e, socket.error):
      # FATAL: no reconnection as error is unknown
      self.close()
  elif not isinstance(e, socket.timeout) and e.errno not in [100, 101, 102, 103, 110, 112, 113]:
      # in this case the error is well known
  ```

Second commit:
- I was referring to the two different thresholds within the patch: [60s](https://github.com/ros/ros_comm/pull/851/files#diff-6b7e261e2132ca820f500ffb84ed38afR332) vs [30s](https://github.com/ros/ros_comm/pull/851/files#diff-a90cf003a88553002fc767ae33135f31R743)

Third commit:
Since `_connect_topic` is an internal function it shouldn't be used from the outside. I think it should be fine to modify the existing logic to do the "unlimited" retrying.

atiderko:
ok, I see now my mistake in first commit ;-(

the connection should be closed in three cases
1. NOT a `socket.error` or 
2. NOT a `socket.timeout` and 
3. NOT a specific `socket.errno`

my pull request should be changed to (+ the rational for the error codes in a comment):

```
if not isinstance(e, socket.error):
    # FATAL: no reconnection as error is unknown
    self.close()
elif not isinstance(e, socket.timeout) and e.errno not in [100, 101, 102, 103, 110, 112, 113]:
    # in this case the error is well known
    self.close()

```

Second commit:
The 30s are because of [timeout=30.](https://github.com/fkie-forks/ros_comm/blob/4c942c4662823882f3b0b9260bd6179c342bede7/clients/rospy/src/rospy/impl/tcpros_base.py#L739) and 60s because of [timeout=60.](https://github.com/fkie-forks/ros_comm/blob/4c942c4662823882f3b0b9260bd6179c342bede7/clients/rospy/src/rospy/impl/tcpros_pubsub.py#L168). No clue why the timeouts are different.

Third commit:
I agree

Currently and for next three weeks I am in holiday and I am not able to change the pull request. Perhaps it is easier if you make the needed changes addressed by this pull request. And then ignore this request.

regards

atiderko:
hi @dirk-thomas,

I fixed/added the changes we discussed

atiderko:
something goes wrong: I changed only python code and `rostest.runner.RosTest.testpubsub_n_fast_udp` for cpp part was not successful. Or I'm misunderstanding the test results?

dirk-thomas:
It could be a flaky test. Let me retrigger the jobs: @ros-pull-request-builder retest this please

dirk-thomas:
The patch looks good to me. But I would merge it into the _kinetic-devel_ branch instead.

@ros/ros_team Can you please review this PR.

clients/rospy/src/rospy/impl/masterslave.py
```diff
@@ -437,20 +437,22 @@ def _connect_topic(self, topic, pub_uri):
         #    individual socket timeouts, but this could potentially
         #    affect user code.
         socket.setdefaulttimeout(60.)
-        tries = 0
-        max_num_tries = 3
         success = False
-        while not success:
-            tries += 1
+        interval = 0.5 # seconds
+        # try to get the topic information until succes or the ROS node is shutdown.
```
wjwwood, line 442: "success" or probably "until successful".

clients/rospy/src/rospy/impl/masterslave.py
```diff
@@ -437,20 +437,22 @@ def _connect_topic(self, topic, pub_uri):
...
         #    affect user code.
         socket.setdefaulttimeout(60.)
-        tries = 0
-        max_num_tries = 3
         success = False
-        while not success:
-            tries += 1
+        interval = 0.5 # seconds
+        # try to get the topic information until succes or the ROS node is shutdown.
+        # -> on connections problems while ROS node start
```
wjwwood, line 443: What does `->` mean in this case, return or exit or something else? I'd prefer an explicit comment, like "while ROS node is running try to connect, exit on connection problems" (not sure if that statement is accurate, but it is clearer than the current comment).

clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -737,8 +750,8 @@ def _reconnect(self):
             except TransportInitError:
                 self.socket = None
                 
-            if self.socket is None:
-                # exponential backoff
+            if self.socket is None and interval < 30.:
```
wjwwood, line 753: nitpick: I know that `30.` is still technically correct to force it as a float literal, but `30.0` is the style used elsewhere, so I'd appreciate updating this for consistency.

wjwwood (approved):
Other than some small stylistic or comment formatting related comments, lgtm.

I agree that this should be target at kinetic. Either only kinetic or at least kinetic first with an option to backport to Indigo if things go well and there is desire for it.

We should solicit testers for this change too, as I could easily imagine it changes behavior in a breaking way for complex systems or even do something like effect `ctrl-c` behavior of simple scripts.

clients/rospy/src/rospy/impl/tcpros_base.py
dirk-thomas: Existing code just a few lines above uses `30.` so I think its fine as is.

clients/rospy/src/rospy/impl/masterslave.py
dirk-thomas: I will address this on cherry-pick.

clients/rospy/src/rospy/impl/tcpros_base.py
wjwwood: Then should new lines like this:

https://github.com/ros/ros_comm/pull/851/files#diff-20362b9a78750b6859885bf4bb7e2364R452
https://github.com/ros/ros_comm/pull/851/files#diff-8582a4a4f1da6f58343838915b4a9b55R176

Use this syntax?

dirk-thomas: I don't mind either or. The ROS 1 code base has pretty bad style and I don't see the way a float is written as important enough to spend more time on it.

wjwwood: Ok, fine by me.

clients/rospy/src/rospy/impl/masterslave.py
dirk-thomas: I updated it to:

> while the ROS node is not shutdown try to get the topic information
> and retry on connections problems after some wait

dirk-thomas:
Thank you for the patch and for iterating on it. I have cherry-picked the patch to the _kinetic-devel_ branch: 14b5c93a1d41bf8f4e73ee3eca1b38ec782c9de0

PR #855, opened 2016-08-09
Fix for #854 - cleaner rosout log rotation

### LINKED ISSUE #854: rosout log rotation is odd
Normally you'd expect that the rosout log file is rotated as usual (https://en.wikipedia.org/wiki/Log_rotation) by increasing the number in the suffix for old log file such that it is always guaranteed that rosout.log.2 is older than rosout.log.1. The current implementation moves rosout.log to rosout.log.(current_back_index % 10) during "rotation", thus the ordering is not preserved as you would expect it.

mgrrx:
This PR fixes #854 by rotating the log files as usual in Linux to guarantee the ordering of log files (https://en.wikipedia.org/wiki/Log_rotation).


dirk-thomas:
Please avoid any kind of unrelated changes and refactorings since it makes it more difficult to read and review the patch as well as more difficult to cherry-picking / backporting between different branches.

Please make future contributions against the latest branch (in this case kinetic-devel). I can cherry pick the patch in this case so no need to open a new PR for this.

dirk-thomas:
Splitting this into two PR: one for Kinetic e.g. moving init code, one for Lunar only with changing order.

Issue #827, opened 2016-07-06
`rosbag filter` should show progress based on uncompressed size of bag (or preserve compression)


benwr:
`rosbag filter` writes to an uncompressed bag, but shows progress based on the raw input bag's filesize. This can be pretty confusing if you're reading from a large, compressed bag - you'll hit 100% progress, but then the output file will keep growing for a long time.

dirk-thomas:
Please consider to contribute a pull request for this enhancement.

PR #857, opened 2016-08-10
update rosbag filter progress meter to use raw uncompressed input size

spaghetti-:
fixes https://github.com/ros/ros_comm/issues/827


dirk-thomas:
Thanks!

PR #858, opened 2016-08-10
Sros

ruffsl:
Following in line with https://github.com/ros-infrastructure/rospkg/pull/104, this is migrating current SROS work on ros_comm to a more longterm branch upstream.


tfoote:
This is fine to merge into the branch to setup your integration testing.

dirk-thomas:
Is the patch in this PR supposed to be reviewed before merging or will we defer this until the branch is considered to be merged into _???-devel_? If the latter please go ahead with the merge anytime.

ruffsl:
I think we'll need to defer the proper review until we are ready to merge upstream (to -devel), i.e, there are still some things about this implementation that need to be finalized after we settle on the security standards in an upcoming REP. For now I'd just like an official home for the code so I can setup some documentation and testing.

PR #859, opened 2016-08-11
[rostest] Add reusable node to test parameters.

130s:
Adding a node `paramtest`, a [reusable `rostest` node](http://wiki.ros.org/rostest/Nodes) that checks if a designated parameter is registered and non-empty. 

**Inspired by**: `hztest`, and the implementation and the usage is similar to hztest.

**Motivation**: It's cumbersome to write your own test node when you simply want to check if a parameter is registered and/or is not null.

**Usage**: Hopefully the usage is obvious enough from [a test file that's also added in this PR](https://github.com/130s/ros_comm/blob/f2fb87e93a1826ef69be041a69b09ce67a74055c/tools/rostest/test/param.test).

### DIFF
--- a/tools/rostest/nodes/paramtest
+++ b/tools/rostest/nodes/paramtest
@@ -1,0 +1,118 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2016, TORK (Tokyo Opensource Robotics Kyokai Association)
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of TORK nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+# Original copied from hztest node
+# https://github.com/ros/ros_comm/blob/24e45419bdd4b0d588321e3b376650c7a51bf11c/tools/rostest/nodes/hztest
+# Integration test node that checks if a designated parameter is already
+# registered at the Parameter Server. Following parameters must be set:
+#
+#  * ~/param_name_target: expected parameter name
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Depending on how it is supposed to be used would it make sense to optionally also check if the parameter has a specific value?

dirk-thomas:
I can cherry-pick this patch manually but please make future PRs against the latest branch, currently _kinetic-devel_ for this repo.

tools/rostest/nodes/paramtest
```diff
@@ -0,0 +1,118 @@
...
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+# Original copied from hztest node
+# https://github.com/ros/ros_comm/blob/24e45419bdd4b0d588321e3b376650c7a51bf11c/tools/rostest/nodes/hztest
```
dirk-thomas, line 35: If the code was copied it should preserve the original copyright.

tools/rostest/nodes/paramtest
```diff
@@ -0,0 +1,118 @@
...
+        # Fetch parameters
+        try:
+            # Param to test
+            param_name_target = rospy.get_param('~param_name_target')
+            # length of test
+            test_duration = float(rospy.get_param('~test_duration'))
+            # time to wait before
+            wait_time = rospy.get_param('~wait_time', 20.)
+        except KeyError as e:
+            self.fail('ParamTest not initialized properly. Parameter [%s] not set. debug[%s] debug[%s]'%(str(e), rospy.get_caller_id(), rospy.resolve_name(e.args[0])))
```
dirk-thomas: Should the double `debug` phrase be changed, e.g. into "caller id" and "resolved name"?

Same below.

tools/rostest/nodes/paramtest
```diff
@@ -0,0 +1,118 @@
...
+
+        print("""Parameter: %s Test Duration: %s""" % (param_name_target, test_duration))
```
dirk-thomas: No need for triple quotes here.

tools/rostest/nodes/paramtest
```diff
@@ -0,0 +1,118 @@
...
+            rospy.loginfo('Content of the obtained parameter: {}'.format(param_obtained))
+        except KeyError as e:
+            self.fail('Designated parameter [%s] is not registered. debug[%s] debug[%s]'%(target_param, rospy.get_caller_id(), rospy.resolve_name(e.args[0])))
+
+        self.assertIsNotNone(param_obtained)
+
+if __name__ == '__main__':
+    # A dirty hack that is taken from hztest and not sure if it's necessary
+    # for paramtest as well.
+    time.sleep(0.75)
```
dirk-thomas: Please avoid keeping a hack like this if it is not for sure required.

tools/rostest/nodes/paramtest
130s: The original is Willow 2008, but I'll just revert.

130s:
I added a few commits that include the following changes:
- addressed comments.
- feature to test specific param value.
- use double quote consistently instead of single quote.

Btw I added the changes without squashing but once it LGT I can rebase (or please rely on github's feature).

> I can cherry-pick this patch manually but please make future PRs against the latest branch, currently kinetic-devel for this repo.

Sure. I'm still mostly on indigo but changes like this can be easily tested on cloud.

dirk-thomas:
The patch currently fails on test.

130s:
Btw where does buildfarm obtain the email that it sends out the build result to? It's using my old one (130s@lateeye.net) that I stopped using awhile ago.

dirk-thomas:
Some tests are still failing.

I assume Jenkins has fetched the email address of your user earlier and has cached that value since then. I think you should be able to update your profile yourself: http://build.ros.org/user/130s/configure If not please let me know what I should update the email address to.

130s:
Tests passed. And thanks, I was able to resume getting notifications from buildfarm.

dirk-thomas:
Thanks for iterating on this. I have squashed and cherry-picked the change to the `kinetic-devel` branch: c485654cd16c130364144a18b3e83e4b01713ed4

130s:
Thank you for the review and acceptance. Do new functionality only go to the newest branch for `ros_comm`? While I now understand the PR should be targetted at the latest branch, I want non-api/abi breaking change like this to be ported to older supported distros.

dirk-thomas:
Any change should go into the latest development branch. From there certain changes can be backported. After they have been available in the latest distribution for a while and increased the confidence that they don't cause regressions.

But backports should happen only when there is a "good reason" to do so (e.g. fix a significant bug). Not every feature should be backported. The reasons not to do it are extra effort and chance of regressions. While I do understand that it would be nice to always have every change all distros the reason why we have distros is to keep already released distros as stable as possible and add new stuff only into new distros.

So each request to backport specific changes needs to be considered carefully.

PR #862, opened 2016-08-11
[rostest] Refactor hztestX.test

wkentaro:
## Why?

The files hztestX.test are the samples for ROS users for testing nodes.
This PR simplify the test xml to demonstrate how to use `<rosparam>` for private params.

### DIFF
--- a/tools/rostest/test/hztest.test
+++ b/tools/rostest/test/hztest.test
@@ -1,14 +1,11 @@
 <launch>
   <node name="talker" pkg="rospy" type="talker.py" />
 
-  <test name="hztest1" test-name="hztest_test" pkg="rostest" type="hztest">
-    <rosparam>
-      topic: chatter
-      hz: 10.0
-      hzerror: 0.5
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please add more information to the pull request describing why you want this to be changed.

wkentaro:
Updated the information.

dirk-thomas:
For new ROS users I would expect the previous way of setting the parameters to be easier and much less error prone. Each parameter is very simple and separate from the others. Using `<rosparam>` for this requires writting valid yaml which some users might not be familiar with. So I think this changes raises the bar to understand the example which I don't think it a good idea.

wkentaro:
Your comments are reasonable, but in the first look, I thought the `<rosparam>` does not work for `hztest` test node. So can I add a comment about the another style of code using rosparam?

wkentaro:
Fixed.

dirk-thomas:
Looks good to me. Thanks!.

PR #864, opened 2016-08-11
[rospy] Refactor talker.py which is rospy test node

wkentaro:
## What
- Advertise after rospy.init_node
- Use rospy.Timer which is desirable implementation of periodical
  publishing.
## Why
- Advertising after rospy.init causes wrong topic name with private name.
- This node seems a sample for ROS new users, and using timer is better than `while not rospy.is_shutdown()`, this PR improves the sample code.

### DIFF
--- a/clients/rospy/test_nodes/talker.py
+++ b/clients/rospy/test_nodes/talker.py
@@ -34,20 +34,19 @@
 ## Simple talker demo that published std_msgs/Strings messages
 ## to the 'chatter' topic
 
+
 import rospy
 from std_msgs.msg import String
 
-def talker():
-    pub = rospy.Publisher('chatter', String, queue_size=10)
-    rospy.init_node('talker', anonymous=True)
-    r = rospy.Rate(10) # 10hz
-    while not rospy.is_shutdown():
-        str = "hello world %s"%rospy.get_time()
-        rospy.loginfo(str)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please add more information to the pull request describing why you want this to be changed.

wkentaro:
Updated.

dirk-thomas:
I have extracted the fix to the order (`init` before `Publisher`) into #873.

For the different second part using a timer instead of a rate I don't think that either one is better. It just depends on which one fits the use case. Also I don't see this file as an important example. It is not referenced anywhere. So I would think the canonical way to find sample code is the ROS wiki or e.g. the `ros_tutorials` package.

tfoote:
I think the simpler rate.sleep() approach is simpler and more explicit. The timer class is more "elegant" with less lines of code, in the script but it actually increases the system complexity. 

The talker.py which is in the tutorials is in rospy_tutorials: https://github.com/ros/ros_tutorials/tree/kinetic-devel/rospy_tutorials Adding a timer based version in the tutorial would be a better place to be an example.

wkentaro:
I see. I will send PR to rospy_tutorials.
Thanks.

wkentaro:
Sent here: https://github.com/ros/ros_tutorials/pull/34

PR #865, opened 2016-08-12
Allow 64-bit sizes to be passed to robag max_size

contradict:
Internally, bagfile size is 64 bits. However, the types used in command parsing are only 32 bits, making the maximum usable split size 2GB. This patch switches the argument parsing to use 64 bit types.


dirk-thomas:

dirk-thomas:

PR #866, opened 2016-08-12
Record a maximum number of splits and then begin deleting old files

contradict:
This patch allows creating a fixed size recording of the most recent data. It add a --max-splits option to rosbag record. When --split is used, after max-splits files have been created, the oldest one is deleted so a fixed size (or fix duration if --duration is used) recording can be made.

### DIFF
--- a/tools/rosbag/include/rosbag/recorder.h
+++ b/tools/rosbag/include/rosbag/recorder.h
@@ -160,7 +160,7 @@ private:
 
     std::string                   target_filename_;
     std::string                   write_filename_;
-    std::list<std::string>         current_files_;
+    std::list<std::string>        current_files_;
 
     std::set<std::string>         currently_recording_;  //!< set of currenly recording topics
     int                           num_subscribers_;      //!< used for book-keeping of our number of subscribers
--- a/tools/rosbag/include/rosbag/recorder.h
+++ b/tools/rosbag/include/rosbag/recorder.h
@@ -45,6 +45,7 @@
 #include <queue>
 #include <string>
 #include <vector>
+#include <list>
 
 #include <boost/thread/condition.hpp>
 #include <boost/thread/mutex.hpp>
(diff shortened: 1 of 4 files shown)


tools/rosbag/src/record.cpp
```diff
@@ -62,6 +62,7 @@ rosbag::RecorderOptions parseOptions(int argc, char** argv) {
       ("bz2,j", "use BZ2 compression")
       ("lz4", "use LZ4 compression")
       ("split", po::value<int>()->implicit_value(0), "Split the bag file and continue recording when maximum size or maximum duration reached.")
+      ("max-splits", po::value<int>()->default_value(0), "Split the bag file at most splits times, then begin erasing the oldest split to keep a constant number of files.")
```
dirk-thomas: Since the new options isn't doing the actual splitting but limits the number of log files I would suggest a help text like this:

> Keep a maximum of N bag files, when reaching the maximum erase the oldest one to keep a constant number of files.

The same / similar for the Python help string.

tools/rosbag/include/rosbag/recorder.h
```diff
@@ -157,6 +160,7 @@ class ROSBAG_DECL Recorder
 
     std::string                   target_filename_;
     std::string                   write_filename_;
+    std::list<std::string>         current_files_;
```
dirk-thomas: Indentations seems to be off by one.

dirk-thomas:
This is a nice feature. The patch looks fine - I only commented inline on some minor things.

dirk-thomas:

Please consider updating the [wiki page](http://wiki.ros.org/rosbag/Commandline) to mention the new command line option (with `<<Version(ROS Kinetic)>>` to indicate that the new option has been added in Kinetic). Please post a link to the diff here for easier discoverability.

dirk-thomas:

Please consider updating the [wiki page](http://wiki.ros.org/rosbag/Commandline) to mention the new command line option (with ```<<Version(ROS Kinetic)>>``` to indicate that the new option has been added in Kinetic). Please post a link to the diff here for easier discoverability.

dirk-thomas:
For the record: http://wiki.ros.org/action/diff/rosbag/Commandline?action=diff&rev1=30&rev2=31 Thanks for updating the page.

PR #377, opened 2014-03-11
roscpp: Adding AsyncSpinner::canStart to check if a spinner can be started

jbohren:
This adds the introspection described in https://github.com/ros/ros_comm/issues/277

Should a PR also be made against hydro?

PR #867, opened 2016-08-15
fix multi-threaded spinning

### LINKED ISSUE #277: roscpp: Multiple AsyncSpinner behavior is confusing and precludes introspection
Currently, if you try to create two AsyncSpinner instances this error is emitted:

> ``` cpp
> ROS_ERROR("MultiThreadeSpinner: You've attempted to call ros::spin "
>           "from multiple threads... "
>           "but this spinner is already multithreaded.");
> ```
> 
> https://github.com/ros/ros_comm/blob/groovy-devel/clients/roscpp/src/libros/spinner.cpp#L75

This is normally not a problem, but if two different systems with ROS plugins are trying to create spinners, there's no way to determine whether spinners currently exist nor is there a way to determine the number of spinning threads.

First, it seems like this should be a warning instead of an error. But more importantly, it seems like the AsyncSpinner class can't determine if it wants to be a handle to a global service or not. You can create any number of AsyncSpinner instances, but they all want to use a single _static_ mutex so only the first one is actually going to function properly.

I'd expect that if I create an AsyncSpinner with `n` threads in one place, and create another AsyncSpinner with `m` threads in another place, that there should subsequently be `n+m` threads processing ROS callbacks. This shouldn't be hard to implement without changing the API. Would such a change be welcome?

rhaschke:
This is an attempt to fix #277 in a more fundamental fashion.

As @po1 pointed out in https://github.com/ros/ros_comm/issues/277#issuecomment-37054768, due to the [global mutex](https://github.com/ros/ros_comm/blob/groovy-devel/clients/roscpp/src/libros/spinner.cpp#L36), only a single thread was allowed to run/start spinners, even if operating on different callback queues.

As @tfoote pointed out in https://github.com/ros/ros_comm/issues/277#issuecomment-238713425, the mutex was probably introduced to prevent interleaving access to a callback queue thus guaranteeing _in-order_ execution of queued callbacks.

Obviously, this protection should be _local per callback queue_ (instead of using a global mutex) and it should only be considered for `SingleThreadedSpinners` as multi-threaded spinners deliberately request asynchronous processing.

This PR attempts to solve that issue by replacing the global mutex with a `SpinnerMonitor` that keeps track of all callback queues that are currently spinning. If a single-threaded spinner wants to spin a queue in parallel to another spinner, an error can be issued. IMHO, this error should be even fatal.

Alternatively, the callback queue should monitor who is spinning it. However, this would require an API change, which is why I decided for the `SpinnerMonitor`.

### DIFF
--- a/clients/roscpp/include/ros/spinner.h
+++ b/clients/roscpp/include/ros/spinner.h
@@ -109,10 +109,10 @@ public:
 
 
   /**
-   * \brief Check if the spinner can be started. A spinner can't be started if
-   * another spinner is already running.
+   * \brief Check if the spinner can be started. The spinner can't be started if
+   * another single-threaded spinner is already operating on the callback queue.
    */
-  bool canStart();
+  ROS_DEPRECATED bool canStart();
   /**
    * \brief Start this spinner spinning asynchronously
    */
--- a/clients/roscpp/include/ros/spinner.h
+++ b/clients/roscpp/include/ros/spinner.h
@@ -109,8 +109,10 @@ public:
 
 
   /**
-   * \brief Check if the spinner can be started. The spinner can't be started if
+   * \brief Check if the spinner can be started. The spinner shouldn't be started if
    * another single-threaded spinner is already operating on the callback queue.
+   *
+   * This function is not necessary anymore. start() will always start spinning.
    */
   ROS_DEPRECATED bool canStart();
   /**
--- a/clients/roscpp/include/ros/spinner.h
+++ b/clients/roscpp/include/ros/spinner.h
@@ -112,9 +112,11 @@ public:
    * \brief Check if the spinner can be started. The spinner shouldn't be started if
    * another single-threaded spinner is already operating on the callback queue.
    *
-   * This function is not necessary anymore. start() will always start spinning.
+   * This function is not necessary anymore. start() will always try to start spinning
+   * and throw a std::runtime_error if it failed.
    */
-  ROS_DEPRECATED bool canStart();
+  // TODO: deprecate in L-turtle
+  bool canStart();
   /**
    * \brief Start this spinner spinning asynchronously
    */
--- a/clients/roscpp/src/libros/spinner.cpp
+++ b/clients/roscpp/src/libros/spinner.cpp
@@ -30,10 +30,82 @@
 #include "ros/callback_queue.h"
 
 #include <boost/thread/thread.hpp>
-#include <boost/thread/recursive_mutex.hpp>
+#include <boost/thread/mutex.hpp>
 
 namespace {
-  boost::recursive_mutex spinmutex;
+
+/** class to monitor running single-threaded spinners.
+ *
+ *  Calling the callbacks of a callback queue _in order_, requires a unique SingleThreadedSpinner
+ *  spinning on the queue. Other threads accessing the callback queue will probably intercept execution order.
+
+ *  To avoid multiple SingleThreadedSpinners (started from different threads) to operate on the same callback queue,
+ *  this class stores a map of all spinned callback queues.
+ *  If the spinner is single threaded, the corresponding thread-id is stored in the map
+ *  and other threads are not allowed to spin on this callback queue too.
+ *
+ *  If the spinner is multi-threaded, the stored thread-id is NULL and future SingleThreadedSpinners
+ *  are not allowed to spin on this queue.
+ */
+struct SpinnerMonitor {
+  /// check whether the queue could be spinned without problems
+  // TODO: Remove! This is only for use with deprecated AsyncSpinner::canStart()
+  // Calling canSpin() and add() in sequence might give different return values
+  // if a single-threaded spinner was started in between.
+  bool canSpin(ros::CallbackQueue* queue, bool single_threaded)
+  {
+    boost::mutex::scoped_lock lock(mutex_);
+
+    boost::thread::id tid;
+    if (single_threaded)
+      tid = boost::this_thread::get_id();
+
+    std::map<ros::CallbackQueue*, boost::thread::id>::const_iterator it = spinning_queues_.find(queue);
+    return it == spinning_queues_.end() || it->second == tid;
+  }
+
+  /// add a queue to the list
+  bool add(ros::CallbackQueue* queue, bool single_threaded)
+  {
+    boost::mutex::scoped_lock lock(mutex_);
+
+    boost::thread::id tid;
+    if (single_threaded)
+      tid = boost::this_thread::get_id();
+
+    std::map<ros::CallbackQueue*, boost::thread::id>::const_iterator it = spinning_queues_.find(queue);
+    bool can_spin = ( it == spinning_queues_.end() || it->second == tid );
+    if (!can_spin)
+      return false;
+
+
+    if (it == spinning_queues_.end())
+      spinning_queues_.insert(it, std::make_pair(queue, tid));
+
+    return true;
+  }
+
+  /// remove a queue to the list
+  void remove(ros::CallbackQueue* queue)
+  {
+    boost::mutex::scoped_lock lock(mutex_);
+    std::map<ros::CallbackQueue*, boost::thread::id>::const_iterator it = spinning_queues_.find(queue);
+    if (it != spinning_queues_.end())
+    {
+      ROS_ASSERT_MSG(it->second == boost::thread::id() || it->second == boost::this_thread::get_id(),
+                     "SpinnerMonitor::remove() called from different thread than add().");
+      spinning_queues_.erase(it);
+    }
+  }
+
+  std::map<ros::CallbackQueue*, boost::thread::id> spinning_queues_;
+  boost::mutex mutex_;
+};
+
+SpinnerMonitor spinner_monitor;
+const std::string DEFAULT_ERROR_MESSAGE =
+    "Attempt to spin a callback queue from two spinners, one of them being single-threaded."
+    "This will probably result in callbacks being executed out-of-order.";
 }
 
 namespace ros
@@ -42,25 +114,24 @@ namespace ros
 
 void SingleThreadedSpinner::spin(CallbackQueue* queue)
 {
-  boost::recursive_mutex::scoped_try_lock spinlock(spinmutex);
-  if(!spinlock.owns_lock()) {
-    ROS_ERROR("SingleThreadedSpinner: You've attempted to call spin "
-              "from multiple threads.  Use a MultiThreadedSpinner instead.");
-    return;
-  }
-
-  ros::WallDuration timeout(0.1f);
-
   if (!queue)
   {
     queue = getGlobalCallbackQueue();
   }
 
+  if (!spinner_monitor.add(queue, true))
+  {
+    ROS_ERROR_STREAM("SingleThreadedSpinner: " << DEFAULT_ERROR_MESSAGE);
+    return;
+  }
+
+  ros::WallDuration timeout(0.1f);
   ros::NodeHandle n;
   while (n.ok())
   {
     queue->callAvailable(timeout);
   }
+  spinner_monitor.remove(queue);
 }
 
 MultiThreadedSpinner::MultiThreadedSpinner(uint32_t thread_count)
@@ -70,13 +141,6 @@ MultiThreadedSpinner::MultiThreadedSpinner(uint32_t thread_count)
 
 void MultiThreadedSpinner::spin(CallbackQueue* queue)
 {
-  boost::recursive_mutex::scoped_try_lock spinlock(spinmutex);
-  if (!spinlock.owns_lock()) {
-    ROS_ERROR("MultiThreadeSpinner: You've attempted to call ros::spin "
-              "from multiple threads... "
-              "but this spinner is already multithreaded.");
-    return;
-  }
   AsyncSpinner s(thread_count_, queue);
   s.start();
 
@@ -97,7 +161,6 @@ private:
   void threadFunc();
 
   boost::mutex mutex_;
-  boost::recursive_mutex::scoped_try_lock member_spinlock;
   boost::thread_group threads_;
 
   uint32_t thread_count_;
@@ -136,8 +199,7 @@ AsyncSpinnerImpl::~AsyncSpinnerImpl()
 
 bool AsyncSpinnerImpl::canStart()
 {
-  boost::recursive_mutex::scoped_try_lock spinlock(spinmutex);
-  return spinlock.owns_lock();
+  return spinner_monitor.canSpin(callback_queue_, false);
 }
 
 void AsyncSpinnerImpl::start()
@@ -145,18 +207,13 @@ void AsyncSpinnerImpl::start()
   boost::mutex::scoped_lock lock(mutex_);
 
   if (continue_)
-    return;
+    return; // already spinning
 
-  boost::recursive_mutex::scoped_try_lock spinlock(spinmutex);
-  if (!spinlock.owns_lock()) {
-    ROS_WARN("AsyncSpinnerImpl: Attempt to start() an AsyncSpinner failed "
-             "because another AsyncSpinner is already running. Note that the "
-             "other AsyncSpinner might not be using the same callback queue "
-             "as this AsyncSpinner, in which case no callbacks in your "
-             "callback queue will be serviced.");
+  if (!spinner_monitor.add(callback_queue_, false))
+  {
+    ROS_ERROR_STREAM("AsyncSpinnerImpl: " << DEFAULT_ERROR_MESSAGE);
     return;
   }
-  spinlock.swap(member_spinlock);
 
   continue_ = true;
 
@@ -172,14 +229,10 @@ void AsyncSpinnerImpl::stop()
   if (!continue_)
     return;
 
-  ROS_ASSERT_MSG(member_spinlock.owns_lock(), 
-                 "Async spinner's member lock doesn't own the global spinlock, hrm.");
-  ROS_ASSERT_MSG(member_spinlock.mutex() == &spinmutex, 
-                 "Async spinner's member lock owns a lock on the wrong mutex?!?!?");
-  member_spinlock.unlock();
-
   continue_ = false;
   threads_.join_all();
+
+  spinner_monitor.remove(callback_queue_);
 }
 
 void AsyncSpinnerImpl::threadFunc()
--- a/clients/roscpp/src/libros/spinner.cpp
+++ b/clients/roscpp/src/libros/spinner.cpp
@@ -153,7 +153,6 @@ void MultiThreadedSpinner::spin(CallbackQueue* queue)
   s.start();
 
   ros::waitForShutdown();
-  s.stop();
 }
 
 class AsyncSpinnerImpl
--- a/clients/roscpp/src/libros/spinner.cpp
+++ b/clients/roscpp/src/libros/spinner.cpp
@@ -70,7 +70,7 @@ struct SpinnerMonitor {
     return true;
   }
 
-  /// remove a queue to the list
+  /// remove a queue from the list
   void remove(ros::CallbackQueue* queue)
   {
     boost::mutex::scoped_lock lock(mutex_);
@@ -89,7 +89,7 @@ struct SpinnerMonitor {
 
 SpinnerMonitor spinner_monitor;
 const std::string DEFAULT_ERROR_MESSAGE =
-    "Attempt to spin a callback queue from two spinners, one of them being single-threaded."
+    "Attempt to spin a callback queue from two spinners, one of them being single-threaded. "
     "This will probably result in callbacks being executed out-of-order.";
 }
 
--- a/clients/roscpp/src/libros/spinner.cpp
+++ b/clients/roscpp/src/libros/spinner.cpp
@@ -76,7 +76,7 @@ struct SpinnerMonitor
       return false;
 
     if (it == spinning_queues_.end())
-      spinning_queues_.insert(it, std::make_pair(queue, Entry(tid)));
+      it = spinning_queues_.insert(it, std::make_pair(queue, Entry(tid)));
 
     if (!single_threaded) // increment number of multi-threaded spinners
       it->second.num_multi_threaded += 1;
--- a/clients/roscpp/src/libros/spinner.cpp
+++ b/clients/roscpp/src/libros/spinner.cpp
@@ -92,7 +92,11 @@ struct SpinnerMonitor
     ROS_ASSERT_MSG(it != spinning_queues_.end(), "Call to SpinnerMonitor::remove() without matching call to add().");
 
     if (it->second.tid != boost::thread::id() && it->second.tid != boost::this_thread::get_id())
-      ROS_ERROR("SpinnerMonitor::remove() called from different thread than add().");
+    {
+      // This doesn't harm, but isn't good practice?
+      // It was enforced by the previous implementation.
+      ROS_WARN("SpinnerMonitor::remove() called from different thread than add().");
+    }
 
     if (single_threaded)
       spinning_queues_.erase(it);
--- a/clients/roscpp/src/libros/spinner.cpp
+++ b/clients/roscpp/src/libros/spinner.cpp
@@ -121,7 +121,7 @@ void SingleThreadedSpinner::spin(CallbackQueue* queue)
 
   if (!spinner_monitor.add(queue, true))
   {
-    ROS_ERROR_STREAM("SingleThreadedSpinner: " << DEFAULT_ERROR_MESSAGE);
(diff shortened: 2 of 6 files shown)


rhaschke:
PR #377 introducing `AsyncSpinner::canStart()` is incomplete: As another spinner might have been started in between the two calls to `canStart()` and `start()`, `start()` can again silently fail.

A proper solution would be to change API and return the success of `start()`. However, to avoid this API change(), I suggest either to `ROS_ISSUE_BREAK` in response to the fatal error or to start spinning in any case, thus making `canStart()` obsolete again. The next commit decided for the latter.

rhaschke:
Instead of starting spinning in any case as suggested in bce1f4b, I decided to throw an exception, when spinning cannot be safely started. This allows for proper feedback as requested in #277, but generally should abort the program and force the developer to fix his program logic.

Having this in place, allows to activate the spinner tests as proper unittests. They were not automatically run, because (i) there was no introspection into spinner state available and (ii) individual tests had to be run independent of each other.

rhaschke:
Friendly ping.

clients/roscpp/include/ros/spinner.h
```diff
@@ -109,10 +109,12 @@ class ROSCPP_DECL AsyncSpinner
...
   /**
-   * \brief Check if the spinner can be started. A spinner can't be started if
-   * another spinner is already running.
+   * \brief Check if the spinner can be started. The spinner shouldn't be started if
+   * another single-threaded spinner is already operating on the callback queue.
+   *
+   * This function is not necessary anymore. start() will always start spinning.
    */
-  bool canStart();
+  ROS_DEPRECATED bool canStart();
```
dirk-thomas: I am not convinced that we should add a deprecation warning to an already released distribution. Maybe the deprecation warning can be added for L-turtle.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -30,10 +30,90 @@
...
+ */
+struct SpinnerMonitor
+{
+  /* store spinner information per callback queue:
+     Only as unique single-threaded spinner is allowed and we store its thread id.
+     As multiple multi-threaded spinners are allowed in parallel, their number is counted.
+  */
+  struct Entry
+  {
+    Entry(const boost::thread::id &tid) : tid(tid) {}
```
dirk-thomas: The member `num_multi_threaded` should be initialized with `0`.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -30,10 +30,90 @@
...
+
+    boost::thread::id tid; // thread id of single-threaded spinner
+    unsigned int num_multi_threaded; // number of multi-threaded spinners
```
dirk-thomas: The current logic to increment and decrement is checking `!single_threaded`. This logic could be simplified if its semantic is changed to `num_spinners`. It can then be increment unconditionally (independent of the callers type) and also the decrement / cleanup becomes easier.

dirk-thomas:
Beside my minor comments this looks good to me. @ros/ros_team Any comments?

wjwwood:
The changes lgtm, +1.

tfoote:
This looks like a good approach to me. +1 for deferring the deprecation warning though.

dirk-thomas:
Thanks. Once the comments have been addressed it can be merged.

clients/roscpp/include/ros/spinner.h
rhaschke: OK. However, if this gets merged, we should directly create an L-turtle branch and apply the deprecation.

clients/roscpp/src/libros/spinner.cpp
rhaschke: Good point.

clients/roscpp/src/libros/spinner.cpp
rhaschke: I actually started with this simpler implementation, but turned to the current implementation on purpose:
We want to allow multiple multi-threaded spinners on a queue, but only one single-threaded spinner - in that case also excluding any other multi-threaded spinners.
This will guarantee proper in-order event processing by the single single-threaded spinner.
If there is a multi-threaded spinner acting on a queue, in-order processing obviously isn't considered important by the programmer and thus multiple (multi-threaded) spinners are allowed.

clients/roscpp/include/ros/spinner.h
dirk-thomas: Since we will not start building from that branch until early 2017  I don't see I reason to create that branch now. We can create a separate ticket so that the deprecation is not forgotten but done as soon as the branch exists.

clients/roscpp/src/libros/spinner.cpp
tfoote: I don't agree that `in-order` processing is not important if the programmer supports multi-threading. Those are very different dimensions. Multi-threading allows parallel processing and is useful for long running things but they can still be stacked and sequenced appropriately. 

Using an example of processing a video stream. You can currently trivially subscribe with a multi-threaded spinner and process the frames in the order they come in and have a book keeping process that locks correctly such that the N parallel threads return in the order they were started or based on their timestamp sorting. If there's a older message in one thread hold it until older ones have been popped out of the parallel processing. 

However if you break the in order assumption you no longer can assume that the number of threads are the latest N messages. And if you care about the order, (say you're processing video frames) you now have to have a cache of the latest messages and insert the frames back into a sorted manner, holding them for some duration at which you expect that the ordering will not be out of order. 

To effectively process data which may come in out of order we need to provide enough information for developers to reliably reorder the data. Off the top of my head how many items might be out of order, how long they might need to wait for the older messages. And the really important thing to be able to tell developers is how much extra computational overhead and latency will this requirement to resort the data cost them?

In order arrival is currently something that can be relied upon in the transport, and since our queue is in order I think it's worth the effort to keep that guarentee alive and make sure that the interleaving queues pop data in order as well.

rhaschke:
Rebased to latest kinetic-devel. buildfarm seems to check for cmake warnings now, which made the old branch fail.

dirk-thomas:
The CMake warnings will still be there. They require a genmsg release to be fixed (ros/rosdistro#12586). If those are the only warnings they won't prevent this PR to be merged.

rhaschke:
Good to know. Looks like all other issues are resolved.

rhaschke:
@dirk-thomas: Friendly ping.

clients/roscpp/src/libros/spinner.cpp
dirk-thomas: The current patch allows more then one multi threaded spinners to handle the same queue. @tfoote based on your last paragraph it sounds like you are not agreeing that this should be the case? Am I reading this right?

I am not sure if I understand what you are saying correctly. @rhaschke point is that once you choose to use a multi threaded spinner (with N > 1) you can't rely on the order anyway which I would agree to.

Consider the following example where the queue contains the messages `A, B, C`. A single threaded spinner will take them in order and process each item using whatever action (e.g. creating an `A'` from the `A`). The results will for sure have the same order `A', B', C'` since the time the process each message is irrelevant. But for a multi threaded spinner I don't see how that would currently be guaranteed. While it would still pick the messages out of the queue in the same order due to processing them in parallel the result order might be different, e.g. `B', C', A'` if `A` took significantly longer to process then `B` and `C`.

@tfoote Can you please clarify your statement and your opinion of the current state of the patch.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -70,17 +149,11 @@ MultiThreadedSpinner::MultiThreadedSpinner(uint32_t thread_count)
...
-    ROS_ERROR("MultiThreadeSpinner: You've attempted to call ros::spin "
-              "from multiple threads... "
-              "but this spinner is already multithreaded.");
-    return;
-  }
   AsyncSpinner s(thread_count_, queue);
   s.start();
 
   ros::waitForShutdown();
+  s.stop();
```
dirk-thomas: This call is redundant and should be removed.

@rhaschke If you could update this that would be great. Otherwise I can apply it during the merge.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -30,10 +30,90 @@
...
+
+    return true;
+  }
+
+  /// remove a queue from the list
+  void remove(ros::CallbackQueue* queue, bool single_threaded)
+  {
+    boost::mutex::scoped_lock lock(mutex_);
+    std::map<ros::CallbackQueue*, Entry>::iterator it = spinning_queues_.find(queue);
+    if (it != spinning_queues_.end())
```
tfoote: It would be good to have a way to catch the error condition of the else for this if. It should at least ROS_ERROR. And this method should probably have a bool return code.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -30,10 +30,90 @@
...
+    {
+      if (it->second.tid != boost::thread::id() && it->second.tid != boost::this_thread::get_id())
+        ROS_ERROR("SpinnerMonitor::remove() called from different thread than add().");
```
tfoote: Is this an error or a warning? If it's an error it should return false and not continue.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -136,27 +208,21 @@ AsyncSpinnerImpl::~AsyncSpinnerImpl()
...
-    ROS_WARN("AsyncSpinnerImpl: Attempt to start() an AsyncSpinner failed "
-             "because another AsyncSpinner is already running. Note that the "
-             "other AsyncSpinner might not be using the same callback queue "
-             "as this AsyncSpinner, in which case no callbacks in your "
-             "callback queue will be serviced.");
-    return;
+  if (!spinner_monitor.add(callback_queue_, false))
+  {
+    ROS_FATAL_STREAM("AsyncSpinnerImpl: " << DEFAULT_ERROR_MESSAGE);
+    throw std::runtime_error("There is already a single-threaded spinner on this queue");
```
tfoote: This is a notable change in behavior to escalate to a runtime_error where the old behavior is to not register the callback queue. We could consider making this change in an upcoming release, but we should not make this change in  an existing distro.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -42,25 +122,24 @@ namespace ros
...
-
   if (!queue)
   {
     queue = getGlobalCallbackQueue();
   }
 
+  if (!spinner_monitor.add(queue, true))
+  {
+    ROS_FATAL_STREAM("SingleThreadedSpinner: " << DEFAULT_ERROR_MESSAGE);
+    throw std::runtime_error("There is already another spinner on this queue");
```
tfoote: See other comment about API stability.

tfoote (commented):
Overall this looks good, but I'm concerned that this changes the behavior significantly due to possibly throwing and tearing down running systems.

clients/roscpp/src/libros/spinner.cpp
tfoote: Indeed once things are in separate threads you're at the whim of the scheduler. I was more making the point that there's a level that we can make sure that the multiple servicing callback queues coordinate cleanly and do things like immediately invoke the callbacks they pop off the queue instead of popping a batch of the incoming queue and then internally queuing them for invocation such that the other multithreaded callback queue gets a batch of later messages and starts invoking callback from the 2nd batch while the first multithreaded callback queue is processing threads from the first batch. I don't think that's the case in our implementation. So I'm not objecting to the current patch but want to make sure that we are doing our best effort to make sure invoke callbacks in order up to the limit of our control.

clients/roscpp/src/libros/spinner.cpp
rhaschke: I did this on purpose: Indeed, the old behavior simply warned on the console, but otherwise silently continued without registering the callback queue. So, if the user / developer doesn't notice the warning in first place, he might assume everything is fine, but events are never processed on this queue!
The code out there either doesn't encounter this error/exception because AsyncSpinner was correctly used beforehand or - if it encountered the warning - it wasn't processing its events at all, which (hopefully) would have been triggered the developer to look for the error too. Hence, probably / hopefully such code was never released.

clients/roscpp/src/libros/spinner.cpp
rhaschke: Will do.

clients/roscpp/src/libros/spinner.cpp
rhaschke: OK, I will turn this into a warning.

clients/roscpp/src/libros/spinner.cpp
rhaschke: This is private code, only used in spinner.cpp. As in the existing code, calls to add() and remove() are always paired, the condition should be always fulfilled and thus should be replaced by a ROS_ASSERT. I will do so.

clients/roscpp/src/libros/spinner.cpp
dirk-thomas: The original behavior was printing an error message and then returned gracefully ignoring the new request. This behavior should be maintained. Therefore the `ROS_FATAL_STREAM` should be replaced with `ROS_ERROR_STREAM` (since its not fatal anymore, but indicates an error in using the API) and `throw` should be replaced with a `return`.

clients/roscpp/src/libros/spinner.cpp
dirk-thomas: I agree with the arguments from both of you. But since Kinetic has already been released for a while any behavior change should be avoided (no matter how unlikely it is that code relies on it). Therefore the same conclusion as above:
- `ROS_FATAL_STREAM` -> `ROS_ERROR_STREAM`
- `throw` -> `return`.

rhaschke: I see. However, this is not as trivial as changing `throw` -> `return`.

Previous behavior was to accept multiple spinners on a queue as long as they were _started_ from the same thread. To retain this wrong behavior, I need to remember which thread initially started spinning on a particular queue and then allow other threads as well.

Hence, the main improvement remaining from this PR in Kinetic, is the ability to have multiple spinners on different queues.

rhaschke:
The last commit 47ed5e91f49b17ce162f2d64dcea29dd813c0bc0 should be reverted on the L-turtle branch. @dirk-thomas If you create an L-turtle branch, I will file a corresponding PR and enable the deprecation warning in the header (which was removed in 835579fcb017b9eb4334a6185c003ed91984789d).

clients/roscpp/src/libros/spinner.cpp
dirk-thomas: I don't see why we need `initial_tid` and the condition check based on it. We only need `tid` to distinguish single threaded and multi threaded spinners. The same queue can only be handled by one single threaded spinner _or_ any number of multi threaded spinners - no matter what the thread id of the multi threaded spinner is.

rhaschke: @dirk-thomas What you describe is the new and intended behavior and I fully agree. However, @tfoote asked to maintain the previous behavior and not throw. Previously, we could have the following situations:
1. `S1` (single-threaded spinner started in `thread 1`): will block `thread 1` until shutdown. Any further spinners in different threads were not allowed (with an error message).
2. `M1` (multi-threaded spinner started in `thread 1`): Further spinners started from _different_ threads were not allowed (with an error message).
3. `M1 ... M1 S1` (multi-threaded spinners started in `thread 1` and afterwards a single-threaded one started): This was accepted without any errors. But the new behavior is to reject `S1`!

Restrictions of case 1 + 2 are relaxed with this PR: Other spinners are allowed as long as the operate on a different queue. Thread doesn't matter.

The tricky part is case 3, which - although nonsense - was perfectly valid code before.
In order to maintain the old behavior, I need to remember, which thread the first M-spinner was started in (the `initial_tid`). If I wouldn't store the `initial_tid`, but allow S on any thread, this would relax unintended behavior even beyond previous behavior.

dirk-thomas: The "collision" in case 3 doesn't matter in my opinion. The fact that from which thread the multi-threaded spinners have been started from shouldn't be considered. The only collision to avoid is which queues they handle and to avoid that a single-threaded spinner handles the same queue as a multi-threaded spinner.

rhaschke: I fully agree. But it wasn't like this before!
To retain old behavior as requested by Tully, I need to introduce `initial_id`. Actually, case 3 might be a common use case: starting some AsyncSpinners before finally entering the `ros::spin` loop.

dirk-thomas: As long as both spinners don't spin on the same queue I think that is totally fine - independent from which thread the multi-threaded one was started from. That should be achieve but just removing the `initial_tid` again and its check, correct?

rhaschke: If spinners operate on different queues, there is no conflict at all and spinners will be started without problems (this is the basic improvement we gain with this PR).

However, I thought we discuss the case, where spinners want to operate on the _same_ queue. There might be code out there, which hits case 3. Removing `initial_id` and the corresponding check, but allowing to start the S\* spinner from any thread after M1 is operating on the queue, will be a _weaker check_ than before: The old code before this PR, at least rejected spinning if S\* and M1 were started from different threads. Of course, we would like to reject always, but Tully requested to not do so.

dirk-thomas:
This is getting really complicated :worried: I will try to summarize before / goal / after for easier understanding. Just writing this summary based on reading the code took me a good amount of time. Hopefully it will help others to get an overview in the future:
1. Without this patch
   1. The queue is not relevant in any decision
   2. The code doesn't consider thread ids at all
   3. Every spinner tries to get a recursive lock on a global mutex
   4. Therefore multiple spinners started from different threads are ignored
   5. Since a multi-threaded spinner does nothing in its blocking thread it was called with it can't be used to start more spinners
   6. Since single-threaded spinners execute the events in the same thread that could be used to start other spinners
   7. The async spinners are not blocking so the same thread can be used to start other spinners
2. Desired change of behavior
   1. The queue should be considered in the decision if multiple spinners can operate concurrently
   2. If spinners operate on different queues there is no need to restrict them
   3. Multiple multi-threaded spinners can handle the same queue
3. With this patch
   1. If the queue doesn't overlap allow new spinners (`bool can_spin = (it == spinning_queues_.end() || ...`)
   2. A multi-threaded spinners can be started for the same queue if the already running spinner is also a multi-threaded spinner (`bool can_spin = (... || it->second.tid == tid);`)
   3. If an additional spinner is added for the same queue and they are of diffeerent type (`it->second.tid == tid` and `it->second.initial_tid == tid`) they are allowed for backward compatibilty and print a message warning the user that events might not be handled in order
   4. If a single-threaded spinner is being started from the same thread as another single-threaded spinner (operating on the same queue) it is being allowed (`it->second.tid == tid`)

For the case 3.iv I see a problem with the bookkeeping. When the second spinner is added it overrides the existing entry in `spinning_queues_` and removes the entry when it finished. I am not sure if this is a relevant case but after the second spinner finished the data structure is inconsistent and doesn't know that the first spinner is still around. Arguably that can only happen when `ok()` returns `false` which might not be a real world problem.

Please let me know if I got something wrong (or also if you agree with the summary).

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -30,10 +30,109 @@
...
+
+    if (!can_spin)
+    {
+      // Previous behavior (up to Kinetic) was to accept multiple spinners on a queue
+      // as long as they were started from the same thread. Although this is wrong behavior,
+      // we retain it here for backwards compatibility, i.e. we allow spinning of a
+      // single-threaded spinner after several multi-threaded ones, given that they
+      // were started from the same initial thread
+      if (it->second.initial_tid == tid)
+        ROS_ERROR_STREAM("SpinnerMonitor: " << DEFAULT_ERROR_MESSAGE);
```
tfoote: This error message should be different to reflect the backwards compatibility incase anyone reads it in depth.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -30,10 +30,109 @@
...
+  /* store spinner information per callback queue:
+     Only one unique single-threaded spinner is allowed and we store its thread id.
+     As multiple multi-threaded spinners are allowed in parallel, their number is counted.
+  */
+  struct Entry
+  {
+    Entry(const boost::thread::id &tid,
+          const boost::thread::id &initial_tid) : tid(tid), initial_tid(initial_tid), num_multi_threaded(0) {}
+
+    boost::thread::id tid; // thread id of single-threaded spinner
```
tfoote: It would be great to have a comment here that it will be the default value which represents 'Not-a-Thread` if multithreaded.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -30,10 +30,109 @@
...
+      // This doesn't harm, but isn't good practice?
+      // It was enforced by the previous implementation.
+      ROS_WARN("SpinnerMonitor::remove() called from different thread than add().");
+    }
+
+    if (single_threaded)
+      spinning_queues_.erase(it);
+    else
+    {
+      ROS_ASSERT(it->second.num_multi_threaded > 0);
```
tfoote: It would be great to have a message for this assert like: "Call to SpinnerMonitor::remove() for a multi-threaded spinner cannot be achieved since reference count is not greater than 0."

tfoote (commented):
That looks like a good summary @dirk-thomas I've been through it too and believe that it's as you've summarized. I had two usability comments on error messages, but otherwise it looks good to me.

tfoote:
With respect to the double entry for the single threaded spinners. It will likely lead to some possible errors on teardown.  It will likely hit [this assert](https://github.com/ros/ros_comm/pull/867/files#diff-54085ad311fb86a9bb4615192321d722R111) and since we know this we could soften that so that we can catch it appropriately.

rhaschke:
I fully agree with 1. and 2, with a minor remark on 1.ii: It's correct that previous code didn't explicitly deal with thread ids. However, this was implicitly done in the recursive mutex remembering the thread that holds the lock.

In principle I also agree to 3. However, I have some remarks:
- 3.ii: it->second.tid and tid can be either 0 (indicating a multi-threaded spinner) or a proper thread-id (indicating a single-threaded spinner). Hence, the condition `bool can_spin = (... || it->second.tid == tid);` evaluates true when existing spinners and the new one are multi-threaded (both tid's == 0) or if the new (single-threaded) spinner originates from the same thread as the existing one.
- 3.iii: The given condition is wrong (description is correct). 
  `!can_spin = (it != spinning_queues_.end() && it->second.tid != tid` (same queue, but different type). Thus, the backwards compatibility condition is `(it->second.tid != tid and it->second.initial_tid == tid)`.
- 3.iv: To be honest, I wasn't (actively) aware of this situation anymore. However, I remember that I thought about that situation. Currently, the `SpinnerMonitor` is tailored towards the existing `SingleThreadedSpinner`, which blocks until ROS is shutdown (`while ros::ok() {...}`). Hence, even if we have nested single-threaded spinners, only the inner-most would remain active. And if he finishes, he finishes because of ROS was shutdown, i.e. the outer spinner won't spin anything anymore. Thus, currently 3.iv is somewhat artificial and wouldn't harm as you noticed as well.
  However, thinking about future single-threaded spinners that might be able to finish() without ROS shutdown() your argument is perfectly valid. I adapted the code to do "spinner counting" for both, single- and multi-threaded spinners. This also simplifies code structure.

clients/roscpp/src/libros/spinner.cpp
rhaschke: I mentioned it would be NULL in the reworked commit.

clients/roscpp/src/libros/spinner.cpp
rhaschke: Done.

clients/roscpp/src/libros/spinner.cpp
rhaschke: I opted for a shorted one: "SpinnerMonitor::remove(): Invalid spinner count (0) encountered."

dirk-thomas:
Thank you for updating the logic. It is definitely easier to follow the flow now. I think this is ready to be merged, Thanks for iterating on this!

rhaschke:
Cool. Thanks for your patience. Could you create a L-turtle branch to revert the backwards compatibility commit there? I will file a corresponding PR then. Or should I simply file an issue as a reminder to revert 91be0e5?

dirk-thomas:
I don't plan to create a branch for the next ROS distro until we start building Debian packages for it on the build farm. Simply because it implies additional overhead. Please go ahead and create an issue to not forget about it. I will also search through the code and look for TODO comments or comments mentioning l-turtle after branching.

PR #872, opened 2016-08-23
[rostopic] Fix rostopic echo for list of ROS message to fix #868

### LINKED ISSUE #868: 'rostopic echo' returning incomplete message data for some topics
it appears that the kinetic-devel branch has a regression, introduced at commit https://github.com/ros/ros_comm/commit/e0f63978fe0dd239ba8002a7e79756cdf1a335d0

`rostopic echo` may not properly return the message data for some topics

this can be reproduced as follows:
- put the latest ros_comm (kinetic-devel branch) in a source overlay workspace (i overlaid this on-top of indigo)
- start roscore in one terminal
- start the sample TF broadcaster node (code below) in another terminal
- run `rostopic echo /tf`. this will return `transforms: None`
- spinning up a rosnode (tested via rospy) and subscribing to the topic does indeed result in the expected message data. this is only a problem using `rostopic echo` on the commandline it seems

repeating the above on a previous commit (i.e. 9ff6955679afc791240aba637dacde6eeb5ae8eb) of ros_comm has no issue and `rostopic echo /tf` returns the expect result

sample TF broadcaster node:

```
#!/usr/bin/env python  
import rospy
import tf
import time

if __name__ == '__main__':
    rospy.init_node('test')

        br = tf.TransformBroadcaster()
        br.sendTransform((0, 0, 0),
                         tf.transformations.quaternion_from_euler(0, 0, 0),
                         rospy.Time.now(),
                         'test',
                         "world")
        time.sleep(1)
```

wkentaro:
Fix #868

The output will be like below:

```
% rostopic echo /tf
transforms: 
  - 
    header: 
      seq: 0
      stamp: 
        secs: 1471928060
        nsecs: 475493655
      frame_id: world
    child_frame_id: base_link
    transform: 
      translation: 
        x: 0.0
        y: 0.0
        z: 0.0
      rotation: 
        x: 0.0
        y: 0.0
        z: 0.0
        w: 1.0

% rostopic echo /tf --noarr
transforms: <array type: geometry_msgs/TransformStamped, length: 1>

% rostopic echo /tf --nostr
transforms: 
  - 
    header: 
      seq: 0
      stamp: 
        secs: 1471928069
        nsecs: 386026988
      frame_id: world
    child_frame_id: base_link
    transform: 
      translation: 
        x: 0.0
        y: 0.0
        z: 0.0
      rotation: 
        x: 0.0
        y: 0.0
        z: 0.0
        w: 1.0
```

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -1362,7 +1362,8 @@ def create_value_transform(echo_nostr, echo_noarr):
                 try:
                     msg_class = genpy.message.get_message_class(t)
                     if msg_class is None:
-                        continue
+                        # happens for list of ROS messages like std_msgs/String[]
+                        raise ValueError
                     nested_transformed = value_transform(f_val)
                     setattr(val_trans, f, nested_transformed)
                 except ValueError:


dirk-thomas:
Thanks!

PR #874, opened 2016-08-24
[rostopic] Get parent topics (upstream in pipeline) with rostopic hz

wkentaro:
Moved from https://github.com/ros/ros_comm/pull/718
to send PR to K

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -241,15 +241,15 @@ def _get_ascii_table(header, cols):
 def _sleep(duration):
     rospy.rostime.wallsleep(duration)
 
-def get_parent_topics(topic, parent_topics=None, impl=False):
-    """Get topics to which the specified topic subscribes to.
+def get_parent_topics(topic, parent_topics=None, recursive=False):
+    """Get topics to which the node of specified topic subscribes to.
 
     @type topic: ``str``
     @param topic: topic name
     @type parent_topics: ``NoneType`` or ``list``
     @param parent_topics: list of parent topics which already got.
-    @type impl: ``bool``
-    @param impl: get parent topics recursively or not
+    @type recursive: ``bool``
+    @param recursive: get parent topics recursively or not
     """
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Should there be a non-recursive option?

I think the documentation needs to be rephrased. I know what you mean with "the node of specified topic" but that is not very clear. Also the term `parent` is not that good imo. @ros/ros_team What do you think? Any better naming suggestions.

wjwwood:
If this is recursive, does it avoid cycles?

I don't have any suggestions for naming. Maybe upstream rather than parent, but upstream implies a tree to me, which again is not necessarily the case.

dirk-thomas:
`upstream` sounds like a better term to me too.

> Get topics to which the node of specified topic subscribes to.

This doc string could be more precise imo, e.g.: "Get topics to which all nodes publishing the passed topic are subscribed to."

When thinking about this more the logic to consider upstream topics (recursively or non-recursively) could be useful beyond `rostopic hz`. Maybe it would be better to have a separate verb (or an existing one with special arguments) which returns that upstream topic list. And then the result of that can be passed to `rostopic hz`?

Anyway before integrating this feature a bug with multiple topics and `rostopic hz` needs to be addressed: https://github.com/ros/ros_comm/pull/712#issuecomment-244490918

wkentaro:
> upstream sounds like a better term to me too.
> 
> Get topics to which the node of specified topic subscribes to.
> This doc string could be more precise imo, e.g.: "Get topics to which all nodes publishing the passed topic are subscribed to."

I agree.

> When thinking about this more the logic to consider upstream topics (recursively or non-recursively) could be useful beyond rostopic hz. Maybe it would be better to have a separate verb (or an existing one with special arguments) which returns that upstream topic list. And then the result of that can be passed to rostopic hz?

+1
Maybe `rostopic upstream /topic`?

dirk-thomas:
The proposed verb name `upstream` sounds ok to me. @ros/ros_team Any thoughts on this?

wjwwood:
Fine by me.

dirk-thomas:

PR #875, opened 2016-08-24
[rospy] Logging function for rospy with node/func named msg

wkentaro:
Below line outputs below image:

``` python
#!/usr/bin/env python

import rospy

rospy.init_node('a')

rospy.loginfo_func_named('a')
rospy.logwarn_func_named('a')
rospy.logfatal_func_named('a')
rospy.logerr_func_named('a')

rospy.loginfo_node_named('a')
rospy.logwarn_node_named('a')
rospy.logfatal_node_named('a')
rospy.logerr_node_named('a')

rospy.loginfo_node_func_named('a')
rospy.logwarn_node_func_named('a')
rospy.logfatal_node_func_named('a')
rospy.logerr_node_func_named('a')

rospy.loginfo_func_named('%d', 1)
rospy.logwarn_func_named('%d', 1)
rospy.logfatal_func_named('%d', 1)
rospy.logerr_func_named('%d', 1)

rospy.loginfo_node_named('%d', 1)
rospy.logwarn_node_named('%d', 1)
rospy.logfatal_node_named('%d', 1)
rospy.logerr_node_named('%d', 1)

def main():
    rospy.loginfo_func_named('a')
    rospy.logwarn_func_named('a')
    rospy.loginfo_node_named('a')
    rospy.logwarn_node_named('a')
    rospy.loginfo_node_func_named('a')
    rospy.logwarn_node_func_named('a')
main()

class Test(object):
    def __init__(self):
        rospy.loginfo_func_named('a')
        rospy.logwarn_func_named('a')
        rospy.loginfo_node_named('a')
        rospy.logwarn_node_named('a')
        rospy.loginfo_node_func_named('a')
        rospy.logwarn_node_func_named('a')
Test()
```

screen shot 2016-08-25 at 12 06 56 am

PR #877, opened 2016-08-25
[rospy] Write log for class method with class name for rospy

wkentaro:
For https://github.com/ros/ros_comm/pull/875

```
$ export ROSCONSOLE_FORMAT='[${severity}] [${time}]: [${node}] [${function}] ${message}'
```

```
$ python spam.py
[INFO] [1472155704.567361]: [/a] [<module>] a
[WARN] [1472155704.568193]: [/a] [<module>] a
[FATAL] [1472155704.568894]: [/a] [<module>] a
[ERROR] [1472155704.569581]: [/a] [<module>] a
[INFO] [1472155704.570351]: [/a] [main] a
[WARN] [1472155704.571100]: [/a] [main] a
[FATAL] [1472155704.571878]: [/a] [main] a
[ERROR] [1472155704.572626]: [/a] [main] a
[INFO] [1472155704.573439]: [/a] [Test::__init__] a
[WARN] [1472155704.574176]: [/a] [Test::__init__] a
[FATAL] [1472155704.574937]: [/a] [Test::__init__] a
[ERROR] [1472155704.575672]: [/a] [Test::__init__] a
```

```
$ cat spam.py

import rospy

rospy.init_node('a')

rospy.loginfo('a')
rospy.logwarn('a')
rospy.logfatal('a')
rospy.logerr('a')

def main():
    rospy.loginfo('a')
    rospy.logwarn('a')
    rospy.logfatal('a')
    rospy.logerr('a')
main()

class Test(object):
    def __init__(self):
        rospy.loginfo('a')
        rospy.logwarn('a')
        rospy.logfatal('a')
        rospy.logerr('a')
Test()
```


dirk-thomas:
This patch seems to not resolve the line number. Doesn't this breaks existing behavior when `${line}` is being used.

Please also add a test for the new behavior.

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -39,12 +39,31 @@
...
+        """
+        Find the stack frame of the caller so that we can note the source
+        file name, line number, and function name with class name if possible.
+        """
+        frame = inspect.stack()[3][0]
+        file_name = inspect.getabsfile(frame)
+        func_name = frame.f_code.co_name
+        try:
+            class_name = frame.f_locals['self'].__class__.__name__
+            func_name = '%s::%s' % (class_name, func_name)
```
dirk-thomas: `%s::%s` is not the Python way to reference a method of a class. Please use `%s.%s` instead.

wkentaro:
Fixed.

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -39,12 +39,31 @@
...
+        file_name = inspect.getabsfile(frame)
+        func_name = frame.f_code.co_name
+        try:
+            class_name = frame.f_locals['self'].__class__.__name__
+            func_name = '%s.%s' % (class_name, func_name)
+        except KeyError:
+            pass
+        return file_name, lineno, func_name
+
+logging.setLoggerClass(RospyLogger)
```
dirk-thomas, line 65: This should ensure not to overwrite custom logger classes already being set by user code. See https://docs.python.org/2/library/logging.html#logging.getLoggerClass on how to achieve this.

Could you please also cover that case in the test.

wkentaro:
Fixed.

dirk-thomas:
Please use the approach described in the link from https://github.com/ros/ros_comm/pull/877#discussion_r77016247 since it also allows to customize the logger if a user defined class has been set before.

wkentaro:
> This should ensure not to overwrite custom logger classes already being set by user code. See https://docs.python.org/2/library/logging.html#logging.getLoggerClass on how to achieve this.

But below code overwrites the custom logger class. Is that ok?

``` python
class MyLogger(logging.getLoggerClass()):
    # ... override behaviour here
```

dirk-thomas:
Yes, but it creates the subclass not of the original logger but of the currently set logger class which could be set by a user before. So it will maintain all other changes from the custom class (except the overridden functionality).

wkentaro:
Ok, fixed.

dirk-thomas:
The patch currently fails one of the new tests.

wkentaro:
Fixed.

tools/rosgraph/test/isolated/test_roslogging_user_logger.py
```diff
@@ -0,0 +1,133 @@
+# Software License Agreement (BSD License)
```
dirk-thomas: Why is this file in a subfolder?

dirk-thomas: This file shouldn't repeat all the tests from the other file but only check the `UserCustomLogger`.

tools/rosgraph/CMakeLists.txt
```diff
@@ -17,4 +17,5 @@ catkin_install_python(PROGRAMS scripts/rosgraph
 
 if(CATKIN_ENABLE_TESTING)
   catkin_add_nosetests(test)
+  catkin_add_nosetests(test/isolated/test_roslogging_user_logger.py)
```
dirk-thomas: This line shouldn't be necessary. The line above already collects all tests in the `test` folder.

tools/rosgraph/test/isolated/test_roslogging_user_logger.py
wkentaro: > This file shouldn't repeat all the tests from the other file but only check the UserCustomLogger.

Ok I will fix it.

tools/rosgraph/CMakeLists.txt
wkentaro: Actually, the test fails on my PC if I run nosetests for test directory.
below shows the cases. The cause seems the logger at running time, but not sure how to fix this at this time.

```
$ ls test
test_ros_logging_user_logger.py  ....py ....py  ....py
$ nosetests test  # fails
$ nosetests test/test_ros_logging_user_logger.py  # success
```

dirk-thomas: I haven't tried it but the problem might be related to the fact that you do things in the module level of the test file. Please try to move all logic into a test function (and eventually a setup and teardown function).

wkentaro: That sounds fine. I will try it.

wkentaro:
Well, .. it still fails

```
% nosetests test/    
........ERROR: Invalid remapping argument ':=:='
ERROR: Invalid remapping argument 'foo:=bar:=baz'
..........invalid ROS_IP (must be a valid IPv4 or IPv6 address)
invalid ROS_IP (must be a valid IPv4 or IPv6 address)
.invalid ROS_IP (must be a valid IPv4 or IPv6 address)
.invalid ROS_IP (must be a valid IPv4 or IPv6 address)
..............................F..
FAIL: test.test_roslogging_user_logger.test_roslogging_user_logger
  File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/wkentaro/ros/ros_comm_ws/src/ros_comm/tools/rosgraph/test/test_roslogging_user_logger.py", line 112, in test_roslogging_user_logger
    assert_regexp_matches(log_actual, log_expected)
AssertionError: Regexp didn't match: 'INFO 127.0.0.1 Hello world. [0-9]*\\.[0-9]* [0-9]* rosout <filename> <lineno> <func_name> /unnamed [0-9]*\\.[0-9]*' not found in 'INFO Hello world. 1477682768.908821 140199866681152 rosout /home/wkentaro/ros/ros_comm_ws/src/ros_comm/tools/rosgraph/test/test_roslogging_user_logger.py 93 test_roslogging_user_logger /unnamed 1477682768.908852'

Ran 53 tests in 0.531s

FAILED (failures=1)

% nosetests test/test_roslogging_user_logger.py
.
Ran 1 test in 0.079s

OK
```

dirk-thomas:
Addressed by #1043.

Issue #876, opened 2016-08-25
Different output of rosconsole message using ROSCONSOLE_FORMAT environmental variable


wkentaro:
Refering to http://wiki.ros.org/rosconsole#Console_Output_Formatting, it says the default value is

```
export ROSCONSOLE_FORMAT='[${severity}] [${time}]: ${message}'
```

But the outputs are different

```
% export ROSCONSOLE_FORMAT='[${severity}] [${time}]: ${message}'

% python spam.py 
[INFO] [1472113962.464240]: a
[WARN] [1472113962.464692]: a
[FATAL] [1472113962.465004]: a
[ERROR] [1472113962.465293]: a
[INFO] [1472113962.465572]: a
[WARN] [1472113962.465838]: a
[ERROR] [1472113962.466106]: a
[FATAL] [1472113962.466405]: a
[INFO] [1472113962.466765]: a
[WARN] [1472113962.467037]: a
[ERROR] [1472113962.467304]: a
[FATAL] [1472113962.467577]: a
```

```
% unset ROSCONSOLE_FORMAT 

% python spam.py         
[INFO] [WallTime: 1472113988.890666] a
[WARN] [WallTime: 1472113988.891116] a
[FATAL] [WallTime: 1472113988.891413] a
[ERROR] [WallTime: 1472113988.891683] a
[INFO] [WallTime: 1472113988.891950] a
[WARN] [WallTime: 1472113988.892203] a
[ERROR] [WallTime: 1472113988.892459] a
[FATAL] [WallTime: 1472113988.892712] a
[INFO] [WallTime: 1472113988.893018] a
[WARN] [WallTime: 1472113988.893272] a
[ERROR] [WallTime: 1472113988.893521] a
[FATAL] [WallTime: 1472113988.893764] a
```

dirk-thomas:
The default value in C++ is as mentioned in the wiki: https://github.com/ros/ros_comm/blob/cc808bd9a266149f2fffd46b623435927e719d20/tools/rosconsole/src/rosconsole/rosconsole.cpp#L108

> [${severity}] [${time}]: ${message}

In Python support for the formatting was implemented in #519. If the environment variable is not set the patch maintained the previous behavior which included the `WallTime:` string.

I am not sure if it is worth changing the default behavior in Python or if the wiki page should just mention it.

dirk-thomas:
[...]

PR #879, opened 2016-08-26
[rosgraph] Fix rospy default rosconsole format for consistency with roscpp

wkentaro:
Close https://github.com/ros/ros_comm/issues/876

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -174,33 +174,28 @@ class RosStreamHandler(logging.Handler):
     def emit(self, record):
         level, color = _logging_to_rospy_names[record.levelname]
         record_message = _defaultFormatter.format(record)
-        if 'ROSCONSOLE_FORMAT' in os.environ.keys():
-            msg = os.environ['ROSCONSOLE_FORMAT']
-            msg = msg.replace('${severity}', level)
-            msg = msg.replace('${message}', str(record_message))
-            msg = msg.replace('${walltime}', '%f' % time.time())
-            msg = msg.replace('${thread}', str(record.thread))
-            msg = msg.replace('${logger}', str(record.name))
-            msg = msg.replace('${file}', str(record.pathname))
-            msg = msg.replace('${line}', str(record.lineno))
-            msg = msg.replace('${function}', str(record.funcName))
-            try:
-                from rospy import get_name
-                node_name = get_name()
-            except ImportError:
-                node_name = '<unknown_node_name>'
-            msg = msg.replace('${node}', node_name)
-            if self._get_time is not None and not self._is_wallclock():
-                t = self._get_time()
-            else:
-                t = time.time()
-            msg = msg.replace('${time}', '%f' % t)
-            msg += '\n'
+        msg = os.environ.get('ROSCONSOLE_FORMAT',
+                             '[${severity}] [${time}]: ${message}')
+        msg = msg.replace('${severity}', level)
+        msg = msg.replace('${message}', str(record_message))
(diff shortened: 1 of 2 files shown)


tfoote:
+1 It looks like this repo is failing the newly added cmake warnings settings.

dirk-thomas:
The CMake warnings have been addressed by ros/genmsg#65.

tfoote:

dirk-thomas:
This is not going to pass until `genmsg` is released.

dirk-thomas:
@ros/ros_team What do you think? Should we merge this and make the default behavior the same or not and keep the currently different default in rospy?

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -174,33 +174,28 @@ def __init__(self, colorize=True):
...
+        msg = msg.replace('${node}', node_name)
+        if self._get_time is not None and not self._is_wallclock():
+            t = self._get_time()
         else:
-            msg = '[%s] [WallTime: %f]' % (level, time.time())
-            if self._get_time is not None and not self._is_wallclock():
-                msg += ' [%f]' % self._get_time()
-            msg += ' %s\n' % record_message
+            t = time.time()
+        msg = msg.replace('${time}', '%f' % t)
```
wjwwood: In the case that simtime is being used, roscpp prints both the wall clock and the sim time, like this:

```
[INFO] [${walltime}, ${simtime}]: ${message}
```

Previously Python did:

```
[INFO] [WallTime: ${walltime}] [${simtime}] ${message}
```

With the `[${simtime}]` part being excluded if simtime is not being used.

This patch doesn't work for me, since it no longer prints wall and sim time at the same time, it now prints wall or sim time. This is both different from the current behavior and different from roscpp (which is the goal of this pr as far as I gather).

wjwwood: This is relevant C++ code for the non-trivial expansion of `${time}`:

https://github.com/ros/ros_comm/blob/kinetic-devel/tools/rosconsole/src/rosconsole/rosconsole.cpp#L206-L213

wjwwood:
I'm ok with changing the logging format output in Kinetic so that it is consistent with roscpp. We might want to let people know in a high profile way, say with an email to ros-users@, and at that point ask if anyone is depending on the format of the log output in any serious way.

I could imagine people who are parsing log output for some purpose like scraping meta data from log files. This hypothetical user would be broken by a change like this.

tfoote:
I think it's worth making the default consistent. In our email we can give the environment variable to set if you need the old behavior. I think it's unlikely that there will be much hard coded to that format since it's only for rospy logging. Roscpp clearly has a different format. Though we should match the time representation as mentioned by @wjwwood

tools/rosgraph/src/rosgraph/roslogging.py
wkentaro: Thanks. I think I fixed it.

wkentaro:
I think I fixed the code following the reviews. Could you please review again?

dirk-thomas:
Please see wkentaro/ros_comm#1 for proposed changes to your PR.

dirk-thomas:
I documented the new unified behavior in http://wiki.ros.org/action/diff/rosconsole?action=diff&rev1=71&rev2=72

@ros/ros_team Does this look ready-to-merge to you?

wkentaro:
@dirk-thomas Thanks.

wjwwood:
lgtm

dirk-thomas:
Thank you. I have cherry-picked the change in 7be2c06bad7a9bfeb31459f053d68c0928bcb0a6.

PR #880, opened 2016-08-26
register nosetests only when testing is enabled

dirk-thomas:
Only register tests when testing is enabled. This will fix one of the CMake warnings in the devel job. The remaining ones should be addressed by ros/genmsg#65.


tfoote:
+1

PR #882, opened 2016-08-30
[roslaunch] Added condition to skip empty ('') <include> paths

jliviero:
Our internal build system runs roslaunch-check against relevant packages, which will fail when it encounters `<include file='' />`. This change allows roslaunch-check to pass, but prints a warning when such an include is encountered.

This allows a file path to be passed in as an arg, and left empty by default, and still have roslaunch-check pass (but with a warning).  For example:

```
<launch>
  <arg name="some_mesh_launch" default=""/>

  <group if="$(eval len(some_mesh_launch) > 0)">
    <include file="$(arg some_mesh_launch)"/>
  </group>

</launch>
```

### DIFF
--- a/tools/roslaunch/src/roslaunch/depends.py
+++ b/tools/roslaunch/src/roslaunch/depends.py
@@ -147,11 +147,12 @@ def _parse_launch(tags, launch_file, file_deps, verbose, context):
             except KeyError as e:
                 raise RoslaunchDepsException("Cannot load roslaunch <%s> tag: missing required attribute %s.\nXML is %s"%(tag.tagName, str(e), tag.toxml()))
 
-            # Check if an empty file is included, and skip if so.  This will allow a default-empty <include> inside a
+	    # Check if an empty file is included, and skip if so.  This will allow a default-empty <include> inside a
             # conditional to pass
             if sub_launch_file == '':
-                print("WARNING: empty <include> in %s. Skipping <include> of %s" % (launch_file,
-                                                                                    tag.attributes['file'].value))
+                if verbose:
+                    print("Empty <include> in %s. Skipping <include> of %s" % (launch_file,
+                                                                           tag.attributes['file'].value))
             else:
                 if verbose:
                     print("processing included launch %s"%sub_launch_file)
@@ -161,7 +162,7 @@ def _parse_launch(tags, launch_file, file_deps, verbose, context):
                 if sub_pkg is None:
                     print("ERROR: cannot determine package for [%s]"%sub_launch_file, file=sys.stderr)
 
-                elif sub_launch_file not in file_deps[launch_file].includes:
+                if sub_launch_file not in file_deps[launch_file].includes:
                     file_deps[launch_file].includes.append(sub_launch_file)
                 if launch_file_pkg != sub_pkg:
                     file_deps[launch_file].pkgs.append(sub_pkg)
--- a/tools/roslaunch/src/roslaunch/depends.py
+++ b/tools/roslaunch/src/roslaunch/depends.py
@@ -146,32 +146,39 @@ def _parse_launch(tags, launch_file, file_deps, verbose, context):
                 sub_launch_file = resolve_args(tag.attributes['file'].value, context)
             except KeyError as e:
                 raise RoslaunchDepsException("Cannot load roslaunch <%s> tag: missing required attribute %s.\nXML is %s"%(tag.tagName, str(e), tag.toxml()))
-                
-            if verbose:
-                print("processing included launch %s"%sub_launch_file)
 
-            # determine package dependency for included file
-            sub_pkg = rospkg.get_package_name(os.path.dirname(os.path.abspath(sub_launch_file)))
-            if sub_pkg is None:
-                print("ERROR: cannot determine package for [%s]"%sub_launch_file, file=sys.stderr)
-            
-            if sub_launch_file not in file_deps[launch_file].includes:
-                file_deps[launch_file].includes.append(sub_launch_file)
-            if launch_file_pkg != sub_pkg:            
-                file_deps[launch_file].pkgs.append(sub_pkg)
-            
-            # recurse
-            file_deps[sub_launch_file] = RoslaunchDeps()
-            try:
-                dom = parse(sub_launch_file).getElementsByTagName('launch')
-                if not len(dom):
-                    print("ERROR: %s is not a valid roslaunch file"%sub_launch_file, file=sys.stderr)
-                else:
-                    launch_tag = dom[0]
-                    sub_context = _parse_subcontext(tag.childNodes, context)
-                    _parse_launch(launch_tag.childNodes, sub_launch_file, file_deps, verbose, sub_context)
-            except IOError as e:
-                raise RoslaunchDepsException("Cannot load roslaunch include '%s' in '%s'"%(sub_launch_file, launch_file))
+            # Check if an empty file is included, and skip if so.  This will allow a default-empty <include> inside a
+            # conditional to pass
+            if sub_launch_file == '':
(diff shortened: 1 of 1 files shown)


tfoote:
Can you explain your use case a little bit more? 

And please consider adding test cases to make sure it behaves in the expected ways. From your example it looks like the modified code path will never be hit.

jliviero:
Certainly.  This covers a case where we want to configure a set of parameters in a Gazebo simulation at launch time. In our specific case, this constitutes a series of additions to the environment (which is why I chose "some_mesh" - the actual arg is arbitrary).  However, we don't need or want these changes on every launch, and we want to be able to quickly swap between configurations of the additions (pose, etc.) to test different solutions.

edited to add: Our internal build system runs roslaunch-check against the relevant package, which will fail when it encounters `<include file=''/>`.  This change allows roslaunch-check to pass, but prints a warning when such an include is encountered.

~~As-written, you are correct - the modified path would not be hit.  However, by providing an arg in the CLI like so, it would:
`roslaunch robot_sim arbitrary_environment.launch some_mesh_launch:=/path/to/config/launch`~~
(edit: the above didn't address the actual code changes, which are relevant to roslaunch-check only - sorry for any confusion)

dirk-thomas:
I just modified an existing launch file and added an optional `arg`, the `group` as well as the nested `include` as in your original comment. Without any modification in `roslaunch` I was able to run the launch file with as well as without the optional argument.

Can you please clarify your use case and describe why you need the modification in this PR to run the launch file with and without.

jliviero:
> Without any modification in `roslaunch` I was able to run the launch file with as well as without the optional argument.

Ah, I see where I haven't been clear, thank you.  This allows rolaunch-check to pass; it does not affect the behaviour of roslaunch.

Without this change, a file with the modifications you've made based on my example will fail:

```
jliviero@CPR00612:~/ws/src$ rosrun roslaunch roslaunch-check -o - launch/custom_map.launch 
checking launch/custom_map.launch
ERROR: cannot determine package for []
  File "/home/jliviero/ws/src/ros_comm/tools/roslaunch/scripts/roslaunch-check", line 86, in <module>
    error_msg = check_roslaunch_file(roslaunch_path)
  File "/home/jliviero/ws/src/ros_comm/tools/roslaunch/scripts/roslaunch-check", line 46, in check_roslaunch_file
    error_msg = roslaunch.rlutil.check_roslaunch(roslaunch_file)
  File "/home/jliviero/ws/src/ros_comm/tools/roslaunch/src/roslaunch/rlutil.py", line 200, in check_roslaunch
    base_pkg, file_deps, missing = roslaunch.depends.roslaunch_deps([f])
  File "/home/jliviero/ws/src/ros_comm/tools/roslaunch/src/roslaunch/depends.py", line 323, in roslaunch_deps
    rl_file_deps(file_deps, launch_file, verbose)
  File "/home/jliviero/ws/src/ros_comm/tools/roslaunch/src/roslaunch/depends.py", line 224, in rl_file_deps
    parse_launch(launch_file, file_deps, verbose)
  File "/home/jliviero/ws/src/ros_comm/tools/roslaunch/src/roslaunch/depends.py", line 209, in parse_launch
    _parse_launch(launch_tag.childNodes, launch_file, file_deps, verbose, context)
  File "/home/jliviero/ws/src/ros_comm/tools/roslaunch/src/roslaunch/depends.py", line 137, in _parse_launch
    _parse_launch(tag.childNodes, launch_file, file_deps, verbose, context)
  File "/home/jliviero/ws/src/ros_comm/tools/roslaunch/src/roslaunch/depends.py", line 181, in _parse_launch
    launch_file))
roslaunch.depends.RoslaunchDepsException: Cannot load roslaunch include '' in 'launch/custom_map.launch'
```

With my change, the output of roslaunch is:

```
jliviero@CPR00612:~/ws/src$ rosrun roslaunch roslaunch-check -o - launch/custom_map.launch
checking launch/custom_map.launch
WARNING: empty <include> in launch/custom_map.launch. Skipping <include> of $(arg apriltags)
...writing test results to -
passed
```

While the failing test could probably be ignored, it is causing our internal build tests to fail, and this seems like a reasonable enough use-case to submit a fix.

Editing my above posts accordingly.

tools/roslaunch/src/roslaunch/depends.py
```diff
@@ -146,32 +146,39 @@ def _parse_launch(tags, launch_file, file_deps, verbose, context):
...
+            else:
+                if verbose:
+                    print("processing included launch %s"%sub_launch_file)
+
+                # determine package dependency for included file
+                sub_pkg = rospkg.get_package_name(os.path.dirname(os.path.abspath(sub_launch_file)))
+                if sub_pkg is None:
+                    print("ERROR: cannot determine package for [%s]"%sub_launch_file, file=sys.stderr)
+
+                elif sub_launch_file not in file_deps[launch_file].includes:
```
dirk-thomas: Why should this be changed from `if` to `elif`?

tools/roslaunch/src/roslaunch/depends.py
```diff
@@ -146,32 +146,39 @@ def _parse_launch(tags, launch_file, file_deps, verbose, context):
...
-                    launch_tag = dom[0]
-                    sub_context = _parse_subcontext(tag.childNodes, context)
-                    _parse_launch(launch_tag.childNodes, sub_launch_file, file_deps, verbose, sub_context)
-            except IOError as e:
-                raise RoslaunchDepsException("Cannot load roslaunch include '%s' in '%s'"%(sub_launch_file, launch_file))
+
+            # Check if an empty file is included, and skip if so.  This will allow a default-empty <include> inside a
+            # conditional to pass
+            if sub_launch_file == '':
+                print("WARNING: empty <include> in %s. Skipping <include> of %s" % (launch_file,
```
dirk-thomas: If this is a valid use cases making it print a warning sounds not good to me. Should this instead just print a message in verbose mode? Like line 157 but with a different message.

jliviero: Works for me.  Because an empty launch include is _potentially_ problematic I think a strong notification would be preferable, but really I'm good either way.

Note however that by running `rosrun roslaunch roslaunch-check /path/to/files`, this would never be triggered, since [roslaunch-check does not take a verbosity arg](https://github.com/ros/ros_comm/blob/kinetic-devel/tools/roslaunch/scripts/roslaunch-check#L62), and `verbose` defaults to false.  Since this command is a part of our pre-commit workflow, I may address this in a later merge request.

tools/roslaunch/src/roslaunch/depends.py
jliviero: It shouldn't; this was an error, thanks for catching.  Will correct.

dirk-thomas:
To avoid the huge diff can you please avoid the change of indentation and instead use `continue` in the case the include is empty. That will also prevent unintentional changes in the block which should only changed the indentation.

tools/roslaunch/src/roslaunch/depends.py
dirk-thomas: The point is that an empty include can also be completely intentional. So always printing a message about it sounds pretty verbose to me.

jliviero:
> To avoid the huge diff can you please avoid the change of indentation and instead use continue in the case the include is empty. That will also prevent unintentional changes in the block which should only changed the indentation.

Done. Still some whitespace/line length fixes in the block, but it's otherwise unchanged.

dirk-thomas:
Any unnecessary whitespace changes just make it more difficult to backport changes between different branches. Therefore I have cherry-picked your patch (without the whitespace changes) to the _kinetic-devel_ branch: b323b8d5987036357fe2b272b57a77cc9ee2647b

Thank you for the PR.

jliviero:
Fair enough re: whitespace.  Thanks.

PR #885, opened 2016-09-02
need to consider `test_depends` for calculate_missing?

k-okada:
see discussion https://github.com/ros-infrastructure/ros_buildfarm/issues/341#issuecomment-244131670


dirk-thomas:
I don't think considering test dependencies is reasonable here. In the common case where you install a launch file and use this check to ensure that "it will work" (aka all its dependencies are available) test dependencies are generally not available. So the test would pass but the launch file would fail if the Debian package (which doesn't depend on test dependencies) is being installed.

k-okada:
ok, how about https://github.com/ros/ros_comm/pull/887, use `-t` or `--use-test-depends` option when we know the target launch file is only for testing and depends on `test_depend` tag.

dirk-thomas:
That sounds much better. Closing this in favor of #887.

PR #886, opened 2016-09-03
[rostopic] Fix for no messages for one topic in two topics

wkentaro:
Close https://github.com/ros/ros_comm/pull/712#issuecomment-244490918


dirk-thomas:
Thanks.

dirk-thomas:
There is yet another case to catch. See #888.

PR #887, opened 2016-09-03
support use_test_depends option for roslaunch-check

k-okada:
following https://github.com/ros/ros_comm/pull/885, when we check roslaunch, sometimes we have test launch code which depends on `test_depend`


dirk-thomas:
Thanks for the patch. I have cherry-picked your commit with some changes to pass keyword arguments always explicitly by specifying them as a keyword arguments (and not as a positional argument) to the kinetic-devel branch: c224dcd258e8a732ecf098f253856e8d67f2898f

Please make future PRs against the latest branch.

PR #888, opened 2016-09-06
fix rostopic hz to print no message when there are no message on any topic

dirk-thomas:
When passing more then one topic but none of them has any messages it should print "no messages" every second.

Follow up of #886.


wkentaro:
Thank you !

PR #890, opened 2016-09-07
Added missing arg to check_roslaunch_dir() definition

jliviero:
`rosrun roslaunch roslaunch-check /path/to/file` fails on kinetic-devel with the following output:

```
  File "/home/jliviero/ws/src/ros_comm/tools/roslaunch/scripts/roslaunch-check", line 93, in <module>
    error_msg = check_roslaunch_dir(roslaunch_path, use_test_depends=options.test_depends)
TypeError: check_roslaunch_dir() got an unexpected keyword argument 'use_test_depends'
```

It seems the new arg was added to the call, but not the method definition, in [this diff](https://github.com/ros/ros_comm/commit/c224dcd258e8a732ecf098f253856e8d67f2898f).

### DIFF
--- a/tools/roslaunch/scripts/roslaunch-check
+++ b/tools/roslaunch/scripts/roslaunch-check
@@ -48,7 +48,7 @@ def check_roslaunch_file(roslaunch_file, use_test_depends=False):
     if error_msg:
         return "[%s]:\n\t%s"%(roslaunch_file,error_msg)
 
-def check_roslaunch_dir(roslaunch_dir):
+def check_roslaunch_dir(roslaunch_dir, use_test_depends=False):
     error_msgs = []
     for f in os.listdir(roslaunch_dir):
         if f.endswith('.launch'):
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Good catch. Can you please update the diff to only contain the changes compared to the kinetic-devel branch. Currently it seems to include many changes which have already been merged.

jliviero:
> Currently it seems to include many changes which have already been merged.

Just noticed that as well; will do.

tools/roslaunch/scripts/roslaunch-check
```diff
@@ -41,14 +41,14 @@ import rospkg
...
+def check_roslaunch_file(roslaunch_file, use_test_depends=False):
     print("checking", roslaunch_file)
-    error_msg = roslaunch.rlutil.check_roslaunch(roslaunch_file)
+    error_msg = roslaunch.rlutil.check_roslaunch(roslaunch_file, use_test_depends=use_test_depends)
     # error message has to be XML attr safe
     if error_msg:
         return "[%s]:\n\t%s"%(roslaunch_file,error_msg)
 
-def check_roslaunch_dir(roslaunch_dir):
+def check_roslaunch_dir(roslaunch_dir, use_test_depends=False):
```
dirk-thomas, line 51: The parameter also needs to be passed to the `check_roslaunch_file` call in line 57.

jliviero: Got it, good catch.

dirk-thomas:
Thank you for the patch.

For the record this fixes a regression of #887. @k-okada FYI

PR #892, opened 2016-09-09
[roslaunch-check] Typo in named arg in call to check_roslaunch()

jliviero:
Corrected named arg option_use_test_depends in call to check_roslaunch()


mikepurvis:
:+1:

dirk-thomas:
Thanks for the PR. I went for a slightly different fix and dropped the `option_` prefix to make the parameter the same as for the other functions: a3a3b2a8c478f4a5a74d01ba6956374d77a82c65

jliviero:
Neat.  Honestly, I prefer that fix anyway - more consistently is more better, but I didn't know if there was a reason for the existing naming pattern.  Thanks!

Issue #891, opened 2016-09-08
ros::init with no __name and name = "" is not tested in roscpp?


furushchev:
I'd like to report a possible bug at roscpp initialization.
In roscpp initializer `ros::init (const M_string &remappings, const std::string &name, uint32_t options=0)`, a node can be initialized even empty string is passed to `name` argument if remappings are passed through from command line arguments.
But if `__name:=foobar` is removed from `remappings` and `name` is empty, ros node can be registered with invalid name, which can occur unexpected errors.
I did experiment in some conditions. the result is followings:

| remap_name | node_name_empty | test_group | this_node::name | this_node::namespace |
| --- | --- | --- | --- | --- |
| true | false | false | /test_node | / |
| true | false | true | /group1/test_node | //group1 |
| true | true | false | /test_node | / |
| true | true | true | /group1/test_node | //group1 |
| false | false | false | /test_node | / |
| false | false | true | /group1/test_node | //group1 |
| false | true | false | / | / |
| false | true | true | //group1 | //group1 |

In the last condition, namespace and name are the same and other nodes cannot resolve the node.

furushchev:
code for reproduce:

``` c++
#include <ros/ros.h>
#include <std_msgs/String.h>
#include <map>

int main(int argc, char** argv) {
  bool remap_name = true;
  bool node_name_empty = true;
  std::map<std::string, std::string> remaps;
  if (remap_name) remaps["__name"] = "test_node";
  std::string node_name = "test_node";
  if (node_name_empty) node_name = "";
  ros::init(remaps, node_name, (uint32_t)0);

  ROS_INFO_STREAM("getName: " << ros::this_node::getName());
  ROS_INFO_STREAM("getNameSpace: " << ros::this_node::getNamespace());

  ros::NodeHandle nh;
  ros::Publisher pub = nh.advertise<std_msgs::String>("string", 1);

  std_msgs::String s;
  ros::Rate r(10);
  for(int i = 0; ros::ok(); ++i) {
    std::ostringstream os;
    os << "published string " << i;
    s.data = os.str();
    pub.publish(s);
    ros::spinOnce();
    r.sleep();
  }

  return 0;
}
```

``` xml
<launch>
  <arg name="test_group" default="false" />
  <group ns="group1" if="$(arg test_group)">
    <node name="nstest_node" pkg="roscpp_nstest" type="nstest" output="screen" />
  </group>
[...]

PR #895, opened 2016-09-11
[rospy] raise error on rospy.init_node with None or empty node name string

furushchev:
according to the discussion at #891 
current rospy shows warning if node name is empty string, so this pull request ensures the empty string for node name on initialization is invalid with raising error.


dirk-thomas:
Thank you for the improvement. I have cherry picked the changes to the _kinetic-devel_ branch: bd3af70520648783da8aa4d3610f234a4d2bd41f

PR #896, opened 2016-09-14
restore API compatibility of rostopic.ROSTopicHz

dirk-thomas:
Restore API compatibility which was broken in https://github.com/ros/ros_comm/pull/712#issuecomment-245989837.


dirk-thomas:
@ros/ros_team Does that look reasonable to maintain API compatibility even if it is pretty verbose with separate members for the data without a topic and the ones with a topic?

wjwwood (approved):

wkentaro:
@dirk-thomas Thank you!

PR #898, opened 2016-09-19
[kinetic] fix check of test type

stonier:
Similar to the update discussed in [ros/pulls/121](https://github.com/ros/ros/pull/121) and merged in [ros/pulls/123](https://github.com/ros/ros/pull/123).


dirk-thomas:
This patch fails a lot of tests.

I applied a different patch to the _kinetic-devel_ branch: 2736a11ad9f851547df9bf25cc949b065aa09c6c

PR #899, opened 2016-09-19
[indigo] fix check of test type

stonier:
Similar to the update discussed in [ros/pulls/121](https://github.com/ros/ros/pull/121) and merged in [ros/pulls/123](https://github.com/ros/ros/pull/123).


dirk-thomas:
This patch fails a lot of tests. I will backport the patches from the kinetic-devel branch before doing new releases.

PR #906, opened 2016-09-28
Use boost::bind to bind the callback function

jspricke:
Example usage:
void callback(const boost::shared_ptr<M const>&)

message_filters::Subscriber<std_msgs::UInt32> sub(nh, "my_topic", 1);
sub.registerCallback(myCallback);

cf. http://wiki.ros.org/message_filters#Example_.28C.2B-.2B-.29


dirk-thomas:
Thank you!

Issue #908, opened 2016-09-29
rostopic echo topic/field no longer works


mikepurvis:
Looks like this might be more fallout from e0f63978fe0dd239ba8002a7e79756cdf1a335d0:

```
$ rostopic echo /rosout/msg -n 1
Traceback (most recent call last):
  File "<ws>/lib/python2.7/dist-packages/rostopic/__init__.py", line 940, in callback
    self.suffix + '\n')
  File "<ws>/lib/python2.7/dist-packages/rostopic/__init__.py", line 876, in custom_strify_message
    val = value_transform(val)
  File "<ws>/lib/python2.7/dist-packages/rostopic/__init__.py", line 1393, in value_transform
    class TransformedMessage(genpy.Message):
  File "<ws>/lib/python2.7/dist-packages/rostopic/__init__.py", line 1396, in TransformedMessage
    __slots__ = val.__slots__[:]
AttributeError: 'str' object has no attribute '__slots__'

$ rostopic echo /rosout/topics -n 1
Traceback (most recent call last):
  File "<ws>/lib/python2.7/dist-packages/rostopic/__init__.py", line 940, in callback
    self.suffix + '\n')
  File "<ws>/lib/python2.7/dist-packages/rostopic/__init__.py", line 876, in custom_strify_message
    val = value_transform(val)
  File "<ws>/lib/python2.7/dist-packages/rostopic/__init__.py", line 1393, in value_transform
    class TransformedMessage(genpy.Message):
  File "<ws>/lib/python2.7/dist-packages/rostopic/__init__.py", line 1396, in TransformedMessage
    __slots__ = val.__slots__[:]
AttributeError: 'list' object has no attribute '__slots__'
```

@wkentaro @dirk-thomas Thoughts on a fix?

dirk-thomas:
Thank you for reporting this. I can confirm that this is a regression :disappointed: I am looking into it since we really want to sync Kinetic before ROSCon...

wkentaro:
Sorry that I could not catching the case, #909 will fix the problem.

```
% rostopic echo /rosout/msg -n1
hello world 1475185572.54
---

% rostopic echo /rosout/topics -n1
['/chatter', '/rosout']
---

% rostopic echo /rosout/topics -n1 --noarr
None
---

% rostopic echo /rosout/msg -n1 --nostr
None
---
```

PR #909, opened 2016-09-29
[rostopic] Fix rostopic echo for non rosmsg field

wkentaro:
Close https://github.com/ros/ros_comm/issues/908

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -1389,17 +1389,21 @@ def _rostopic_cmd_echo(argv):
 
 def create_value_transform(echo_nostr, echo_noarr):
     def value_transform(val, type_information=None):
+        def transform_field_value(value, value_type, echo_nostr, echo_noarr):
+            if echo_noarr and '[' in value_type:
+                return '<array type: %s, length: %s>' % \
+                    (value_type.rstrip('[]'), len(value))
+            elif echo_nostr and value_type == 'string':
+                return '<string length: %s>' % len(value)
+            elif echo_nostr and value_type == 'string[]':
+                return '<array type: string, length: %s>' % len(value)
+            return value
+
         if not isinstance(val, genpy.Message):
             if type_information is None:
                 return val
-            if echo_noarr and '[' in type_information:
-                return ('<array type: %s, length: %s>' %
-                        (type_information.rstrip('[]'), len(val)))
-            elif echo_nostr and type_information == 'string':
-                return '<string length: %s>' % len(val)
-            elif echo_nostr and type_information == 'string[]':
-                return '<array type: string, length: %s>' % len(val)
-            return val
+            return transform_field_value(val, type_information,
+                                         echo_nostr, echo_noarr)
 
         class TransformedMessage(genpy.Message):
             # These should be copy because changing these variables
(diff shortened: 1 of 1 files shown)


dirk-thomas:
If you run `roslaunch rospy_tutorials talker_listener.launch` and `rostopic echo /chatter --nostr` it prints:

> `data: <string length: 25>`

Therefore I would expect `rostopic echo /chatter --nostr -n 1` to print:

> `<string length: 25>`

but it actually prints:

> `None`

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1389,6 +1389,12 @@ def eval_fn(m):
 
 def create_value_transform(echo_nostr, echo_noarr):
     def value_transform(val):
+        if not isinstance(val, genpy.Message):
+            if echo_nostr and isinstance(val, str):
+                return None
```
dirk-thomas: What about this instead:

`return '<string length: %s>' % len(val)`

wkentaro: Nice.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1389,6 +1389,12 @@ def eval_fn(m):
...
+            elif echo_noarr and isinstance(val, list):
+                return None
```
dirk-thomas: Here I don't see how we could print the type information (`<array type: ??? ...>`):

`return '<array length: %s>' % len(val)`

wkentaro:
With 47adb6d, 

```
% rostopic echo /rosout/msg --nostr
<string length: 24>
---

% rostopic echo /rosout/topics --noarr
<array type: string, length: 2>
---
```

wkentaro:
With 9dd94cf, 

**Before**

```
% rostopic echo /rosout --nostr -n1
header:
  seq: 13891
  stamp:
    secs: 1475187530
    nsecs: 501231908
  frame_id: <string length: 0>
level: 2
name: <string length: 9>
msg: <string length: 41>
file: <string length: 11>
function: <string length: 8>
line: 43
topics: <string length: 2>
```

**After**

```
% rostopic echo /rosout --nostr -n1
header:
  seq: 14060
  stamp:
    secs: 1475187547
    nsecs: 301131010
  frame_id: <string length: 0>
level: 2
name: <string length: 9>
msg: <string length: 41>
file: <string length: 11>
function: <string length: 8>
line: 43
topics: <array type: string, length: 2>
---
```

dirk-thomas:
The current patch duplicates a lot of code / logic. I would recommend to create a function like this:

```
def _value_transform(type_information, val, echo_nostr, echo_noarr):
    if echo_noarr and '[' in type_information:
        return '<array type: %s, length: %s>' % \
            (type_information.rstrip('[]'), len(val))
    elif echo_nostr and type_information == 'string':
        return '<string length: %s>' % len(val)
    elif echo_nostr and type_information == 'string[]':
        return '<array type: string, length: %s>' % len(val)
    return None
```

and call it from the two locations accordingly.

dirk-thomas:
This looks good to me. Thank you for providing this patch so fast.

mikepurvis:
Thanks for the quick turnaround, all.

PR #911, opened 2016-10-04
Add '#Include <vector>' to fix building on GCC-6

ghost:
Building with GCC-6 fails with:

`/var/tmp/portage/dev-ros/rosconsole-1.12.4/work/ros_comm-1.12.4/tools/rosconsole/include/ros/console.h:124:14: error: 'vector' in namespace 'std' does not name a template type
 typedef std::vector<TokenPtr> V_Token;`

'tools/rosconsole/include/ros/console.h' doesn't have  '#Include <vector>'.

It shouldn't rely on transitive header inclusion for <vector> and needs to explicitly include it.

### DIFF
--- a/tools/rosconsole/include/ros/console.h
+++ b/tools/rosconsole/include/ros/console.h
@@ -40,6 +40,7 @@
 #include <cstdarg>
 #include <ros/macros.h>
 #include <map>
+#include <vector>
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you.

PR #916, opened 2016-10-21
[rostopic] Fix typo of arg for _str_plot function

### LINKED ISSUE #915: rostopic echo -p no longer works
There seems to be a new bug with rostopic echo, when printing in the csv format.

It looks like the call at https://github.com/ros/ros_comm/blob/kinetic-devel/tools/rostopic/src/rostopic/__init__.py#L936 expects value_transform as the parameter name. Meanwhile, the function defined at https://github.com/ros/ros_comm/blob/kinetic-devel/tools/rostopic/src/rostopic/__init__.py#L725 takes value_transform_fn as a parameter. Changing either seems to work as long as they're consistent, but I'm not sure which is better to change.

wkentaro:
Fix #915

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -722,7 +722,7 @@ def _sub_str_plot_fields(val, f, field_filter):
     return None
 
 
-def _str_plot(val, time_offset=None, current_time=None, field_filter=None, type_information=None, fixed_numeric_width=None, value_transform_fn=None):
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@geoffviola Can you please try this proposed patch and comment if it resolves the problem you have.

geoffviola:
I tested it on my machine. This patch fixes the issue for me

dirk-thomas:
Thank you, both.

PR #918, opened 2016-10-28
Add node name to the output in rosout.log

### LINKED ISSUE #912: Add node name to rosout.log
Can I suggest we output the node name in the log message to rosout.log? https://github.com/ros/ros_comm/blob/kinetic-devel/tools/rosout/rosout.cpp#L144

patrickcjh:
Fixes #912.

### DIFF
--- a/tools/rosout/rosout.cpp
+++ b/tools/rosout/rosout.cpp
@@ -141,7 +141,8 @@ public:
       ss << msg->level << " ";
     }
 
-    ss << "[" << msg->file << ":" << msg->line << "(" << msg->function << ") ";
+    ss << msg->name << " ";
+    ss << "[" << msg->file << ":" << msg->line << "(" << msg->function << ")] ";
 
     ss << "[topics: ";
     std::vector<std::string>::const_iterator it = msg->topics.begin();


dirk-thomas:

IanTheEngineer:
@commaster90 can you give us a sample of what the text output was before, and a sample of what your change will produce?

tools/rosout/rosout.cpp
```diff
@@ -141,7 +141,8 @@ class Rosout
       ss << msg->level << " ";
     }
 
-    ss << "[" << msg->file << ":" << msg->line << "(" << msg->function << ") ";
+    ss << msg->name << " ";
+    ss << "[" << msg->file << ":" << msg->line << "(" << msg->function << ")] ";
```
dirk-thomas, line 145: Just to make this change more visible. In this line the missing `]` has been added. Without the change the brackets are unbalanced.

patrickcjh:
This is an example output in ~/.ros/latest/rosout.log before the change.

```
1477675270.252137195 INFO [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main) [topics: /rosout, /chatter] hello world 2
1477675270.351831526 INFO [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main) [topics: /rosout, /chatter] hello world 3
1477675270.451983685 INFO [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main) [topics: /rosout, /chatter] hello world 4
1477675270.551896309 INFO [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main) [topics: /rosout, /chatter] hello world 5
1477675270.651884679 INFO [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main) [topics: /rosout, /chatter] hello world 6
```

And this will be the output in ~/.ros/latest/rosout.log after the change.

```
1477674754.711808141 INFO /talker [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main)] [topics: /rosout, /chatter] hello world 2
1477674754.811829213 INFO /talker [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main)] [topics: /rosout, /chatter] hello world 3
1477674754.911385469 INFO /talker [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main)] [topics: /rosout, /chatter] hello world 4
1477674755.011573405 INFO /talker [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main)] [topics: /rosout, /chatter] hello world 5
1477674755.110618167 INFO /talker [/tmp/binarydeb/ros-indigo-roscpp-tutorials-0.5.5/talker/talker.cpp:111(main)] [topics: /rosout, /chatter] hello world 6
```

IanTheEngineer:
That makes sense, but I think I'm a bit confused. I thought the output format was dictated by the `rosconsole` config file:
http://wiki.ros.org/rosconsole#Console_Output_Formatting
It appears that it was never used in `rosout.log` at all. Your change seems like a sensible default, but could break any code that attempts to parse this file. I've written a script like this in the past, but I'm not particularly bothered, as it would be easy to fix. However,  if we obeyed the console output formatting in the `rosout` log, then users could easily configure to add / remove console output formatting fields as they desire.

patrickcjh:
@IanTheEngineer It is my concern too that this might break other code that parse rosout.log. Therefore, I would like to seek opinion here.

I notice that the _rosconsole config file_ you mentioned affects only the output shown on the console, and the config can be different for every node. It however does not log any output to file. It is also a bit confusing because for Python nodes, individual log files for each nodes are generated in the ~/.ros/latest folder (whereas C++ nodes do not have individual logs).  Meanwhile, _rosout_ is running within _roscore_, and it aggregates the log from every nodes and write it to a single rosout.log

IanTheEngineer:
Ah, that is nearly true. The rosconsole var only affects the `stdout` if you are running with `rosrun`, or with a roslaunch node set to `output=console`. But if you are setting roslaunch `output=log`, then this `rosconsole` config environment var dictates the output fields of each line in the `*-stdout.log`. 
Anyway, that's getting into the weeds a bit. I have observed that the `rosout.log` ignores this configuration regardless, as is now obvious by looking at this file that you updated. I am not opposed to your change, but think we should open a ticket to address this inconsistency between log formatting.

dirk-thomas:
I have cherry-picked your patch to the newly created `lunar-devel` branch: bc9222ab9b639ab71cd426d8ecf34fa2ff74ad68

Thank you!

PR #924, opened 2016-11-02
Fix WallTimerEvent class -> struct

mikepurvis:
Aligns the forward decl with its definition, silencing a clang warning:

https://github.com/ros/ros_comm/blob/kinetic-devel/clients/roscpp/include/ros/forwards.h#L149


dirk-thomas:
Thank you!

PR #925, opened 2016-11-02
Drop epydoc from rosbag_storage

mikepurvis:
It doesn't generate anyway: http://docs.ros.org/jade/api/rosbag_storage/html/python/

### DIFF
--- a/tools/rosbag_storage/rosdoc.yaml
+++ b/tools/rosbag_storage/rosdoc.yaml
@@ -1,5 +1,3 @@
- - builder: epydoc
-   output_dir: python
  - builder: doxygen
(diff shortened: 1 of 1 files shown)


tfoote (approved):

dirk-thomas (changes requested):
Since the doxygen entry is the only one remaining I think the file can be removed completely.

mikepurvis:
That will change the existing URLs, I assumed we wouldn't want that.

dirk-thomas:
Good point. While cross references from other packages will be update automatically manual links would certainly break.

dirk-thomas (approved):

PR #926, opened 2016-11-03
Update console.h

tomekent:
Fix for errors using std::vector when not included, this matches the fix of the kinetic branch


helenol:
:+1:, this is fixing a broken build for me on OS X 10.11 with indigo and boost 1.62.

dirk-thomas:
Duplicate of #930.

PR #929, opened 2016-11-10
In write_header() use poll if available.

deng02:
We are hitting the select bug where fileno > 1023
causing an "out of range" error.  Changing the select to a
poll where available to support higher numbered fileno.

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -613,11 +613,29 @@ class TCPROSTransport(Transport):
             return
         fileno = sock.fileno()
         ready = None
-        while not ready:
-            _, ready, _ = select.select([], [fileno], [])
+        poller = None
+        if hasattr(select, 'poll'):
+            poller = select.poll()
+            poller.register(fileno, select.POLLOUT)
+            while not ready:
+                events = poller.poll()
+                for _, flag in events:
+                  if flag & select.POLLOUT:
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -613,11 +613,29 @@ def write_header(self):
...
+                for _, flag in events:
+                  if flag & select.POLLOUT:
+                        ready = True
+        else:
+            while not ready:
+                try:
+                    _, ready, _ = select.select([], [fileno], [])
+                except ValueError as e:
+                    logger.error("[%s]: select fileno '%s': %s", self.name, str(fileno), str(e))
+                    raise e
```
mikepurvis: I believe you can just `raise` and it will automatically re-raise the current exception.

mikepurvis:
Looks good to me. 👍

deng02:
ping @tfoote @dirk-thomas

mikepurvis:
Looks like this is a similar kind of change to what's going on in #833. @mgrrx, can you comment?

Am curious to better understand the relationship between this code and the `Poller` abstraction class seen here: https://github.com/ros/ros_comm/blob/674263142a0be1124b273bdb1e0a79c3c956e259/clients/rospy/src/rospy/topics.py#L185

dirk-thomas:
Just a though: the code path if `poll` is not available likely needs to be tested in Windows.

PR #930, opened 2016-11-10
rosconsole: Adding std::vector include

patrickelectric:
Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/tools/rosconsole/include/ros/console.h
+++ b/tools/rosconsole/include/ros/console.h
@@ -40,6 +40,7 @@
 #include <cstdarg>
 #include <ros/macros.h>
 #include <map>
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please add a description why you are proposing the patch. I assume it is related to GCC 6?

patrickelectric:
@dirk-thomas yes, Message updated.
I am trying to upgrade the packages that are not compiling in Arch Linux aur system.

dirk-thomas:
This has already been addressed on the `kinetic-devel` branch in #911. The fix will be consider for backporting before the next Indigo / Jade release.

PR #931, opened 2016-11-14
Fixing logic issue causing cache not to be saved properly.

bponsler:
The returns were causing the logic to save the cache never execute when a executable/resource was successfully located. This change reduced the XML parsing/loading process for my launch file from ~15 seconds to < 1 second.


dirk-thomas:
Sorry for the late response. Thank you for your patch. I finally got around to take a closer look. I created #966 as a replacement which targets the `kinetic-devel` branch instead and moves the additional code to persist the data out of the try block. It would be great if you could try it and confirm if it works for you too.

PR #938, opened 2016-11-23
Return result of client execute

efernandez:
This replaces https://github.com/ros/ros_comm/pull/928

All test passing. Note that with `return ok;` some rostests fail.

### DIFF
--- a/clients/roscpp/src/libros/master.cpp
+++ b/clients/roscpp/src/libros/master.cpp
@@ -186,9 +186,9 @@ bool execute(const std::string& method, const XmlRpc::XmlRpcValue& request, XmlR
   bool printed = false;
   bool slept = false;
   bool ok = true;
+  bool b = false;
   do
   {
-    bool b = false;
     {
 #if defined(__APPLE__)
       boost::mutex::scoped_lock lock(g_xmlrpc_call_mutex);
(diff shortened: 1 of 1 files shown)


efernandez:
FYI @afakihcpr @mikepurvis @jasonimercer @dirk-thomas

mikepurvis:
LGTM. +1

jasonimercer:
+1

efernandez:
ping @tfoote @dirk-thomas

mikepurvis:
@jspricke @dirk-thomas We've been running this at Clearpath for several months; hasn't caused any regressions in our use-case.

jspricke:
Woah, that function looks quite entangled already. I would prefer to simplify it. But otherwise +1 from me for the fix.

dirk-thomas:
@efernandez Sorry for the long delay and thank you for the patch.

@mikepurvis Thank you for testing this change.

@jspricke Thanks for the review.

efernandez:
Thanks @dirk-thomas 

BTW, @afakihcpr has an additional fix for the case when the ROS initialization is done from a separate thread and it gets shutdown while still initializing. I'm not sure if there's a PR upstream for that one or not.
It solves crashes on nodes like `move_base`, that does the ROS initialization from a thread other than the main one.

PR #939, opened 2016-11-23
Add a missing #include <vector> statement

romainreignier:
It was not possibe to build with GCC 6.2.1 without it.


dirk-thomas:
Duplicate of #930.

romainreignier:
Oh sorry, I have checked before but missed this one :/

PR #942, opened 2016-11-29
[rostopic] Rostopic pub option -1 delay reduced

javierdiazp:
The expected behaviour of "rostopic pub -1 ..." is to send a message without a delay after sending the message, but there is a delay of 3 seconds. If the user wanted the delay, the default option should be used (without -1).

This PR reduces the delay to an imperceptible amount of time.

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -1488,7 +1488,6 @@ def publish_message(pub, msg_class, pub_args, rate=None, once=False, verbose=Fal
     try:
         
         if rate is None:
-            s = "publishing and latching [%s]"%(msg) if verbose else "publishing and latching message"
             if once:
                 s = "[%s] published"%(msg) if verbose else "message published."
             else:
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -1451,7 +1451,7 @@ def _publish_at_rate(pub, msg, rate, verbose=False, substitute_keywords=False, p
         pub.publish(msg)
         r.sleep()
 
-_ONCE_DELAY = 3.
+_ONCE_DELAY = 0.1
 def _publish_latched(pub, msg, once=False, verbose=False):
     """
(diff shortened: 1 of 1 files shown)


jfabry:
This is joint work of Javier with @OmniSliver (credit where credit is due) as part of a course I am teaching at U. Chile. They have a blog about their work (in Spanish) here: https://javierdiazp.github.io/robotica2/

mikepurvis:
I'm not sure this is desirable— the reason for the 3 second delay is that it takes a moment for the underlying XMLRPC activity to occur which sets up the publisher/subscriber socket. 0.1s may not be long enough for the message to actually get through, particularly if the master is on a different host.

jfabry:
The problem that Javier and Patricio had with -1 is that 3 seconds is a really long time for the command to hang on to the console, for no apparent reason. It is a bad user experience that they felt should be avoided.

In their tests with a local master on a low-end machine, 0.1 was enough. They did not test with a remote master. I understand that this scenario will take longer, but I think it would be exceptional to take more than a second. So maybe change it to a 1 second delay?

jspricke:
Why wouldn't you leave the program running? As the topic is latched anyhow, I would do it the other way round. Don't end the program but add a stdout that you can end it with Ctrl-C but it will stop the latched message as well.
-1 for the current implementation because it could brake if you start the subscriber after the publisher.

jfabry:
The point is that if you do want to leave the program running, you can already do a `rostopic pub` without the -1 option.  This addresses your comments: you kill it with ctrl-c and it latches, so you can start the subscriber after the publisher. 

Moreover it's kind of obvious that ctrl-c will also kill `rostopic pub -1` I see my students discovering this all the time. What they don't get and don't like is the 3 seconds waiting.

dirk-thomas:
I agree with the arguments brought up by @mikepurvis and @jspricke. The delay is there for a good reason. Reducing it will break existing use cases where it takes that amount of time to publish the latched message.

An improved message to the user might be helpful though.

mikepurvis:
A new PR to update the console message would be welcome!

PR #943, opened 2016-11-29
[rostopic] Unified list with subscribers and publishers

javierdiazp:
The current verbose option displays two lists with many duplicated topics between lists (those with publishers and subscribers).  This leads to having to search twice the same topic in some cases.

This PR implements a new option (-V or --verbose2) to display an unified list with publishers and subscribers information for each topic. In addition the list is sorted by topic name.

Example of current -v and new -V behaviour:
img

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -991,14 +991,62 @@ def _rostopic_list_bag(bag_file, topic=None):
                 if rospy.is_shutdown():
                     break
 
-def _sub_rostopic_list(master, pubs, subs, publishers_only, subscribers_only, verbose, indent=''):
+def _sub_rostopic_list(master, pubs, subs, publishers_only, subscribers_only, verbose, verbose2, indent=''):
     def topic_type(t, topic_types):
         matches = [t_type for t_name, t_type in topic_types if t_name == t]
         if matches:
             return matches[0]
         return 'unknown type'
 
-    if verbose:
+    if verbose2:
+        topic_types = _master_get_topic_types(master)
+        
+        def getTopic(item): return item[0]
+        sortedPubs = sorted(pubs, key=getTopic)
+        sortedSubs = sorted(subs, key=getTopic)
+        sortedAll = []
+        
+        pubsPos = 0
+        subsPos = 0
+        while pubsPos < len(sortedPubs) and subsPos < len(sortedSubs):
+            pubItem = sortedPubs[pubsPos]
+            subItem = sortedSubs[subsPos]
+            if pubItem[0] == subItem[0]:
+                sortedAll.append([pubItem[0], len(pubItem[1]), len(subItem[1])])
+                pubsPos += 1
+                subsPos += 1
+            elif pubItem[0] < subItem[0]:
+                sortedAll.append([pubItem[0], len(pubItem[1]), 0])
+                pubsPos += 1
+            else:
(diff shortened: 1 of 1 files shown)


jfabry:
This is joint work of Javier with @OmniSliver (credit where credit is due) as part of a course I am teaching at U. Chile. They have a blog about their work (in Spanish) here: https://javierdiazp.github.io/robotica2/

mikepurvis:
This looks good to me, but I question the need for a separate flag. Can we just make the `-v` behaviour be this?

jfabry:
Javier and Patricio did not want to change existing behavior for fear of breaking any other tools that may consume the input of -v and/or the need to change all of the documentation.

I am all for changing -v to this implementation, as I consider it to be much better. I think it is a choice for the maintainers to make since there might be unpleasant side-effects.

jspricke:
+1 for making this `-v`.

evenator:
Maybe make it `-V` in current versions and move it to '-v` for L-turtle and above? I agree that changing output so dramatically in a released version might be disruptive.

dirk-thomas:
When the PR is being made against `kinetic-devel` I would prefer making the existing option `-v` use this new style.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -991,14 +991,74 @@ def _rostopic_list_bag(bag_file, topic=None):
...
             return matches[0]
         return 'unknown type'
 
-    if verbose:
+    if verbose2:
+        topic_types = _master_get_topic_types(master)
+        
+        def getTopic(item): return item[0]
+        sortedPubs = sorted(pubs, key=getTopic)
+        sortedSubs = sorted(subs, key=getTopic)
```
mikepurvis, line 1006: Please use a lambda here instead of a separately defined function.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -991,14 +991,74 @@ def _rostopic_list_bag(bag_file, topic=None):
...
+            
+            pubStr = ("%" + str(pubsDigits) + "d pub") % (p)
+            subStr = ("%" + str(subsDigits) + "d sub") % (s)
+            
+            if p != 1: pubStr += "s,"
+            else: pubStr += ", "
+            if s != 1: subStr += "s,"
+            else: subStr += ", "
+            
+            print(indent+" * %s %s %s [%s]"%(pubStr, subStr, t, topic_type(t, topic_types)))
```
mikepurvis, line 1058: Whitespace, please. See: https://www.python.org/dev/peps/pep-0008/#other-recommendations

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -991,14 +991,74 @@ def _rostopic_list_bag(bag_file, topic=None):
...
+            if (publishers_only and p == 0) or (subscribers_only and s == 0) :
+                continue
+            
+            pubStr = ("%" + str(pubsDigits) + "d pub") % (p)
+            subStr = ("%" + str(subsDigits) + "d sub") % (s)
+            
+            if p != 1: pubStr += "s,"
+            else: pubStr += ", "
+            if s != 1: subStr += "s,"
+            else: subStr += ", "
```
mikepurvis, line 1056: Separate lines, please.

mikepurvis:
@javierdiazp I'd love to merge this; are you willing to fix up the style issues and make the new behaviour use the original flag?

dirk-thomas:
Closing in favor of #1674.

PR #947, opened 2016-12-08
Implemented rate-control-topic and rate-control-max-delay.

EliteMasterEric:
Added the ability to add a rate-control-topic, which allows you to specify a topic where, if it has not been published to in the last <rate-control-max-delay> seconds, playback will be delayed (distinguished from paused) until the topic continues.


EliteMasterEric:
I closed the pull request to make some fixes, and now it is open again.

Below is a zip file containing:
Test.msg, a message type with a timestamp and a string message.
test_control_in.py, a python script that simply prints a timestamped message 20 times a second, used to create a bag.
2016-12-15-12-17-30.bag, a bag file created by recording the output of the above script for 60 seconds.
test_control.py, a python script that simulates a processing function by waiting 100 ms every time it waits a message, before it can receive another. 

To test, run test_control.py, then run 
```
rosrun rosbag play --rate-control-topic /test/out --rate-control-max-delay 1.0 --bags /home/eric/ros_workspace/2016-12-15-12-17-30.bag
```

If the test_control.py script becomes more than 1 second behind, the rosbag playback will be automatically delayed until the test_control.py script catches up.

This is a useful feature to allow rosbag playback that prioritizes processing every message, rather than running at the required rate.

[RateControlTopicTest.zip](https://github.com/ros/ros_comm/files/655414/RateControlTopicTest.zip)

kmhallen:
# Added arguments

```--rate-control-topic /some/topic```
* Watch the given topic, and if the last publish was more than [rate-control-max-delay] ago, wait until the topic publishes again to continue playback
* Defaults to empty string

```--rate-control-max-delay 12.34```
* Maximum time difference in seconds from the timestamp in  [rate-control-topic] to bag time before pausing
* Defaults to 1.0

# Use Case

Playing back recorded lidar data to build a map offline, but how fast can the bag run without dropping messages?

# Example

The bag was recorded at 20Hz. A simulated system can only handle 10Hz. This shows the benefits of the added rate-control arguments.

* Start a roscore
    * ```roscore```
* Simulate 100ms processing delay with queue size of 10
    * ```./test.py```
* Play bag with messages every 50ms, some messages will be dropped
    * ```rosrun rosbag play test.bag --clock -r 1.0```
* Play bag with rate control, bag will pause to prevent dropped messages
    * ```rosrun rosbag play test.bag --clock -r 1.0 --rate-control-topic /test/out --rate-control-max-delay 0.5```


[test.tar.gz](https://github.com/ros/ros_comm/files/655899/test.tar.gz)

kmhallen (changes requested):
Doesn't compile, please update the boost::bind that references updateRateTopicTime(), and check that it compiles.

dirk-thomas:
@MasterEric Thanks for addressing the compilation problem. In the mean time I have resolved the merge conflict to make the patch mergable again. @kmhallen Can you please try this again and update your review status.

kmhallen (approved):

dirk-thomas:
This change does break ABI. @ros/ros_comm-maintainers what do you think? Should this target Lunar instead?

mikepurvis:
We already broke `rosbag::Player` ABI recently, see discussion in https://github.com/ros/ros_comm/pull/949#discussion_r96080225, so I don't think it's a concern.

Bhupesh19:
Is it designed to work with only one topic in the bag file? What happens in the following scenario:
-  Bag file has multiple topics 
-  Node subscribe to one topic and publish one topic (rate-control-topic)

kmhallen:
Yes, it was designed to work with one topic. The bag playback will pause, so all topics will stop until the rate-control-topic catches up.

PR #948, opened 2016-12-12
Add named loggers to rospy 

ggallagher01:
This change will create (named) child loggers under rospy.rosout. This allows for each named logger to have seperate verbosity setting, this will allow for two different verboisty settings for logging in a single node.  Two named loggers can log to rosout at two different verbosity setings, which can help noisy debug logs.

### DIFF
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -72,6 +72,7 @@ from rospy.names import *
 from rospy.impl.validators import ParameterInvalid
 
 from rosgraph_msgs.msg import Log
+from functools import partial
 
 _logger = logging.getLogger("rospy.core")
 
@@ -143,17 +144,38 @@ def rospywarn(msg, *args):
     """Internal rospy client library warn logging"""
     _rospy_logger.warn(msg, *args)
     
-logdebug = logging.getLogger('rosout').debug
 
-logwarn = logging.getLogger('rosout').warning
+def _base_logger(msg, *args, **kwargs):
 
-loginfo = logging.getLogger('rosout').info
+    name = kwargs.pop('logger_name', None)
+    throttle = kwargs.pop('logger_throttle', None)
+    level = kwargs.pop('logger_level', None)
+                    
+    rospy_logger = logging.getLogger('rosout')
+    if name:
+        rospy_logger = rospy_logger.getChild(name)
+    logfunc = getattr(rospy_logger, level)
+
+    if throttle:
+        caller_id = _frame_to_caller_id(inspect.currentframe().f_back.f_back)
+        _logging_throttle(caller_id, logfunc, throttle, msg, *args)
+    else:
+        logfunc(msg, *args)
+
+
+loginfo = partial(_base_logger, logger_level='info')
+
 logout = loginfo # alias deprecated name
 
-logerr = logging.getLogger('rosout').error
+logdebug = partial(_base_logger, logger_level='debug')
+
+logwarn = partial(_base_logger, logger_level='warn')
+
+logerr = partial(_base_logger, logger_level='error')
+
 logerror = logerr # alias logerr
 
-logfatal = logging.getLogger('rosout').critical
+logfatal = partial(_base_logger, logger_level='critical')
 
 
 class LoggingThrottle(object):
@@ -191,29 +213,19 @@ def _frame_to_caller_id(frame):
 
 
 def logdebug_throttle(period, msg):
-    caller_id = _frame_to_caller_id(inspect.currentframe().f_back)
-    _logging_throttle(caller_id, logdebug, period, msg)
-
-
+    logdebug(msg, logger_name=None, logger_throttle=period)
+      
 def loginfo_throttle(period, msg):
-    caller_id = _frame_to_caller_id(inspect.currentframe().f_back)
-    _logging_throttle(caller_id, loginfo, period, msg)
-
-
+    loginfo(msg, logger_name=None, logger_throttle=period)
+              
 def logwarn_throttle(period, msg):
-    caller_id = _frame_to_caller_id(inspect.currentframe().f_back)
-    _logging_throttle(caller_id, logwarn, period, msg)
-
+    logwarn(msg, logger_name=None, logger_throttle=period)
 
 def logerr_throttle(period, msg):
-    caller_id = _frame_to_caller_id(inspect.currentframe().f_back)
-    _logging_throttle(caller_id, logerr, period, msg)
-
-
+    logerr(msg, logger_name=None, logger_throttle=period)
+                              
 def logfatal_throttle(period, msg):
-    caller_id = _frame_to_caller_id(inspect.currentframe().f_back)
-    _logging_throttle(caller_id, logfatal, period, msg)
-
+    logfatal(msg, logger_name=None, logger_throttle=period)
 
 class LoggingOnce(object):
 
--- a/test/test_rospy/test/rostest/test_rospy_client_online.py
+++ b/test/test_rospy/test/rostest/test_rospy_client_online.py
@@ -182,6 +182,24 @@ class TestRospyClientOnline(unittest.TestCase):
                     rospy.sleep(rospy.Duration(2))
                 else:
                     self.assert_("test 4" in sys.stderr.getvalue())
+            
+            rospy.loginfo("test child logger 1", logger_name="log1")
+            self.assert_("test child logger 1" in sys.stdout.getvalue())
+            
+            rospy.logwarn("test child logger 2", logger_name="log2")            
+            self.assert_("[WARN]" in sys.stderr.getvalue())
+            self.assert_("test child logger 2" in sys.stderr.getvalue())
+
+            sys.stderr = StringIO()
+            rospy.logerr("test child logger 3", logger_name="log3")            
+            self.assert_("[ERROR]" in sys.stderr.getvalue())
+            self.assert_("test child logger 3" in sys.stderr.getvalue())
+            
+            sys.stderr = StringIO()
+            rospy.logfatal("test child logger 4", logger_name="log4")            
+            self.assert_("[FATAL]" in sys.stderr.getvalue())            
+            self.assert_("test child logger 4" in sys.stderr.getvalue()) 
+
         finally:
             sys.stdout = real_stdout
             sys.stderr = real_stderr
--- a/test/test_rospy/test/unit/test_rospy_core.py
+++ b/test/test_rospy/test/unit/test_rospy_core.py
@@ -79,7 +79,13 @@ class TestRospyCore(unittest.TestCase):
         rospy.logout('out')
         rospy.logerr('err')
         rospy.logfatal('fatal')
-        
+        #basic named logger test
+        rospy.logdebug('debug', logger_name="child1")
+        rospy.logwarn('warn', logger_name="child1")
+        rospy.logout('out', logger_name="child1")
+        rospy.logerr('err', logger_name="child1")
+        rospy.logfatal('fatal', logger_name="child1")
+
     def test_add_shutdown_hook(self):
         def handle(reason):
             pass


clients/rospy/src/rospy/core.py
```diff
@@ -142,20 +143,37 @@ def rospyerr(msg, *args):
 def rospywarn(msg, *args):
     """Internal rospy client library warn logging"""
     _rospy_logger.warn(msg, *args)
-    
-logdebug = logging.getLogger('rosout').debug
```
dirk-thomas: These variables are public API and can't be changed without breaking compatibility.

clients/rospy/src/rospy/core.py
```diff
@@ -109,7 +110,7 @@ def parse_rosrpc_uri(uri):
     @raise ParameterInvalid: if uri is not a valid ROSRPC URI
     """
     if uri.startswith(ROSRPC):
-        dest_addr = uri[len(ROSRPC):]            
+        dest_addr = uri[len(ROSRPC):]
```
dirk-thomas: Please avoid any unrelated white space changes in your PR (even if the current file might be "ugly"). It makes backporting changes across branches much more difficult.

dirk-thomas (commented):
Currently the patch fails to run the tests. Please iterate on this and run the tests locally.

clients/rospy/src/rospy/core.py
mikepurvis: I think this is replaced in line 181 of the new file:

https://github.com/ros/ros_comm/pull/948/files#diff-84739190beb4f06aecd26a816f126e29R181

ggallagher01: I'll take a look at it, I included a bunch of different commit here by accident. I'm cleaning up now

dirk-thomas:
My previous comment still applies to the updated change. Please avoid any unrelated white space changes in your PR.

clients/rospy/src/rospy/core.py
dirk-thomas: Please take a look at the diff on the [GitHub PR](https://github.com/ros/ros_comm/pull/948/files) and update the PR to avoid any unnecessary changes.

ggallagher01: sorry I thought I got them, I'll fix it up

clients/rospy/src/rospy/core.py
```diff
@@ -143,78 +146,114 @@ def rospywarn(msg, *args):
...
+
+        if (last.time is None or (now - last.time) > rospy.Duration(period) or digest != last.digest):
+            logging_func(msg, *args)
+            self.last_log_entry[caller_id] = self.LogEntry(time=now, digest=digest)
+
+
+def _get_digest(*args):
+    m = hashlib.md5()
+    for arg in args:
+        m.update(str(args))
```
dirk-thomas: Should this call `update` with `arg` instead of `args`?

clients/rospy/src/rospy/core.py
```diff
@@ -143,78 +146,114 @@ def rospywarn(msg, *args):
...
+    except KeyError:
+        level = None
+
+    rospy_logger = logging.getLogger('rosout')
+    if name:
+        rospy_logger = rospy_logger.getChild(name)
+    logfunc = getattr(rospy_logger, level)
+
+    if throttle:
+        caller_id = _frame_record_to_caller_id(inspect.stack()[1])
```
dirk-thomas: Is `1` still correct now that the function is one level more nested?

dirk-thomas (commented):
Can you please add some unit tests (https://github.com/ros/ros_comm/blob/kinetic-devel/test/test_rospy/test/unit/test_rospy_core.py) covering the new funtionality.

clients/rospy/src/rospy/core.py
ggallagher01: I believe 2 is correct now because of the one more level down, I'll verify and fix that up

clients/rospy/src/rospy/core.py
ggallagher01: yes, it should be arg instead of args.  I'll fix that up.

mikepurvis: Using MD5 for this seems like overkill. What about something like `hash(tuple(*args + [str(msg)]))`?

(Python's `hash`doesn't do the right thing, so you can't use it directly, see https://github.com/ros/genpy/issues/73)

ggallagher01: That should work, we only want to know the content of the print has changed. I'll test it out tonight.

clients/rospy/src/rospy/core.py
ggallagher01: I'll fix this again, I think the file has mixed line endings and my editor is trying to be helpful

ggallagher01:
I resolved all the white space conflicts that my editor injected.  Is there a standard line ending that each file should follow?  I see this core.py has mixed line endings.  I should have unit tests up tomorrow, I have a test program that I used for develpoment I'll port to the test framework.

clients/rospy/src/rospy/core.py
```diff
@@ -143,78 +145,107 @@ def rospywarn(msg, *args):
...
+
+    try:
+        throttle = kwargs.pop('logger_throttle')
+    except KeyError:
+        throttle = None
+
+    try:
+        level = kwargs.pop('logger_level')
+    except KeyError:
+        level = None
```
mikepurvis: For my curiosity, what does this achieve over, eg:

```
def _base_logger(msg, name=None, throttle=None, level=None, *args, **kwargs):
    rospy_logger = logging.getLogger('rosout')

    if name:
        rospy_logger = rospy_logger.getChild(name)

    logfunc = getattr(rospy_logger, level)

    if throttle:
        caller_id = _frame_record_to_caller_id(inspect.stack()[2])
        _logging_throttle(caller_id, logfunc, throttle, msg, *args)
    else:
        logfunc(msg, *args)
```

ggallagher01: I had the function using default values in one iteration but I believe i started failing tests and decided to use the keyword arguments instead.  I'll rerun the tests and see if I can remember why I switched to using keyword args instead.

ggallagher01: Looking back through my testing, I had a test case where using the named arguments for name, throttle and level (like above) would interfer with variable length arguments.  This happened when testing out code that was using the existing public API, the named variables would end up taking some of the values from the multi length arguments.

mikepurvis: Oh of course, that makes sense. In that case, maybe pass the default to `pop` rather than catching the exception. Eg:

```
def _base_logger(msg, *args, **kwargs):
    name = kwargs.pop('logger_name', None)
    throttle = kwargs.pop('logger_throttle', None)
    level = kwargs.pop('logger_level', None)
```

ggallagher01: much cleaner, will change that

ggallagher01:
I believe all the the comments have been resolved, please let me know if there is anything else needed

mikepurvis:
There seem to be a lot of whitespace-only changes here. With those reverted, I think this can go in.

http://stackoverflow.com/questions/3515597/add-only-non-whitespace-changes

ping @ros/ros_comm-maintainers

clients/rospy/src/rospy/core.py
```diff
@@ -143,78 +145,96 @@ def rospywarn(msg, *args):
...
+logfatal = partial(_base_logger, logger_level='critical')
+
+
+class LoggingThrottle(object):
+
+    LogEntry = namedtuple("LogEntry", "time digest")
+
+    last_log_entry = {}
+
+    def __call__(self, caller_id, logging_func, period, msg, *args):
```
dirk-thomas: `*args` should be documented in the docblock.

clients/rospy/src/rospy/core.py
```diff
@@ -143,78 +145,96 @@ def rospywarn(msg, *args):
...
+
+        - caller_id (str): Id to identify the caller
+        - logging_func (function): Function to do logging.
+        - period (float): Period to do logging in seconds.
+        - msg (object): Message to do logging.
+        """
+        now = rospy.Time.now()
+
+        last = self.last_log_entry.get(caller_id, self.LogEntry(time=None, digest=None))
+        digest = hash(*args + tuple([str(msg)]))
```
dirk-thomas: That seems to change the throttling behavior significantly and goes beyond just adding named loggers. The PR description doesn't mention this at all.

dirk-thomas (commented):
Please make sure that the patch applies to the latest devel branch (`lunar-devel`). You can edit the existing PR to target that branch instead.

clients/rospy/src/rospy/core.py
ggallagher01: This throttles messages that don't change, if the message payload changes then it allows it through.  I can add this to the PR comments if this is something that adds benefit but if it doesn't I will remove.

ggallagher01:
Backed out the changes to the log throttling and now targeting lunar-devel.  My local tests are passing.

dirk-thomas:
Closing and reopening so that the CI job for Lunar picks it up.

The diff shows many lines as modified despite the fact that the content is the same. Please check your patch for any differences (like line endings) and try to make the diff only show actually changed lines.

ggallagher01:
cleaned up the review, looks better now.  Tests are passing on my local machine.

ggallagher01:
Looks good now, white space should be fixed to only what I changed and tests are now passing.

ggallagher01:
ping @ros/ros_comm-maintainers 

All comments I believe are resolved

DLu:
:+1:

dirk-thomas:
Sorry for the late response. This looks almost ready. Recently some concerns regarding performance came up though which have been addressed in #1091 for the other logging function. Can you please update the PR to use a similar approach. Thank you.

This will also need to be rebased.

dirk-thomas:
Thank you for this new feature. Please consider adding some information to the wiki page to mention this to be available as of Lunar and post the link here for future readers.

kartikmohta:
This breaks the `Location` of the message shown in `rqt_console`. Now the location of the message is always shown to be `core.py:_base_logger:163` instead of the user code location.

dirk-thomas:
> This breaks the `Location` of the message shown in `rqt_console`. Now the location of the message is always shown to be `core.py:_base_logger:163` instead of the user code location.

@ggallagher01 Can you please look into this.

mikepurvis:
I think the issue here is that we jump up two stack frames in the throttle case, but in the other case we need to jump up one frame (currently nothing):

```
    if throttle:
         caller_id = _frame_to_caller_id(inspect.currentframe().f_back.f_back)
         _logging_throttle(caller_id, logfunc, throttle, msg, *args)
     else:
         logfunc(msg, *args)
```

I can verify this and throw up a PR for it.

mikepurvis:
Okay, I did dig into this a little more, and it's not quite as simple— the code I pasted only has to do with identifying the caller for the purposes of enforcing throttling. The actual logic which identifies the caller was added only recently in #1043, see:

https://github.com/ros/ros_comm/blob/lunar-devel/tools/rosgraph/src/rosgraph/roslogging.py#L52-L80

It looks like this is a case where the underlying logic built into `logging` is basically doing the right thing as far as identifying call site, but since that API has been wrapped, it's now pointing at the wrapper rather than user code. AFAICT, there are a couple ways this could be handled:

Since the amount of wrapping is not always consistent:

- Put a blacklist in `findCaller` (eg, if you see `rospy/core.py`, keep going, or if you see `base_logger`, keep going, etc).
- In the `partial` and other pass-thru calls, curry in a number for how many steps out to take.

See some discussion here as well: https://stackoverflow.com/questions/19615876/showing-the-right-funcname-when-wrapping-logger-functionality-in-a-custom-class

mikepurvis:
Okay, I've opened #1141 with a potential fix, so we can discuss that and other options there.

gavanderhoorn:
This is a great addition, but [wiki/rospy/Overview/Logging](http://wiki.ros.org/rospy/Overview/Logging) does not mention it. Is that an oversight? @ggallagher01: would you be willing to add that?

From [rospy equivalent of ROS_*_NAMED](https://answers.ros.org/question/275295) on ROS Answes.

ggallagher31:
I'll have some time this weekend to add this

yli-cpr:
I found that http://wiki.ros.org/rospy/Overview/Logging doesn't contain the features this commit adds. It may need to be updated.

Another confusing thing is that on the wiki page, `rospy.log*()` methods don't take `kwargs`. But it seems before this change, they did accept `kwargs`.

I am actually looking at the support of `exc_info`, one of the 2 `kwargs` that python Logger supports (the other is `extra`). See https://docs.python.org/2/library/logging.html.

This ticket probably broke the python logger's `kwargs` like `exc_info`, because it doesn't forward `kwargs` to the logger. Should `rospy` logging support `exc_info` and `extra`?

mgrrx:
@yli-cpr checkout https://github.com/ros/ros_comm/pull/1289 for the support of kwargs.

dirk-thomas:
> I'll have some time this weekend to add this

> I found that http://wiki.ros.org/rospy/Overview/Logging doesn't contain the features this commit adds. It may need to be updated.

@ggallagher31 Can you please follow up on this and add documentation about this feature to the wiki page.

timonegk:
Ping. This feature is still not documented in the ROS wiki. I would be willing to provide the documentation but the wiki site is immutable for me.

gavanderhoorn:
@timonegk: the wiki is read-only, unless you have an account.

You can register for an account [here](http://wiki.ros.org/action/newaccount/Documentation?action=newaccount). After that, you'd have to request write-access (spam protection measure) in ros-infrastructure/roswiki#258.

timonegk:
@gavanderhoorn Thank you. I was not aware of the whitelist.

gavanderhoorn:
I believe it's mentioned after you register.

timonegk:
This feature is now documented (http://wiki.ros.org/rospy/Overview/Logging#Named_Logging).

PR #949, opened 2016-12-14
Pause service for rosbag player

ayrton04:
We've recently required the ability to pause playback in `rosbag play` via service calls within code. The changes include the addition of a `spinOnce()` call within `doPublish()` so that we can process the service request. I'm not sure if having callbacks in the `Player` class is acceptable, or if this functionality would be useful for anyone else, but I thought it worth a PR.

### DIFF
--- a/tools/rosbag/include/rosbag/player.h
+++ b/tools/rosbag/include/rosbag/player.h
@@ -47,6 +47,7 @@
 
 #include <ros/ros.h>
 #include <ros/time.h>
+#include <std_srvs/SetBool.h>
 
 #include "rosbag/bag.h"
 
@@ -177,6 +178,7 @@ private:
 
     void printTime();
 
+    bool pauseCallback(std_srvs::SetBool::Request &req, std_srvs::SetBool::Response &res);
 
 private:
 
@@ -184,9 +186,15 @@ private:
 
     ros::NodeHandle node_handle_;
 
+    ros::ServiceServer pause_service_;
+
     bool paused_;
 
     bool pause_for_topics_;
+
+    bool pause_change_requested_;
+
+    bool requested_pause_state_;
 
     ros::WallTime paused_time_;
 
--- a/tools/rosbag/include/rosbag/player.h
+++ b/tools/rosbag/include/rosbag/player.h
@@ -180,6 +180,8 @@ private:
 
     bool pauseCallback(std_srvs::SetBool::Request &req, std_srvs::SetBool::Response &res);
 
+    void processPause(const bool paused, ros::WallTime &horizon);
+
 private:
 
     PlayerOptions options_;
--- a/tools/rosbag/src/player.cpp
+++ b/tools/rosbag/src/player.cpp
@@ -110,7 +110,8 @@ Player::Player(PlayerOptions const& options) :
     requested_pause_state_(false),
     terminal_modified_(false)
 {
-  pause_service_ = node_handle_.advertiseService("pause", &Player::pauseCallback, this);
+  ros::NodeHandle private_node_handle("~");
+  pause_service_ = private_node_handle.advertiseService("pause", &Player::pauseCallback, this);
 }
 
 Player::~Player() {
@@ -290,7 +291,7 @@ bool Player::pauseCallback(std_srvs::SetBool::Request &req, std_srvs::SetBool::R
 
   if (res.success)
   {
-    res.message = (requested_pause_state_ ? "Pause" : "Resume") + std::string(" requested.");
+    res.message = std::string("Playback is now ") + (requested_pause_state_ ? "paused" : "resumed");
   }
   else
   {
--- a/tools/rosbag/src/player.cpp
+++ b/tools/rosbag/src/player.cpp
@@ -286,8 +286,16 @@ bool Player::pauseCallback(std_srvs::SetBool::Request &req, std_srvs::SetBool::R
   pause_change_requested_ = (req.data != paused_);
   requested_pause_state_ = req.data;
 
-  res.success = true;
-  res.message = (requested_pause_state_ ? "Pause" : "Resume") + std::string(" requested.");
+  res.success = pause_change_requested_;
+
+  if (res.success)
+  {
+    res.message = (requested_pause_state_ ? "Pause" : "Resume") + std::string(" requested.");
+  }
+  else
+  {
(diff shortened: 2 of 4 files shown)


tools/rosbag/src/player.cpp
```diff
@@ -335,6 +358,30 @@ void Player::doPublish(MessageInstance const& m) {
...
+            {
+              paused_ = requested_pause_state_;
+
+              if (paused_)
+              {
+                paused_time_ = ros::WallTime::now();
+              }
+              else
+              {
+                ros::WallDuration shift = ros::WallTime::now() - paused_time_;
```
ayrton04: I realize this logic is duplicated in the keyboard input processing below, and can put it in a method if preferred.

tappan-at-git:
<s>I think having the service be namespaced is a good call, even if we need node_handle_ to be at the base namespace of the node. Otherwise you'll have issues if you run more than one rosbag record instance in the same namespace.</s> Tom has rightly pointed out that 1) this is for playback, not recording, and 2) having a service namespaced under an anonymous node is ridiculously unhelpful.

That said, it may be good to rename the service to something a little less generic if we can't conveniently namespace it. `rosbag_pause` could be a good match with the existing style.

tools/rosbag/src/player.cpp
```diff
@@ -278,6 +281,25 @@ void Player::printTime()
...
+bool Player::pauseCallback(std_srvs::SetBool::Request &req, std_srvs::SetBool::Response &res)
+{
+  pause_change_requested_ = (req.data != paused_);
+  requested_pause_state_ = req.data;
+
+  res.success = pause_change_requested_;
+
+  if (res.success)
+  {
+    res.message = (requested_pause_state_ ? "Pause" : "Resume") + std::string(" requested.");
```
mikepurvis: Wondering if this should be past tense? Eg, `Bag playback is now paused/resumed.`

tools/rosbag/src/player.cpp
```diff
@@ -106,8 +106,11 @@ Player::Player(PlayerOptions const& options) :
     // If we were given a list of topics to pause on, then go into that mode
     // by default (it can be toggled later via 't' from the keyboard).
     pause_for_topics_(options_.pause_topics.size() > 0),
+    pause_change_requested_(false),
+    requested_pause_state_(false),
     terminal_modified_(false)
 {
+  pause_service_ = node_handle_.advertiseService("pause", &Player::pauseCallback, this);
```
mikepurvis: Obviously a different service name here.

Perhaps another option might be:

```
ros::NodeHandle private_node_handle('~');
pause_service_ = private_node_handle.advertiseService("pause", &Player::pauseCallback, this);
```

This should get you a `rosbag/pause` service name, but it will be trivial to rename/remap it from roslaunch or wherever else.

mikepurvis:
This seems simple and straightforward. I'd support a merge if the service name can be sorted out.

tools/rosbag/src/player.cpp
locusroboticsdev: Will fix.

ayrton04: Will fix.

tools/rosbag/src/player.cpp
ayrton04: I wasn't sure what the best route was here. The problem with a private node handle is that if you're running `rosbag play` outside a launch file, it will by default be given an anonymous name, and so calling the service is a pain (e.g., `rosservice call /play_1484077899473435649/pause`). As I couldn't envisage a use case wherein we'd be running more than one `rosbag play` instance, I figured a service in the global namespace would work.

However, I also can't really see using this outside of a launch file, and even if we did, we could assign a name via the command line. I'll move it back to the private node handle.

mikepurvis:
LGTM.

tools/rosbag/src/player.cpp
```diff
@@ -106,8 +106,12 @@ Player::Player(PlayerOptions const& options) :
     // If we were given a list of topics to pause on, then go into that mode
     // by default (it can be toggled later via 't' from the keyboard).
     pause_for_topics_(options_.pause_topics.size() > 0),
+    pause_change_requested_(false),
+    requested_pause_state_(false),
     terminal_modified_(false)
 {
+  ros::NodeHandle private_node_handle("~");
+  pause_service_ = private_node_handle.advertiseService("pause", &Player::pauseCallback, this);
```
jspricke: maybe name it "pause_playback"?

tools/rosbag/include/rosbag/player.h
```diff
@@ -177,17 +178,24 @@ class ROSBAG_DECL Player
...
 
+    bool pauseCallback(std_srvs::SetBool::Request &req, std_srvs::SetBool::Response &res);
 
 private:
 
     PlayerOptions options_;
 
     ros::NodeHandle node_handle_;
 
+    ros::ServiceServer pause_service_;
```
jspricke, line 195: I guess this changes the ABI, but I don't think anyone is relying on that @dirk-thomas ?

tools/rosbag/src/player.cpp
jspricke: +1 for merging methods

jspricke (commented):
Thanks! +1 apart from some comments.

tools/rosbag/src/player.cpp
ayrton04: OK, can do.

tools/rosbag/include/rosbag/player.h
dirk-thomas: If we can avoid breaking the ABI than we should for sure aim for that. While we don't "guarantee" it we try to make an effort where possible.

ayrton04: OK. Shall I make this static?

mikepurvis: I think ABI would be maintained if all the new fields (and method) were appended to the class rather than inserted.

However, IMO this is not worth it for the loss in source readability— there are unlikely to be external packages linking to this class.

ayrton04: OK, I can go either way. Let me know what you'd all prefer.

jspricke: leave it as it is.

mikepurvis:
Looks great— thanks Tom!

PR #189, opened 2013-03-13
Rosbag windows port

stonier:
Fixes applied to groovy ros_comm with the patches originally outlined in
- https://github.com/ros/ros_comm/issues/106
- https://github.com/ros/ros_comm/issues/114

Issue #114, opened 2013-01-09
rosbag: Rosbag play and record for windows patch


ablasdel:
This follows on from  #106 (adds the library export macros for windows).

This patch:
- Windows console code replaces termios code - Boost filesystem v3 (if available) drops in for disk space checks. - Boost date_time for timestamping the name of the rosbag file. - Minor posix-win32 cross platform hacks.

rosbag record and play both work. Compress still seems to have a few problems.

Tested also on an ubuntu catkin build.

migration of trac ticket 4006:
  https://code.ros.org/trac/ros/ticket/4006

```
diff -NBaur -x .svn -x *.pyc ros_comm/tools/rosbag/CMakeLists.txt ros_comm_rosbag_working/tools/rosbag/CMakeLists.txt
--- ros_comm/tools/rosbag/CMakeLists.txt    Fri Jun 22 11:06:36 2012
+++ ros_comm_rosbag_working/tools/rosbag/CMakeLists.txt Thu Jul 12 08:20:11 2012
@@ -5,7 +5,7 @@
   )

 find_package(ROS REQUIRED COMPONENTS cpp_common roscpp_serialization roscpp rosconsole xmlrpcpp topic_tools rosunit rostest)
-find_package(Boost REQUIRED COMPONENTS regex program_options)
+find_package(Boost REQUIRED COMPONENTS date_time regex program_options filesystem)
 find_package(BZip2 REQUIRED)
 # Support large bags (>2GB) on 32-bit systems
 add_definitions(-D_FILE_OFFSET_BITS=64)
diff -NBaur -x .svn -x *.pyc ros_comm/tools/rosbag/include/rosbag/player.h ros_comm_rosbag_working/tools/rosbag/include/rosbag/player.h
--- ros_comm/tools/rosbag/include/rosbag/player.h   Thu Jul 12 11:45:15 2012
+++ ros_comm_rosbag_working/tools/rosbag/include/rosbag/player.h    Thu Jul 12 08:51:11 2012
@@ -36,9 +36,11 @@
 #define ROSBAG_PLAYER_H

 #include <sys/stat.h>
-#include <termios.h>
+#if !defined(_MSC_VER)
+  #include <termios.h>
+  #include <unistd.h>
+#endif
 #include <time.h>
-#include <unistd.h>

 #include <queue>
 #include <string>
@@ -176,8 +178,13 @@

     // Terminal
     bool    terminal_modified_;
+#if defined(_MSC_VER)
[...]

PR #951, opened 2016-12-16
terminate underlying 'rosbag {play,record}' on SIGTERM

jwm:
https://github.com/ros/ros_comm/pull/189 (which is merged, but its underlying bug https://github.com/ros/ros_comm/issues/114 is not closed) migrated from `os.execv()` to `subprocess.call()` to run the `play` and `record` nodes.

Unfortunately, when you send `rosbag` a `SIGTERM`, such as with control-C, the parent `rosbag` process exits, but leaves the child node running, playing or recording the bag.

This traps `SIGTERM` and propagates the signal to the underlying node process. Then, a control-C ends the process as the user would expect.

I don't have a Windows machine to test on, and it looks like I'd have to compile ROS from the source. However, the Python documentation says this approach is compatible with Windows:

https://docs.python.org/2/library/signal.html
> On Windows, signal() can only be called with SIGABRT, SIGFPE, SIGILL, SIGINT, SIGSEGV, or SIGTERM. A ValueError will be raised in any other case.

https://docs.python.org/2/library/subprocess.html
> Popen.terminate()
>     Stop the child. On Posix OSs the method sends SIGTERM to the child. On Windows the Win32 API function TerminateProcess() is called to stop the child.

### DIFF
--- a/tools/rosbag/src/rosbag/rosbag_main.py
+++ b/tools/rosbag/src/rosbag/rosbag_main.py
@@ -65,6 +65,13 @@ def handle_split(option, opt_str, value, parser):
         print("Use of \"--split <MAX_SIZE>\" has been deprecated.  Please use --split --size <MAX_SIZE> or --split --duration <MAX_DURATION>", file=sys.stderr)
         parser.values.size = int(parser.rargs.pop(0))
 
+
+def _stop_process(signum, frame, old_handler, process):
+    process.terminate()
+    if old_handler:
+        old_handler(signum, frame)
+
+
 def record_cmd(argv):
     parser = optparse.OptionParser(usage="rosbag record TOPIC1 [TOPIC2 TOPIC3 ...]",
                                    description="Record a bag file with the contents of specified topics.",
@@ -124,9 +131,15 @@ def record_cmd(argv):
 
     cmd.extend(args)
 
+    old_handler = signal.signal(
+        signal.SIGTERM,
+        lambda signum, frame: _stop_process(signum, frame, old_handler, process)
+    )
     # Better way of handling it than os.execv
     # This makes sure stdin handles are passed to the process.
-    subprocess.call(cmd)
+    process = subprocess.Popen(cmd)
+    process.wait()
+
 
 def info_cmd(argv):
     parser = optparse.OptionParser(usage='rosbag info [options] BAGFILE1 [BAGFILE2 BAGFILE3 ...]',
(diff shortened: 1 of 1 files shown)


jwm:
Example process tree:

```
vagrant@local-driver-ui:~$ ps auwwwxf | grep pts/0
vagrant   9033  0.0  0.2 105644  4940 ?        S    17:42   0:00  |   \_ sshd: vagrant@pts/0
vagrant   9034  0.0  0.3  24504  7384 pts/0    Ss   17:42   0:00  |       \_ -bash
vagrant   9438  2.0  1.0  73328 20532 pts/0    S+   17:50   0:00  |           \_ /usr/bin/python /opt/ros/indigo/bin/rosbag play -l driver_ui_test.bag
vagrant   9441  1.0  0.5 413664 11368 pts/0    Sl+  17:50   0:00  |               \_ /opt/ros/indigo/lib/rosbag/play --loop --queue 100 --rate 1.0 --delay 0.2 --start 0.0 driver_ui_test.bag
```

With the patch, after hitting control-C:

```
vagrant@local-driver-ui:~$ ps auwwwxf | grep pts/0
vagrant   9033  0.0  0.2 105644  4940 ?        S    17:42   0:00  |   \_ sshd: vagrant@pts/0
vagrant   9034  0.0  0.3  24504  7384 pts/0    Ss+  17:42   0:00  |       \_ -bash
vagrant   9473  0.0  0.1  11752  2244 pts/1    S+   17:50   0:00              \_ grep --color=auto pts/0
```

jspricke:
+1

dirk-thomas:
@jwm Can you please provide a short sequence of command to reproduce the problem before / the improved behavior after the patch.

jwm:
@dirk-thomas Sure, it's pretty straightforward.

Without the patch, run `rosbag play some-bag.bag` and hit `Ctrl-C`. The `rosbag` Python process exits, leaving the underlying `play` node running and emitting output, even though I'm back at a command prompt.

With the patch, `Ctrl-C` terminates both the Python process and the underlying node, leaving me at a command prompt with no further output from the `play` node.

dirk-thomas:
I don't see any difference in behavior when hitting `Ctrl-C`. Looking at the patch that is not surprising since it only handles `SIGTERM` but `Ctrl-C` emits a `SIGINT`.

jwm:
Ah, it's been a few weeks, I was misremebering.

We were running `rosbag play` from a `pytest` fixture that begins playback, yields to the test function, then cleans up by killing the child. This results in the node process remaining after the Python interpreter exits.

dirk-thomas:
Can you please share a reproducible example ([SSCCE](http://sscce.org/)).

jwm:
- Run `rosbag play some-bag.bag`
- Send the Python process a `SIGTERM`. See the Python interpreter process disappear from the process table.
- Watch the node process continue to play the bag, emitting output.

dirk-thomas:
I can reproduce that case and also that the patch fixes the problem.

But currently the patch overwrites the signal handler. Instead it should make sure to also call any previously registered signal handle from within the lambda in order to not ignore previously registered signal handler.

jwm:
Force-pushed an updated patch.

dirk-thomas:

jwm:
Thanks for merging, Dirk!

Issue #950, opened 2016-12-15
rostopic: pluggable message printers


mikepurvis:
Following on from #724 and related changes, I'd like to explore making rostopic able to scan the workspace and find printer modules able to handle specific message types. There could always be a `--raw` flag to suppress this behaviour, but in general there are a number of cases where the default message presentation is less than ideal:

- Messages like `Log` and `DiagnosticArray` could be presented much better than they presently are.
- Messages with bitfields/enums could show the correct constant name(s) rather than just the number.
- Covariance matrices could be shown in a properly aligned tabular format.
- In general, `rostopic` becomes smarter without needing a lot of built-in knowledge about specific message types.

In terms of an implementation, it would probably be something that would show up in the `package.xml`, eg:

```
<package>
  <name>my_fancy_package</name>
  ...
  <export>
    <rostopic>
      <echo_plugin type="my_fancy_package.RostopicFanciness" />
    </rostopic>
  </export>
</package>
```

So `rostopic` (when not in `--raw` mode) would check the workspace package XML files for such export tags and import the specified classes, passing messages through them for processing as required (probably subject to an optional blacklist).

Would like to get a temperature on whether this interests people. Part of the work would probably involve [refactoring this 2000+ line monster](https://github.com/ros/ros_comm/blob/kinetic-devel/tools/rostopic/src/rostopic/__init__.py) into some separate files, and adding some tests to cover core functionality.

gavanderhoorn:
I like the idea in general, but am worried about performance impact. `rostopic` should be fast, and especially on resource-restricted systems, crawling the file-system is something that can take ages. I'm not saying that should stop you from doing this (as it definitely seems like a good idea), but it's something to keep in mind when implementing this.

`catkin_tools` has something similar with the *verbs*, but afaik that uses `setup_tools` entry points, correct? `rospkg` would seem to cache things, so may be more suitable (but would perhaps need a `--force-discover` like `rqt`).

mikepurvis:
Yeah, definitely something to watch out for.

I like the catkin_tools approach, and have some exposure to it (I'm the author of a WIP [documentation plugin](https://github.com/mikepurvis/catkin_tools_document) that Clearpath is already using in production internally). It's possible that something similar would work for catkin-wrapped Python packages, but it will take experimenting to see.

[...]

PR #955, opened 2016-12-31
rostopic: Break up __init__.py

mikepurvis:
Laying the groundwork to begin addressing #950. I've done basic local smoke testing, consisting of manually running and validating the output of:

```
roscore &
rostopic pub /foo geometry_msgs/TwistStamped "{}" -r 20 &
rostopic pub /bar std_msgs/String "data: 'test'" -r 1 &

rostopic list
rostopic list --host
rostopic echo bar
rostopic echo bar/data
rostopic hz foo
rostopic delay foo
rostopic bw foo
rostopic info bar
rostopic type bar
rostopic find geometry_msgs/TwistStamped
```

There's a reasonable set of tests already present in the `rostopic` and `test_rostopic` packages which end to end validation of captured output, although I'm having trouble with some failures there which I don't understand and will need to dig into further.

The big issue will be with other packages which depend on being able to import random functions from the `rostopic` module. This is be no means an exhaustive list, but a workspace search of stuff in use at Clearpath quickly surfaces a number of examples:

- https://github.com/ros-visualization/rqt_robot_plugins/search?q=rostopic 
- https://github.com/ros-visualization/rqt_common_plugins/search?q=rostopic
- https://github.com/ros/ros_comm/search?q=rostopic.get_topic_class
- https://github.com/RobotWebTools/rosbridge_suite/search?q=from+rostopic+import

To not break these usages we'll need identify the names which are being used externally, and have a block of legacy imports at the bottom of the new `__init__.py` which pulls them into their old locations.

Finally, this PR is kind of unreviewable in its current state, and `git diff -M` isn't making it any better. Can someone recommend an external diff tool which would be better able to identify code that has moved and changes which occurred on top of the move? (eg renaming `rostopic._check_master` -> `rostopic.util.check_master`). I tried to minimize such changes, leaving any necessary refactoring for a later pass, but some was required and/or obvious.

At least for now, the primary purpose of this PR is mostly to stimulate discussion— in particular:

- Does the overall directory structure make sense? Should each verb have a folder rather than just a file, or does it matter? (intended next step is for each verb module to expose a description dict, similar to how [catkin_tools verbs work](https://github.com/catkin/catkin_tools/blob/master/catkin_tools/verbs/catkin_build/__init__.py#L21).)
- Does the composition of `util.py` make sense? Should there be a separate `msgeval.py` which has just the message post-processing stuff in it? Does the `sleep` function need to exist or can it be inlined where used?
- There was a TODO in the code about lazy-importing rospy. Is this still something anyone cares about? If so, are there ways I can organize the source to make that objective more likely to be achievable?

@wjwwood, @dirk-thomas, @gavanderhoorn, @NikolausDemmel


tools/rostopic/src/rostopic/__init__.py
```diff
@@ -30,2037 +29,39 @@
...
-        field_types = val._slot_types
-        for f, t in zip(val.__slots__, val._slot_types):
-            if echo_noarr and '[' in t:
-                continue
-            elif echo_nostr and 'string' in t:
-                continue
-            yield f
-    return field_filter
 
 def _optparse_topic_only(cmd, argv):
```
mikepurvis, line 53: This function doesn't appear to be referenced anywhere; any reason not to remove it?

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -2079,13 +80,25 @@ def _fullusage():
...
+
+from rostopic.verbs.echo import _rostopic_cmd_echo
+from rostopic.verbs.hz import _rostopic_cmd_hz
+from rostopic.verbs.type import _rostopic_cmd_type
+from rostopic.verbs.list import _rostopic_cmd_list
+from rostopic.verbs.info import _rostopic_cmd_info
+from rostopic.verbs.pub import _rostopic_cmd_pub
+from rostopic.verbs.bw import _rostopic_cmd_bw
+from rostopic.verbs.find import _rostopic_cmd_find
+from rostopic.verbs.delay import _rostopic_cmd_delay
```
mikepurvis, line 92: Obviously temporary, just maintaining functionality without making more changes than necessary upfront.

dirk-thomas:
Let me try to answer some of your questions and add some other random comments and thoughts:

> Does the overall directory structure make sense? Should each verb have a folder rather than just a file, or does it matter?

Luckily this is transparent for the user. For now while the amount of code for a single verb is not too much I think it is ok to keep it in files. If any of them ever grows too much in can be moved into a subfolder without affecting the usage (if the new `subfolder/__init__.py` ensures to expose the previous symbols). Important is that all verbs are in a subfolder `verbs` in order to not collide with modules on the higher level (like `util`). 

> Does the composition of `util.py` make sense? Should there be a separate `msgeval.py` which has just the message post-processing stuff in it?

I would leave this up to you. I have no preference for either.

> Does the `sleep` function need to exist or can it be inlined where used?

This boils down to backward compatibility. In order to ensure that existing code continues to work I think the function should still be there. In general the new `__init__.py` should ensure that all symbols (except "hidden" ones beginning with an underscore) are still available which existed in that file before (simply because we don't know who uses what). To make this more obvious I would recommend to move necessary imports for backward compatibility into a `legacy.py` file and import that in the `__init__.py` file with `from legacy import *`.

> There was a TODO in the code about lazy-importing rospy. Is this still something anyone cares about? If so, are there ways I can organize the source to make that objective more likely to be achievable?

I would assume the opposite here. Until someone can "prove" that we don't care about this anymore the previous behavior of importing `rospy` lazily should be preserved.

One suggestion for making this PR more reviewable. When moving or refactoring larger pieces of code I suggest doing the following steps in separate commits for easier diffs:

* Copy the entire content to the new location(s) with no modification. That will allow git to identify (based on the same hash) that nothing has changed.
* Then remove obsolete parts from the original file and all new locations.
* Then address changed includes / imports to make the refactored code work again.
* Then start with actual modification.

dirk-thomas:

PR #956, opened 2017-01-05
Add '#Include <vector>' to fix building on GCC-6 (cherry-pick #911)

v4hn:
Don't rely on transitive header inclusion to declare std::vector as building with GCC-6 fails due to no '#Include <vector>' statement.

### DIFF
--- a/tools/rosconsole/include/ros/console.h
+++ b/tools/rosconsole/include/ros/console.h
@@ -40,6 +40,7 @@
 #include <cstdarg>
 #include <ros/macros.h>
 #include <map>
+#include <vector>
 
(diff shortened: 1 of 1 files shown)


v4hn:
This is a harmless cleanup and there's no reason not to apply it to the indigo branch too.

This allows to compile rosconsole-indigo with gcc 6.

dirk-thomas:
Duplicate of #930 and #939.

v4hn:
Sorry for the dup.

When will be the next release of ros_comm into indigo then?
Clearly there are a number of people hitting this bug, so it would be great to see it in there!

dirk-thomas:
I understand it is an urgent request. I can't promise a specific time frame at the moment though. Sorry.

PR #957, opened 2017-01-09
fixed UDP block number when EAGAIN or EWOULDBLOCK

minsuu:
I found a bug in `TransportUDP::write`. When this branch is taken, `num_bytes` is initialized to retry the write operation. However, `this_block` stays increased. This causes block number in the datagram head to increase to an arbitrary amount, and the subscriber cannot receive the expected block. This commit just decreases `this_block` and retry the write properly.


jspricke:
+1, but could you remove the unrelated whitespace changes from the patch?

jspricke:
Thx!

PR #959, opened 2017-01-12
rosbag play: added option wait-for-subscriber

cwecht:
Added an option `--wait-for-subscriber`. If this option is chosen, `rosbag play` will wait before publishing until every topic has at least one subscriber. This is pretty useful in tests using `rostest` if one wants to make sure that no message is dropped because of missing subscribers.

### DIFF
--- a/tools/rosbag/src/play.cpp
+++ b/tools/rosbag/src/play.cpp
@@ -62,7 +62,7 @@ rosbag::PlayerOptions parseOptions(int argc, char** argv) {
       ("topics", po::value< std::vector<std::string> >()->multitoken(), "topics to play back")
       ("pause-topics", po::value< std::vector<std::string> >()->multitoken(), "topics to pause playback on")
       ("bags", po::value< std::vector<std::string> >(), "bag files to play back from")
-      ("wait-for-subscribers", "wait for subscribers before publishing");
+      ("wait-for-subscribers", "wait for at least one subscriber on each topic before publishing");
     
     po::positional_options_description p;
     p.add("bags", -1);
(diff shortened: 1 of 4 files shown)


tools/rosbag/src/play.cpp
```diff
@@ -61,7 +61,8 @@ rosbag::PlayerOptions parseOptions(int argc, char** argv) {
       ("try-future-version", "still try to open a bag file, even if the version is not known to the player")
       ("topics", po::value< std::vector<std::string> >()->multitoken(), "topics to play back")
       ("pause-topics", po::value< std::vector<std::string> >()->multitoken(), "topics to pause playback on")
-      ("bags", po::value< std::vector<std::string> >(), "bag files to play back from");
+      ("bags", po::value< std::vector<std::string> >(), "bag files to play back from")
+      ("wait-for-subscribers", "wait for subscribers before publishing");
```
dirk-thomas: Maybe update the text to `wait for at least one subscriber on each topic before publishing` to make the condition clearer to the user.

Same for the Python help text.

tools/rosbag/src/player.cpp
```diff
@@ -278,6 +284,20 @@ void Player::printTime()
...
+{
+    bool all_topics_subscribed = false;
+    std::cout << "Waiting for subscribers." << std::endl;
+    while (!all_topics_subscribed) {
+        all_topics_subscribed = true;
+        typedef std::pair<std::string, ros::Publisher> value_type;
+        foreach(const value_type& pub, publishers_) {
+            all_topics_subscribed &= pub.second.getNumSubscribers() > 0;
+        }
+    }
```
dirk-thomas, line 297: Is this loop busy waiting? If yes, something less resource intensive might be good.

tools/rosbag/src/player.cpp
```diff
@@ -278,6 +284,20 @@ void Player::printTime()
     }
 }
 
+void Player::waitForSubscribers() const
+{
+    bool all_topics_subscribed = false;
+    std::cout << "Waiting for subscribers." << std::endl;
+    while (!all_topics_subscribed) {
+        all_topics_subscribed = true;
+        typedef std::pair<std::string, ros::Publisher> value_type;
```
mikepurvis: Might be cleaner to `typedef std::map<std::string, ros::Publisher> PublisherMap` in the header, and then here you can use `PublisherMap::value_type` directly in the for loop without needing this duplicated definition.

dirk-thomas:
Thank you for the contribution!

PR #961, opened 2017-01-18
Add default ROS_MASTER_URI

jspricke:


tools/rosgraph/src/rosgraph/rosenv.py
```diff
@@ -69,7 +69,10 @@ def get_master_uri(env=None, argv=None):
                 if not val:
                     raise ValueError("__master remapping argument '%s' improperly specified"%arg)
                 return val
-        return env[ROS_MASTER_URI]
+        if ROS_MASTER_URI in env:
```
mgrrx: You could use a one-liner instead:
`env.get(ROS_MASTER_URI, 'http://localhost:11311')`

mikepurvis:
I have no objection to this change, but I'm curious the motivation— are you trying to eliminate the need for any env setup at all in some baseline installed-into-`/usr` scenario?

jspricke:
Exactly, it's required by the Debian policy (and implemented in the packages in Debian already).

mikepurvis:
IMO, the only risk here is possible confusion for a user who tries to directly run a roscpp binary _without_ sourcing the environment in a situation where they should have, and get a partially functional system rather than this error message which explicitly tells them what is wrong and how to fix it.

LGTM +1

tools/rosgraph/src/rosgraph/rosenv.py
```diff
@@ -69,7 +69,7 @@ def get_master_uri(env=None, argv=None):
                 if not val:
                     raise ValueError("__master remapping argument '%s' improperly specified"%arg)
                 return val
-        return env[ROS_MASTER_URI]
+        env.get(ROS_MASTER_URI, 'http://localhost:11311')
```
evenator: Isn't this missing a return statement? I think it should be:
```python
        return env.get(ROS_MASTER_URI, 'http://localhost:11311')
```

jspricke:
Oups, thx @evenator!

mikepurvis:
Since we're eliminating a potentially helpful user prompt here, I'd appreciate commentary from @tfoote or @dirk-thomas. Then again, my experience is that it's typically much more important to have explicitly set ROS_IP/ROS_HOSTNAME than ROS_MASTER_URI, so maybe this prompt is not really that important.

A middle ground might be to leave a non-fatal warning? Not sure if that's really complying with the spirit of the policy. Particularly since the message will be emitted once per node starting up, so it will get worse and worse the more of them are started.

dirk-thomas:
In order to consider using a default value for `ROS_MASTER_URI` if it is not set I think it would be crucial to define a constant in Python as well as C++ which every code can use. Otherwise the actual default value will be spread throughout the code base.

@ros/ros_team Please comment on the idea of having a default if `ROS_MASTER_URI` is not set.

wjwwood:
I'm fine with a default, as there is a reasonable default for `ROS_MASTER_URI`.

As @dirk-thomas said, this default should be replicated in as few places as possible. Also the documentation needs to be updated, for example: http://wiki.ros.org/ROS/EnvironmentVariables#ROS_MASTER_URI

It would be great to preserve the behavior for people installing to `/opt` or to a local folder since running without sourcing the setup file is almost always not what you meant to do, but I guess it's more trouble than it's worth.

dirk-thomas:
Sounds good. @jspricke Can you please update the PR accordingly. And check for other locations which use the environment variable and need to be updated to use the new default.

jspricke:
Great idea, I will proposes a patch on the weekend (sorry for being
silent, but we have integration week this week).

jspricke:
I did a grep through all my ROS packages, and this is all I came up with. As we have an API to get the URI, I think it doesn't make sense to put it somewhere else.

dirk-thomas:
The API (e.g. the `init()` function) already wrap some logic around the retrieval of the environment variable (e.g. remapping). A global constant would therefore be good to have that any other code can use it as is (e.g. to show the default value for a command line interface in the help string). Does that make sense?

jspricke:
superseeded by #1666.

PR #964, opened 2017-01-23
Add note for dict usage in rospy set_param

AlexReimann:
I think it is kind of important to have that information that an set_param with a dictionary kicks out other parameters.

The text is basically a copy of [the documentation in the master_api](https://github.com/ros/ros_comm/blob/a8c5f9e95e6823417b67ee0da2fa0a91acd97ef5/tools/rosmaster/src/rosmaster/master_api.py#L346)


dirk-thomas:

PR #965, opened 2017-01-24
Added the "export_all_args" attribute to the include tag.

bponsler:
This enables a parent launch file to automatically export all args defined in an included launch file. This means that any arg defined in the included launch file will be automatically passed to the include tag, and also added as args to the parent launch file. Additionally, any documentation strings set for the arg will be exported as well.

This eliminates the need for args to be added to each file in a chain of launch files in order to pass an argument down the chain.

For example,

**file1.launch**:
```
<launch>
    <include file="file2.launch" export_all_args="true"/>
</launch>
```

**file2.launch**:
```
<launch>
    <arg name="this_is_my_arg" default="one"/>
</launch>
```

The use of export_all_args="true" makes file1.launch equivalent to:

```
<launch>
    <arg name="this_is_my_arg" default="one"/>
    <include file="file2.launch">
        <arg name="this_is_my_arg" value="$(arg this_is_my_arg)"/>
    </include>
</launch>
```

### DIFF
--- a/tools/roslaunch/src/roslaunch/arg_dump.py
+++ b/tools/roslaunch/src/roslaunch/arg_dump.py
@@ -45,7 +45,7 @@ from roslaunch.core import RLException
 from roslaunch.config import load_config_default
 
 def get_args(roslaunch_files):
-    loader = roslaunch.xmlloader.XmlLoader(resolve_anon=False)
+    loader = roslaunch.xmlloader.XmlLoader(resolve_anon=False, dumping_args=True)
     config = load_config_default(roslaunch_files, None, loader=loader, verbose=False, assign_machines=False)
     return loader.root_context.resolve_dict.get('arg_doc', {})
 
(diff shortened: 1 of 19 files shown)


dirk-thomas:
Can you please describe how this is different from the existing attribute `pass_all_args` (see [roslaunch wiki page](http://wiki.ros.org/roslaunch/XML/include)).

bponsler:
pass_all_args automatically passes all args from the parent to the child include's scope. But, the parent has to have all of the args it wishes to pass down already defined.

export_all_args re-defines all of the args defined in the child include inside the parent launch file's scope, and then automatically passes only those arguments to the child (which is similar to the pass_all_args functionality).

In the example I provided before, the "this_is_my_arg" is defined in the child launch file (file2.launch). If you wanted to use pass_all_args you would need to define an arg in file1.launch called this_is_my_arg. 

The use of export_all_args makes it so you do not have to duplicate the definition of the args in the parent's launch file. This allows the parent to automatically inherit the default value of the child's arguments, and allows the parent to pass only the supported args to the child (pass_all_args could pass arguments that the child does not support).

dirk-thomas:
Thank you for clarifying and for the contribution. It might take a bit until someone will have the time to review the contribution. `roslaunch` is sadly not the easiest code to modify / review.

dirk-thomas:
With Indigo being EOL please retarget the latest default branch (currently `melodic-devel`) to be considered for review and merge.

dirk-thomas:

PR #966, opened 2017-01-26
Issue947

dirk-thomas:
Replaces #931.

Fixing the original implementing of the caching: #682

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -144,27 +144,24 @@ def _find(resolved, a, args, context):
     rp = _get_rospack()
     if path:
         source_path_to_packages = rp.get_custom_cache('source_path_to_packages', {})
+        res = None
         try:
             res = _find_executable(
                 resolve_without_path, a, [args[0], path], context,
                 source_path_to_packages=source_path_to_packages)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@ros/ros_team Please review.

tfoote:
Should the update be triggered on a miss too? Previously it only updated on a miss, now it only updates on a success. Otherwise lgtm.

bponsler:
Thanks -- I just confirmed these changes work for me. Reduced XmlLoader.load time from ~12 seconds to ~0.5 seconds (for my specific launch tree).

Do you have any plans to make these same changes in the _indigo-devel_ branch?

dirk-thomas:
@tfoote Good point. While that case might be rare there is no reason to not persist the data in the case of a miss. I modified the code to persist the data in both cases: 1f09fd15ad250f546548751cf84ee7616db92bf2

@bponsler Thank you for checking. Could I maybe ask you to check again (with the additional modification) just to be sure :wink:) Once the changes are merged and have been released into Kinetic (without regressing) they will be considered for backporting into Jade and Indigo.

bponsler:
@dirk-thomas -- no problem. Just checked your latest changes. Works for me. Thanks!

dirk-thomas:
@bponsler Thanks for making sure!

PR #971, opened 2017-02-01
throw exception instead of accessing invalid memory

### LINKED ISSUE #969: rosbag: Recorder throws a segmentation fault if options.prefix is not set
If I construct a ``rosbag::Recorder`` object with a ``rosbag::RecorderOptions`` where the member 'prefix' is not set, the ``recorder::run()`` method will segfault.

For instance, this code will segfault if the line ```options.prefix = "test";``` is commented out:

```cpp
#include <ros/ros.h>
#include <rosbag/recorder.h>

{
  ros::init(argc, argv, "data_logger_test_node");
  ros::NodeHandle n;

  rosbag::RecorderOptions options;
  options.append_date = false;
  options.trigger = false;
  options.min_space = 0;
  options.verbose = true;
  options.prefix = "test";
  options.split = false;
  options.regex = false;
  options.topics.push_back("/rosout");
  options.max_duration = ros::Duration(1.0);

  rosbag::Recorder recorder(options);
  recorder.run();
}
```

dirk-thomas:
Fixes #969.

@ros/ros_team Please review.

### DIFF
--- a/tools/rosbag/src/recorder.cpp
+++ b/tools/rosbag/src/recorder.cpp
@@ -348,6 +348,11 @@ void Recorder::updateFilenames() {
     if (options_.split)
         parts.push_back(boost::lexical_cast<string>(split_count_));
 
+    if (parts.size() == 0)
+    {
+      throw BagException("Bag filename is empty (neither of these was specified: prefix, append_date, split)");
+    }
+
     target_filename_ = parts[0];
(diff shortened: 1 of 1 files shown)


jspricke:
+1

mikaelarguedas:
looks good to me. typo on `split` in the error message though

dirk-thomas:
Doh, fixed the wrong spelling. Thank you both for the review.

PR #973, opened 2017-02-06
[rostest] Check /clock publication neatly in publishtest

wkentaro:
- Use time.sleep because rospy.sleep(0.1) hangs if /clock is not published
- Add timeout for clock publication

### DIFF
--- a/tools/rostest/nodes/publishtest
+++ b/tools/rostest/nodes/publishtest
@@ -111,12 +111,12 @@ class PublishTest(unittest.TestCase):
         t_start = time.time()
         while not rospy.is_shutdown() and \
                 use_sim_time and (rospy.Time.now() == rospy.Time(0)):
-            rospy.logwarn('/use_sim_time is specified and rostime is 0, '
-                          '/clock is published?')
+            rospy.logwarn_throttle(
+                1, '/use_sim_time is specified and rostime is 0, /clock is published?')
             if time.time() - t_start > 10:
                 self.fail('Timed out (10s) of /clock publication.')
             # must use time.sleep because /clock isn't yet published, so rospy.sleep hangs.
-            time.sleep(1)
+            time.sleep(0.1)
         # subscribe topics
         checkers = []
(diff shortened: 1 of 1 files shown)


wkentaro:
Could you please review this, maintainers?

mikepurvis:
Added your comment about `time.sleep` to the source. Otherwise LGTM!

tools/rostest/nodes/publishtest
```diff
@@ -107,10 +108,15 @@ class PublishTest(unittest.TestCase):
...
+        t_start = time.time()
+        while not rospy.is_shutdown() and \
+                use_sim_time and (rospy.Time.now() == rospy.Time(0)):
             rospy.logwarn('/use_sim_time is specified and rostime is 0, '
                           '/clock is published?')
-            rospy.sleep(0.1)
+            if time.time() - t_start > 10:
+                self.fail('Timed out (10s) of /clock publication.')
+            # must use time.sleep because /clock isn't yet published, so rospy.sleep hangs.
+            time.sleep(1)
```
dirk-thomas: Why does it sleep for 1s now and not for 0.1s as before? I think it is valuable to continue quickly once the clock is available.

mikepurvis: In that case, the warning should only be printed once to avoid flooding the console unnecessarily.

dirk-thomas: Either once or using the throttled logging functions to print it only every N seconds.

wkentaro:
I fixed it. Could u please review the change?

wkentaro:
Should I change the base branch from kinetic-devel to lunar-devel?

dirk-thomas:
Sorry for the late response. The patch looks good as is. Yes, please retarget *lunar-devel* and it can be merged.

dirk-thomas:
Changing the target branch was already enough (no need to rebase). Once CI passed I will go ahead and merge this... Thanks.

PR #976, opened 2017-02-07
Revert changes from #960.

mikepurvis:
As discussed in https://github.com/ros/ros_comm/pull/960#issuecomment-278045035.


dirk-thomas:
With this being merged can you please create a new PR with a comment that the patch needs to be updated to address the problems you notices.

mikepurvis:
See #977.

PR #565, opened 2015-02-09
Small fixes for Python 3 compatibility.

aballier:
Here is a set of small independent patches to improve python 3 compatibility. These can be squashed if desired.

I have a few more pending but their correctness is more debatable.

PR #978, opened 2017-02-08
Fixes for Python 3 compatibility.

mikepurvis:
tcpros_service.py: encode error messages as bytes in python3 before packing them.
test_rospy_tcpros_base.py: check against bytes equality for python3 compatibility.
test_rospy_tcpros_pubsub.py: use bytes data instead of str for python3 compatibility.
test_rospy_topics.py: check for bytes equality for python3 compatibility
test_rosservice_command_line_online.py: pipes are bytes with python3, compare or decode accordingly. Fixes this test with python3.
test_bag.py: use range() instead of xrange() so that it works with python 2 & 3. Since it is a test, performance does not matter much.

Replaces #565.

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_service.py
+++ b/clients/rospy/src/rospy/impl/tcpros_service.py
@@ -608,6 +608,8 @@ class ServiceImpl(_Service):
         @param err_msg: error message to send to client
         @type  err_msg: str
         """
+        if sys.hexversion > 0x03000000: #Python3
+            err_msg = bytes(err_msg, 'utf-8')
         transport.write_data(struct.pack('<BI%ss'%len(err_msg), 0, len(err_msg), err_msg))
 
     def _handle_request(self, transport, request):
--- a/test/test_rosbag/test/test_bag.py
+++ b/test/test_rosbag/test/test_bag.py
@@ -287,7 +287,7 @@ class TestRosbag(unittest.TestCase):
     def test_get_message_count(self):
         fn = '/tmp/test_get_message_count.bag'
         with rosbag.Bag(fn, mode='w') as bag:
-            for i in xrange(100):
+            for i in range(100):
                 bag.write("/test_bag", Int32(data=i))
                 bag.write("/test_bag", String(data='also'))
                 bag.write("/test_bag/more", String(data='alone'))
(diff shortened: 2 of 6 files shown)


mikepurvis:
@dirk-thomas Does the Kinetic build somehow test Python 2 and 3, or just one of them?

mikepurvis:
CI failures appear to be all related to roscpp parameter logic, which has nothing to do with this change.

@ros-pull-request-builder retest this please.

dirk-thomas:
It appears similar tests failed on the dev job already: http://build.ros.org/view/Kdev/job/Kdev__ros_comm__ubuntu_xenial_amd64/96/ Likely related to the recent merges...

mikepurvis:
Seems hard to believe— only three changes went in this morning, and none had anything to do with parameters, unless the select/poll changes are exposing races. But they all passed pre-merge, so that seems weird too.

dirk-thomas:
I am running a few more dev jobs for different commit hashes to figure out where the problem is. Locally I can't reproduce the failing tests :worried:

dirk-thomas:
See #981 for the result of the search.

dirk-thomas:

dirk-thomas:
Sorry, I meant #981.

dirk-thomas:

mikepurvis:
We have a checkmark here now, but I'm still hazy on what exactly the buildfarm is able to test. AFAICT, even in Ubuntu Zesty, `python` is still 2.7.11. So is there any actual way to verify that this is correct short of building it from source and running the tests on a platform where `python` is 3.5+?

Wondering if Python 3 compatibility might be an argument for setting up a Travis job on this repo?

dirk-thomas:
We currently don't run CI with Python 3 and the buildfarm doesn't even have an option for it. It would be great to run this locally with Python 3. That being said reading the diff it looks good. Both kind of changes (range without X and popen returning bytes instead of strings) are common updates necessary. So I am fine merging this as is.

mikepurvis:
Sounds good. Thanks for this contribution, @aballier!

mikepurvis:
I repushed the squashed commit to retain original authorship, since github stripped it out when executing the merge (since I was the creator of the PR).

PR #979, opened 2017-02-08
Add more logging to publisher update calls.

deng02:
Because the publisher update calls are critical for setting
up topic connections between subscribers and publishers,
have added more detailed logging messages.  The publishers
are included so you know what the subscriber was told, as
well as when the call succeeded or failed and if so why.
The rosout topic is filtered out to reduce noise.

### DIFF
--- a/tools/rosmaster/src/rosmaster/master_api.py
+++ b/tools/rosmaster/src/rosmaster/master_api.py
@@ -202,22 +202,18 @@ def publisher_update_task(api, topic, pub_uris):
     @param pub_uris: list of publisher APIs to send to node
     @type  pub_uris: [str]
     """
-    is_rosout_topic = 'rosout' in topic
     start_sec = time.time()
     msg = "publisherUpdate[%s] -> %s %s" % (topic, api, str(pub_uris))
-    if not is_rosout_topic:
-        mloginfo(msg)
+    mloginfo(msg)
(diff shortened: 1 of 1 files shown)


mikepurvis:
Filtering events for the rosout topics seems a bit arbitrary. I can appreciate that this could be noisy with every single node bringing up a rosout publisher, but I'm not sure it's worth it.

mikepurvis:

deng02:
The rosout topic is "special" because every node automatically publishes to it and it's brought up automatically by the system, hence the special treatment. So I'd argue less arbitrary and more intentional.  However I can agree that special treatment in code is not always desirable and after testing, I don't feel this buys us much in terms of data volume savings.  I will remove it.

dirk-thomas:
I added the commit 34c57ac6ce874db4961747c67eb0409b45800348 which moves some code to keep variables and comments closer to where they are used as well as reduce duplication. Since I only wrote the patch in the GitHub web interface it would be great if you could confirm that it still achieves your goal.

deng02:
@dirk-thomas Confirmed: output with commit https://github.com/ros/ros_comm/pull/979/commits/34c57ac6ce874db4961747c67eb0409b45800348 achieves my goal.

mikepurvis:
LGTM 👍

dirk-thomas:

PR #980, opened 2017-02-08
Add logger roscpp_internal.connections.

deng02:
We need more connection leve diagnostics in our logs but turning
on DEBUG for roscpp_internal generates a flood of information. We
only need visibility into when tcpros connections are created/dropped
for publishers and subscribers. The rospy nodes currently do this
but the roscpp ones do not.

A DEBUG macro for logging messages to the named logger
roscpp_internal.connections has been added.  This will
allow it to inherit the roscpp_internal log level or be
turned on separately to just get connection-related messages.


mikepurvis:
+1

PR #981, opened 2017-02-08
Revert "Using poll() in favor of select() in the XmlRPCDispatcher"

dirk-thomas:
Currently the head of kinetic-devel fails some tests. I ran several dev jobs (http://build.ros.org/view/Kdev/job/Kdev__ros_comm__ubuntu_xenial_amd64/ builds 102 - 105) for specific hashes to bisect when the problem was introduced and it looks like that PR #833 is the culprit. Therefore this PR revert that change.

We will see if CI will agree and pass again. Then I would propose to merge this in order to test and merge other pending PRs. The change needs to then be applied after it passes CI again.


mikepurvis:
LGTM

PR #990, opened 2017-02-16
[rosout] Moved code from init to initializer

mgrrx:
Splitted #855 into two parts. This one cleans up the init method by using the initializer of the class.


mikepurvis:
Since `init` was a public method, I could see this being used to reset an object to its original state. But of course, there's no header for any of this, so presumably nothing is linking to or using this object outside of the immediate file.

+0 from me.

dirk-thomas:
Thanks!

PR #991, opened 2017-02-16
[rosout] Proper logrotation for rosout

### LINKED ISSUE #854: rosout log rotation is odd

mgrrx:
This PR fixes #854 by rotating the log files as usual in Linux to guarantee the ordering of log files (https://en.wikipedia.org/wiki/Log_rotation).
As discussed previously, this PR does not touch the initializers of the class and targets the Luna release.

### DIFF
--- a/tools/rosout/rosout.cpp
+++ b/tools/rosout/rosout.cpp
@@ -91,7 +91,7 @@ public:
       std::cout << "logging to " << log_file_name_.c_str() << std::endl;
 
       std::stringstream ss;
-      ss <<  "\n\n" << ros::Time::now() << "  Node Startup\n";
+      ss << "\n\n" << ros::Time::now() << "  Node Startup\n";
       int written = fprintf(handle_, "%s", ss.str().c_str());
       if (written < 0)
       {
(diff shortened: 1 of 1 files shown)


mikepurvis:
Summoning @gregwjacobs who is likely to have an opinion on this.

It'd be great to have some tests covering this behaviour but otherwise LGTM.

mgrrx:
I also thought about writing a test for this functionality but with the current implementation we always need to write at least 100MB to the disk to see that the log rotates once - which I think is too much for a test. I'd therefor expose the max file size parameter as a ros parameter or command line argument and write the test with a reduced file size.
Do you prefer to have the test together with this PR or should I create a new ticket and PR the test?

dirk-thomas:
I have cherry-picked your patch to the newly created lunar-devel branch: ce4d7263138719590ca826fce02cae9f5bc3904a

In that process I have removed the unrelated whitespace change (https://github.com/ros/ros_comm/pull/991/files#diff-c21b75d2d000be7cceac00629a0a9e23R94) as well as readded the output to `cout` informing the user that the log is being rotated (https://github.com/ros/ros_comm/commit/ce4d7263138719590ca826fce02cae9f5bc3904a#diff-c21b75d2d000be7cceac00629a0a9e23R181).

Thank you!

dirk-thomas:
It would be great if you could create a separate PR for making the limit configurable as well as cover the feature with tests.

PR #992, opened 2017-02-17
[rostest] trivial typo.

130s:
>     self.assert_(len(topic), "bad parameter (topic")

During hours of debugging, even a trivial typo like this could fluctuate me between finally-found-a-bug joy and oh-nevermind sorrow.


mikepurvis:
LGTM.

PR #996, opened 2017-02-23
Fix ROS message migration rule tool

efernandez:
# Description of the buggy corner case not handled on the rosbag migration tool

Consider the following scenario for a message G with the following versions:

> G0 -- rule.bmr --> G1 -- auto-generated rule --> G3

where `G3` is the **current** msg, and `G2` is the **bag** message that needs migration.

We need to check for a migration rule like this first:

> G2 --> G3

instead of the backwards ones that were generated before this bug fix:

> G2 --> G1

Note that the **lazy** check to mark the rules as valid in the `make_update_rule` (see https://github.com/clearpathrobotics/ros_comm/blob/kinetic-devel-cpr/tools/rosbag/src/rosbag/migration.py#L1123) method makes that rule pass initially, but later it fails because there's no rule for the sub fields.

This has been adapted from the commit https://github.com/clearpathrobotics/ros_comm/commit/ec8ec371fb06ea92b0a92f6fdd56d40940aab5d4 description.

# Improvement to the validation check for the rules

The other commit https://github.com/clearpathrobotics/ros_comm/commit/97fdf48274788a0a1265d515b8cc34d55014dd64 simply allows to improve the **lazy** check in `make_update_rule` and therefore allows to find a better placement/position for the message to be migrated inside of the chain rules. This helps to avoid marking rules as valid, that later would fail because there's no rule for the sub fields. The more precise check that fails later happens at the sub fields level, and there's already a comment on the code that suggest that the way this is handled should be improved (see https://github.com/clearpathrobotics/ros_comm/blob/kinetic-devel-cpr/tools/rosbag/src/rosbag/migration.py#L307).

This replaces https://github.com/clearpathrobotics/ros_comm/pull/10

I recommend installing this chrome add-on to have a better diff: https://chrome.google.com/webstore/detail/github-diff-whitespace/lhbcdehjihmbiafeodkfnbndleijnnhp

FYI @mikepurvis @dirk-thomas @afakihcpr

### DIFF
--- a/tools/rosbag/src/rosbag/migration.py
+++ b/tools/rosbag/src/rosbag/migration.py
@@ -968,7 +968,7 @@ class MessageMigrator(object):
                 found_stop = True
                 break
 
-        # Next see if we can create a valid rule
+        # Next see if we can create a valid rule, including the sub rules
         if not found_stop:
             for (ind, tmp_sn) in reversed(list(zip(range(len(sn_range)), sn_range))):
                 if (tmp_sn.new_class._type != new_class._type):
@@ -977,12 +977,13 @@ class MessageMigrator(object):
                 R = new_rule(self, 'GENERATED.' + new_rule.__name__)
                 if R.valid:
                     R.find_sub_paths()
-                    sn = ScaffoldNode(tmp_sn.new_class, new_class, R)
-                    self.extra_nodes.append(sn)
-                    sn_range = sn_range[:ind+1]
-                    sn_range.append(sn)
-                    found_stop = True
-                    break
+                    if R.sub_rules_valid:
+                        sn = ScaffoldNode(tmp_sn.new_class, new_class, R)
+                        self.extra_nodes.append(sn)
+                        sn_range = sn_range[:ind+1]
+                        sn_range.append(sn)
+                        found_stop = True
+                        break
 
         # If there were no valid implicit rules, we suggest a new one from to the end
         if not found_stop:
@@ -1018,7 +1019,7 @@ class MessageMigrator(object):
                     self.found_paths[key] = [sn]
                     return [sn]
 
-        # Next see if we can create a valid rule
+        # Next see if we can create a valid rule, including the sub rules
         if not found_start:
             for (ind, tmp_sn) in reversed(list(zip(range(len(sn_range)), sn_range))):
                 if (tmp_sn.old_class._type != old_class._type):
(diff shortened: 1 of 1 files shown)


mikepurvis:
I've never created or even used bag migration rules, so I haven't a clue what's going on here, despite the effort at explaining the problem and this solution.

Is there some way that the [bag migration tests](https://github.com/ros/ros_comm/tree/lunar-devel/test/test_rosbag/bag_migration_tests) could be modified to illustrate the problem, so that we have a failing test corrected by this patch?

efernandez:
@mikepurvis I didn't know about those test, although I'm happy to see that something is exercising this code. Upstream (publicly) only two package use rules, and they're only for one generation/change.

I need to look into those tests and add one for the use case they didn't covered. It'd take me some time to get it done.
The good thing is that my change didn't break any of those tests :smiley:

efernandez:
@dirk-thomas it seems you wrote most of those tests, could you please orient me on writing a new one that exercise the use case fixed on this PR?

efernandez:
I've figured out myself what's needed to test this corner case.

Essentially, the `MigratedMixed.msg` almost covers this use case, except for a subtle thing: it's "version" on the `gen3` already appears as the new message on the rule from `gen2` to `gen3`, so the proper rule is created.

Therefore, I need to create a new generation `gen4` with no explicit changes on `MigratedMixed.msg`. That should expose the bug and show how this PR fixes it.

However, I'd need access to upload new .bag files for the new `gen4` messages into http://download.ros.org/data/test_rosbag/
@dirk-thomas @mikepurvis Does any of you have access there, or give me access?

For now I'd put the new .bag files in the repo, inside the `test` folder.

dirk-thomas:
@efernandez Atm we don't have a way to provide access to other people. I can upload additional files to download.ros.org. Please either send me the bag files by email or post some links where I can download them.

efernandez:
@dirk-thomas OK, thanks. I'd upload the .bag files somewhere and paste the links here. Once you upload the .bag files I'd update the links on the `CMakeLists.txt`.

I'm still working on creating the .bag files and setting up the new tests. I hope to finish either EOD today or tomorrow.

efernandez:
Closing in favour of https://github.com/ros/ros_comm/pull/1011 (which also depends on https://github.com/ros/ros_comm/pull/1009 and https://github.com/ros/ros_comm/pull/1010).

PR #997, opened 2017-02-23
Copy headers into devel

paulbovbel:
Fixes including xmlrpccpp headers from another package in develspace.

FYI @mmunawar01 @mikepurvis


paulbovbel:
Silly mistake, disregard

Issue #993, opened 2017-02-18
[roslaunch] roslaunch-check doesn't respect if/unless


mikepurvis:
We're having a case where we have some launch files which optionally include other launch files based on an environment variable being set. These are failing the test created by `roslaunch_add_file_check`, since it doesn't respect the conditional and attempts to resolve all include elements without regard for context.

This can be reproduced trivially by a launch file such as this, which launches just fine, but will (wrongly, in my opinion) fail the roslaunch check:

```
<launch>
  <include if="0" file="does/not/exist.launch" />
</launch>
```

----

Interestingly, `roslaunch_add_file_check` totally does resolve `$(arg)`, `$(env)`, `$(optenv)`, etc, so this _really_ does seem to be an issue with just the conditionals.

mikepurvis:
I believe the issue is here:

https://github.com/ros/ros_comm/blob/674263142a0be1124b273bdb1e0a79c3c956e259/tools/roslaunch/src/roslaunch/depends.py#L144-L148

It needs logic to explicitly check for if/unless, similar to how the args work:

https://github.com/ros/ros_comm/blob/674263142a0be1124b273bdb1e0a79c3c956e259/tools/roslaunch/src/roslaunch/depends.py#L99-L106

What's mysterious to me is that this works correctly for a normal launch but not for `roslaunch-check`. Is there some preprocessing going on that strips out tags based on if/unless attributes?

----

Looks like for a normal launch, this is handled by `xmlloader.py`, and the `@ifunless` decorator:

https://github.com/ros/ros_comm/blob/caab9fddc66dfa076dae6ac204c013fc49e2e67d/tools/roslaunch/src/roslaunch/xmlloader.py#L69-L96

@dirk-thomas It seems like there are some parallel implementation issues going on here, but I don't have the desire or energy to do a wholesale fixup of it. I'd prefer to fix the immediate bug as best I can and at least add a unit test which would help support a future refactoring. Unless you have some particular insight into what the relationship is between xmlloader and these other pieces?

dirk-thomas:
Addressed by #998. Thank you!

PR #998, opened 2017-02-24
Make roslaunch-check respect if/unless attribute on <include>

mikepurvis:
As discussed in #993. With the fix removed, the updated test XML fails with:

```
ERROR: unit.test_roslaunch_depends.test_roslaunch_deps
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/administrator/rocksteady_robot_ws/src/roslaunch/test/unit/test_roslaunch_depends.py", line 157, in test_roslaunch_deps
    base_pkg, file_deps, missing = roslaunch_deps(files, verbose=v)
  File "/home/administrator/rocksteady_robot_ws/src/roslaunch/src/roslaunch/depends.py", line 331, in roslaunch_deps
    rl_file_deps(file_deps, launch_file, verbose)
  File "/home/administrator/rocksteady_robot_ws/src/roslaunch/src/roslaunch/depends.py", line 225, in rl_file_deps
    parse_launch(launch_file, file_deps, verbose)
  File "/home/administrator/rocksteady_robot_ws/src/roslaunch/src/roslaunch/depends.py", line 210, in parse_launch
    _parse_launch(launch_tag.childNodes, launch_file, file_deps, verbose, context)
  File "/home/administrator/rocksteady_robot_ws/src/roslaunch/src/roslaunch/depends.py", line 182, in _parse_launch
    raise RoslaunchDepsException("Cannot load roslaunch include '%s' in '%s'"%(sub_launch_file, launch_file))
RoslaunchDepsException: Cannot load roslaunch include 'does/not/exist.launch' in '/home/administrator/rocksteady_robot_ws/src/roslaunch/resources/example.launch
```

### DIFF
--- a/tools/roslaunch/src/roslaunch/depends.py
+++ b/tools/roslaunch/src/roslaunch/depends.py
@@ -45,6 +45,7 @@ from xml.dom import Node as DomNode
 
 import rospkg
 
+from .loader import convert_value
 from .substitution_args import resolve_args
 
 NAME="roslaunch-deps"
@@ -94,20 +95,16 @@ def _get_arg_value(tag, context):
     else:
         raise RoslaunchDepsException("No value for arg [%s]"%(name))
 
-def _parse_arg(tag, context):
-    name = tag.attributes['name'].value
+def _check_ifunless(tag, context):
     if tag.attributes.has_key('if'):
         val = resolve_args(tag.attributes['if'].value, context)
-        if val == '1' or val == 'true':
-            return (name, _get_arg_value(tag, context))
+        if not convert_value(val, 'bool'):
+            return False
     elif tag.attributes.has_key('unless'):
         val = resolve_args(tag.attributes['unless'].value, context)
-        if val == '0' or val == 'false':
-            return (name, _get_arg_value(tag, context))
-    else:
-        return (name, _get_arg_value(tag, context))
-    # nothing to return (no value, or conditional wasn't satisfied)
-    return None
+        if convert_value(val, 'bool'):
+            return False
+    return True
 
 def _parse_subcontext(tags, context):
     subcontext = {'arg': {}}
@@ -116,12 +113,8 @@ def _parse_subcontext(tags, context):
        return subcontext
     
     for tag in [t for t in tags if t.nodeType == DomNode.ELEMENT_NODE]:
-        if tag.tagName == 'arg':
-            # None is returned for args with if/unless that evaluate to false
-            ret = _parse_arg(tag, context)
-            if ret is not None:
-                (name, val) = ret
-                subcontext['arg'][name] = val
+        if tag.tagName == 'arg' and _check_ifunless(tag, context):
+            subcontext['arg'][tag.attributes['name'].value] = _get_arg_value(tag, context)
     return subcontext
 
 def _parse_launch(tags, launch_file, file_deps, verbose, context):
@@ -130,6 +123,8 @@ def _parse_launch(tags, launch_file, file_deps, verbose, context):
             
     # process group, include, node, and test tags from launch file
     for tag in [t for t in tags if t.nodeType == DomNode.ELEMENT_NODE]:
+        if not _check_ifunless(tag, context):
+            continue
 
         if tag.tagName == 'group':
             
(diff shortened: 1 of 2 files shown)


mikepurvis:
@dirk-thomas Getting a dependency failure here on the Lunar buildfarm:
```
KeyError: "The cache has no package named 'ros-lunar-rosbuild'"
```
I'm assuming we should be targeting Lunar with new changes now, and considering Kinetic for backport?

----

Oh blah, I see that there are a number of broken/missing packages at the moment, including rosbuild: http://repositories.ros.org/status_page/ros_lunar_default.html?q=rosbuild

tools/roslaunch/src/roslaunch/depends.py
```diff
@@ -142,6 +143,14 @@ def _parse_launch(tags, launch_file, file_deps, verbose, context):
...
                 context['arg'][name] = val
         elif tag.tagName == 'include':
+            if tag.attributes.has_key('if'):
+                val = resolve_args(tag.attributes['if'].value, context)
+                if not convert_value(val, 'bool'):
+                    continue
+            elif tag.attributes.has_key('unless'):
+                val = resolve_args(tag.attributes['unless'].value, context)
+                if convert_value(val, 'bool'):
+                    continue
```
mikepurvis: Another possibility would be moving this logic up a level so that it would also do the right thing on `<group>` and `<node>`— not currently the case.

mikepurvis: This is what I've done instead in df0dfb8acd67bf8dbcb6765a5777515c2af3059b.

mikepurvis:
@dirk-thomas I think this is good to go, the failed tests appear to be unrelated to this change.

dirk-thomas:
The failing tests might be related to ros/rospack#70. I will retrigger the PR jobs once the new version of rospack has been synced to the testing repo.

dirk-thomas:

tools/roslaunch/src/roslaunch/depends.py
dirk-thomas: Is this block still necessary with `_check_ifunless` being called at the beginning of the loop?

mikepurvis: Oh— no it isn't, but I forgot to remove it. Fixed now.

Looks like it's still rospack 2.4.0 in the shadow repo, so this isn't going to pass yet.

dirk-thomas: `rospack` 2.4.1 has propagated to shadow-fixed already: http://repositories.ros.org/status_page/ros_lunar_default.html?q=rospack

Other jobs have turned green.

mikepurvis:
Most recent test got rospack 2.4.1 and ran clean.

tools/roslaunch/src/roslaunch/depends.py
```diff
@@ -94,20 +95,16 @@ def _get_arg_value(tag, context):
...
-    name = tag.attributes['name'].value
+def _check_ifunless(tag, context):
     if tag.attributes.has_key('if'):
         val = resolve_args(tag.attributes['if'].value, context)
-        if val == '1' or val == 'true':
-            return (name, _get_arg_value(tag, context))
-    elif tag.attributes.has_key('unless'):
+        if not convert_value(val, 'bool'):
+            return False
+    if tag.attributes.has_key('unless'):
```
dirk-thomas: Previously if a tag had both - an `if` as well as an `unless` attribute the second one wasn't considered. Is there a reason to change this?

mikepurvis: Oh I suppose not; I'll change it back.

mikepurvis:
@ros/ros_comm-maintainers This should be good to go in. Anyone got thoughts?

IanTheEngineer:
Not a maintainer, but changes look good on my end. Is this a candidate for backporting into Kinetic?

dirk-thomas:
@mikepurvis Looks good to me :+1:

@IanTheEngineer First it will be released into Lunar to give it some soak time. It can be backported if there is a strong need for it, the chance for regressions are low, and it doesn't change API.

PR #1003, opened 2017-02-26
add MonotonicTimer

flixr:
based on MonotonicTime (which uses the CLOCK_MONOTONIC).
This timer is not influenced by time jumps of the system time,
so ideal for things like periodic checks of timeout/heartbeat, etc...

Depends on ros/roscpp_core#55

### DIFF
--- a/clients/roscpp/CMakeLists.txt
+++ b/clients/roscpp/CMakeLists.txt
@@ -118,6 +118,7 @@ add_library(roscpp
   src/libros/poll_set.cpp
   src/libros/service.cpp
   src/libros/this_node.cpp
+  src/libros/monotonic_timer.cpp
   )
 
 add_dependencies(roscpp ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPORTED_TARGETS})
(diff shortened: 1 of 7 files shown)


flixr:
So while in theory this should work well, it seems that there is a bug in the boost condition timed_wait:
Even if you pass it a duration it will simply add it to the current system time and wait until then.
It seems to be fixed in boost 1.61 if you specify `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC`, see https://svn.boost.org/trac/boost/ticket/6377

So that means that currently the MonotonicTimer can nicely deal with forward jumps in time, but not with backwards jumps.

Possible solutions:
- also only wait for 1ms just like in the sim time case and re-check the sleep_end predicate
  - drawback is that you up a lot more often than needed
  - to do this only for the MonotonicTimer we either need to add a member function `isMonotonicTime` or use a typeid check?
- replace the boost condition variable with C++11 equivalent where you can specify it to use a monotonic clock

flixr:
Closing in favor of #1014

Issue #146, opened 2013-01-31
roscpp: Cannot have multiple latched publishers on the same topic  (ros ticket #2356)


tfoote:
In rosplay, to actually mirror the appearance of multiple nodes, I need to create multiple latched publishers on the same topic.

The expected behavior would be for every advertiser to publish its own most recent message when a new subscriber shows up, which is what independent nodes would do.  However, only a single message gets published.

trac data:
- Owner: **straszheim**
- Reporter: **leibs**
- Reported at: **Tue Jan 19 03:50:03 2010**
- URL: https://code.ros.org/trac/ros/ticket/2356

tfoote:
[jfaust] This isn't how we spec'ed out latched publishers to work, and it's not completely trivial to change.  Not going to happen for 1.0.

tfoote:
[jfaust] Ken, I assume this wouldn't work in rospy either?

tfoote:
[kwc] Right, the latch is stored with the publisher singleton, not the publisher instance

tfoote:
[leibs] Pushing is fine.  I don't think it's a very common use-case given that latched recording wasn't working at all previously and nobody was complaining.

Is there a reason that latched publishers were spec'ed to be per-topic rather than per-advertiser?

If I have multiple advertisers from within the same node how do their subscription callbacks interact?  Does each get its own subscription callback at least?

tfoote:
[jfaust] subscription callbacks work per advertiser, but latched topics are really meant more for topics that only a single publisher is publishing on (like the map).

tfoote:
[tfoote] With people using nodelets more this can start to produce unexpected errors, and it would enable libraries from using latched topics.  In particular we are looking at libraries which used latched topics for bulk updates with a parallel channel with delta updates.

tfoote:
[jbohren] I also encountered this "feature" when I was trying to use latched publishers in smach last year.

wjwwood:
This came up again in the context of recording `/tf_static`.

One solution would be to have rosbag advertise as a non-latched topic but keep the latest message for each publisher, watch for new subscribers to come up and when they do publish the latest message from each publisher. The upside is that this approach would ensure that all subscribers would receive the latest message from each publisher, but the downside is that existing subscribers will get repeated messages when new subscribers come up.

Short of refactoring how publishers and latching works I don't think we can fix this...

mkhansenbot:
Any chance this will be fixed in an upcoming release?

[...]

PR #1004, opened 2017-02-26
Fix for rosbag play not publishing multiple latched messages on the same topic

racko:
Proof of concept for a solution to  #146 (in the rosbag context).

- Starts one child-process per latched topic callerid to publish the
  messages of this sender.
- Shutdown behavior is strange: First Ctrl+C kills child-processes,
  second kills rosbag play.
- Does not solve the issue of latched topics not being present in split
  bags after the first one.

### DIFF
--- a/tools/rosbag/CMakeLists.txt
+++ b/tools/rosbag/CMakeLists.txt
@@ -6,7 +6,7 @@ if(NOT WIN32)
 endif()
 
 find_package(catkin REQUIRED COMPONENTS rosbag_storage rosconsole roscpp std_srvs topic_tools xmlrpcpp)
-find_package(Boost REQUIRED COMPONENTS date_time regex program_options filesystem)
+find_package(Boost REQUIRED COMPONENTS date_time regex program_options filesystem thread)
 find_package(BZip2 REQUIRED)
 
 catkin_python_setup()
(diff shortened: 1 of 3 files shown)


dirk-thomas:

dirk-thomas:
The problem is that the spawned child process will operate completely independent from the original player instance. The player instance has an API to pause / resume etc. All of these commands wouldn't affect the spawned child process.

racko:
I finally found time to try and work out the kinks in this commit, but I 
get it now. I give up :)

For everyone trying something similar, here are the reasons why spawning
a child process does not solve the problem (completely):

If the topics are more "regular" (i.e. more like /tf than like
/tf_static), and the latching is just "an extra", the child process
would have to be paused and unpaused with the parent process, contrary
to my original assumptions. This could be done, but a simple,
unsynchronized implementation might behave less deterministically then
expected, while a synchronized version (all processes make sure that
they publish their messages in order) could be less performant.

The bigger problem is that it only works if every node publishes at most 
one message on the latched topic: If one node has more than one
publisher latching messages on the same topic, we again run into the 
issue that instead of one message per publisher we would get one message
per node. Sounds contrived, but imagine a node with multiple
tf2::StaticTransformBroadcasters. I guess it would be beneficial if the
callerid would contain a "unique ros::Publisher id". Then we could get
to the point of "one message per publisher". But if, for example, a
single StaticTransformBroadcaster is used to publish a number of
different static transformations, we'd have no way to latch every single
message because it doesn't even work that way in the recorded system:
For each StaticTransformBroadcaster we only get a single latched
message.

So in the end, providing a launch file full of 
static_transform_publisher calls (or something similar) still seems to
works better than the best workaround, making it something of a waste to
keep working on it.

P.S. here's a TODO for the current code, If anybody cares:

Only works for a single latched topic per node: if we have /tf_static
and /rosout, we launch pub_tf_static twice per callerid and each will
publish both topics. Would be easy to fix by replacing threads with a
"callerid -> child" map (one process per callerid is fine because they
publish all latched topics for that callerid)

PR #1006, opened 2017-02-27
fix rosmsg show from bag

### LINKED ISSUE #1002: rosmsg: Error when showing message from a bag file
I am trying to show a custom message from a bag file by calling `rosmsg show -b my.bag msg_here`, which fails with the following error message:

```
  File "xxxx/ros_catkin_ws/install_isolated/bin/rosmsg", line 35, in <module>
    rosmsg.rosmsgmain()
  File "xxxx/ros_catkin_ws/install_isolated/lib/python2.7/site-packages/rosmsg/__init__.py", line 747, in rosmsgmain
    sys.exit(rosmsg_cmd_show(ext, full, command))
  File "xxxx/ros_catkin_ws/install_isolated/lib/python2.7/site-packages/rosmsg/__init__.py", line 607, in rosmsg_cmd_show
    print(get_msg_text(datatype, options.raw, pytype._full_text))
  File "xxxx/ros_catkin_ws/install_isolated/lib/python2.7/site-packages/rosmsg/__init__.py", line 425, in get_msg_text
    for p in rospack.list():
AttributeError: 'str' object has no attribute 'list'
```

Looking at the source code, it seems that the [message's full text](https://github.com/ros/ros_comm/blob/kinetic-devel/tools/rosmsg/src/rosmsg/__init__.py#L606) (as a string) is being passed to a [method that expects a rospack object](https://github.com/ros/ros_comm/blob/kinetic-devel/tools/rosmsg/src/rosmsg/__init__.py#L414).

dirk-thomas:
Fixes #1002.

### DIFF
--- a/tools/rosmsg/package.xml
+++ b/tools/rosmsg/package.xml
@@ -21,6 +21,7 @@
 
   <run_depend version_gte="0.6.4">catkin</run_depend>
   <run_depend>genmsg</run_depend>
+  <run_depend version_gte="0.6.5">genpy</run_depend>
   <run_depend>python-rospkg</run_depend>
   <run_depend>rosbag</run_depend>
   <run_depend>roslib</run_depend>
--- a/tools/rosmsg/src/rosmsg/__init__.py
+++ b/tools/rosmsg/src/rosmsg/__init__.py
@@ -48,6 +48,7 @@ from catkin.find_in_workspaces import find_in_workspaces
 
 import rospkg
 import genmsg
+from genpy.dynamic import generate_dynamic
 
 import roslib.message
 import rosbag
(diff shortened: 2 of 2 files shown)


kvalev:
This fix did not work for me - I dont have the type locally, so RosPack cannot find it:
```
$ rosmsg show -b my.bag packageYYY/messageZZZ
Unable to load msg [packageYYY/messageZZZ]: Cannot locate message [messageZZZ]: unknown package [packageYYY] on search path [....]
```

What does fix the problem for me is just printing the full text of pytype:
 ```
           datatype, _, _, _, pytype = msg
            if datatype == arg:
                print(pytype._full_text)
                break
```

dirk-thomas:
I only tried the patch with a message which was available on the system.

The snippet you posted does not take the `--raw` option into account. I updated the patch to support messages not available on the system and handle `--raw` but the change depends on a change to `genpy` to expose the specs for the dynamically created message (see ros/genpy#75).

kvalev:
Right, I forgot about the `--raw` option... I tested the changes locally (both with and without `--raw`) and it works as expected. :+1:

dirk-thomas:
Thanks for trying the patch and reporting back!

dirk-thomas:

dirk-thomas:
This needs to wait until ros/genpy#75 has been released and can be backported to Kinetic.

PR #989, opened 2017-02-16
[roslaunch] Indicating <node> tag when it is actually a <test> tag is confusing.

130s:
For example, [this warning msg](https://travis-ci.org/wg-perception/people/jobs/202019288#L3737) (in this [PR](https://github.com/wg-perception/people/pull/49)):

 ```
 * WARN: unrecognized 'group' tag in <node> tag. Node xml is <test name="$(arg testnode_name)" pkg="rostest" test-name="hztest_test" type="hztest">
    <group unless="$(arg expected_success)">
      <node args="-r 0.5 $(find face_detector)/test/face_detector_noface_test_diamondback.bag" name="play" pkg="rosbag" type="play"/>
      <param name="hz" value="0.0"/>
    </group>
```

This is confusing because the tag in question is actually `<test>`, not `<node>`. Although the warning message refers to the exact <test> tag, I didn't try to read it because I was told the issue is with `<node>`, not `<test>`.

With this PR the message becomes a bit verbose but hope the readers get better idea.

PR #985, opened 2017-02-13
add test code for multibyte string

dirk-thomas:
Replaces #530 and target kinetic-devel branch.

PR #958, opened 2017-01-10
[rospy] made get_published_topics threadsafe

jabbenseth:
With multiple threads trying to get the currently published topics at least one of them crashes with 

```
Exception in thread Thread-4:
Traceback (most recent call last):
    topics = rospy.get_published_topics()
  File "/ros_comm/clients/rospy/src/rospy/client.py", line 376, in get_published_topics
    code, msg, val = get_master().getPublishedTopics(namespace)
  File "/ros_comm/clients/rospy/src/rospy/msproxy.py", line 106, in wrappedF
    return f(*args, **kwds)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1233, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1587, in __request
    verbose=self.__verbose
  File "/usr/lib/python2.7/xmlrpclib.py", line 1273, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1303, in single_request
    response = h.getresponse(buffering=True)
  File "/usr/lib/python2.7/httplib.py", line 1077, in getresponse
    raise ResponseNotReady()
ResponseNotReady

Exception in thread Thread-5:
Traceback (most recent call last):

    topics = rospy.get_published_topics()
  File "/ros_comm/clients/rospy/src/rospy/client.py", line 376, in get_published_topics
    code, msg, val = get_master().getPublishedTopics(namespace)
  File "/ros_comm/clients/rospy/src/rospy/msproxy.py", line 106, in wrappedF
    return f(*args, **kwds)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1233, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1587, in __request
    verbose=self.__verbose
  File "/usr/lib/python2.7/xmlrpclib.py", line 1273, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1298, in single_request
    self.send_request(h, handler, request_body)
  File "/usr/lib/python2.7/xmlrpclib.py", line 1400, in send_request
    connection.putrequest("POST", handler, skip_accept_encoding=True)
  File "/usr/lib/python2.7/httplib.py", line 906, in putrequest
    raise CannotSendRequest()
CannotSendRequest
```

the acquisition of `f` is thread safe, but the final call to `f` seems not.

PR #1008, opened 2017-03-01
changes between 1.11.20 and 1.12.7 for backporting

dirk-thomas:
The following list of changes has been integrated into ros_comm 1.12.7 (Kinetic) since the last Indigo / Jade release (1.11.20).

**Backported:** (these changes are part of this PR)

* improve stacktrace #811

  * helpful improvement, low chance of regression

* rospy.logXXX_throttle #812

  * new feature, backport requested, low chance of regression

* rostopic doc #816

  * helpful improvement

* rostest target dependency #830

  * bug fix

* fix copyright #844

  * helpful improvement

* rosbag filter progress #857

  * helpful improvement

* fix target dependency #869

  * helpful improvement

* Timer doc #878

  * helpful improvement

* init values #884

  * bug fix

* test type handling e34122564c254252f201a778dd01466509bb325a

  * bug fix

* remove invalid export 4468f862d3c80dbba1bb61b5a8cc6c9a86de641b

  * bug fix

* include for gcc6 #911

  * bug fix

* increase rpc queue size #849

  * improvement for OS X

* Fix BagMigrationException #917

  * helpful improvement

* rospy param doc #964

  * helpful improvement

* fix rospack caching in launch #966

  * significant performance improvement

* prevent segfault #971

  * helpful improvement

* typo #992

  * helpful improvement

* improve error message #989

  * helpful improvement

* get_published_topics threadsafe #958

  * bug fix, low risk of regression?

* fix UDP block number #957

  * bug fix, low risk of regression?

* rosbag terminate subprocesses #951

  * bug fix, low risk of regression?

**Not backported:**

* make destructor virtual #773 

  * breaking compatibility, not backported

* add parameter #774

  * breaking compatibility, not backported

* change return value #775

  * breaking compatibility, not backported

* remove global variables #776 #871

  * breaking compatibility, not backported

* reenable tests 33a5a277a90d75eced6d27d6e67d9da2ebccb7f3 22a7dc62176223324a7469d057f692bce2d25ae6

  * no need to backport

* add test #780

  * no need to backport

* defused xml #782

  * requires new dependency, not backported

* remove duplicate function #783

  * depends on change in genpy only available in Kinetic and higher

* roslaunch expressions #784 #793

  * new feature, not backported

* rosnode info quiet #809

  * new feature, not backported

* timestamp-less cache #806

  * new feature, not backported

* compiler warning #817

  * not necessary

* edge list contains #754

  * not necessary

* rostopic hz multiple #712 #843 #886 #888 #896

  * new feature, not backported

* abstract connection based transport #713

  * new feature, not backported

* enable tests #826

  * not necessary

* change queue operation order to match msg #818

  * not necessary

* explicit constructor #835

  * not necessary

* compiler warning #839 #842

  * not necessary

* Python 3 #840 #847 #978

  * not supported, not necessary

* lz4_FOUND c14b9428e63fc4a4591851c642d814f80940ec0b

  * depends on catkin change only available in Kinetic

* logging formatter #828

  * new feature, not backported

* Darwin #846

  * not necessary

* rostopic echo --noarr/nostr #724 #909 #916

  * new feature, not backported

* fast approximate time synchronization #802

  * new feature, not backported

* 64bit rosbag #865

  * new feature, not backported

* record max number of splits #866

  * new feature, not backported

* specific to kinetic only #872

  * does not apply

* example / comment #873 #862

  * not necessary

* rostopic type #860

  * new feature, not backported

* CMake warning #880

  * not necessary

* paramtest, publishtest #863 #859

  * new feature, not backported

* use_test_depends #887 #890 660684352f3fbb90ebf3a8a2278667fc4649d0c9

  * new feature, not backported

* gennodejs #02794b1760aaaef8c7683615aa792b8d57d1956d

  * does not apply to Indigo / Jade

* roslaunch include file empty #882

  * new feature, not backported

* empty node name #894 #895

  * new check, potential break existing code, not backported

* roslaunch #897

  * depends on patch in ros only available in Kinetic

* fix multi-threaded spinning #867

  * significant change, not backported

* boost bind callback #906

  * new feature, not backported

* roslaunch USE_TEST_DEPENDENCIES #910

  * new feature, not backported

* default rospy logging format #879

  * changes behavior, not backported

* rospy reconnect #851

  * changes behavior, not backported

* class -> struct #924

  * not necessary

* drop epydoc #925

  * not necessary

* use poll in write_header #929

  * changes implementation, risk of regression, not backported

* change return value #938

  * changes behavior, risk of regression, not backported

* rosbag plat wait for subscribers #959

  * new feature, not backported

* rosbag player pause service #949

  * new feature, not backported

* move xmlrpc headers #962

  * not necessary

* poll -> epoll #831

  * changes implementation, risk of regression, not backported

* separate logging #980

  * changing logger name, not backported

* additional tests #772 #985

  * not necessary

* more logging #979

  * new feature, not backported

* rosmsg info #941

  * new feature, not backported

* rate control topic #947

  * new feature, not backported

@ros/ros_team @ros/ros_comm-maintainers Please comment on the decision which changes to (not) backport. Especially for the three backported changes marked with a "?" I wasn't sure about the regression risk.


wjwwood (approved):
lgtm, I especially looked at the "low risk of regression?" labeled ones.

dirk-thomas:
@wjwwood Thank you for going through this lengthy list as well as all the other backport PRs. :bowing_man:

gavanderhoorn:
@dirk-thomas: I found the list extremely informative - not only because it provides an overview of bugs that were fixed, but also because it shows what is not backported. +10.

Do you intend to provide such as list for future backport PRs?

Would it be an idea to link to these PRs somewhere or post the list to a more visible location?

dirk-thomas:
> Do you intend to provide such as list for future backport PRs?

Due to the amount of patches to be considered and the advantage of having the backports tested by CI, yes, I think I will do it the same way in the future in this repo.

> Would it be an idea to link to these PRs somewhere or post the list to a more visible location?

Why not. What do you have in mind?

tfoote:
It would be great to queue #962 for backport. I have ros/geometry/pull/145 against geometry to use the new path. However since I haven't forked development I'm holding the PR until #962 is backported. Since the goal is for downstream packages to use the new path it would be great to let that happen sooner rather than waiting for indigo to EOL or development to fork.

dirk-thomas:
#962 is intentionally not backported as described above since it is not necessary for older distributions. So I don't think any of the rational had changed to justify backporting it.

tfoote:
Indeed #962 is "not necessary" for older distributions. It's also "not necessary" on any distribution. It's a style cleanup flagged by @jspricke as a potential conflict on Debian. And the goal of fixing this way was that downstream code should be easy to fix. The benefit of this change is not achieved until everyone has switched.

Looking at the [comment](https://github.com/ros/ros_comm/pull/962/files#diff-37f0e72f6015fee29b627c357f9794beR10) stating that it should be removed in lunar @mightyCelu helpfully submitted ros/geometry#145 to update it, but I cannot merge it because the development is not forked for the new distro(s). And especially as geometry is now mostly a backwards compatibility place holder I don't plan on forking it since there's no new feature development going on. Not backporting this change would force forking the repo as soon as the old header locations are removed. I can see that it's the same situation for diagnostics. And if there is any other moderately stable package out there they will have the same issue.

Thus I'm asking that this be queued as it will be convenient for downstream developers, such as myself, as well as anyone else using the xmlrpcpp headers directly in a package not under active development but still being maintained. Alternatively tocking it can wait until the Indigo rosdistro is EOL, but that's quite a long way out still.

gavanderhoorn:
@dirk-thomas wrote:
>> Would it be an idea to link to these PRs somewhere or post the list to a more visible location?
> 
> Why not. What do you have in mind?

I only stumbled on this PR and the description of the backports by accident. Perhaps we could create some page on the wiki where these kind of PRs are listed chronologically?

I was first thinking of perhaps adding a link to the repo sync mails / posts that @tfoote sends out, but that would possibly be hard to correlate with a particular backport PR.

mikepurvis:
In terms of overall visibility/discoverability, I'm not sure that another wiki page to have to maintain would be any better than linking to a Github search, eg:

https://github.com/ros/ros_comm/pulls?q=backport%20in%3Atitle

gavanderhoorn:
I don't really care how or where, as long these backport PRs receive some more attention. Keeping up-to-speed with all the changes in (core) packages is quite involved, so having a single overview of backports like this would seem like a valuable resource, that seems to deserve a bit more exposure.

mikepurvis:
Is the concern primarily about breakages, or new features, or something else? These are intentionally selected on the basis of being low-risk changes/fixes that have been specifically requested by users of the LTS release.

gavanderhoorn:
I'm not too worried about breakages: the really important pkgs seem to have sufficient unit tests to catch regressions most of the time.

Keeping up with backports of new features is definitely nice, but knowing what has changed (even if it was supposed to be an orthogonal change that shouldn't have introduced any changes to existing behaviour) is invaluable to me when trying to debug something.

Also: being able to point users on ROS Answers to PRs that have either changed something that they were not aware of or introduced features they thought / assumed they could not use on their version of ROS would be another rationale.

PR #1009, opened 2017-03-02
Fix test generation

efernandez:
This PR fixes the data generation, which was completely broken for the messages generations other than the current one. It was broken mainly because of the message generation from the CMakeLists.txt, especially because some messages are moved/renamed from one generation to the other, so we need to provide the specific message list for each generation.

I've done that by providing a cmake macro called `MSG_DIRECTORY`.

I also simplified and fix the script to generate the data/bag files, which works completely unattended.

I also created a script to generate the rules, but note that this one only aims to help in the process, but the user still needs to provide some input, e.g. moved/renamed messages, remove rules that we intentionally don't want for the tests, e.g. unmigrated messages, and organize the rules properly.

@mikepurvis @dirk-thomas


test/test_rosbag/bag_migration_tests/scripts/generate_data
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+# Function that builds a message generation:
+build_generation()
+{
+    catkin build --no-deps --no-notify --no-status --no-summary --cmake-args -DMSG_DIRECTORY="msg_$1" -- test_rosbag \
+        >/dev/null
```
mikepurvis: We probably don't want a `catkin_tools` dependency here. Given the simplicity of what's being done, consider just using `cmake; make` directly?

test/test_rosbag/bag_migration_tests/CMakeLists.txt
```diff
@@ -5,22 +5,82 @@
...
+    MigratedExplicit.msg
+    MigratedImplicit.msg
+    MigratedMixed.msg
+    PartiallyMigrated.msg
+    Renamed1.msg
+    Simple.msg
+    SimpleMigrated.msg
+    SubUnmigrated.msg
+    Unmigrated.msg
+  )
```
mikepurvis: `add_message_files` can be called more than once. Would it make sense to have a single block for all the common ones, and then only do conditionally the ones that are different (eg `Renamed1`/`Renamed2`)?

Note also that CMake lets you do fun stuff like this:

```
set(msglist_msg_gen1 Constants.msg Converged.msg ...)
add_message_files(
  DIRECTORY ${MSG_DIRECTORY}
  FILES ${msglist_${MSG_DIRECTORY}}
)
```

efernandez: Ok, that makes sense... will do

test/test_rosbag/bag_migration_tests/scripts/generate_data
efernandez: I could try, but remember this is only to generate the data. In other words, it's an utility script that the developer would run to provide/update the tests, but it's not a dependency of the package strictly speaking.

I'd try doing that though, but where should I run `cmake` from? I'd like to keep things tidy and clean, with the `build` directory in the `build/PACKAGE` folder under the workspace.

dirk-thomas: You can use `catkin_make` or `catkin_make_isolated` since this package already depends on `catkin`. It doesn't depend on `catkin_tools` (and shouldn't) so that command is not available (if the user hasn't installed it manually).

efernandez: Fair enough, I'll change it to `catkin_make` or `catkin_make_isolated`, whatever is faster to execute.

efernandez: Done, using `catkin_make`.

test/test_rosbag/bag_migration_tests/CMakeLists.txt
efernandez: Done.

I create a list of common messages, and separate list for each generation. Then I use the `MSG_DIRECTORY` macro to select the list directly, with no if-else statements. Now is much more compact and cleaner.

efernandez:
Review comments addressed in this PR.
Changes are already applied in this PR and I also rebased #1010 and #1011 .

Tests are running again, but they won't test the data and rules generation tools. I tested them again locally (now that they use `catkin_make`, and I need to run it from the root of the workspace)

mikepurvis:
LGTM 👍

efernandez:
ping @dirk-thomas

dirk-thomas:
The code changes look reasonable on a first look. Can you please give some examples how to invoke these scripts.

efernandez:
@dirk-thomas I've added a `README.md` file explaining how to use the two scripts **generate_data** and **generate_rules**: https://github.com/clearpathrobotics/ros_comm/blob/c75c704143ec875405e86e8f50aed214456df869/test/test_rosbag/bag_migration_tests/README.md

Please remember that **generate_rules** only aims to help in the process. It's not a perfect thing, but it's better than nothing (as before), and still requires an expert user to run it and understand what's going on with the rules. If the user has no idea, he/she can set the generated rules to `valid = True` and that should be enough to generate all the rules templates, but they still need to be organized because the rules inside the files can be for messages that changed for fields in the messages that were being migrated.

test/test_rosbag/bag_migration_tests/README.md
```diff
@@ -0,0 +1,349 @@
...
+ * **generate_data** : Generates *.bag* files with messages (data) to run the bag migration tests. 
+ * **generate_rules** : Help in the process of creating rules between two message generations.
+
+The next sections explain how to use these scripts in detail. The rest are auxiliary scripts that should NOT be used directly.
+
+Before running any of these scripts you need to build the **test_rosbag** package at least once in your workspace and `source` the environment workspace setup file. This can be done running the following from the root of the workspace:
+
+``` bash
+$ catkin_make
+$ source devel/setup.bash
```
dirk-thomas: This new doc is great. Thank you for putting this together.

Since not all workspaces can be built by `catkin_make` it might be better to use `catkin_make_isolated`?

efernandez: That's why I started with `catkin build`, but I see your point. :smiley: 

I'd need some time to re-adapt it to `cakin_make_isolated`, but hopefully it's a single line change.

efernandez: I see you're talking about the `README.md`, but I'm also using `catkin_make` on the scripts... so they should be same because we can't mix different building tools.

efernandez: I honestly don't know which one is a better option. I prefer `catkin_make` because I can pass `--only-pkg-with-deps test_rosbag` to build it faster. I could update the `README.md` to do the same.

With `catkin_make_isolated` the `--only-pkg-with-deps test_rosbag` option needs `--install` and a quick test has failed for me because of missed dependencies (like `generate_messages`). I don't use `catkin_make_isolated` often, so I'm not very familiar with the environment setup I need to make it work properly.

dirk-thomas: Yes, the scripts should use `catkin_make_isolated` too.

The problem is that `catkin_make` doesn't work for arbitrary workspaces (even if each package is "correct"). `catkin_make_isolated` scales to work with any workspace (if each package is "correct"). Therefore I think we should use that. The usage is pretty much the same - the setup file to source will just be in `devel_isolated/setup.bash` - no need to use `--install`.

I am happy to provide feedback if you post more information about your workspace and how it fails.

efernandez: Never mind, I got it to work. This is what I did wrong:

1. I called `catkin_make_isolated` with `--pkg` from a clean ws. And according with the documentation I need to call `cakin_make_isolated` w/o `--pkg` once first.
``` bash
  --pkg PKGNAME [PKGNAME ...]
                        Invoke 'make' on specific packages (only after
                        catkin_make_isolated has been invoked before with the
                        same install flag)
```

Otherwise I have an error saying that `cmake` fails, which makes sense:

``` bash
==> cmake /home/eperdomo/dev/ws/clearpath_ws/src/ros_comm/test/test_rosbag -DCATKIN_DEVEL_PREFIX=/home/eperdomo/dev/ws/clearpath_ws/devel_isolated/test_rosbag -DCMAKE_INSTALL_PREFIX=/home/eperdomo/dev/ws/clearpath_ws/install_isolated -G Unix Makefiles in '/home/eperdomo/dev/ws/clearpath_ws/build_isolated/test_rosbag'
Unhandled exception of type 'OSError':
  File "/opt/ros/indigo/lib/python2.7/dist-packages/catkin/builder.py", line 965, in build_workspace_isolated
    number=index + 1, of=len(ordered_packages)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/catkin/builder.py", line 665, in build_package
    destdir=destdir, use_ninja=use_ninja
  File "/opt/ros/indigo/lib/python2.7/dist-packages/catkin/builder.py", line 397, in build_catkin_package
    run_command_colorized(cmake_cmd, build_dir, quiet, add_env=add_env)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/catkin/builder.py", line 187, in run_command_colorized
    run_command(cmd, cwd, quiet=quiet, colorize=True, add_env=add_env)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/catkin/builder.py", line 205, in run_command
    raise OSError("Failed command '%s': %s" % (cmd, e))
OSError: Failed command '['/home/eperdomo/dev/ws/clearpath_ws/devel_isolated/rosbag/env.sh', 'cmake', '/home/eperdomo/dev/ws/clearpath_ws/src/ros_comm/test/test_rosbag', '-DCATKIN_DEVEL_PREFIX=/home/eperdomo/dev/ws/clearpath_ws/devel_isolated/test_rosbag', '-DCMAKE_INSTALL_PREFIX=/home/eperdomo/dev/ws/clearpath_ws/install_isolated', '-G', 'Unix Makefiles']': [Errno 2] No such file or directory
<== Failed to process package 'test_rosbag': 
  Failed command '['/home/eperdomo/dev/ws/clearpath_ws/devel_isolated/rosbag/env.sh', 'cmake', '/home/eperdomo/dev/ws/clearpath_ws/src/ros_comm/test/test_rosbag', '-DCATKIN_DEVEL_PREFIX=/home/eperdomo/dev/ws/clearpath_ws/devel_isolated/test_rosbag', '-DCMAKE_INSTALL_PREFIX=/home/eperdomo/dev/ws/clearpath_ws/install_isolated', '-G', 'Unix Makefiles']': [Errno 2] No such file or directory
Command failed, exiting.
```

2. I was trying to pass `--only-pkg-with-deps`, which as you're saying I don't need. In fact, I can now use `--pkg`.
``` bash
  --only-pkg-with-deps ONLY_PKG_WITH_DEPS [ONLY_PKG_WITH_DEPS ...]
                        Only consider the specific packages and their
                        recursive dependencies and ignore all other packages
                        in the works
```

With this one the error was a missed dependency.

Either way, I've just updated the PR to use `catkin_make_isolated`.

efernandez:
ping @dirk-thomas 

I've updated scripts and documentation to use `catkin_make_isolated` as suggested. All related PRs rebased and updated.

efernandez:
ping @dirk-thomas

efernandez:
ping @dirk-thomas

efernandez:
@dirk-thomas Do you have any other comments or suggestions? I'd like to see this PR merged, so we can starting discussing the test in #1010 and the fix in #1011 .

dirk-thomas:
@efernandez I get notifications for every comment on repositories I maintain. Sometimes my other responsibilities have a higher priority and therefore I can't always respond immediately. Repeatedly pinging me will certainly not speed up my response time. I will try to take a look at the updated patch as soon as possible.

efernandez:
I understand, sorry for that. I just didn't want this to get lost now that I addressed all your comments.

I'd wait for your reply. Thanks!

dirk-thomas:
I just ran through the examples and it works fine for me. This is a great improvement. Thank you for you effort on this!

PR #1010, opened 2017-03-02
Add test migration rule

efernandez:
This adds a test for https://github.com/ros/ros_comm/pull/996, which I'm going to close a create a new one rebased on top of this to show it actually fixes the issue (see https://github.com/ros/ros_comm/pull/1011). I'd also target `lunar-devel`.

As a reminder, this explains the corner case this test covers now.

# Description of the buggy corner case not handled on the rosbag migration tool

Consider the following scenario for a message G with the following versions:

> G0 -- rule.bmr --> G1 -- auto-generated rule --> G3

where `G3` is the **current** msg, and `G2` is the **bag** message that needs migration.

We need to check for a migration rule like this first:

> G2 --> G3

instead of the backwards ones that were generated before this bug fix:

> G2 --> G1

Note that the **lazy** check to mark the rules as valid in the `make_update_rule` (see https://github.com/clearpathrobotics/ros_comm/blob/kinetic-devel-cpr/tools/rosbag/src/rosbag/migration.py#L1123) method makes that rule pass initially, but later it fails because there's no rule for the sub fields.

@mikepurvis @dirk-thomas 

Note this PR is on top of https://github.com/ros/ros_comm/pull/1009


test/test_rosbag/bag_migration_tests/CMakeLists.txt
```diff
@@ -5,47 +5,136 @@
...
 
 catkin_download_test_data(download_data_test_constants_gen1.bag http://download.ros.org/data/test_rosbag/constants_gen1.bag FILENAME test/constants_gen1.bag MD5 77ec8cb20e823ee3f3a87d07ea1132df )
 catkin_download_test_data(download_data_test_constants_gen2.bag http://download.ros.org/data/test_rosbag/constants_gen2.bag FILENAME test/constants_gen2.bag MD5 4efd8459bf9403d1171dff06abcd6548 )
 catkin_download_test_data(download_data_test_converged_gen1.bag http://download.ros.org/data/test_rosbag/converged_gen1.bag FILENAME test/converged_gen1.bag MD5 8e3524157d31b5761ac951fe16e03e12 )
 catkin_download_test_data(download_data_test_converged_gen2.bag http://download.ros.org/data/test_rosbag/converged_gen2.bag FILENAME test/converged_gen2.bag MD5 0ad4041d2e3bab8262c12020ec3e048e )
 catkin_download_test_data(download_data_test_converged_gen3.bag http://download.ros.org/data/test_rosbag/converged_gen3.bag FILENAME test/converged_gen3.bag MD5 90dd16cea5c51fca65be617654fb6b76 )
+#catkin_download_test_data(download_data_test_converged_gen4.bag http://download.ros.org/data/test_rosbag/converged_gen4.bag FILENAME test/converged_gen4.bag MD5 c88ac1f4a9a5eebd667ec55224988e40 )
```
efernandez: @dirk-thomas These 6 bag files are now in Dropbox. Could you please move them to `download.ros.org`, please? Then I'd update the PR.

efernandez:
Please note that this PR should fail.

Here I just want to:

1. Show the test on its own
2. Show that this test fails with the current code

I'm going to create another PR on top of this one that should pass.
The intention is to close this one and merge the other one.

I'd prefer that you review the test changes here, rather than on the PR I'm going to create with the fix.

Also remember that https://github.com/ros/ros_comm/pull/1009 should be reviewed and merged before this one.

Indeed, these two tests should fail:

``` bash
FAIL: test_converged_gen4 (migrate_test.MigrationTest)
  File "/home/eperdomo/dev/ws/clearpath_ws/build/test_rosbag/test/migrate_test.py", line 324, in test_converged_gen4
    self.do_converged(4)
  File "/home/eperdomo/dev/ws/clearpath_ws/build/test_rosbag/test/migrate_test.py", line 294, in do_converged
    self.assertTrue(not False in [m[1] == [] for m in res], 'Bag not ready to be migrated')
AssertionError: Bag not ready to be migrated

FAIL: test_migrated_mixed_gen4 (migrate_test.MigrationTest)
  File "/home/eperdomo/dev/ws/clearpath_ws/build/test_rosbag/test/migrate_test.py", line 243, in test_migrated_mixed_gen4
    self.do_migrated_mixed(4)
  File "/home/eperdomo/dev/ws/clearpath_ws/build/test_rosbag/test/migrate_test.py", line 215, in do_migrated_mixed
    self.assertTrue(not False in [m[1] == [] for m in res], 'Bag not ready to be migrated')
AssertionError: Bag not ready to be migrated

```

And now that the PR test jenkins job finished you can see it's exactly failing on those two, as expected: http://build.ros.org/job/Lpr__ros_comm__ubuntu_xenial_amd64/25/testReport/

And http://build.ros.org/job/Lpr__ros_comm__ubuntu_xenial_amd64/26/ shows that #1011 fixes it.

test/test_rosbag/bag_migration_tests/CMakeLists.txt
efernandez: ping @dirk-thomas

dirk-thomas: Sorry, for the delay. The 6 files are available via http://download.ros.org/data/test_rosbag/ now.

efernandez: Thanks. PR updated to use them.

Also rebased/updated #1011 that should pass the test; remember that this PR (#1010 it's only here to show that the current code fails to handle the new test provided)

dirk-thomas: :+1:

PR #1014, opened 2017-03-06
Add SteadyTimer

flixr:
Add a SteadyTimer which is based on SteadyTime (which uses the CLOCK_MONOTONIC).
This timer is not influenced by time jumps of the system time, so ideal for things like periodic checks of timeout/heartbeat, etc...

The difficulty is that using `wait_for` or `wait_until` on a boost condition variable always uses the system clock until version 1.61 (see https://svn.boost.org/trac/boost/ticket/6377).
Which obviously doesn't help for the SteadyTimer where we explicitly want the steady clock.
    
So as a workaround, include the backported versions of the boost condition variable if boost version is not recent enough.

This depends on ros/roscpp_core#57 where SteadyTime was added.

This also replaces #1003

### DIFF
--- a/clients/roscpp/CMakeLists.txt
+++ b/clients/roscpp/CMakeLists.txt
@@ -22,7 +22,7 @@ list(GET roscpp_VERSION_LIST 2 roscpp_VERSION_PATCH)
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/include/ros/common.h.in ${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_INCLUDE_DESTINATION}/ros/common.h)
 
-find_package(Boost REQUIRED COMPONENTS signals filesystem system)
+find_package(Boost REQUIRED COMPONENTS signals filesystem system chrono)
 
 include_directories(include ${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_INCLUDE_DESTINATION}/ros ${catkin_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS})
 
--- a/clients/roscpp/CMakeLists.txt
+++ b/clients/roscpp/CMakeLists.txt
@@ -22,7 +22,7 @@ list(GET roscpp_VERSION_LIST 2 roscpp_VERSION_PATCH)
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/include/ros/common.h.in ${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_INCLUDE_DESTINATION}/ros/common.h)
 
-find_package(Boost REQUIRED COMPONENTS signals filesystem system chrono)
+find_package(Boost REQUIRED COMPONENTS chrono filesystem signals system)
 
 include_directories(include ${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_INCLUDE_DESTINATION}/ros ${catkin_INCLUDE_DIRS} ${Boost_INCLUDE_DIRS})
 
--- a/clients/roscpp/CMakeLists.txt
+++ b/clients/roscpp/CMakeLists.txt
@@ -118,6 +118,7 @@ add_library(roscpp
   src/libros/poll_set.cpp
   src/libros/service.cpp
   src/libros/this_node.cpp
+  src/libros/steady_timer.cpp
   )
 
 add_dependencies(roscpp ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPORTED_TARGETS})
--- a/clients/roscpp/include/ros/steady_timer.h
+++ b/clients/roscpp/include/ros/steady_timer.h
@@ -1,0 +1,127 @@
+/*
+ * Copyright (C) 2017, Felix Ruess, Roboception GmbH
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *   * Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *   * Neither the names of Stanford University or Willow Garage, Inc. nor the names of its
+ *     contributors may be used to endorse or promote products derived from
+ *     this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef ROSCPP_STEADY_TIMER_H
+#define ROSCPP_STEADY_TIMER_H
+
+#include "common.h"
+#include "forwards.h"
+#include "steady_timer_options.h"
+
+namespace ros
+{
+
+/**
+ * \brief Manages a steady-clock timer callback
+ *
+ * A SteadyTimer should always be created through a call to NodeHandle::createSteadyTimer(), or copied from one
+ * that was. Once all copies of a specific
+ * SteadyTimer go out of scope, the callback associated with that handle will stop
+ * being called.
+ */
+class ROSCPP_DECL SteadyTimer
+{
+public:
+  SteadyTimer() {}
+  SteadyTimer(const SteadyTimer& rhs);
+  ~SteadyTimer();
+
+  /**
+   * \brief Start the timer.  Does nothing if the timer is already started.
+   */
+  void start();
+  /**
+   * \brief Stop the timer.  Once this call returns, no more callbacks will be called.  Does
+   * nothing if the timer is already stopped.
+   */
+  void stop();
+
+  /**
+   * \brief Returns whether or not the timer has any pending events to call.
+   */
+  bool hasPending();
+
+  /**
+   * \brief Set the period of this timer
+   */
+  void setPeriod(const WallDuration& period, bool reset=true);
+
+  bool isValid() { return impl_ && impl_->isValid(); }
+  operator void*() { return isValid() ? (void *) 1 : (void *) 0; }
+
+  bool operator<(const SteadyTimer& rhs)
+  {
+    return impl_ < rhs.impl_;
+  }
+
+  bool operator==(const SteadyTimer& rhs)
+  {
+    return impl_ == rhs.impl_;
+  }
+
+  bool operator!=(const SteadyTimer& rhs)
+  {
+    return impl_ != rhs.impl_;
+  }
+
+private:
+  SteadyTimer(const SteadyTimerOptions& ops);
+
+  class Impl
+  {
+  public:
+    Impl();
+    ~Impl();
+
+    bool isValid();
+    bool hasPending();
+    void setPeriod(const WallDuration &period, bool reset=true);
+
+    void start();
+    void stop();
+
+    bool started_;
+    int32_t timer_handle_;
+
+    WallDuration period_;
+    SteadyTimerCallback callback_;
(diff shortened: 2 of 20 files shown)


clients/roscpp/CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ CHECK_FUNCTION_EXISTS(trunc HAVE_TRUNC)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/libros/config.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h)
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
-add_library(roscpp 
+add_library(roscpp
```
dirk-thomas: Please avoid any unrelated whitespace changes in your PR.

clients/roscpp/include/ros/steady_timer.h
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright (C) 2009, Willow Garage, Inc.
```
dirk-thomas: This should not be "2009, Willow Garage, Inc.".

dirk-thomas (commented):
Beside the inline comments I do have a few more:

* Why should the boost headers be pulled into this repo?

* `steady_clock` seems to be only used in the `.cpp` files but not being exposed thorugh any idea. If that is the case why not use the native type available as of C++11?

* Please add unit tests covering the new functionality.

* The CI tests failed for what looks like a simple compiler error.

clients/roscpp/include/ros/steady_timer.h
wjwwood: Help @flixr out, what should it be (at least some guidelines)?

wjwwood:
> If that is the case why not use the native type available as of C++11?

That's probably my fault. I suggested in another pr that we might not want to use C++11 features in the core yet, but that was mostly because I didn't know how you felt about it. For instance would we need to add the `-std=c++11` option to the `libroscpp` compile options and some other externally affecting changes. Since this is to Lunar's branch, I'd say it's fine. It's a different question for indigo or something like that.

dirk-thomas:
If the C++11 specific features are only used in `.cpp` files the packages doesn't need to expose the compiler option. In that case I would say we can definitely use C++11 features.

wjwwood:
What I meant is that you can have ABI problems if roscpp is compiled with c++11 but other libraries are not and they both use STL types across the boundary. This usually isn't an issue for gcc/clang on unix systems, but it's not guaranteed either (at least as far as I know).

clients/roscpp/include/ros/steady_timer.h
dirk-thomas: The year the code snippet was written and the author / company who wrote it.

flixr:
> Why should the boost headers be pulled into this repo?

The only working implementation I could find was the one in boost > 1.61, see description.

I'm not particularly fond of this kludge to include the boost headers from 1.61 if the system version is not recent enough, but I couldn't think of another way (short of providing our own condition variable implementation). Any better ideas?

> steady_clock seems to be only used in the .cpp files but not being exposed thorugh any idea. If that is the case why not use the native type available as of C++11?

Sorry, don't quite understand which part you mean here... `SteadyTime` is added in rostime analogous to `WallTime`, see https://github.com/ros/roscpp_core/pull/57

Regarding C++11:
At least for the condition variable `wait_for` or `wait_until` it doesn't currently work with the gcc stdc++ implementation, as they also always use the system clock (contrary to what the standard says).
See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=41861 for the bug report and discussion on why it's still not fixed.

> The CI tests failed for what looks like a simple compiler error.

The CI tests obviously fail because https://github.com/ros/roscpp_core/pull/57 is not yet merged and hence `SteadyTime` is not available.

flixr:
Added some unit tests, but to make (the existing) tests compile I had to add the `-pthread` flag...
I'm still on indigo, so this might not actually be needed... if it's not, disregard the last commit that adds the flag.

flixr:
Also please note that the added unit tests (basically the same as for WallTimer) don't cover the actually important aspect of the SteadyTimer: changing the system time doesn't/shouldn't affect it.
I have no idea how that could be tested in the unit/ros tests though...

dirk-thomas:
> The CI tests obviously fail because ros/roscpp_core#57 is not yet merged and hence SteadyTime is not available.

Sorry, I missed this obvious reason. It indeed requires `roscpp_core` to be released with `SteadyTime`.

> > steady_clock seems to be only used in the .cpp files but not being exposed thorugh any idea. If that is the case why not use the native type available as of C++11?

> Sorry, don't quite understand which part you mean here... SteadyTime is added in rostime analogous to WallTime, see ros/roscpp_core#57

> Regarding C++11:
> At least for the condition variable wait_for or wait_until it doesn't currently work with the gcc stdc++ implementation, as they also always use the system clock (contrary to what the standard says).
> See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=41861 for the bug report and discussion on why it's still not fixed.

Let me try to rephrase my question and ask a few more specific ones:

* I was referring to these lines: https://github.com/ros/ros_comm/pull/1014/files#diff-3decb5e6adcb333e676de1a4812e1378R580 Is this the only location which uses the "new" `wait_until` from Boost 1.61?
* Is it correct that the boost headers are only used for these few lines? Then they would only be used for building this package but don't need to be exposed in any headers.
* Would it be possible to replace the boost `wait_until` with an equivalent from C++11 (http://www.cplusplus.com/reference/condition_variable/condition_variable/wait_until/) or doesn't the C++11 API provide the ability to operate on a monotonic clock (I haven't looked at it long enough to answer the question myself)?

wjwwood:
> Would it be possible to replace the boost `wait_until` with an equivalent from C++11 (http://www.cplusplus.com/reference/condition_variable/condition_variable/wait_until/) or doesn't the C++11 API provide the ability to operate on a monotonic clock (I haven't looked at it long enough to answer the question myself)?

I think @flixr said that these don't work in some version of gcc's standard library (that presumably we're using in Indigo?):

> At least for the condition variable `wait_for` or `wait_until` it doesn't currently work with the gcc stdc++ implementation, as they also always use the system clock (contrary to what the standard says).
See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=41861 for the bug report and discussion on why it's still not fixed.

flixr:
>     I was referring to these lines: https://github.com/ros/ros_comm/pull/1014/files#diff-3decb5e6adcb333e676de1a4812e1378R580 Is this the only location which uses the "new" wait_until from Boost 1.61?

Yes.

>     Is it correct that the boost headers are only used for these few lines? Then they would only be used for building this package but don't need to be exposed in any headers.

Basically yes... but since this code is the generic timer_manager.h (which is actually also installed), the boost headers also need to be exposed/installed...

> > Would it be possible to replace the boost wait_until with an equivalent from C++11 (http://www.cplusplus.com/reference/condition_variable/condition_variable/wait_until/) or doesn't the C++11 API provide the ability to operate on a monotonic clock (I haven't looked at it long enough to answer the question myself)?
> 
> I think @flixr said that these don't work in some version of gcc's standard library (that presumably we're using in Indigo?):

That is actually what I implemented at first, but it didn't work:
While the standard says that `wait_for` should always use a steady clock and you can specify `wait_until` to use a steady/monotonic clock, this is currently not the case in _any_ version of gcc's standard library. See bug report I linked before.
The "problem" here seems to be that in pthread you have to set the clock you want to use before you initialize the condition variable, but the standard says that you should be able to choose which clock you want to use for waiting after the condition variable was already created...

The boost implementation from 1.61 simply always uses CLOCK_MONOTONIC for initializing the condition variable if you have `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` defined.

Also I added another commit that actually uses SteadyTime for some timeouts and the same fix for using a steady/monotonic timer for the internal timer manager.

dirk-thomas:
@flixr Thank you for the quick answers.

> I'm not particularly fond of this kludge to include the boost headers from 1.61 if the system version is not recent enough, but I couldn't think of another way (short of providing our own condition variable implementation). Any better ideas?

If we want to target this for Lunar we need to put the headers in this repo since Xenial doesn't have a new enough boost version (they could be placed in a location though which isn't being installed / exposed to downstream packages).

The alternative would be to only target ROS M-turtle for this change (which is over a year out, May 2018, I know a long time).

@ros/ros_team @ros/ros_comm-maintainers Please let me know what you think.

flixr:
> If we want to target this for Lunar we need to put the headers in this repo since Xenial doesn't have a new enough boost version (they could be placed in a location though which isn't being installed / exposed to downstream packages).

That would work if `timer_manager.h` is not needed in any downstream packages, which I'm not sure of, since it is currently installed...
If we can also not install `timer_manager.h`, then it would indeed be nicer to not expose the new boost headers.

> The alternative would be to only target ROS M-turtle for this change (which is over a year out, May 2018, I know a long time).

I would strongly recommend to backport this to all active ROS versions as IMHO this makes it impossible to build robust real world ROS systems (where the clock might be adjusted by NTP, PTP, etc..).
Such a SteadyTimer is needed e.g. for fixing bond so that nodelets don't crash when the time jumps forward: https://github.com/ros/bond_core/issues/16

There are probably also a couple of other places where we should really use SteadyTime or a SteadyTimer instead of WallTime(r).

mikaelarguedas:
The implementation of SteadyTimer will be very useful to prevent nodelets from crashing on time jumps (c.f. https://github.com/ros/bond_core/issues/16). So it would be awesome to get this in before ROS-M.
If not (which would be totally understandable given the overhead of storing headers here + backporting) bond must implement an alternative solution to be robust to time jumps and fix the nodelet issue for all distros before ROS-M.

clients/roscpp/include/ros/timer_manager.h
```diff
@@ -560,10 +574,21 @@ void TimerManager<T, D, E>::threadFunc()
...
       else
       {
-        // On system time we can simply sleep for the rest of the wait time, since anything else requiring processing will
-        // signal the condition variable
-        int32_t remaining_time = std::max((int32_t)((sleep_end - current).toSec() * 1000.0f), 1);
-        timers_cond_.timed_wait(lock, boost::posix_time::milliseconds(remaining_time));
+        // we have to distinguish between SteadyTime and WallTime here,
+        // because boost timed_wait with duration just adds the duration to current system time
+        // this however requires boost 1.61, see: https://svn.boost.org/trac/boost/ticket/6377
+        if (typeid(T) == typeid(SteadyTime))
```
dirk-thomas: This check prevents subclassing `SteadyTime`. I am not sure if that is a valid use case but directly comparing `typeid` here looks problematic to me.

flixr: Yeah, but I don't have a better idea.... anyone else?

dirk-thomas: If you want to check the type `T` and only accept exactly `SteadyTime`. That could be done with some template meta programming since the function is already templated on `T`.

Or the `SteadyTime` class could have a method which return `true` for that class (and `false` for the other variations). It might not be possible to introduce that in an ABI compatible fashion which would prevent backporting the change.

flixr: Thanks, added a specialization for the SteadyTimer in the cpp... also fixed it for the case when the SteadyTimer is created via the node handle (somehow didn't use the backported boost version in that case before)

dirk-thomas:

dirk-thomas:
The latest CI build is using the new release of `rostime`. It is failing one of the new tests though.

flixr:
Hm... maybe a timing issue?
At least if I run these tests on indigo here they all pass...

dirk-thomas:

test/test_roscpp/test/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
```
dirk-thomas: Why is this necessary now? (Same below)

Does that affect downstream packages? Do they need to apply the same change?

flixr: As I noted in https://github.com/ros/ros_comm/pull/1014#issuecomment-285128889 this was needed to make the existing tests compile on indigo...

dirk-thomas: Is that also necessary for other ROS packages which use this modified version of `roscpp`?

flixr: No, this was only needed to make the tests compile... without that tests unrelated to this change didn't compile on my machine. If it is not needed with more recent ROS versions than indigo anymore, you can also remove it again.

dirk-thomas: Can you please try to remove them for now to see if CI will pass.

dirk-thomas:
I think this looks good. I will check with other pending tickets to coordinate a Lunar release with this soon.

flixr:
Thanks! And don't forget https://github.com/ros/bond_core/pull/18 which was the main reason for adding this ;-)

mikaelarguedas:
Given that this is merged and will be released only for Lunar, ros/bond_core#18 will need some extra work to support it

flixr:
ping, what is the holdup here? Would be great to finally get this merged...

dirk-thomas:
I just need to find the time to do ROS 1 specific work. Go through PRs, merge them, and release a new version. But currently I have to focus on other projects. I can't tell when I will get to this - hopefully within a month.

dirk-thomas:
I finally was able to get back to this. I created flixr/ros_comm#1 to update the patch to specify the minimum version of `rostime` required for this patch. Once that has been merged I will go ahead and merge this PR.

Thank you for you effort on this great improvement.

kmactavish:
At least one of the tests (`multipleSteadyTimeCallbacks`) is failing, see [this comment](https://github.com/ros/ros_comm/pull/1113#issuecomment-318729822) in #1113.

flixr:
@mikaelarguedas @dirk-thomas how about porting this back to at least kinetic and preferably even indigo?

dirk-thomas:
Backporting to Kinetic sounds good to me. I don't think this needs to go into Indigo. Indigo has been released 3.5 yrs ago, new features are not necessary to be backported to it.

mikaelarguedas:
Same thing for the corresponding PR on bond_core, I'm fine backporting it to kinetic because it touches only private functions signature. I agree with @dirk-thomas and wouldn't plan on backporting it to indigo

PR #1016, opened 2017-03-10
Fix BZip2 inclusion

vrabaud:


vrabaud:
I have a very specific setup where BZip2 is in its own folder.
The include_dir makes sense. for the _libraries, I am not sure why it worked before while bzip2 was not linked ...

dirk-thomas:
To retrigger the Lunar CI job...

dirk-thomas:
> The include_dir makes sense. for the _libraries, I am not sure why it worked before while bzip2 was not linked

That was actually my first thought too...

dirk-thomas:
Thank you for the patch.

PR #1017, opened 2017-03-16
fix rosbag::View::iterator copy assignment operator

sbarthelemy:
the compiler-generated copy assignment operator did lead to segfault and
memory leaks.

Side notes:
* could this fix be backported to indigo?
* a boost::scoped_ptr would have avoided this bug.

### DIFF
--- a/test/test_rosbag_storage/src/create_and_iterate_bag.cpp
+++ b/test/test_rosbag_storage/src/create_and_iterate_bag.cpp
@@ -10,71 +10,71 @@
 #include "boost/foreach.hpp"
 #include <gtest/gtest.h>
 
-
-TEST(rosbag_storage, create_and_iterate_bag)
+void create_test_bag(const std::string &filename)
 {
-  const char* bag_filename = "/tmp/rosbag_storage_create_and_iterate_bag.bag";
+  rosbag::Bag bag;
+  bag.open(filename, rosbag::bagmode::Write);
+
+  std_msgs::String str;
+  str.data = std::string("foo");
+
+  std_msgs::Int32 i;
+  i.data = 42;
+
+  bag.write("chatter", ros::Time::now(), str);
+  bag.write("numbers", ros::Time::now(), i);
+
+  bag.close();
+}
+
+const char* bag_filename = "/tmp/rosbag_storage_create_and_iterate_bag.bag";
+
+TEST(rosbag_storage, iterate_bag)
+{
+  rosbag::Bag bag;
+  bag.open(bag_filename, rosbag::bagmode::Read);
+
+  std::vector<std::string> topics;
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Please avoid any unrelated whitespace changes in your patch. And use consistent indentation in new code blocks.

sbarthelemy:
@dirk-thomas: thank you for the quick review.

If fixed the indentation. Please point the unrelated whitespace changes that I shall undo.

dirk-thomas:
Please compare the current diff (https://github.com/ros/ros_comm/pull/1017/files) with the diff with whitespace changes ignored (https://github.com/ros/ros_comm/pull/1017/files?w=1) to determine the difference.

sbarthelemy:
@dirk-thomas: I gave it a try, and added back 3 trailing spaces to lines that I'm still touching anyway. 

The two diffs are still different.

I'm not sure to understand what you want.

evenator:
IMHO, the diff with whitespace changes is actually a more accurate diff. The file was structurally modified. The fact that there are several lines with the same content in the same order is overridden by the fact that the functions they're part of now have completely different signatures.

test/test_rosbag_storage/src/create_and_iterate_bag.cpp
```diff
@@ -10,71 +10,104 @@
...
+  rosbag::View::const_iterator it1;
+  it1 = it0;
+  EXPECT_EQ(it0, it1);
+  EXPECT_EQ(42, it1->instantiate<std_msgs::Int32>()->data);
+  ++it1;
+  EXPECT_NE(it0, it1);
+  EXPECT_EQ(42, it0->instantiate<std_msgs::Int32>()->data);
+}
+
+TEST(rosbag_storage, create_and_iterate_bag)
```
dirk-thomas: Since the creation has been moved out can you please rename the test to just `iterate_bag`.

dirk-thomas (commented):
I misinterpreted the diff as indentation changes and missed that part of the code was refactored into the `create_test_bag` function. (which then resulted in the indentation changes).

test/test_rosbag_storage/src/create_and_iterate_bag.cpp
sbarthelemy: done.
I also took the opportunity to re-remove the 3 trailing space :).

dirk-thomas:

sbarthelemy:
thank you for merging!
I'm stuck on indigo, is there any chance to see this fix back-ported in an indigo maintenance release?

dirk-thomas:
It can certainly be backports. The latest Indigo release has just happened last week so it will be quite a while till the next one. We try to only release such core packages every few months.

PR #1023, opened 2017-03-28
Add subscriber to connection log messages.

deng02:
When roscpp_internal.connection is set to DEBUG for a node, we get  a log message when connection to a Publisher is made, bu tnot when a Subscriber connects.  This change fixes that.


mikepurvis:
LGTM. Anyone else have concerns about this change?

IanTheEngineer (approved):
Looks good to me as well. Was this supposed to be a `ROSCPP_CONN_LOG_DEBUG` all along?

mikepurvis:
These were only added recently by @deng02 in #980.

PR #1025, opened 2017-04-06
Sleep in rospy wait_for_service even if exceptions raised

AlexReimann:
If a service is not actually up, wait_for_service will throw an exception at the [contact_service call](https://github.com/ros/ros_comm/blob/247459207e20c1da109fc306e58b84d15c4107bd/clients/rospy/src/rospy/impl/tcpros_service.py#L129) and skip the sleeping.

This leads to the loop being executed without sleep, burning a lot of CPU.

This makes sure we always sleep. Sleep is good for your health.

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_service.py
+++ b/clients/rospy/src/rospy/impl/tcpros_service.py
@@ -128,7 +128,6 @@ def wait_for_service(service, timeout=None):
             try:
                 if contact_service(resolved_name, timeout_t-time.time()):
                     return
-                time.sleep(0.3)
             except KeyboardInterrupt:
                 # re-raise
                 rospy.core.logdebug("wait_for_service: received keyboard interrupt, assuming signals disabled and re-raising")
@@ -137,6 +136,7 @@ def wait_for_service(service, timeout=None):
                 if first:
                     first = False
                     rospy.core.logerr("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
+            time.sleep(0.3)
         if rospy.core.is_shutdown():
             raise ROSInterruptException("rospy shutdown")
         else:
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/impl/tcpros_service.py
```diff
@@ -127,8 +127,7 @@ def contact_service(resolved_name, timeout=10.0):
         while not rospy.core.is_shutdown() and time.time() < timeout_t:
             try:
                 if contact_service(resolved_name, timeout_t-time.time()):
-                    return
-                time.sleep(0.3)
+                    return
```
dirk-thomas: Please remove the added trailing whitespaces.

Same below.

clients/rospy/src/rospy/impl/tcpros_service.py
```diff
@@ -137,6 +136,7 @@ def contact_service(resolved_name, timeout=10.0):
                 if first:
                     first = False
                     rospy.core.logerr("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
+            time.sleep(0.3)
```
dirk-thomas, line 139: With the sleep being moved here the call is not retried immediately but during the 0.3s sleep the call is not being retried yet. That seems to be not the idea of the code. Which exception is being reused in your case and what timeout are you using?

AlexReimann: The ``rosgraph.MasterException`` from the nested ``contact_service`` function.
The problem is the immediately re-trying of the call. This spams the ros master with service lookups without any sleep.

For me the original logic seems:
* If can contact the service, return immediately. Else sleep.
* Ohh, we should add some exception checking somewhere here

dirk-thomas:
Thanks for the patch.

PR #1026, opened 2017-04-06
Enable service unavailable log in rospy wait_for_service

AlexReimann:
There is no log published if the service is down and we are waiting for it.

Enables (the code was already there) logging a message.

I changed the error message to a info in case we have a timeout and a warning in case we have no timeout.
In case of no timeout this will be the only message telling you that you are stuck in an endless loop in case the service does not show up. That's why I thought a warning would be better than just a log.


clients/rospy/src/rospy/impl/tcpros_service.py
```diff
@@ -136,7 +136,7 @@ def contact_service(resolved_name, timeout=10.0):
             except: # service not actually up
                 if first:
                     first = False
-                    rospy.core.logerr("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
+                    rospy.core.loginfo("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
```
dirk-thomas: This message would be printed every 0.3 s. I think that is too verbose. Please consider using a throttled log call (http://wiki.ros.org/rospy/Overview/Logging#Logging_Periodically). Maybe once every 10s is ok?

clients/rospy/src/rospy/impl/tcpros_service.py
```diff
@@ -154,7 +154,7 @@ def contact_service(resolved_name, timeout=10.0):
             except: # service not actually up
                 if first:
                     first = False
-                    rospy.core.logerr("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
+                    rospy.core.logwarn("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
```
dirk-thomas: When no timeout is specified the default value is 10s. So the message would get printed repeatedly - not only once. Therefore I would argue both log levels should be the same. Not sure if `info` or `warn` make more sense though.

dirk-thomas (commented):
Please target the latest development branch (currently `lunar-devel`) for your patch. You should be able to edit the existing PR and change the target branch name.

clients/rospy/src/rospy/impl/tcpros_service.py
AlexReimann: It gets printed only once because of the ``first`` flag.

dirk-thomas: Does the user get any feedback once the service has become available?

AlexReimann: No, can add that.

AlexReimann: Anyways, I two ways for this (and also the [other comment above](https://github.com/ros/ros_comm/pull/1026#discussion_r112511092)):

1. Publish a warning once when we failed to contact service and then publish a info when it becomes available
2. Publish a throttled info continuously  (and maybe publish a info when it becomes available?)

1\. is easy to miss, but 2. might be a bit spamy. I personally prefer spamy over missing stuff, so I would go for two (with the "now available" info)

clients/rospy/src/rospy/impl/tcpros_service.py
```diff
@@ -121,21 +121,22 @@ def contact_service(resolved_name, timeout=10.0):
...
             except KeyboardInterrupt:
                 # re-raise
                 rospy.core.logdebug("wait_for_service: received keyboard interrupt, assuming signals disabled and re-raising")
                 raise 
             except: # service not actually up
-                if first:
-                    first = False
-                    rospy.core.logerr("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
+                contact_failed = True
+                rospy.core.loginfo("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
```
dirk-thomas: Should this call `logwarn` instead?

clients/rospy/src/rospy/impl/tcpros_service.py
```diff
@@ -145,15 +146,16 @@ def contact_service(resolved_name, timeout=10.0):
...
             except KeyboardInterrupt:
                 # re-raise
                 rospy.core.logdebug("wait_for_service: received keyboard interrupt, assuming signals disabled and re-raising")
                 raise 
             except: # service not actually up
-                if first:
-                    first = False
-                    rospy.core.logerr("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
+                contact_failed = True
+                rospy.core.logwarn("wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
```
dirk-thomas: With #1091 being merged this could be updated to call `rospy.core.logwarn_throttle(10, "...")` to reduce the verbosity of this line.

Same for the other warning above.

dirk-thomas (commented):
Sorry for the late response. The good news is that the throttled logging is now available and this patch could benefit from it.

dirk-thomas:

mathias-luedtke:
This patch introduced a bug:
```
  File "/opt/ros/lunar/lib/python2.7/dist-packages/rospy/core.py", line 520, in signal_shutdown
    h()
  File "/root/catkin_ws/install/lib/controller_manager/spawner", line 65, in shutdown
    unload_controller.wait_for_service(timeout=shutdown_timeout)
  File "/opt/ros/lunar/lib/python2.7/dist-packages/rospy/impl/tcpros_service.py", line 420, in wait_for_service
    wait_for_service(self.resolved_name, timeout=timeout)
  File "/opt/ros/lunar/lib/python2.7/dist-packages/rospy/impl/tcpros_service.py", line 139, in wait_for_service
    rospy.core.logwarn_throttle(10, "wait_for_service(%s): failed to contact [%s], will keep trying"%(resolved_name, uri))
NameError: global name 'uri' is not defined
```
(https://travis-ci.org/ros-controls/ros_control/jobs/295104386#L4754)

Is `uri` accessible from the outer function at all?

dirk-thomas:
@ipa-mdl Thank you for reporting this. Can you please confirm that #1213 fixes the problem.

PR #1027, opened 2017-04-06
Removed "ROS time moved backwards" log message

peci1:
The log message doesn't make sense IMO, because there is still the exception being thrown.

So the user can either catch it (and then he knows best whether it is an error or not), or he'll not catch it, and then he'll see the message printed in the stack trace.

This is a followup of comment https://github.com/ros/ros_comm/pull/492#discussion_r109936788 .

My use-case: I have a ROS node running alongside Gazebo, and resetting Gazebo always prints this error to the log, even if it is no error for me, since I expect the time to be restarted.


dirk-thomas:
Thank you for creating the patch.

Issue #1030, opened 2017-04-12
roslaunch fails if ROS_LOG_DIR is set and ~/.ros was not yet created


TimSchneider42:
Hi,

if ROS_LOG_DIR is set to something other than "\~/.ros", this directory is not created. This causes the subprocess.Popen call in nodeprocess.py:298 to fail if its cwd is set to "\~/.ros". That is the case for example during the automatic creation of rosmaster.

Steps to reproduce:
```
rm -r ~/.ros
export ROS_LOG_DIR=~/log
mkdir $ROS_LOG_DIR
roslaunch <any launchfile>
```

I used this launchfile to produce the error:
```
<launch>
  <param name="robot_description"
    value="&lt;robot&gt;&lt;/robot&gt;" />

  <node name="robot_state_publisher" pkg="robot_state_publisher"
    type="robot_state_publisher" />
</launch>
```

Output:
```
$ roslaunch test.launch
... logging to /home/*****/log/4695aafa-1f12-11e7-9ad6-64006a5db393/roslaunch-state-ub-ph-23654.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

started roslaunch server http://*********/

SUMMARY
========

PARAMETERS
 * /robot_description: <robot></robot>
 * /rosdistro: kinetic
 * /rosversion: 1.12.7

NODES
  /
    robot_state_publisher (robot_state_publisher/robot_state_publisher)

ROS_MASTER_URI=http://localhost:11311

core service [/rosout] found
Roslaunch got a 'No such file or directory' error while attempting to run:

/opt/ros/kinetic/lib/robot_state_publisher/robot_state_publisher __name:=robot_state_publisher __log:=/home/*****/log/4695aafa-1f12-11e7-9ad6-64006a5db393/robot_state_publisher-1.log

Please make sure that all the executables in this command exist and have
executable permission. This is often caused by a bad launch-prefix.
The traceback for the exception was written to the log file
[robot_state_publisher-1] killing on exit

```

Homan

dirk-thomas:
I don't think it is necessary to update ROS to create it on demand.  The simple solution to this is to make sure the directory you are pointing to in the environment variable exists.

Please feel free to create a PR to support this or contribute to the [wiki](http://wiki.ros.org/ROS/EnvironmentVariables) clarifying that the directory needs to exist

TimSchneider42:
Hi Dirk,

[...]

PR #1031, opened 2017-04-14
ensure cwd exists

dirk-thomas:
This can happen when `~/.ros` doesn't exist and a custom `ROS_LOG_DIR` is being used. See ros/ros_comm#1030.


wjwwood (approved):
lgtm

PR #1033, opened 2017-04-19
now allowing python_logging.yaml

asmodehn:
Small PR to enable support for yaml configuration for python logging.
Do not merge yet, this is still in testing...

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -40,6 +40,8 @@ import time
 import logging
 import logging.config
 
+import yaml
+
 import rospkg
 from rospkg.environment import ROS_LOG_DIR
 
@@ -103,16 +105,17 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
     else:
         # search for logging config file in /etc/.  If it's not there,
         # look for it package-relative.
-        fname = 'python_logging.conf'
-        rosgraph_d = rospkg.RosPack().get_path('rosgraph')
-        for f in [os.path.join(rospkg.get_ros_home(), 'config', fname),
-                  '/etc/ros/%s'%(fname),
-                  os.path.join(rosgraph_d, 'conf', fname)]:
-            if os.path.isfile(f):
-                config_file = f
+        config_file = None
+        for fname in ['python_logging.conf', 'python_logging.yaml']:
+            rosgraph_d = rospkg.RosPack().get_path('rosgraph')
+            for f in [os.path.join(rospkg.get_ros_home(), 'config', fname),
+                      '/etc/ros/%s'%(fname),
+                      os.path.join(rosgraph_d, 'conf', fname)]:
+                if os.path.isfile(f):
+                    config_file = f
+                    break
+            if config_file is not None:
                 break
-        else:
-            config_file = None
 
     if config_file is None or not os.path.isfile(config_file):
         # logging is considered soft-fail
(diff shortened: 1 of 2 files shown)


mikepurvis:
~~I'm not a major user of rospy, but the log4cxx logger in roscpp is [configured via the `ROSCONSOLE_CONFIG_FILE` environment variable](http://wiki.ros.org/ROS/EnvironmentVariables). Should there be something similar to allow setting the config location for Python?~~ Wait a minute, nevermind— there totally is `ROS_PYTHON_LOG_CONFIG_FILE`. Is this change just allowing yaml in addition to conf as the format?

FYI @paulbovbel

asmodehn:
@mikepurvis yes just that. I figured it was simple enough to add and would allow more detailed configuration using [recent python 2.7 logging API](https://docs.python.org/2/library/logging.config.html#logging.config.dictConfig). Plus we are already using yaml everywhere for configuration in our ROS software so why not for logging as well.

But AFAIK there is no test for this behaviour it and I haven't used it yet, so this patch might need a few days/weeks to mature...

asmodehn:
I tested part of that new code (by setting `ROS_PYTHON_LOG_CONFIG_FILE` to `python_logging.yaml` in my environment) with this file : 
```yaml
---
version: 1
disable_existing_loggers: False
formatters:
    default:
        format: "[%(name)s][%(levelname)s] %(asctime)s: %(message)s"

handlers:

    rosconsole:
        class: rosgraph.roslogging.RosStreamHandler
        level: DEBUG
        formatter: default
        colorize: True

    roslog:
        class: logging.handlers.RotatingFileHandler
        level: DEBUG
        formatter: default
        filename: os.environ['ROS_LOG_FILENAME']
        maxBytes: 50000000
        backupCount: 4

loggers:

    root:
        level: INFO
        handlers: [roslog]

    rosout:
        level: INFO
        handlers: [rosconsole]
        propagate: yes
        qualname: rosout

```

I get one error that I didn't have with python_logging.conf : 
```
No handlers could be found for logger "roslaunch"
```
As well as for my other custom loggers in my code that didn't add `NullHandler` as explained here : https://docs.python.org/2/howto/logging.html#library-config

So I understand it means that no logger is configured for `roslaunch` in `python_logging.conf`, but I don't know what is the behavior in that case, where and how it will send log messages...

We should add a `NullHandler` into `roslaunch`.
But do we want to add a specific config for roslaunch logger ?

asmodehn:
Or maybe we do not want a `NullHandler` for roslaunch, since it is not a library after all, and that error is expected.

And we should define (for example) : 
- a default console handler for roslaunch in code
- an extra file handler in configuration file (if needed, or even just as an example).

Any idea/advice for this ?

asmodehn:
I solved roslaunch warning on my side by adding this to the config : 
```
    roslaunch:
        level: WARNING
        handlers: [rosconsole]
        propagate: no
        qualname: roslaunch
```

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -122,10 +125,18 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
...
-    logging.config.fileConfig(config_file, disable_existing_loggers=False)
+    if config_file.endswith('.conf'):
+        # #3625: disabling_existing_loggers=False
+        logging.config.fileConfig(config_file, disable_existing_loggers=False)
+    elif config_file.endswith('.yaml'):
+        with open(config_file) as f:
+            dict_conf = yaml.load(f)
+            dict_conf.setdefault('version', 1)
+            logging.config.dictConfig(dict_conf)
+
```
dirk-thomas: Currently this block doesn't have an `else` part. Also if the user was using a different filename before it might not work with the modified code anymore. Therefore I would suggest the logic to be updated to:

```
if yaml:
    # do yaml stuff
else:
    # do stuff as before
```

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -40,6 +40,8 @@
 import logging
 import logging.config
 
+import yaml
```
dirk-thomas, line 43: The package currently doesn't depend on yaml. Please add the additional dependency to the package manifest.

dirk-thomas (commented):
Please retarget the latest development branch (lunar-devel).

tools/rosgraph/src/rosgraph/roslogging.py
asmodehn: I see it there : https://github.com/ros/ros_comm/blob/lunar-devel/tools/roslaunch/package.xml#L25

asmodehn: Ah I get it now rosgraph, not roslaunch. my bad...

PR #1041, opened 2017-04-30
[rospy] Add rospy.logXXX_once

wkentaro:
Following https://github.com/ros/ros_comm/pull/812


dirk-thomas (commented):
I don't think the `once` logic should be built on top of the `throttle` logic. E.g. there is no call `Time.now` in order to implement `once`.

dirk-thomas:
Thank you. Can you please add the new logging functions to the wiki page mentioning that they are available as of ROS Lunar and post the link to the diff here.

wkentaro:
Updated: http://wiki.ros.org/rospy/Overview/Logging#Logging_Once

dirk-thomas:
Thanks!

PR #1040, opened 2017-04-28
logging "both"

asmodehn:
Improved PR #551 for issue #548

PR #1042, opened 2017-05-02
roslaunch node output "both"

asmodehn:
#1040 Rebased on lunar-devel

### DIFF
--- a/tools/roslaunch/src/roslaunch/nodeprocess.py
+++ b/tools/roslaunch/src/roslaunch/nodeprocess.py
@@ -51,7 +51,10 @@ from roslaunch.pmon import Process, FatalProcessLaunch
 
 from rosmaster.master_api import NUM_WORKERS
 
+import threading
+
 import logging
+from logging import handlers
 _logger = logging.getLogger("roslaunch")
 
 _TIMEOUT_SIGINT  = 15.0 #seconds
@@ -145,7 +148,25 @@ def create_node_process(run_id, node, master_uri):
     _logger.debug('process[%s]: returning LocalProcess wrapper')
     return LocalProcess(run_id, node.package, name, args, env, log_output, \
             respawn=node.respawn, respawn_delay=node.respawn_delay, \
-            required=node.required, cwd=node.cwd)
+            required=node.required, cwd=node.cwd, max_logfile_size=node.max_logfile_size,\
+            logfile_count=node.logfile_count)
+
+
+def stream_reader(stream, level, logger, popen):
+    while popen.poll() is None:
+        line = stream.readline()
+        if level == logging.INFO:
+            logger.info('%s', line.strip())
+        elif level == logging.ERROR:
+            logger.error('%s', line.strip())
+    #logger.handlers[0].flush()
+    #logger.handlers[0].close()
+    # it may be some data into th stream at the end of the process :
+    #line = stream.read()
+    #if level == logging.INFO:
+    #    logger.info('%s', line.strip())
+    #elif level == logging.ERROR:
+    #     logger.error('%s', line.strip())
 
 
 class LocalProcess(Process):
@@ -155,7 +176,7 @@ class LocalProcess(Process):
     
     def __init__(self, run_id, package, name, args, env, log_output,
             respawn=False, respawn_delay=0.0, required=False, cwd=None,
-            is_node=True):
+            is_node=True, max_logfile_size=None, logfile_count=None):
         """
         @param run_id: unique run ID for this roslaunch. Used to
           generate log directory location. run_id may be None if this
@@ -179,6 +200,10 @@ class LocalProcess(Process):
         @type  cwd: str
         @param is_node: (optional) if True, process is ROS node and accepts ROS node command-line arguments. Default: True
         @type  is_node: False
+        @param max_logfile_size: (optional) Maximum Size (in bytes) of the node's logfile. 0 mean unlimitted (default value),
+        this value must >= 0``int``
+        @param logfile_count: (optional) If max_logfile_size > 0, and logfile_count > 0, the system will save old log
+        files by appending the extensions .1, .2 etc... This is an optional parameter, default value is 2.
         """    
         super(LocalProcess, self).__init__(package, name, args, env,
                 respawn, respawn_delay, required)
@@ -191,6 +216,10 @@ class LocalProcess(Process):
         self.log_dir = None
         self.pid = -1
         self.is_node = is_node
+        self.max_logfile_size = max_logfile_size
+        self.logfile_count = logfile_count
+        self.logThreadError = None
+        self.logThreadInfo = None
 
     # NOTE: in the future, info() is going to have to be sufficient for relaunching a process
     def get_info(self):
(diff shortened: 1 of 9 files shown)


asmodehn:
From the buildfarm log, all tests are passing , as far as I can tell...

asmodehn:
Found the failing test, did a simple fix.

dirk-thomas:
Still failing two tests.

tools/roslaunch/src/roslaunch/nodeprocess.py
```diff
@@ -94,8 +97,7 @@ def create_master_process(run_id, type_, ros_root, port, num_workers=NUM_WORKERS
         raise RLException("unknown master typ_: %s"%type_)
 
     _logger.info("process[master]: launching with args [%s]"%args)
-    log_output = False
-    return LocalProcess(run_id, package, 'master', args, os.environ, log_output, None)
+    return LocalProcess(run_id, package, 'master', args, os.environ, log_output=False, screen_output=True)
```
dirk-thomas, line 100: `screen_output=True` seems to be the default and can therefore be omitted.

tools/roslaunch/src/roslaunch/nodeprocess.py
```diff
@@ -212,7 +241,7 @@ def _configure_logging(self):
         @return: stdout log file name, stderr log file
         name. Values are None if stdout/stderr are not logged.
         @rtype: str, str
-        """    
+        """
```
dirk-thomas, line 244: Please avoid any unrelated whitespace changes.

tools/roslaunch/src/roslaunch/nodeprocess.py
```diff
@@ -272,15 +331,25 @@ def start(self):
...
             try:
-                logfileout, logfileerr = self._configure_logging()
+                process_logger = self._configure_logging() # Must always return an object even if we log only on screen.
             except Exception as e:
                 _logger.error(traceback.format_exc())
                 printerrlog("[%s] ERROR: unable to configure logging [%s]"%(self.name, str(e)))
                 # it's not safe to inherit from this process as
                 # rostest changes stdout to a StringIO, which is not a
                 # proper file.
+                #logfileout, logfileerr = subprocess.PIPE, subprocess.PIPE
```
dirk-thomas, line 343: Please remove obsolete code rather than commenting it out.

dirk-thomas (commented):
I have one concern for this patch. If I understand it correctly even if the user doesn't need logging to files the output to the console is still being handled using the new thread machinery. That will introduce the potential of breaking the order of the output between `stdout` and `stderr`. Do you agree that this is currently the case with the patch?

I think it is crucial to keep the order of the output of these two streams unchanged in that case.

asmodehn:
From my perspective, I do not think one should be concerned by different order of message between different streams (I have no idea if the underlying OS / kernel / terminal guarantee an order between different streams, even when coming from the same process, or if it just happens to be in a specific order because of system implementation choices).
But in any case, introducing threads here adds some entropy, and I would also like to avoid it.

I also have other reasons : In my personal, albeit limited, experience, a ROS system very easily scale to have too many threads (that ends up being much more than CPUs available), so I'd rather not add 2 more threads by node, only for logging.

That being said, I also feel the need of being able to configure logging output (at least from python) much more than what is possible at the moment, I am just not sure if this is the best way...

I still need to understand all details of this patch. I ll have a deeper look into it to see if I can come up with a way that doesn't require threads...

asmodehn:
That thread-less experiment being unsuccessful, I now think this is the best way to achieve the goal with python2.

If we want to keep the order of messages between stdout and stderr, maybe that means that we want only one stream, and not two, after redirecting stderr to stdout. This would maintain the previous ordering of messages, but maybe it would break some existing feature ?

IMHO I think our (mid-term) best bet here is to : 
- Have only one stream, handled by one thread for processes launched by roslaunch (merging stderr in stdout - which resembles what the terminal usually does anyway - see the nice diagram at http://ptyprocess.readthedocs.io/en/latest/ ).
- eventually start to use two main (ROS-wide) loggers (rosout and roserr) with corresponding configuration to allow the user to customize where the INFO level for the whole ROS system goes, and where the ERROR level for the whole ROS system goes (instead of the currently hard coded setup documented in the table here http://wiki.ros.org/roscpp/Overview/Logging#Output, barely customizable via launch file)

This way, the logging behavior for each node is deterministic and simpler to manage from roslaunch, and the whole multiprocess system has two outputs, similar to what people are used to : normal output here and errors there. We can then eventually review our configuration strategy (less custom things in launch files and more defacto standard things in usual logging config files), to make it simpler and friendlier for everyone.

I believe this would allow processes to communicate via pipes/pseudo-tty/loggers/buffered streams (1 IN 1 OUT), while providing simple but standard logging configuration of logging for the quick'n'dirty use case. And if a user desires more power out of logging, named loggers for each node can provide additional configuration possibilities.

This is too heavy for this PR, but let me know what you think and if you see some parts that are doable and useful along these lines (merging stderr into stdout for all nodes ?) , I might be able to spend a bit of time on it.

dirk-thomas:
Imo the goal here should be to aim for "you don't pay for it if you don't use it". What I mean with that is:

* If a user doesn't need the different sinks for logging messages he shouldn't neither need the additional threads nor face the downside of potentially even more shuffled order of messages.

In the other direction:

* If a user want's to keep `stdout` and `stderr` in separate streams there should be nothing which forces the merge of these streams.

A solution which forces certain "cons" onto the user which doesn't even want the correlated "pros" is not a good implementation in my opinion. Does that make any sense?

asmodehn:
It makes sense and I also agree, as long as it is possible...
I'm thinking there is something more fundamental going on here, with ordering, logging as side-effects, and having multiple streams between distributed processes.

I believe it is not possible (by definition dare I say) to guarantee the order of messages arriving in a process context from multiple (unsynchronized) streams, even when coming from the same process, without adding extra cost (synchronizing algorithm like a [vector clock](https://en.wikipedia.org/wiki/Vector_clock) or so...). I would say holding this expectation could be a symptom of [the fallacies of distributed computing](https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing) number 2, however that might depend on OS implementation...
So, except for the ordering, I agree with the other points.

To avoid imposing constraints to the user, then each process should be responsible for its own logging, as usual, and roslaunch would have nothing to do with it, except redirect the subprocess output to the terminal when launching it. Messages on the terminal then would be disordered.
Fixing this means guaranteeing some kind of ordering via some algorithm, which would induce some costs on all processes, including roslaunch process which would need to proactively reestablish some order before outputting on the terminal, but I don't think it can be avoided...

In any case python2 is quite ill-equipped to deal with these kind of problems (asyncio would be very helpful), which is why this implementation currently uses threads. It would be possible to implement something similar to corountines but it would likely make the code more complex than it already is...

I am fine with declaring this PR as "not good" and closing. I needed this feature on indigo, but recent PR merges in lunar (named loggers especially) might make the logging framework usable enough (at least for my usecase - but I still have to check)

Further improvements to be able to customize terminal log output and sinks from roslaunch would likely require one or more of : 
- asyncio (trollius for py2)
- python logging library extended to deal with distributed communicating processes (non trivial - [bigraphs](https://en.wikipedia.org/wiki/Bigraph) anyone ?)
- design changes for the ROS logging framework (any combination of : "one outstream per process", "roslaunch not handling any subprocess logging", "roslaunch reordering messages for the terminal", etc. )
which are quite heavy changes, and might not be feasible for ROS1...

dirk-thomas:

PR #1043, opened 2017-05-02
[rospy] Write log for class method with class name for rospy

asmodehn:
#877 Rebased on lunar-devel


asmodehn:
Not sure why the test failed, no tests are marked as failed, as far as I can see.. any hint ?

dirk-thomas:
The referenced job has one failing test: http://build.ros.org/job/Lpr__ros_comm__ubuntu_xenial_amd64/94/testReport/junit/test/test_roslogging_user_logger/test_roslogging_user_logger/

asmodehn:
Ah thanks, I was digging through the log and didn't pay attention to the main build page...

I can confirm the test behaves as expected, when launched alone, on my local Xenial machine.

It seems that the problem in the test might come from the logging module from the python library.
In Manager.getLogger(), there is this code : 
```python
            if name in self.loggerDict:
                rv = self.loggerDict[name]
                if isinstance(rv, PlaceHolder):
                    ph = rv
                    rv = (self.loggerClass or _loggerClass)(name)
                    rv.manager = self
                    self.loggerDict[name] = rv
                    self._fixupChildren(ph, rv)
                    self._fixupParents(rv)
            else:
                rv = (self.loggerClass or _loggerClass)(name)
                rv.manager = self
                self.loggerDict[name] = rv
                self._fixupParents(rv)
```

So I suspect that when launched with other tests, our current test will fail because the `rosout` logger is already set and the custom logger class is not used.

Multiple solutions : 
1. Instantiate a Manager in the test, to be used a separate from the default (already used) logging manager. and somehow use that one from 'rosout' logger...
1. force nose to reinitialize its python interpreter for each tests (or run in isolation somehow)
1. launch nose tests one by one from CMakeLists.txt (instead on the full test directory)
1. make a special log for this test to force the custom logger class to be used (using the logging feature that will be wrapped by rospy in https://github.com/ros/ros_comm/pull/948 )

The latter seems the cleaner to me so I ll work on that.

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -39,12 +39,31 @@
...
+        """
+        frame, _, lineno, _, _, _ = inspect.stack()[3]
+        file_name = inspect.getabsfile(frame)
+        func_name = frame.f_code.co_name
+        try:
+            class_name = frame.f_locals['self'].__class__.__name__
+            func_name = '%s.%s' % (class_name, func_name)
+        except KeyError:
+            pass
+        return file_name, lineno, func_name
```
dirk-thomas: I think it would be good to call the implementation of the base class here and then only "enhance" the function name with additional information.

dirk-thomas:

dirk-thomas:

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -39,12 +39,43 @@
...
 import rospkg
 from rospkg.environment import ROS_LOG_DIR
 
 class LoggingException(Exception): pass
 
+class RospyLogger(logging.getLoggerClass()):
+    def findCaller(self):
+        """
+        Find the stack frame of the caller so that we can note the source
+        file name, line number, and function name with class name if possible.
```
dirk-thomas, line 53: While this is a good description of the function itself would it be better to document what this specific variation does (extending the function name with the class name if applicable)?

asmodehn:
I believe it is already written there ?

--
AlexV

wkentaro:
@asmodehn Thank you for addressing this feature. :+1:

dirk-thomas:
I am sorry for the late response. This looks good. Merging...

dirk-thomas:
@asmodehn Please extend the tests from this PR to also over other logging functions beside `loginfo`, especially named and throttled loggers. Currently they are partially broken in the latest state, see #1141.

asmodehn:
@dirk-thomas As far as I understand the code : 
- rospy depend on rosgraph
- rosgraph uses raw python logging (and test the logging format)
- rospy uses rosgraph logging, and extends it with throttled and named loggers.

So I believe we don't want to extend rosgraph tests to tests throttled and named loggers... We probably want to extend rospy logging tests to verify the logging format in the throttled and named cases instead ?
I ll start working on something like that, see if I get somewhere...

asmodehn:
How about something like that : https://github.com/ros/ros_comm/pull/1144 ?

PR #1044, opened 2017-05-02
Abort topic lookup on connection refused

guillaumeautran:
In a multimaster environment where a topic has multiple publishers,
when a node drops out abruptly (host is shutdown), a single subscriber update on
that topic will cause multiple threads to be created (one for each host) in order to
resolve the topic location. This cause a thread leak as host which are turned off
will not respond and when they come back online, the xmlrpc URI is changed causing a
connection refused error at the socket layer.

This fix catches the connection refused error and terminate the thread with the understanding
that if the connection is refused, the rosnode cannot be reached now or never. This effectively
prevents thread leak.

Note: if the remote host where the rosnode is thought to be never comes back up,
then the thread will still be leaked as the exception received is a host unreachable type.
This is intentional to avoid abruptly terminating the thread in case of a temporary DNS failure.

### DIFF
--- a/clients/rospy/src/rospy/impl/masterslave.py
+++ b/clients/rospy/src/rospy/impl/masterslave.py
@@ -59,6 +59,7 @@ import socket
 import threading
 import traceback
 import time
+import errno
 
 try:
     #py3k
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/impl/masterslave.py
```diff
@@ -441,13 +442,19 @@ def _connect_topic(self, topic, pub_uri):
...
             try:
                 code, msg, result = \
                       xmlrpcapi(pub_uri).requestTopic(caller_id, topic, protocols)
                 success = True
             except Exception as e:
-                if not is_shutdown():
+                if getattr(e, 'errno', None) == errno.ECONNREFUSED:
+                    code = errno.ECONNREFUSED
+                    msg = str(e)
+                    break
```
mikepurvis: The early return down below only triggers when `code <= 0`, but you're assigning it a positive value. What is the expected behaviour here, when we break out of the while loop?

guillaumeautran: Good catch, the intention is to return at the early return. But the errno will be positive. Let me fix that.

guillaumeautran:
Ping for review...

mikepurvis:
@dirk-thomas We've been running this on our bots for the week; would like to get it merged if there's no objection.

dirk-thomas:
Sounds good to me. Thank you both for your effort on this.

Issue #1047, opened 2017-05-08
Rosbag API is not compatible with Python 3


nstiurca:
Due to changes in handling of files with regards to `str` vs `bytes` between Python 2 and 3, the `rosbag` API doesn't work in Python 3. I have a very simple, elegant fix which I will set up a PR for shortly, and I hope it can be included in Kinetic.

For reference, my system is Mint 18.1 (ie Ubuntu 16.04) running Kinetic, and we definitely need to use Python 3 for most of what we do.

ehannigan:
I apologize if this is the wrong place to ask this question, but was this fix added to kinetic? Both merges show this only being added to ros:Lunar. It would be really useful to have in kinetic. 

If so, how do I go about getting the update? 

Thank you,
E

dirk-thomas:
@ehannigan All fixes are being committed to the branch of the latest ROS distro first. Once they have been released into the latest ROS distro (which isn't the case for #1150 yet) and that hasn't resulted in any regressions the patch is being considered for backporting into older ROS distros.

You can either wait for this to happen and become available from Debian packages in Kinetic or you can build the specific package from source and apply the same patch locally.

ehannigan:
@dirk-thomas 
Thanks for the quick response! I will look into installing the patch locally.

PR #1048, opened 2017-05-08
Fix rosbag API for Python 3

nstiurca:
Closes ros/ros_comm#1047.

It would be great to get this backported to Kinetic as well.

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -1580,6 +1580,7 @@ def _read_uint32(f): return _unpack_uint32(f.read(4))
 def _read_uint64(f): return _unpack_uint64(f.read(8))
 def _read_time  (f): return _unpack_time  (f.read(8))
 
+def _unpack_str(v):    return v.decode()
 def _unpack_uint8(v):  return struct.unpack('<B', v)[0]
 def _unpack_uint32(v): return struct.unpack('<L', v)[0]
 def _unpack_uint64(v): return struct.unpack('<Q', v)[0]
@@ -1628,8 +1629,7 @@ def _read_field(header, field, unpack_fn):
     
     return value
 
-def _read_str_field   (header, field):
-    return _read_field(header, field, lambda v: v)
+def _read_str_field   (header, field): return _read_field(header, field, _unpack_str)
 def _read_uint8_field (header, field): return _read_field(header, field, _unpack_uint8)
 def _read_uint32_field(header, field): return _read_field(header, field, _unpack_uint32)
 def _read_uint64_field(header, field): return _read_field(header, field, _unpack_uint64)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please provide a reproducible example which failed before / works with the patch.

nstiurca:
```
#!/usr/bin/env python3
import rosbag

with rosbag.Bag('any-bagfile.bag', 'r') as bag:
    for topic, msg, t in bag.read_messages():
        pass
```

crashes with

```
  File "./foo.py", line 5, in <module>
    for topic, msg, t in bag.read_messages():
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rosbag/bag.py", line 2331, in read_messages
    yield self.seek_and_read_message_data_record((entry.chunk_pos, entry.offset), raw)
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rosbag/bag.py", line 2440, in seek_and_read_message_data_record
    raise ROSBagException('unsupported compression type: %s' % chunk_header.compression)
rosbag.bag.ROSBagException: unsupported compression type: b'none'
```

------

Or just run the test suite with Python 3.

dirk-thomas:
Currently some of the test are failing. I will try to rerun them (in case they are flaky). @ros-pull-request-builder retest this please

If that doesn't resolve them please look into why they started to fail.

nstiurca:
Hmm, they didn't fail locally, but I only tested against Kinetic
admittedly. I'm not sure how easy it is for me to test against anything
else.

-- Nicu Știurcă (sent from my phone)

nstiurca:
This patch only addresses reading bag files. The test code first creates a bag file, and I found today that there might be some Python 3 related issues there as well. I will update on this as I can.

Although I presume the failed tests above were run using Python 2, so I'm still not sure about that.

dirk-thomas:
> Although I presume the failed tests above were run using Python 2

Yes, the CI builds only use Python 2.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -1580,6 +1580,7 @@ def _read_uint32(f): return _unpack_uint32(f.read(4))
 def _read_uint64(f): return _unpack_uint64(f.read(8))
 def _read_time  (f): return _unpack_time  (f.read(8))
 
+def _unpack_str(v):    return v.decode()
```
dirk-thomas: The function name doesn't make much sense in the case of Python 3. Since you invoke `decode` on `v` it can't be a `str` but must be `bytes`?

nstiurca:
So it looks like the build is still failing, but if I'm reading the logs right it's an issue with rosdep having timed out during the build rather than anything to do with this patch. Is there a way to re-run the build?

dirk-thomas:
I accidentally assumed that #1150 would address the remaining comment on this PR but relied on other parts of the patch. Now I see that the new PR is "complete" on its own. Therefore I have merged #1150 and will close this as a duplicate. Anyway thanks for working on this.

PR #1050, opened 2017-05-09
Fix bug in transport_tcp

bxwllzz:
Fix about [Subscribing & Publishing from ROS nodes does not work · Issue #1391 · Microsoft/BashOnWindows](https://github.com/Microsoft/BashOnWindows/issues/1391)

Line 316 in [ros_comm/roscpp/src/libros/transport/transport_tcp.cpp](http://docs.ros.org/kinetic/api/roscpp/html/transport__tcp_8cpp_source.html)

```
  312   int ret = ::connect(sock_, (sockaddr*) &sas, sas_len);
  313   // windows might need some time to sleep (input from service robotics hack) add this if testing proves it is necessary.
  314   ROS_ASSERT((flags_ & SYNCHRONOUS) || ret != 0);
  315   if (((flags_ & SYNCHRONOUS) && ret != 0) || // synchronous, connect() should return 0
  316       (!(flags_ & SYNCHRONOUS) && last_socket_error() != ROS_SOCKETS_ASYNCHRONOUS_CONNECT_RETURN)) // asynchronous, connect() should return -1 and WSAGetLastError()=WSAEWOULDBLOCK/errno=EINPROGRESS
  317   {
  318     ROSCPP_CONN_LOG_DEBUG("Connect to tcpros publisher [%s:%d] failed with error [%d, %s]", host.c_str(), port, ret, last_socket_error_string());
  319     close();
  320 
  321     return false;
  322   }
```
It assumes that the `connect` method of non-blocking scoket should return -1 and `last_socket_error()` should return `ROS_SOCKETS_ASYNCHRONOUS_CONNECT_RETURN`(=`EINPROGRESS`). 
But a non-blocking `connect` can return 0 when TCP connection to 127.0.0.1 (localhost).
[http://stackoverflow.com/questions/14027326/can-connect-return-0-with-non-blocing-socket](http://stackoverflow.com/questions/14027326/can-connect-return-0-with-non-blocing-socket)

In Bash on Ubuntu on Windows, when connecting to 127.0.1.1 non-blocking, it returns 0 which raise an error 
 `[WARN] [1479567965.827686]: Inbound TCP/IP connection failed: connection from sender terminated before handshake header received. 0 bytes were received. Please check sender for additional details.`

After patch, topic tests successfully in [ROS/Tutorials/WritingPublisherSubscriber(python)](http://wiki.ros.org/ROS/Tutorials/WritingPublisherSubscriber%28python%29),  [ROS/Tutorials/ExaminingPublisherSubscriber](http://wiki.ros.org/ROS/Tutorials/ExaminingPublisherSubscriber) and [turtlesim](http://wiki.ros.org/ROS/Tutorials/UnderstandingTopics)

### DIFF
--- a/clients/roscpp/src/libros/transport/transport_tcp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_tcp.cpp
@@ -311,9 +311,10 @@ bool TransportTCP::connect(const std::string& host, int port)
 
   int ret = ::connect(sock_, (sockaddr*) &sas, sas_len);
   // windows might need some time to sleep (input from service robotics hack) add this if testing proves it is necessary.
-  ROS_ASSERT((flags_ & SYNCHRONOUS) || ret != 0);
+  // ROS_ASSERT((flags_ & SYNCHRONOUS) || ret != 0);
   if (((flags_ & SYNCHRONOUS) && ret != 0) || // synchronous, connect() should return 0
-      (!(flags_ & SYNCHRONOUS) && last_socket_error() != ROS_SOCKETS_ASYNCHRONOUS_CONNECT_RETURN)) // asynchronous, connect() should return -1 and WSAGetLastError()=WSAEWOULDBLOCK/errno=EINPROGRESS
+      (!(flags_ & SYNCHRONOUS) && // asynchronous, connect() may return 0 or -1. When return -1, WSAGetLastError()=WSAEWOULDBLOCK/errno=EINPROGRESS
+      (ret != 0 && last_socket_error() != ROS_SOCKETS_ASYNCHRONOUS_CONNECT_RETURN))) 
   {
     ROSCPP_CONN_LOG_DEBUG("Connect to tcpros publisher [%s:%d] failed with error [%d, %s]", host.c_str(), port, ret, last_socket_error_string());
     close();
--- a/clients/roscpp/src/libros/transport/transport_tcp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_tcp.cpp
@@ -313,8 +313,8 @@ bool TransportTCP::connect(const std::string& host, int port)
   // windows might need some time to sleep (input from service robotics hack) add this if testing proves it is necessary.
   // ROS_ASSERT((flags_ & SYNCHRONOUS) || ret != 0);
   if (((flags_ & SYNCHRONOUS) && ret != 0) || // synchronous, connect() should return 0
-      (!(flags_ & SYNCHRONOUS) && // asynchronous, connect() may return 0 or -1. When return -1, WSAGetLastError()=WSAEWOULDBLOCK/errno=EINPROGRESS
-      (ret != 0 && last_socket_error() != ROS_SOCKETS_ASYNCHRONOUS_CONNECT_RETURN))) 
+      (!(flags_ & SYNCHRONOUS) && ret != 0 && last_socket_error() != ROS_SOCKETS_ASYNCHRONOUS_CONNECT_RETURN)) 
+      // asynchronous, connect() may return 0 or -1. When return -1, WSAGetLastError()=WSAEWOULDBLOCK/errno=EINPROGRESS
   {
     ROSCPP_CONN_LOG_DEBUG("Connect to tcpros publisher [%s:%d] failed with error [%d, %s]", host.c_str(), port, ret, last_socket_error_string());
     close();


gavanderhoorn:
This is related to [Installing ROS on Ubuntu Bash in Windows 10](http://answers.ros.org/question/238646/installing-ros-on-ubuntu-bash-in-windows-10/#238666) on ROS Answers.

I've tested this briefly on my own Win10/WSL/Trusty/Indigo installation and it seems to work (meaning: both publishing and subscribing work from the 'WSL host'), although I'm not entirely convinced this is *the* way to implement the patch.

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -311,9 +311,10 @@ bool TransportTCP::connect(const std::string& host, int port)
 
   int ret = ::connect(sock_, (sockaddr*) &sas, sas_len);
   // windows might need some time to sleep (input from service robotics hack) add this if testing proves it is necessary.
-  ROS_ASSERT((flags_ & SYNCHRONOUS) || ret != 0);
+  // ROS_ASSERT((flags_ & SYNCHRONOUS) || ret != 0);
   if (((flags_ & SYNCHRONOUS) && ret != 0) || // synchronous, connect() should return 0
-      (!(flags_ & SYNCHRONOUS) && last_socket_error() != ROS_SOCKETS_ASYNCHRONOUS_CONNECT_RETURN)) // asynchronous, connect() should return -1 and WSAGetLastError()=WSAEWOULDBLOCK/errno=EINPROGRESS
+      (!(flags_ & SYNCHRONOUS) && // asynchronous, connect() may return 0 or -1. When return -1, WSAGetLastError()=WSAEWOULDBLOCK/errno=EINPROGRESS
+      (ret != 0 && last_socket_error() != ROS_SOCKETS_ASYNCHRONOUS_CONNECT_RETURN)))
```
dirk-thomas: The added condition `ret != 0` makes sense. Checking `last_socket_error` when `ret` is `0` doesn't make sense.

Can you please indent this line by another space (since the expression is nested within the open parenthesis from the line before). Also the additional parenthesis around this line are not necessary.

bxwllzz: I hava modified the code format .

dirk-thomas:
Thanks!

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -311,9 +311,10 @@ bool TransportTCP::connect(const std::string& host, int port)
 
   int ret = ::connect(sock_, (sockaddr*) &sas, sas_len);
   // windows might need some time to sleep (input from service robotics hack) add this if testing proves it is necessary.
-  ROS_ASSERT((flags_ & SYNCHRONOUS) || ret != 0);
+  // ROS_ASSERT((flags_ & SYNCHRONOUS) || ret != 0);
```
gavanderhoorn, line 314: Should this commented line be left in?

There is no comment near it to explain what is going on here.

dirk-thomas: It could indeed be removed since it doesn't match the behavior on Windows.

janbernloehr:
Is it possible to make this fix available for kinetic as well?

gonzalocasas:
It would be reaaaally good if this fix could be ported to Kinetic as well.

gavanderhoorn:
@dirk-thomas: I couldn't find this (easily, or didn't try hard enough): can you confirm whether or not this fix has been back-ported to Kinetic?

gonzalocasas:
I think it hasn't. I also tried finding it but no luck.

dirk-thomas:
The last release into Kinetic is from before this bug has been fixed. It will be backported before the next Kinetic release.

bxwllzz:
> https://github.com/ros/ros_comm/blob/lunar-devel/clients/roscpp/src/libros/transport/transport_tcp.cpp

I found another bug in `transport_tcp.cpp` with `Windows Subsystem for Linux`.
`TransportTCP::read()` did not check if socket is conneted. If socket is asynchronous, it may fails at recv() if its slow to connect (e.g. happens with wireless).
`TransportTCP::read()` and `TransportTCP::write()` needs to check if its connected or not when it's asynchronous.

# Detailed debug log
Test Environment: 
- PC 1 (Windows Subsystem for Linux, wired): RViz
- PC 2 (Ubuntu 16.04, wireless): roscore and other nodes  
```
...
[DEBUG] [1508756684.267337700]: Retrying connection to [pc2:60159] for topic [/kinect/rgb/image_color]
[DEBUG] [1508756684.387001100]: Resolved publisher host [pc2] to [192.168.123.211] for socket [24]
[DEBUG] [1508756684.387432900]: setsockopt failed to set SO_KEEPALIVE on socket [24] [pc2:60159 on socket 24]
[DEBUG] [1508756684.387574800]: setsockopt failed to set TCP_KEEPCNT on socket [24] [pc2:60159 on socket 24]
[DEBUG] [1508756684.388021100]: Adding tcp socket [24] to pollset
[DEBUG] [1508756684.388515700]: Async connect() in progress to [pc2:60159] on socket [24]
[DEBUG] [1508756684.388849400]: recv() on socket [24] failed with error [传输端点尚未连接] (ENGLISH TRANSLATION: The transmission endpoint is not connected yet)
[DEBUG] [1508756684.389211800]: TCP socket [24] closed
[DEBUG] [1508756684.389618800]: Connection::drop(0)
[DEBUG] [1508756684.389992600]: Connection to publisher [TCPROS connection on port 3678 to [pc2:60159 on socket 24]] to topic [/kinect/rgb/image_color] dropped
...
```

# Fix

> man page [connect(2)](http://man7.org/linux/man-pages/man2/connect.2.html)
>        EINPROGRESS
>              The socket is nonblocking and the connection cannot be
>              completed immediately.  It is possible to select(2) or poll(2)
>              for completion by selecting the socket for writing.  After
>              select(2) indicates writability, use getsockopt(2) to read the
>              SO_ERROR option at level SOL_SOCKET to determine whether
>              connect() completed successfully (SO_ERROR is zero) or
>              unsuccessfully (SO_ERROR is one of the usual error codes
>              listed here, explaining the reason for the failure).

I'm trying to add some code to check if its connected or not.

dirk-thomas:
@bxwllzz Please either open a new ticket about the new bug or propose a PR with the changes you described. Since this ticket is different and already closed it is unlikely that anybody will look at your comment in the future.

bxwllzz:
@dirk-thomas I have proposed a new PR https://github.com/ros/ros_comm/pull/1202 about this problem.

elahia:
Hi guys,

I have the same problem, and i want to update the transport_tcp.cpp file, but i cannot find it!! 
Can anyone please help me to find it!

Thanks

gonzalocasas:
@elahia this is already solved, I am using Kinetic on WSL without any issues. Did you try updating kinetic?

elahia:
@gonzalocasas  thank you for your fast reply, No, actually i'm not using kinetic, my ros pack is so small  (at least by now) and i only need to publish and subscribe to a topic! but when i check the topic by : rostopic echo /topicname

 i get this error:
[WARN] [1532505967.859467]: Inbound TCP/IP connection failed: timed out   


so even for working with messages, i have to install kinetic?


Regards

gavanderhoorn:
`kinetic` is a release name, not a particular set of packages (well .. it sort of is actually).

What @gonzalocasas means is: as long as you have a version of `ros_comm` installed that is part of the `kinetic` release, things should work.

elahia:
@gavanderhoorn  i tried  [this](https://janbernloehr.de/2017/06/10/ros-windows)  to install WSL on my windows 10. and now i have lunar installed on my win 10.
I'm not sure if by installing WSL the ros_comm also is installed or not! 

if i have to install it separately, would you please suggest me some link?
Thanks

gavanderhoorn:
Lunar is newer than Kinetic, and the packages in `ros_comm` are a base dependency / integral part of ROS, so if you followed the instructions on that page it should work.

If it doesn't, then either something changed in WSL (that happens often, seeing as it's still being developed), a regression was introduced by a later commit to `ros_comm` or something else is wrong (firewall issue perhaps?).

In any case: the changes in this PR (ie: the one you're commenting on) were already merged, so manually changing `transport_tcp.cpp` should not be needed.

elahia:
oh my bad! I  was just confused between Kinect and kinetic!
btw how can i fix my publisher ( .py file) to publish a topic without this error!

gonzalocasas:
@elahia if you have ROS Lunar installed, you should **not** have this error. Are you sure the issue you are seeing is really this same one? Could you post some more details about your setup and the log you are getting?

gavanderhoorn:
Let's not start diagnosing this on this thread.

@elahia: please post a question on ROS Answers. If it turns out something is not right in `ros_comm`, you can always post a new issue while referring to this one and your ROS Answers question.

elahia:
@gavanderhoorn  @gonzalocasas  I just switched to Linux 14.04 ( with ros indigo) to test my code (to be sure the problem is with WSL or my own code)

interestingly i get the same error:

[WARN] [WallTime: 1532521089.338552] Inbound TCP/IP connection failed: connection from sender terminated before handshake header received. 0 bytes were received. Please check sender for additional details.

So i think this is not related to this topic even by same error!

gonzalocasas:
ROS Indigo definitely doesn't work properly with WSL, only Kinetic onwards. But as @gavanderhoorn recommended, please post the details on ROS Answers.

elahia:
Ok i will create a new post, but just to clarify, my ros indigo is installed in another pc and i just ran my code on another pc without any WSL. but got same error...

PR #1052, opened 2017-05-10
Add option publish_hz to publish current rate.

dornhege:
This makes it easy to use this in other tools such as rqt_plot.

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -223,6 +223,8 @@ def _rostopic_hz(topic, window_size=-1, filter_expr=None, use_wtime=False, publi
         _sleep(1.0)
         rt.print_hz()
         if pub:
+            if not rt.times:
+                continue
             rate, min_delta, max_delta, std_dev, n = rt.compute_hz()
             pub.publish(rate)
 
@@ -349,10 +351,11 @@ class ROSTopicBandwidth(object):
             except:
                 traceback.print_exc()
 
-    def print_bw(self):
-        """print the average publishing rate to screen"""
-        if len(self.times) < 2:
-            return
+    def compute_bw(self):
+        """
+        Compute the band width based on the recorded times.
+        :returns: a tuple bytes per sec, mean, min, max, n_data
+        """
         with self.lock:
             n = len(self.times)
             tn = time.time()
@@ -368,6 +371,13 @@ class ROSTopicBandwidth(object):
             max_s = max(self.sizes)
             min_s = min(self.sizes)
 
+        return (bytes_per_s, mean, min_s, max_s, n)
+
+    def print_bw(self):
+        """print the average publishing rate to screen"""
+        if len(self.times) < 2:
+            return
+        bytes_per_s, mean, min_s, max_s, n = self.compute_bw()
         #min/max and even mean are likely to be much smaller, but for now I prefer unit consistency
         if bytes_per_s < 1000:
             bw, mean, min_s, max_s = ["%.2fB"%v for v in [bytes_per_s, mean, min_s, max_s]]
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Wouldn't it be better to use the Python API to get the current publish rate rather then publishing that information on another topic?

dornhege:
This would require someone to write code to use this. In this way, one can just start rostopic hz + rqt_plot and plot the current rate.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -62,6 +62,7 @@
 import rosgraph
 #TODO: lazy-import rospy or move rospy-dependent routines to separate location
 import rospy
+import std_msgs
```
dirk-thomas, line 65: This new dependency needs to be listed in the package manifest.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -106,12 +107,12 @@ def __init__(self, window_size, filter_expr=None, use_wtime=False):
         self.times =[]
         self.filter_expr = filter_expr
         self.use_wtime = use_wtime
-        
+
```
dirk-thomas: Please avoid any unrelated whitespace changes.

Several below.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1297,10 +1307,13 @@ def _rostopic_cmd_hz(argv):
     parser.add_option("--wall-time",
                       dest="use_wtime", default=False, action="store_true",
                       help="calculates rate using wall time which can be helpful when clock isnt published during simulation")
+    parser.add_option("--publish-hz",
+                    dest="publish_hz", default=False, action="store_true",
+                    help="publishes the rate on ~hz to be used by other program (e.g. for plotting)")
```
dirk-thomas: Would it make sense to pass a topic name here? Otherwise if the user runs multiple of these commands it becomes difficult to choose the "right" one.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -210,7 +218,9 @@ def _rostopic_hz(topic, window_size=-1, filter_expr=None, use_wtime=False):
     while not rospy.is_shutdown():
         _sleep(1.0)
         rt.print_hz()
-
+        if pub:
+            rate, min_delta, max_delta, std_dev, n = rt.compute_hz()
+            pub.publish(rate)
```
dirk-thomas: Would it make sense to publish more than just the rate? Maybe the user wants to plot the other values too.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -145,25 +146,17 @@ def callback_hz(self, m):
...
-    def print_hz(self):
-        """
-        print the average publishing rate to screen
-        """
-        if not self.times:
-            return
-        elif self.msg_tn == self.last_printed_tn:
-            print("no new messages")
-            return
+    def compute_hz(self):
```
dirk-thomas: Please add a docblock for this function.

dirk-thomas (commented):
Now I understand the use case. That would be a good feature. I added a few comments inline.

Please also change the targeted branch of the PR to be `lunar-devel` (you can do that by clicking the edit button beside the title but might need to ensure that your changes apply or are based ontop of that branch).

dornhege:
@dirk-thomas I addressed the minor issues. Not sure, what you meant with "the other values". Min/max, etc. could also be published. For now, I added the same kind of option for band width. A merge into lunar will require additional changes as the implementation has changed.

dirk-thomas:
I was more thinking about publishing all printed values in a single message. E.g. `rostopic hz --publish /foo` would publish a `Hz` message contains all fields usually printed: `rate`, `min_delta`, `max_delta`, `std_dev`, and (maybe) `n`. That way the user can plot any of the values and is not limited to `rate`.

Please update the PR to target `lunar-devel`. Patches are by default only accepted into the latest branch (except if they fix a bug only present in older distros).

dornhege:
Closing this in favor of #1072 for lunar.

PR #1054, opened 2017-05-11
Fix race condition that lead to miss first message

kamibo:
Callback queue waits for callback from "callOne" method.
When internal queue is not empty this last method succeeded even if id
info mapping does not contains related callback's id.
In this case, first callback (for one id) is never called since
"addCallback" method first push callback into internal queue and *then*
set info mapping.

So id info mapping has to be set before push callback into interal
queue. Otherwise first message could be ignored.

Attached file highlights this issue.
[ros_callback_queue_bug.cpp.zip](https://github.com/ros/ros_comm/files/993526/ros_callback_queue_bug.cpp.zip)

### DIFF
--- a/clients/roscpp/src/libros/callback_queue.cpp
+++ b/clients/roscpp/src/libros/callback_queue.cpp
@@ -102,17 +102,6 @@ void CallbackQueue::addCallback(const CallbackInterfacePtr& callback, uint64_t r
   info.removal_id = removal_id;
 
   {
-    boost::mutex::scoped_lock lock(mutex_);
-
-    if (!enabled_)
-    {
-      return;
-    }
-
-    callbacks_.push_back(info);
-  }
-
-  {
     boost::mutex::scoped_lock lock(id_info_mutex_);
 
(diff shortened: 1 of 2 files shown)


esteve:
@kamibo I've pushed https://github.com/kamibo/ros_comm/pull/1 which includes your fix and a test for the race condition. I've removed the C++11 and C++14 parts from the test and replaced them with Boost so it builds fine on a strict Indigo.

dirk-thomas:
@kamibo Can you please incorporate the test provided by @esteve and target the latest development branch with this PR (`lunar-devel`).

esteve:
@kamibo I've created https://github.com/ros/ros_comm/pull/1058 which is just a rebase of this branch on top of `lunar-devel`

@dirk-thomas do you need a PR for Kinetic as well or you'll back/forward port it yourself?

kamibo:
Thanks @esteve it's perfect !

dirk-thomas:
Closing this in favor of #1058. The patch will be considered for backporting into older distros from the `lunar-devel` branch.

PR #1056, opened 2017-05-16
rosconsole: remove extra semicolon in definition of macro ROSCONSOLE_PRINTF_ATTRIBUTE(a, b)

meyerj:
Another extra semicolon that triggers compiler warnings with `-Wpedantic`, like in https://github.com/ros/ros_comm/pull/817.


dirk-thomas:
Thanks.

PR #1057, opened 2017-05-18
roslaunch: ensure pid file is removed on exit

furushchev:
If `--pid` option is set, pid file is written at https://github.com/ros/ros_comm/blob/lunar-devel/tools/roslaunch/src/roslaunch/__init__.py#L257 but codes for removing written pid file is under `if-else` condition (so there can be conditions that `roslaunch` process dies but pid file is left without removal)
This pull request moves the code for pid file removal to main `try-except-finally` block so to ensure pid file is removed if exists.

### DIFF
--- a/tools/roslaunch/src/roslaunch/__init__.py
+++ b/tools/roslaunch/src/roslaunch/__init__.py
@@ -296,21 +296,15 @@ def main(argv=sys.argv):
             # This is a roslaunch parent, spin up parent server and launch processes.
             # args are the roslaunch files to load
             from . import parent as roslaunch_parent
-            try:
-                # force a port binding spec if we are running a core
-                if options.core:
-                    options.port = options.port or DEFAULT_MASTER_PORT
-                p = roslaunch_parent.ROSLaunchParent(uuid, args, roslaunch_strs=roslaunch_strs,
-                        is_core=options.core, port=options.port, local_only=options.local_only,
-                        verbose=options.verbose, force_screen=options.force_screen,
-                        num_workers=options.num_workers, timeout=options.timeout)
-                p.start()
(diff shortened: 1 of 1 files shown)


tools/roslaunch/src/roslaunch/__init__.py
```diff
@@ -306,11 +306,8 @@ def main(argv=sys.argv):
                         num_workers=options.num_workers, timeout=options.timeout)
                 p.start()
                 p.spin()
-            finally:
-                # remove the pid file
-                if options.pid_fn:
-                    try: os.unlink(options.pid_fn)
-                    except os.error: pass
+            except:
+                pass
```
dirk-thomas: Why should this catch and ignore arbitrary exceptions?

dirk-thomas (commented):
Please edit the PR to target the latest development branch `lunar-devel`.

tools/roslaunch/src/roslaunch/__init__.py
furushchev: @dirk-thomas Thank you for review! I have no reason. Removed `try`.

dirk-thomas:
The patch looks good now. Can you please still edit the PR to target the latest development branch lunar-devel ("edit" button right beside the issue title).

furushchev:
@dirk-thomas Changed base branch to `lunar-devel`, but it seems to be many conflicts. Is it still ok, or should I cherry-pick to lunar-devel?

dirk-thomas:
You will need to rebase your patch to the new target branch.

furushchev:
@dirk-thomas Done!

dirk-thomas:
Thank you!

PR #1058, opened 2017-05-18
Fix race condition that lead to miss first message (lunar)

esteve:
Forward-port of https://github.com/ros/ros_comm/pull/1054 to Lunar

Original author @kamibo


dirk-thomas:
Looks great: http://build.ros.org/job/Lpr__ros_comm__ubuntu_xenial_amd64/121/testReport/(root)/CallbackQueue/raceConditionCallback/

Thanks!

esteve:
@dirk-thomas thanks for the prompt review! Let @kamibo and me know if we can help with getting the change merged into `indigo-devel` and `kinetic-devel`

PR #1062, opened 2017-05-23
Both logging rebased

asmodehn:
A new attempt at https://github.com/ros/ros_comm/pull/1042, using custom LogStreams instead of threads for logging.


asmodehn:
This doesn't work as I expected it to. Looking at https://docs.python.org/2/library/subprocess.html#subprocess.Popen stdin, stdout and stderr args will not simply accept any logging or io derived classes (only open file handles or sys.stdout/stderr).

We could hack something together, but the code would likely be more complex than needed.
So I think the original #1042 is the way to go with python2 (python3 has coroutines which would allow to do all this in only one thread, after some heavy refactoring...)

Closing this.

Issue #1063, opened 2017-05-23
Better error output when logging cannot create folder


nordmoen:
Summary:
In the function [`configure_logging`](https://github.com/ros/ros_comm/blob/247459207e20c1da109fc306e58b84d15c4107bd/tools/rosgraph/src/rosgraph/roslogging.py#L86) it would be nice if in addition to printing that the folder cannot be created print the exception being thrown.

Longer explanation:
I am having a problem where I need to run ROS on a limited environment. To allow for logging the system lets me create a folder and expose that to ROS. The problem I have been having is that when the system creates the folder it is created with different group than ROS is running under. This resulted in the situation where ROS was able to create the logging directory, but not change the folders permissions (parent folder had group `root`). I have used quite some time debugging why this was happening and it took some time before I found the function responsible since the folder was created successfully, but ROS still stopped with "cannot create directory".

dirk-thomas:
That sounds like a good idea. Please consider to provide a pull request for this improvement (against the latest development branch, currently `lunar-devel`).

dirk-thomas:
Addressed by #1068.

PR #1068, opened 2017-05-26
Added logging output when `roslogging` cannot change permissions

nordmoen:
Added better differentiated logging output to `roslogging` so that
problems with permission is made clear to the user. Accompanying test
have also been added.

See #1063 for motivation.


tools/rosgraph/test/test_roslogging.py
```diff
@@ -0,0 +1,55 @@
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2009, Willow Garage, Inc.
```
dirk-thomas: The copyright is wrong. That should refer to you instead.

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -83,7 +83,13 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
             makedirs_with_parent_perms(logfile_dir)
         except OSError:
             # cannot print to screen because command-line tools with output use this
-            sys.stderr.write("WARNING: cannot create log directory [%s]. Please set %s to a writable location.\n"%(logfile_dir, ROS_LOG_DIR))
+            if os.path.exists(logfile_dir):
+                # We successfully created the logging folder, but could not change
+                # permissions of the new folder to the same as the parent folder
+                sys.stderr.write("Could not change permissions for folder [{!s}], make sure that the parent folder has correct permissions.\n".format(logfile_dir))
```
dirk-thomas: Should this message also start with `WARNING: `?

Please use `%` instead of `format` to be consistent with the surrounding code.

tools/rosgraph/test/test_roslogging.py
```diff
@@ -0,0 +1,55 @@
...
+            "test_logging_dir_root")
+    test_log = os.path.join(
+            test_folder,
+            "test_subfolder",
+            "test.log")
+    test_env = {"ROS_LOG_DIR": test_folder}
+    log_filename = roslogging.configure_logging("test", filename=test_log, env=test_env)
+    assert log_filename is None
+    assert os.path.isdir(test_folder)
+    os.rmdir(os.path.join(test_folder, "test_subfolder"))
```
dirk-thomas: Tests shouldn't write into the source folder of the package. Please use a temporary path for that.

nordmoen:
Updated the pull request with your feedback. I removed the testing as I don't know how to test the failure in a transient way. In any case the simple change should not have that big an impact and we know it works.

If you have a good idea on how to test this I will gladly update the pull request again.

dirk-thomas:
I am sorry for the late response. This looks good now. Merging...

PR #1072, opened 2017-06-09
Add options to publish hz and bw.

dornhege:


dornhege:
@dirk-thomas With regard to publishing all values (rate, min, etc.) in a single message: A custom message seems maybe too much (where should that be located?) and a Float64MultiArray is missing the semantics. Alternatively there could be multiple Float64 sub-topics, i.e., if one choses /odo_hz to publish, there will be /odo_hz/rate, /odo_hz/min, etc.

dirk-thomas:
The separate topics (`/odo_hz/rate`, `/odo_hz/min`) would be problematic if a subscriber wants to get multiple of these and wants to use them together.

The package `ros_comm_msgs` might be a good place for that message. It already contains something similar: https://github.com/ros/ros_comm_msgs/blob/indigo-devel/rosgraph_msgs/msg/TopicStatistics.msg

m-tosch:
At least the function `compute_bw` should be added. Currently only a `print_bw` exists and there is no way to get the parameters bytes_per_s, mean, min_s, max_s, n 

For RosTopicHz and RosTopicDelay this already exists but strangely not for RosTopicBandwidth

dirk-thomas:

PR #1076, opened 2017-06-12
speedier lookups with rosbag info -y -k

olgalupuleac:


dirk-thomas:
Please update the PR to not include any generated files. Also the README in the root of the repo seems to be a bad place for the information. Maybe place it in `tools/rosbag/doc` instead?

dirk-thomas:
While the latest commits pass CI the previous comment still applies: do not include any generated files and move the added README file.

cwecht:
Well, as far as I can see, this is basically a C++-reimplementation of ''rosbag infor -y -k".  I think, this should be done in a similar manner as play or record are implemented.

dirk-thomas:

Issue #819, opened 2016-06-20
rosbag: create option to filter for topics published by a node


Petrox:
Like   --node=NODE           record all topics subscribed to by a specific node, but for published messages.

dirk-thomas:
Please consider to contribute a pull request implementing this improvement.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #1077, opened 2017-06-12
Issue #819

GeorgeChernyshov:
SPBAU Teaching. This is for Kirill Krinkin


mikepurvis:
This seems like reasonable functionality to me, though I've never used the `--node` option, so I haven't experienced the need for `--publisher`.

You'll need to drop the readme changes, and figure out what's going wrong in the CI job.

PR #1083, opened 2017-06-26
[rospy] Add option to reset timer when time moved backwards

wkentaro:
## Why?

Sometimes I use a node which uses timer playing rosbag which is played with options `--clock --loop`. I also set `/use_sim_time` to `true`.
In this case, timer raises ROSTimeMoveBackwardsException, and the node which uses timer will stop working.
I'd like to add option to change this behavior with resetting time such like in roscpp.

### DIFF
--- a/clients/rospy/src/rospy/timer.py
+++ b/clients/rospy/src/rospy/timer.py
@@ -57,10 +57,10 @@ class Rate(object):
         @param reset: if True, timer is reset when rostime moved backward. [default: False]
         @type  reset: bool
         """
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch. I updated the exception handling logic slightly in 6af1e30724780b04e47abd66f9e588bceafdbdce. Mostly removing the warning since the caller explicitly asked for the reset there shouldn't be a warning printed.

wkentaro:
@dirk-thomas thanks.

PR #1086, opened 2017-06-26
Close CLOSE_WAIT sockets (#610)

otamachan:
Hi,

I'm having the same issue as in #610 and I found that some of the sockets in caches of ServerProxy([``_proxies``](https://github.com/ros/ros_comm/blob/lunar-devel/tools/rosmaster/src/rosmaster/util.py#L52)) are remained CLOSE_WAIT when a slave(xmlrpc server) close its port suddenly.
I added a function to close all the CLOSE_WAIT sockets in caches(f06bdf1), made it an option to call the function in ``xmlrpcapi`` (d3e7921) and added ``--close-sockets`` option to roscore(00d8f73).

I think calling the ``close_half_closed_sockets`` periodically is enough, but I couldn't find the right way so I call it in ``xmlrpcapi``.
I'm happy if closing CLOSE_WAIT sockets is enabled by default, could you review this and sugget a better way?

Thanks in advance.

### DIFF
--- a/tools/roslaunch/scripts/roscore
+++ b/tools/roslaunch/scripts/roscore
@@ -56,6 +56,10 @@ def _get_optparse():
     parser.add_option("-t", "--timeout",
                       dest="timeout",
                       help="override the socket connection timeout (in seconds).", metavar="TIMEOUT")
+    parser.add_option("--close-sockets",
+                      dest="close_sockets",
+                      action="store_true", default=False,
+                      help="enable checking and closing CLOSE_WAIT sockets.")
     return parser
 
 
--- a/tools/roslaunch/src/roslaunch/__init__.py
+++ b/tools/roslaunch/src/roslaunch/__init__.py
@@ -180,6 +180,10 @@ def _get_optparse():
     parser.add_option("-t", "--timeout",
                       dest="timeout",
                       help="override the socket connection timeout (in seconds). Only valid for core services.", metavar="TIMEOUT")
+    parser.add_option("--close-sockets",
+                      dest="close_sockets",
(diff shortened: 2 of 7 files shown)


dirk-thomas:
The pull request #834 already tries to address the problem described in the issue. The complexity of changes has so far prevented the previous pull request to be merged.

Maybe @mgrrx @mikepurvis @jspricke @furushchev can comment here how this patch interacts with the existing PR.

furushchev:
This is more elegant way to close `CLOSE_WAIT` socket.
I tested this patch on a real robot and checked it works well without any trouble at this time (at least no side effect like the one I encountered in the previous pull request.)
+1 to be merged (also I strongly recommend to set this option `True` by default since it's now clear that the existing ROS system has the bug in master that loses communication among all the nodes.)

otamachan:
Thanks for your comments and feedback. @furushchev 

For anyone who is interested in testing this, I prepared the branch to close ``CLOSE_WAIT`` sockets by default. https://github.com/ros/ros_comm/compare/lunar-devel...otamachan:fix-610-by-default

dirk-thomas:
> For anyone who is interested in testing this, I prepared the branch to close CLOSE_WAIT sockets by default. lunar-devel...otamachan:fix-610-by-default

I created #1104 from that branch. This looks very promising since it seems to resolve the current problem as well as does it in a very "simple" way which should not introduce any side effects.

otamachan:
@dirk-thomas Thank you very much for moving this forward.

dirk-thomas:
Thank you for coming up with this much simpler patch!

PR #1092, opened 2017-07-08
A fix to a critical stack buffer overflow vulnerability which leads to direct control flow hijacking

dingelish:
Hi there, I find a stack buffer overflow in `Bag::readVersion`. When reading from malicious inputs, which contains an over-length log type name in the version line, the stack may be smashed and return address would be tampered. Attacker could hi-jack the system using a malicious input file.

### DIFF
--- a/tools/rosbag_storage/src/bag.cpp
+++ b/tools/rosbag_storage/src/bag.cpp
@@ -225,29 +225,18 @@ void Bag::readVersion() {
 
     file_header_pos_ = file_.getOffset();
 
-    size_t version_line_len = version_line.length();
-
-    if(version_line_len == 0)
-        throw BagIOException("Error reading version line");
-
-    char * logtypename = (char *) malloc(version_line_len + 1);
-
-    if(logtypename == NULL)
-        throw BagIOException("Error reading version line");
-
+    char logtypename[100];
     int version_major, version_minor;
 #if defined(_MSC_VER)
     if (sscanf_s(version_line.c_str(), "#ROS%s V%d.%d", logtypename, sizeof(logtypename), &version_major, &version_minor) != 3)
 #else
-    if (sscanf(version_line.c_str(), "#ROS%s V%d.%d", logtypename, &version_major, &version_minor) != 3)
+    if (sscanf(version_line.c_str(), "#ROS%99s V%d.%d", logtypename, &version_major, &version_minor) != 3)
 #endif
         throw BagIOException("Error reading version line");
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for raising the issue for this potential buffer overflow. The current patch leaks the allocated memory in case a `BagIOException` is thrown in line 244 though.

Instead of allocating memory I would suggest to keep using the stack allocated `logtypename` variable and pass a specific `width` for the string part in the format specified. This will ensure that the buffer won't be exceeded and no dynamic memory needs to be allocated / deallocated.

dingelish:
You are correct. My patch does include a memory leakage vulnerability. 
In *nix, I cannot find something like 'sscanf_s' as MSVC provides. So I recommend dynamically allocating the buffer to avoid from implement it by yourself. 
So how should we deal with it? Should I submit another pull-request? Thanks!

dirk-thomas:
> In *nix, I cannot find something like 'sscanf_s' as MSVC provides. So I recommend dynamically allocating the buffer to avoid from implement it by yourself.

You don't need a different function for this. You can limit the extracted string by using a format specified with a *width* (e.g. `%42s`) which ensures that the variable doesn't exceed 42 characters (plus `\0`).

So as mentioned before there is no need to dynamically allocate memory.

> So how should we deal with it? Should I submit another pull-request?

You can simply commit to the branch this PR is coming from. The PR will always show the latest state on that branch compared to the targeted branch.

dingelish:
You are right. `scanf` series functions do take width with `%s`. But the problem is, over-length string will cause `sscanf`'s early return, which leads to un-assigned version number.
For example, 
`sscanf("aaaaaaaaaa 200", "%9s %d", &buff, &n);` is not correctly assigning 200 to `n`. It just leaves `n` unchanged. 
I don't know if this is a desired behavior. If so, I'd happy to use the buff on stack with a width specifier in format string :)

dirk-thomas:
You can either check the length of `buff` and when it reaches the maximum length error out. Or you could check the value of `n` and error out if it hasn't changed from the initial value (which could be `-1`). I would think the first approach is better since it doesn't narrow the value range of `n`.

**Update:** the return value of `sscanf` / `sscanf_s` will indicate if all three arguments have been extracted. If that is not the case an exception is already thrown.

dingelish:
I think it's beautiful now :)

dirk-thomas:
Looking great. Thank you for the report and for iterating on it.

PR #1096, opened 2017-07-11
rostopic ignores 64-bit values when printing in plotting friendly format

### LINKED ISSUE #1095: rostopic ignores 64-bit values when printing in plotting friendly format
When running the attached launch file you can see that the 64-bit value isn't printed.
[rostopic64bit.zip](https://github.com/ros/ros_comm/files/1139573/rostopic64bit.zip)

mstegmaier:
Fixes #1095.

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -687,7 +687,8 @@ def _sub_str_plot_fields(val, f, field_filter):
     """recursive helper function for _str_plot_fields"""
     # CSV
     type_ = type(val)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
`long` doesn't exist in Python 3. Please update the patch to work with both Python 2 as well as 3.

dirk-thomas:
In order to simplify the patch I would suggest to define `long` as being the same as `int` in the case of Python 3. That way the conditions stay shorter, more readable and less repetitive.

dirk-thomas:
Please see #1110 for a different patch to address the problem.

PR #1098, opened 2017-07-12
only launch core nodes if master was launched by roslaunch

### LINKED ISSUE #1097: Race condition between `roscore` and `roslaunch --wait`
**Steps to reproduce**

```
echo "<launch></launch>" > launch.xml
roscore & roslaunch --wait launch.xml
```
Rarely, both roscore and roslaunch will start the rosout node. If this happens, the rosout nodes will conflict forever as they are both started with `respawn="true"`.

To make this more likely to happen, you can do the following:

```
edit _setup() in roslaunch/launch.py:
        self._launch_master()
        import random
        time.sleep(random.random())
        self._launch_core_nodes() 
edit _wait_for_master() in roslaunch/rlutil.py:
    while not is_running:
        # time.sleep(0.1)
        is_running = m.is_running()
run this script and look for multiple "started core service" lines:
    #!/bin/bash
    export PYTHONUNBUFFERED=yes
    roscore | sed -n "/core service/ {s/^/roscore: /; p}" &
    for i in $(seq 1 100); do
      roslaunch --wait launch.xml | sed -n "/core service/ {s/^/roslaunch $i: /; p}" &
    done
    sleep 5
    pkill -P $$
    wait
```

**Explanation**

If you run `roslaunch --wait`, it will wait for the master before starting, but the `--wait` flag has no effect on the starting of core nodes. In roslaunch/launch.py, we have:

```
        # start up the core: master + core nodes defined in core.xml
        self._launch_master()
        self._launch_core_nodes() 
```

unconditionally of whether `--core` or `--wait` is used. In the `--wait` case, `rlutil._wait_for_master` has already been called, and `_launch_master()` notices that the master is already running so does nothing. However, `_launch_core_nodes` may still try to launch `rosout` if the roscore hasn't started it yet.

**Workaround**

Run this command before `roslaunch --wait`:

```
until rosnode info rosout | grep Pid; do sleep 1; done
```

dirk-thomas:
Fixes #1097.

### DIFF
--- a/tools/roslaunch/src/roslaunch/launch.py
+++ b/tools/roslaunch/src/roslaunch/launch.py
@@ -382,6 +382,9 @@ class ROSLaunchRunner(object):
     def _launch_master(self):
         """
         Launches master if requested. 
+        @return: True if a master was launched, False if a master was
+        already running.
+        @rtype: bool
         @raise RLException: if master launch fails
         """
         m = self.config.master
@@ -425,6 +428,8 @@ class ROSLaunchRunner(object):
             hostname = _hostname_to_rosname(hostname)
             self.logger.info("setting /roslaunch/uris/%s__%s' to %s"%(hostname, port, self.server_uri))
             param_server.setParam(_ID, '/roslaunch/uris/%s__%s'%(hostname, port),self.server_uri)
+
+        return not is_running
 
     def _check_and_set_run_id(self, param_server, run_id):
         """
@@ -627,8 +632,9 @@ class ROSLaunchRunner(object):
             self.remote_runner.add_process_listener(self.listeners)            
 
         # start up the core: master + core nodes defined in core.xml
-        self._launch_master()
-        self._launch_core_nodes()        
+        launched = self._launch_master()
+        if launched:
+            self._launch_core_nodes()
         
         # run exectuables marked as setup period. this will block
         # until these executables exit. setup executable have to run


mikepurvis:
LGTM 👍

paulbovbel:
Thanks, +1

dirk-thomas:

dirk-thomas:

dirk-thomas:

dirk-thomas:
Passing CI after merge: http://build.ros.org/view/Ldev/job/Ldev__ros_comm__ubuntu_xenial_amd64/45/

k-okada:
Just for note..., This patch changes rostest behavior, where we can find `rosout` node before 1.12.8, but now 'rosout' node does not started anymore, and also we do not have `rosout_agg` topic.  Just in case someone run into same trouble...
https://github.com/jsk-ros-pkg/jsk_roseus/pull/551/commits/23579baa665c44a632a9ecb1567618dcb61c7c7b

dirk-thomas:
> This patch changes rostest behavior, where we can find `rosout` node before 1.12.8, but now 'rosout' node does not started anymore, and also we do not have `rosout_agg` topic.  Just in case someone run into same trouble...

I am sorry but I don't understand you comment. Can you please elaborate the problem you are experiencing. If it is a problem with this patch we need enough information to address it. What are you running, what is the expected / desired behavior. What has changes in which version. Otherwise we can't act on it.

PR #1101, opened 2017-07-14
improve roswtf tests

dirk-thomas:
Trying to get `roswtf` tests to pass on CI as well as locally...


dirk-thomas:
The additional dependency on `std_srvs` was introduced in #949.

PR #1103, opened 2017-07-18
Add $(dirname) to get directory of current launch file.

mikepurvis:
I'm working on adding some test coverage for this, but the basic idea for usage is allowing stuff like:

```
# test1.launch
<launch>
  <include file="$(dir)/baz/test2.launch" />
</launch>
```

```
# baz/test2.launch
<launch>
  <param name="foo" value="$(dir)/foo/bar" />
  <param name="bar" value="$(eval dir() + '/foo/bar')" />
</launch>
```

Where the output is:

```
$ cd /tmp
$ roslaunch ~/roslaunch_ws/test1.launch  --dump-params
{/bar: /Users/mikepurvis/roslaunch_ws/baz/foo/bar, /foo: /Users/mikepurvis/roslaunch_ws/baz/foo/bar}
```

Some possible use-cases for this:

- When including another launch file from within the same directory, stay DRY and don't repeat the package name everywhere as with `$(find)`— instead just `$(dir)/other.launch`.
- Using `eval` to make the a launch file do different things depending on where it's installed. For example `<node .... if="$(eval dir().startswith('/opt'))" />`
- For launch files which aren't part of packages. We use free-floating launch files as part of a configuration scheme on our robots, and we'd like them to be able to include yamls from their own directory.

FYI @guillaumeautran @Jin-Myung

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -119,20 +119,20 @@ def _anon(resolved, a, args, context):
     anon_context = context['anon']
     return resolved.replace("$(%s)" % a, _eval_anon(id=args[0], anons=anon_context))
 
-def _eval_dir(filename):
+def _eval_dirname(filename):
     if not filename:
-        raise SubstitutionException("Cannot substitute $(dir), no file/directory information available.")
+        raise SubstitutionException("Cannot substitute $(dirname), no file/directory information available.")
     return os.path.abspath(os.path.dirname(filename))
 
-def _dir(resolved, a, args, context):
+def _dirname(resolved, a, args, context):
     """
-    process $(dir)
+    process $(dirname)
     @return: updated resolved argument
     @rtype: str
     @raise SubstitutionException: if no information about the current launch file is available, for example
            if XML was passed via stdin, or this is a remote launch.
     """
-    return resolved.replace("$(%s)" % a, _eval_dir(context.get('filename', None)))
+    return resolved.replace("$(%s)" % a, _eval_dirname(context.get('filename', None)))
 
 def _eval_find(pkg):
     rp = _get_rospack()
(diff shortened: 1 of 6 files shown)


IanTheEngineer:
I support this proposal. Well done, Mike!

My only thought is regarding the first bullet point:

> When including another launch file from within the same directory, stay DRY and don't repeat the package name everywhere as with $(find)— instead just $(dir)/other.launch

Does this imply that `$(find rospkg_name)` changes the directory into the package? I always assumed that it functioned the same way as the commandline tool `rospack find rospkg_name` and returned an absolute path string. Or does it mean that `$(find)` actually caches the path string for `$(dir)` to use later? How does the mechanism work?

Either way, this can be solved with a sentence in the docs :)

dirk-thomas:
The patch looks good to me. Can you please add some tests (probably to `tools/roslaunch/test/unit/test_substitution_args.py`) for this new feature.

mikepurvis:
@IanTheEngineer Nope, `$(find PKG)` does exactly what you think it does. My comment about avoiding repetition would be for a [scenario like this](https://github.com/turtlebot/turtlebot/blob/244fbb5f7269ae42aa7853dec41fe4a94782df0c/turtlebot_bringup/launch/minimal.launch#L14-L25), where you currently have to have:

```
  <include file="$(find turtlebot_bringup)/launch/includes/robot.launch.xml" />
```

But might prefer to have something more like:

```
  <include file="$(dir)/includes/robot.launch.xml" />
```

@dirk-thomas Yup, tests in progress. I'll add to the unit test, but I'm thinking it would also be worthwhile to add something a bit more end-to-end to the test_roslaunch package, basically an automated version of the demonstration posted in the proposal.

One other bit of bikeshedding to do— the `dir` name is fine with me and the obvious one to use, but it does conflict with the `dir` builtin in the eval case. If we care about that, it might be best to pick a different name. If not, `$(dir)` should stick.

dirk-thomas:
> One other bit of bikeshedding to do— the dir name is file with me and the obvious one to use, but it does conflict with the dir builtin in the eval case. If we care about that, it might be best to pick a different name. If not, $(dir) should stick.

Very good point. I would rather avoid overriding Python keywords in that case. Maybe `dirname` instead?

mikepurvis:
Changed to `$(dirname)`, tests added. I think this is good to go now.

dirk-thomas:
Great, thank you.

When you have a chance can you please mention this new feature in the wiki and post a link here for future readers to find it.

mikepurvis:
Done: http://wiki.ros.org/roslaunch/XML#line-77

PR #1104, opened 2017-07-18
Close CLOSE_WAIT sockets by default

dirk-thomas:
Same as #1086 but always trying to close half closed sockets.


dirk-thomas:
@mikepurvis Did you have a chance to try this out?

dirk-thomas:
I will go ahead and merge this with the hope it fixes the long standing problem but without adding regressions :crossed_fingers:

mikepurvis:
Yes, unfortunately I didn't get a chance to try it, sorry. But we're into dev again, so you'll hear squeals soon enough if there are issues.

PR #1105, opened 2017-07-19
[roscpp] add missing header for writev().

rojkov:
After an update of gcc and glibc roscpp started to fail builds with the error:
```
    /home/rojkov/work/ros/build/tmp-glibc/work/i586-oe-linux/roscpp/1.11.21-r0/ros_comm-1.11.21/clients/roscpp/src/libros/transport/transport_udp.cpp:579:25: error: 'writev' was not declared in this scope
         ssize_t num_bytes = writev(sock_, iov, 2);
                             ^~~~~~
```
According to POSIX.1-2001 the function writev() is declared in sys/uio.h.

The patch includes the missing header for POSIX compliant systems.


dirk-thomas:
Thank you.

PR #1112, opened 2017-07-28
Add shm_transport to roscpp/libros

Jrdevil-Wang:
Add 3 cpp files and 3 hpp files, change the CMakeLists.txt under clients/roscpp to make these files compile.


Jrdevil-Wang:
We have publish an independent package.

PR #1113, opened 2017-07-28
Avoids buffer overflow in xmlrpcpp if fd > 1024

kmactavish:
A temporary fix to avoid buffer overflow in XmlRpcDispatch, since select() only works for file descriptors < `FD_SETSIZE` (1024). If `fd >= FD_SETSIZE`, there is buffer overflow, memory corruption, and all-round undefined behaviour. The real solution is to switch to poll as in #833, but since that PR was reverted in #981, this avoids catastrophic failure until that can be worked out.

It throws an exception since:
- this behaviour is no less intrusive than a buffer overflow
- returning an XmlRpcError was handled silently, and just looked like ros comm not being connected 
- this is a temporary solution to fail in a more predictable way.

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -6,6 +6,7 @@
 #include <math.h>
 #include <errno.h>
 #include <sys/timeb.h>
+#include <stdexcept>
 #if defined (__ANDROID__)
 #include <sys/select.h>
 #endif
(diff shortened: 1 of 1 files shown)


kmactavish:
The build of 582984fb3d394fa7e39acd341a8e28e1b7872aa8 failed due to a test added by #1014. Even with this entire change commented out.
```
08:55:05 [test_roscpp.rosunit-timer_callbacks/multipleSteadyTimeCallbacks][FAILURE]------
08:55:05 /tmp/catkin_workspace/src/ros_comm/test/test_roscpp/test/src/timer_callbacks.cpp:220
08:55:05 Failed
```

kmactavish:
This was superseded by #1133 .

PR #1114, opened 2017-07-28
Remove preprended '|' from pretty-printed strings

christian-rauch:
Calling:
`rosparam get -p /robot_description > robot.urdf`
results in a malformed URDF file because it contains a prepended `|` in the first line.
E.g.:
```
$ cat robot.urdf | head -n2
|
  <?xml version="1.0" ?>
```
What is the `|` used for?

### DIFF
--- a/tools/rosparam/src/rosparam/__init__.py
+++ b/tools/rosparam/src/rosparam/__init__.py
@@ -266,11 +266,7 @@ def _rosparam_cmd_get_param(param, pretty=False, verbose=False):
         if type(val) == dict:
             _pretty_print(val)
         else:
-            if '\n' in val:
-                for l in val.split('\n'):
-                    print('  '+l)
-            else:
-                print(val)
+            print(val)
     else:
         dump = yaml.dump(val)
         # #1617
--- a/tools/rosparam/src/rosparam/__init__.py
+++ b/tools/rosparam/src/rosparam/__init__.py
@@ -267,7 +267,6 @@ def _rosparam_cmd_get_param(param, pretty=False, verbose=False):
             _pretty_print(val)
         else:
             if '\n' in val:
-                print('|')
                 for l in val.split('\n'):
                     print('  '+l)
             else:


gavanderhoorn:
> What is the `|` used for?

it's standard yaml and is used to indicate that the text following it should be interpreted as one single multi-line scalar value (from [here](http://www.yaml.org/spec/1.2/spec.html#id2795688)).

Not sure whether that can be removed from there, as it would probably break things that need that `|` there to correctly parse the output.

christian-rauch:
Why is the output interpreted as YAML? The parameter `robot_description` was originally parsed from an XML file. Fetching the parameter should reproduce the exact file content.

Can someone give an example for the case that removing the pipe `|` would break documented behaviour?

gavanderhoorn:
yaml is used all over ROS. In essence, the parameter server is a large yaml dict. Exporting parameters is also done as yaml, so that the output can be loaded onto the parameter server again.

The wiki page for [rosparam](http://wiki.ros.org/rosparam) also notes this:

> rosparam contains the rosparam command-line tool for getting and setting ROS Parameters on the [Parameter Server](http://www.ros.org/wiki/Parameter%20Server) using YAML-encoded files.

This PR would seem to break that documented behaviour.

christian-rauch:
The removal of the `|` should only effect the pretty printing, not the usual import/export of parameters. Is pretty printing used within the parameter server?
E.g. is there an example where `_rosparam_cmd_get_param` is called with `pretty=True`?

gavanderhoorn:
I have no idea. Pretty printing is stated to be "*not yaml safe*", so perhaps the `|` could be removed there.

dirk-thomas:
I can only speculate what the pretty print was intended for when it was originally implemented. This code area hasn't been touched for many years.

In case `rosparam get -p` is being called with an argument identifying a dictionary of parameters the output is still going to be yaml-style. But for your case of printing one specific parameter I think it is safe to assume that the caller isn't expecting yaml since then he would have called it without the pretty print option. Therefore I think it is ok to remove the pipe as being done in the current patch.

But I would argue that the output shouldn't add the two space indentation for values with newlines either. Instead the value should be printed as is. Does that sounds reasonable? If yes, please update the PR to remove the `if` block (line 269-273 in the original file).

christian-rauch:
I agree, the printed value should be the original without indentation.
I removed the indentation and rebased on `lunar-devel`.

dirk-thomas:
Thank you for the patch as well as the quick update.

PR #1118, opened 2017-08-02
Improves the stability of SteadyTimerHelper.

kmactavish:
Due to scheduling / resource contention, `sleep`s and `wait_until`s may be delayed. The `SteadyTimerHelper` test class was not robust to these delays, which was likely the cause of a failing test (`multipleSteadyTimeCallbacks` in `timer_callbacks.cpp`:220).

### DIFF
--- a/test/test_roscpp/test/src/timer_callbacks.cpp
+++ b/test/test_roscpp/test/src/timer_callbacks.cpp
@@ -80,7 +80,10 @@ class SteadyTimerHelper
 
       if (!first)
       {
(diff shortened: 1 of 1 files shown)


kmactavish:
This is a wild goose chase :duck:. This PR was supposed to fix tests so #1113 could get pulled in. It seems to have fixed the problem, but there are more failing tests downstream.

kmactavish:
So the problem seems to be caused by ros/roscpp_core#61. Not sure if the bug is in the test code here, or the range check there.

kmactavish:
Closing in favour of #1119.

PR #1119, opened 2017-08-03
Upstream compatibility

kmactavish:
There are a number of upstream changes that broke tests. This PR aims to get lunar-devel passing tests again so PRs aren't failing.
- ros/genpy#79 ros/roscpp_core#61 changed python stringification of messages, breaking a test.
- #1014 introduced timing-related stochastic test failures. Due to scheduling / resource contention, `sleep`s and `wait_until`s may be delayed. The `SteadyTimerHelper` test class was not robust to these delays, which was likely the cause of a failing test in #1113 (multipleSteadyTimeCallbacks in timer_callbacks.cpp:220), and forcing retests when #1014 was being considered.
- There is still another failing test, related to ros/roscpp_core#61, I haven't tracked that one down. Maintainers, feel free to add that fix to this branch before merging.

### DIFF
--- a/test/test_roscpp/test/src/stamped_topic_statistics_empty_timestamp.cpp
+++ b/test/test_roscpp/test/src/stamped_topic_statistics_empty_timestamp.cpp
@@ -45,18 +45,77 @@ TEST(TopicStatistics, empty_timestamp_crash_check)
   ros::Publisher pub = nh.advertise<test_roscpp::TestWithHeader>("test_with_empty_timestamp", 0);
   ros::Subscriber sub = nh.subscribe("test_with_empty_timestamp", 0, callback);
 
-  ros::Time start = ros::Time::now();
-  ros::Duration time_to_publish(10.0);
-  while ( (ros::Time::now() - start) < time_to_publish )
-  {
-    test_roscpp::TestWithHeader msg;
-    msg.header.frame_id = "foo";
-    // Don't fill in timestamp so that it defaults to 0.0
-
-    pub.publish(msg);
-    ros::spinOnce();
-    ros::WallDuration(0.01).sleep();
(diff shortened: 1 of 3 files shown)


test/test_roscpp/test/src/stamped_topic_statistics_empty_timestamp.cpp
```diff
@@ -45,18 +45,77 @@ TEST(TopicStatistics, empty_timestamp_crash_check)
...
+      }
+    }
+  } catch(const std::runtime_error & e) {
+    ROS_FATAL_STREAM("Uncaught outside the loop on iter " << i << " with: " << e.what());
+    FAIL();
+  } catch(...) {
+    ROS_FATAL_STREAM("Uncaught outside the loop on iter " << i);
+    FAIL();
   }
+  ROS_FATAL_STREAM("Done testing the message");
```
dirk-thomas, line 118: I think the problem you are trying to identify here has been addressed by #1127.

dirk-thomas:

dirk-thomas:
Thank you for the patch. It was very helpful in addressing the test failures.

* The ros/genpy#79 related issues have been addressed by #1125 (there was another case not covered by this patch).
* The flakyness of the steady timer test has been improved by #1128 (using 9dae982f2a26fd7c850529b25bbd570538d28e0b from this PR).
* The ros/roscpp_core#61 related issue has been addressed by #1127.

kmactavish:
@dirk-thomas Awesome, I couldn't track that one down. Thanks.

dirk-thomas:
> I couldn't track that one down.

I just ran the test executable in `gdb` and the stacktrace immediately revealed the culprit.

PR #1120, opened 2017-08-03
improve SteadyTimer tests

flixr:
Instead of checking when the callback was actually called, check for when it was added to the callback queue.
(Not sure why it was done differently in the WallTimer tests in the first place)...

If I understand things correctly this *should* make the test more reliable and be a better solution than #1118 
So if the tests fail because the callbacks in the queue are delayed, this should fix it (or at least be better). If the source of the sporadic test failures is that the thread actually sleeps too long, this doesn't change anything and thresholds might still need to be loosened up a bit like in #1118 or #1119

### DIFF
--- a/test/test_roscpp/test/src/timer_callbacks.cpp
+++ b/test/test_roscpp/test/src/timer_callbacks.cpp
@@ -74,15 +74,13 @@ class SteadyTimerHelper
     void callback(const SteadyTimerEvent& e)
     {
       bool first = last_call_.isZero();
-      SteadyTime last_call = last_call_;
-      last_call_ = SteadyTime::now();
-      SteadyTime start = last_call_;
+      last_call_ = e.current_real;
 
       if (!first)
       {
-        if (fabsf(expected_next_call_.toSec() - start.toSec()) > 0.1f)
+        if (fabsf(e.current_expected.toSec() - e.current_real.toSec()) > 0.1f)
         {
-          ROS_ERROR("Call came at wrong time (%f vs. %f)", expected_next_call_.toSec(), start.toSec());
+          ROS_ERROR("Call came at wrong time (%f vs. %f)", e.current_expected.toSec(), e.current_real.toSec());
           failed_ = true;
         }
       }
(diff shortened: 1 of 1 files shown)


flixr:
> So if the tests fail because the callbacks in the queue are delayed, this should fix it (or at least be better).

Actually this is not quite correct... `current_real` is also set right before the callback is called and not when it was added to the queue.
Code wise I still think this is nicer than duplicating this info in the test code...

If I see this correctly there are two possible causes for the flaky tests:
- thread sleeps too long and adds the callback to the queue too late
- callbacks were added to the queue in time (within the 0.1s threshold that is checked), but since there are possible quite a few callbacks in the queue it sometimes takes too long until they are actually called.

IMHO there is no way to distinguish the two without changing `SteadyTimerEvent` to add the time when the callback is added to the queue and setting this in `TimerQueueCallback`.

test/test_roscpp/test/src/timer_callbacks.cpp
```diff
@@ -74,15 +74,13 @@ class SteadyTimerHelper
...
       bool first = last_call_.isZero();
-      SteadyTime last_call = last_call_;
-      last_call_ = SteadyTime::now();
-      SteadyTime start = last_call_;
+      last_call_ = e.current_real;
 
       if (!first)
       {
-        if (fabsf(expected_next_call_.toSec() - start.toSec()) > 0.1f)
+        if (fabsf(e.current_expected.toSec() - e.current_real.toSec()) > 0.1f)
```
kmactavish: `fabsf` is for floats, but `toSec` returns a double, use `fabs(...) > 0.1` instead.

kmactavish: I still like splitting up too early, too late into separate checks. Even if you don't relax the too-late threshold. It should never be called too early, so the check can be really tight.

kmactavish (commented):
LGTM. I was trying to minimize my impact on the existing code, but this is definitely cleaner. I think the duration check could still be improved a little though.

dirk-thomas:

kmactavish:
It's not going to build, dependencies broke devel. See #1119 for some of the fixes (it's incomplete). Feel free to push to my branch on that PR to finish it up. I ran out of time over the next few days.

dirk-thomas:
> It's not going to build, dependencies broke devel.

What exactly do you mean with that? It built successfully before and only failed some tests.

I have addressed 2 of the 4 failing tests in #1125 already. Do you think this patch is going to fix one of the two remaining test failures (I thought that was the idea based on the description)?

**Update:** the latest PR build has the same failing tests as the latest devel job. So this patch doesn't seem to improve the failing tests.

kmactavish:
I listed the issues I found in #1119, they are changes in roscpp, which aren't version controlled by this repo.

dirk-thomas:

dirk-thomas:
I used the change from #1119 to reduce the flakyness of the steady timer test and merged it in #1128. Therefore I will close this PR for now. If you think specific parts of the patch are better than the merged patch please open a new PR. Thank you.

dirk-thomas:
I actually rebased this patch in #1129 in the hope that it reduces the flakyness of the test.

PR #1126, opened 2017-08-09
Fix rostopic hz and bw in Python 3

kartikmohta:
The window_size argument was converted to int using string.atoi but it's not present in Python3. Just use int(.) for both Python 2 & 3.

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -426,9 +426,11 @@ class ROSTopicBandwidth(object):
         self.lock = threading.Lock()
         self.last_printed_tn = 0
         self.sizes =[]
-        self.times =[]        
-        self.window_size = window_size or 100
-                
+        self.times =[]
+        if window_size < 0:
+            window_size = 100
+        self.window_size = window_size
+
     def callback(self, data):
         """ros sub callback"""
         with self.lock:
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -1528,7 +1528,7 @@ def _rostopic_cmd_bw(argv=sys.argv):
     from optparse import OptionParser
     parser = OptionParser(usage="usage: %prog bw /topic", prog=NAME)
     parser.add_option("-w", "--window",
-                      dest="window_size", default=None,
+                      dest="window_size", default=-1,
                       help="window size, in # of messages, for calculating rate", metavar="WINDOW")
     options, args = parser.parse_args(args)
     if len(args) == 0:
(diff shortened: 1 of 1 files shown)


dirk-thomas:

dirk-thomas:

dirk-thomas:

dirk-thomas:

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1532,19 +1530,15 @@ def _rostopic_cmd_bw(argv=sys.argv):
     from optparse import OptionParser
     parser = OptionParser(usage="usage: %prog bw /topic", prog=NAME)
     parser.add_option("-w", "--window",
-                      dest="window_size", default=None,
+                      dest="window_size", default=-1,
```
dirk-thomas: Why is this change necessary for Python 3 support? `None` seems to be a perfectly reasonable default value.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -426,9 +426,11 @@ def __init__(self, window_size=100):
         self.lock = threading.Lock()
         self.last_printed_tn = 0
         self.sizes =[]
-        self.times =[]        
-        self.window_size = window_size or 100
-                
+        self.times =[]
+        if window_size < 0:
+            window_size = 100
+        self.window_size = window_size
```
dirk-thomas: Why is this logic change necessary for Python 3 support?

tools/rostopic/src/rostopic/__init__.py
kartikmohta: It's not required but I did this for simplification of the following logic for converting `window_size` to int since `int(None)` is not allowed. I can check for None there instead if that seems better.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1540,11 +1536,7 @@ def _rostopic_cmd_bw(argv=sys.argv):
     if len(args) > 1:
         parser.error("you may only specify one input topic")
     try:
-        if options.window_size:
-            import string
-            window_size = string.atoi(options.window_size)
-        else:
-            window_size = options.window_size
+        window_size = int(options.window_size) if options.window_size else None
```
dirk-thomas: This will set the variable to `None` in case the option is `0`. So this should check for `None` explicitly:

```
int(options.window_size) if options.window_size is not None else None
```

kartikmohta: Does a window_size of 0 even make sense for `rostopic bw`? Just falling back to the default might be better in that case.

dirk-thomas: If that is what the user asked for why do something else? This magic would then also need to be documented.

kartikmohta: The user is always right 🙂

dirk-thomas: > The user is always right

Oh no, but we give him what was asked for anyway :wink: :fire:

dirk-thomas:

PR #1134, opened 2017-08-12
xmlrpc_manager: use SteadyTime for timeout

flixr:
Use the SteadyTime instead of WallTime when checking if a cached client is dead.

### DIFF
--- a/clients/roscpp/include/ros/xmlrpc_manager.h
+++ b/clients/roscpp/include/ros/xmlrpc_manager.h
@@ -80,7 +80,7 @@ public:
   }
 
   bool in_use_;
-  ros::WallTime last_use_time_; // for reaping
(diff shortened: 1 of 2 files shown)


dirk-thomas:
How will this affect the behavior when e.g. using sim time? I could imagine that existing connections would be ended when simulation is paused.

flixr:
> How will this affect the behavior when e.g. using sim time?

Not at all as long as the system time doesn't change.

> I could imagine that existing connections would be ended when simulation is paused.

I'm not exactly sure when a new xmlrpc client is needed, but since it was using WallTime before, it would have had the same effect.

So the only difference is that you don't get a timeout if you set the system time forward.

dirk-thomas:
Nvm. Since it was using wall time before I wasn't affected by sim time anyway.

Thanks.

PR #1136, opened 2017-08-12
Replaced deprecated lz4 function call

racko:


racko:
lz4 1:1.7.5-1

/usr/include/lz4.h:
`LZ4_DEPRECATED("use LZ4_compress_default() instead") int LZ4_compress_limitedOutput (const char* source, char* dest, int sourceSize, int maxOutputSize);`

dirk-thomas:
Thanks.

PR #1137, opened 2017-08-12
Removed deprecated dynamic exception specifications

racko:
```
.../ros_comm/tools/rosbag_storage/include/rosbag/bag_player.h:80:42: warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]
   BagPlayer(const std::string &filename) throw(BagException);
                                          ^~~~~
.../ros_comm/tools/rosbag_storage/src/bag_player.cpp:8:48: warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]
 BagPlayer::BagPlayer(const std::string &fname) throw(BagException) {
                                                ^~~~~
```

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2010/n3051.html says it all:

>Exception specifications have proven close to worthless in practice, while adding a measurable overhead to programs. The feature should be deprecated. The one exception to the rule is the empty throw specification which could serve a legitimate optimizing role if the requirement to call the runtime unexpected mechanism was relaxed in this case.

### DIFF
--- a/tools/rosbag_storage/include/rosbag/bag_player.h
+++ b/tools/rosbag_storage/include/rosbag/bag_player.h
@@ -77,7 +77,7 @@ class BagPlayer
 {
 public:
   /* Constructor expecting the filename of a bag */
-  BagPlayer(const std::string &filename) throw(BagException);
+  BagPlayer(const std::string &filename);
 
   /* Register a callback for a specific topic and type */
   template<class T>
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please provide a description with your PR to clarify why this patch is desired. The exception specification has some value.

racko:
Updated

I'm not sure if you get notified automatically :)

dirk-thomas:
> I'm not sure if you get notified automatically :)

GitHub only sends emails for new comments and new commits. Edited comments and amended commits don't generate a notification. So this comment was good to notify me :+1:

I agree that it makes sense to remove the exception specification. I also removed it from `rostime`: ros/roscpp_core#64

PR #1138, opened 2017-08-12
Improved whitespace to fix g++ 7 warning

racko:
.../ros_comm/tools/rosbag_storage/src/view.cpp:249:5: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
     if ((bag.getMode() & bagmode::Read) != bagmode::Read)
     ^~
.../ros_comm/tools/rosbag_storage/src/view.cpp:252:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
  boost::function<bool(ConnectionInfo const*)> query = TrueQuery();
  ^~~~~


dirk-thomas:
Please limit the patch to required changes to get rid of warning but avoid all other whitespace changes. While it would improve the code style it adds unnecessary extra effort when backporting patches between different branches.

racko:
Updated

dirk-thomas:
Thanks.

PR #1140, opened 2017-08-14
Don't direct users to build rosout with rosmake.

mikepurvis:


tools/roslaunch/src/roslaunch/launch.py
```diff
@@ -540,7 +540,7 @@ def launch_node(self, node, core=False):
         except roslaunch.node_args.NodeParamsException as e:
             self.logger.error(e)
             if node.package == 'rosout' and node.type == 'rosout':
-                printerrlog("\n\n\nERROR: rosout is not built. Please run 'rosmake rosout'\n\n\n")
+                printerrlog("\n\n\nERROR: rosout is not built. Please run 'catkin_make rosout'\n\n\n")
```
dirk-thomas: Is it a good idea to recommend calling `catkin_make` with the target `rosout`? The user could be using any other build tool (e.g. `catkin_make_isolated`, `catkin build`, etc.).

asmodehn: It seems to me that this message is addressed at a developer building ROS core libraries. I would think the official way to build ROS core libraries (and the only supported way) is `catkin_make_isolated` ?
Ref : http://wiki.ros.org/Installation/Source

dirk-thomas: That might be the one way mentioned in the wiki but there are many different ways. Since we don't know what the user is using I would rather see the message being generic and telling the user to build that package without a reference to a specific command.

mikepurvis: In that case, let's just drop the special case; the generic one is just as applicable to rosout being missing.

dirk-thomas: Sounds good to me.

PR #1141, opened 2017-08-14
Fix stack frame identification in rospy logging.

mikepurvis:
A possible solution to the issue introduced in #948.

Definitely hacky; should almost certainly be covered by some tests.

----

Manual verification was by means of the following trivial script:

```
import rospy

rospy.init_node('test')

r = rospy.Rate(10)
   rospy.loginfo('regular logging')
   rospy.loginfo_throttle(0.5, 'throttled logging')
   rospy.loginfo('named logging', logger_name='foobar')
   r.sleep()
```

Running this, I can `rostopic echo /rosout` and verify that all the `file:` fields point to this script name.

However, in the course of playing with this, I noted that the following does _not_ work:

```
rospy.loginfo('named throttled logging', logger_name='foobar', logger_throttle=0.5)
```

It crashes on account of the throttle code's assumption that the caller is two stack frames up, whereas in this case it is only one.

FYI @ggallagher01

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -83,7 +83,7 @@ class RospyLogger(logging.getLoggerClass()):
             pass
 
         if sys.version_info > (3, 2):
-            # Dummy last argument to match Python3 return type.
+            # Dummy last argument to match Python3 return type
             return co.co_filename, f.f_lineno, func_name, None
         else:
             return co.co_filename, f.f_lineno, func_name
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -68,10 +68,10 @@ class RospyLogger(logging.getLoggerClass()):
             f = f.f_back
 
         # Jump up two more frames, as the logger methods have been double wrapped.
-        if f.f_back:
+        if f.f_back and f.f_code and f.f_code.co_name == '_base_logger':
             f = f.f_back
-        if f.f_back:
-            f = f.f_back
+            if f.f_back:
+                f = f.f_back
         co = f.f_code
         func_name = co.co_name
 
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -67,17 +67,31 @@ class RospyLogger(logging.getLoggerClass()):
             if filename != file_name or f.f_lineno != lineno or co.co_name != func_name:
                 f = f.f_back
                 continue
-            # we found  the correct frame, now extending func_name with class name
-            try:
-                class_name = f.f_locals['self'].__class__.__name__
-                func_name = '%s.%s' % (class_name, func_name)
-            except KeyError:  # if the function is unbound, there is no self.
-                pass
             break
+
+        # Jump up one more frame, as the underlying logger functions have been wrapped.
+        f = f.f_back
+        co = f.f_code
+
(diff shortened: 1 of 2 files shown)


mikepurvis:
Okay, so this is I believe in a shippable state as far as unbreaking the head of devel and passing the smoke test laid out in the first post, but it's still not super great. The new deal is:

- All the `logXXX[_YYYY]` methods now call `_base_logger`.
- Therefore, the "user" method is always two stack frames away from the underlying method.
- So we can start from the logging-derived one, and jump up two frames, and be good to go.

However, it sucks because:

- We're now assuming that everything we process is coming through these layers of indirection, which will especially not be the case if we're picking up logging from a third-party module (but isn't even the case internally— there are direct log method calls right in rospy itself, hence the if-checks).

I think it's probably mergeable as-is, but I would propose one or more of the following future directions:

- Move all these wrapper functions off to their own module.
- Generate them using some kind of function that can maybe tag or track them somehow.
- Use that information when walking up the stack, and if no tagged (or logging module) method is found, fall back on the logic from the original class.
- Eliminate the `_base_logger` wrapper altogether, and figure out a way to supply throttling/once-ing functionality using the filtering scheme built into to python's logging method.

dirk-thomas:
The updated approach looks good to me to get this working and :ship: again.

I added 8021d9e6077272ffdc422468b83b2014a453a6f7 to fix the `log*_once` functions.

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -60,24 +60,33 @@ def findCaller(self, dummy=False): # Dummy second arg to match Python3 function
...
-            break
+            if filename == file_name and f.f_lineno == lineno and co.co_name == func_name:
+                break
+            f = f.f_back
+
+        # Jump up two more frames, as the logger methods have been double wrapped.
+        if f.f_back:
+            f = f.f_back
+        if f.f_back:
+            f = f.f_back
```
dirk-thomas: How about making this "move-two-frames-up" conditional on being "our" use case with the `_base_logger`?

```
if f.f_back and f.f_code and f.f_code.co_name == '_base_logger':
    f = f.f_back
    if f.f_back:
        f = f.f_back
```

dirk-thomas:
`test_roslogging.test_rosconsole__logging_format` still fails.

tools/rosgraph/src/rosgraph/roslogging.py
mikepurvis: That's a better idea. FWIW, I can't really see a case where it would be only one layer of wrapping, so the following _should_ be safe:

```
if f.f_code and f.f_code.co_name == '_base_logger':
    f = f.f_back.f_back
```

dirk-thomas: Better be safe than sorry: 64e284ed2c8105657e391a96e7782dab79abc697

mikepurvis:
Looks like a legit test failure, though. Expected regex was:

```
INFO on module [0-9]*\\.[0-9]* [0-9]* rosout /tmp/catkin_workspace/src/ros_comm/tools/rosgraph/test/test_roslogging.py [0-9]* <module> /unnamed [0-9]*\\.[0-9]*
```

Actual was:

```
INFO on module 1502806685.841191 139998642550528 rosout /usr/lib/python2.7/dist-packages/nose/importer.py 47 Importer.importFromPath /unnamed 1502806685.841208
```

This is likely to be fixed by 64e284ed2c8105657e391a96e7782dab79abc697.

dirk-thomas:
Thank you @mikepurvis. This looks good to go.

PR #1142, opened 2017-08-14
only use CLOCK_MONOTONIC if not on osx

ffurrer:
I'm not sure if this is the cleanest solution, but it resolves the following compilation error:
> /usr/local/include/boost/thread/pthread/condition_variable_fwd.hpp:42:13: error: use of undeclared identifier 'pthread_condattr_setclock'; did you mean 'pthread_condattr_setpshared'?
            pthread_condattr_setclock(&attr, CLOCK_MONOTONIC);
            ^~~~~~~~~~~~~~~~~~~~~~~~~
            pthread_condattr_setpshared

### DIFF
--- a/clients/roscpp/src/libros/internal_timer_manager.cpp
+++ b/clients/roscpp/src/libros/internal_timer_manager.cpp
@@ -25,8 +25,10 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-// make sure we use CLOCK_MONOTONIC for the condition variable
+// Make sure we use CLOCK_MONOTONIC for the condition variable if not Apple.
+#ifndef __APPLE__
 #define BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
+#endif
 
 #include "ros/timer_manager.h"
 #include "ros/internal_timer_manager.h"
--- a/clients/roscpp/src/libros/steady_timer.cpp
+++ b/clients/roscpp/src/libros/steady_timer.cpp
@@ -25,8 +25,10 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-// make sure we use CLOCK_MONOTONIC for the condition variable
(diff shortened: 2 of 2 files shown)


dirk-thomas:
@flixr since you have contributed this code part recently can you please comment on this patch.

flixr:
Hm.. this actually seems to be a "bug" in boost, it doesn't handle the case that `pthread_condattr_setclock` is apparently not available under OSX...
This is a real shame, as the SteadyTimer then won't work (or rather use the system time instead of monotonic time).
But in the absence of a better solution, this seems ok as it will at least make it compile again...

@ffurrer what boost version are you using?

dirk-thomas:
Falling back to system clock when a steady clock is requested on OSX doesn't sounds like a good behavior. If there is no "simple" fix it might require a custom implementation for OSX to achieve feature parity.

clalancette:
FWIW, I just did some research on this topic for ROS2: https://github.com/ros2/rcutils/issues/43#issuecomment-320954506 .  We could use one of the solutions I pointed out there.

dirk-thomas:
I will merge this to fix the regression introduced by the switch to the steady time. Please create a follow up PR to support steady time on OS X (#1143).

flixr:
Actually it would make sense to also `#ifdef` out the specialization in steady_timer.cpp, so it uses the same implementation as for WallTime (with the time jump detection) on OSX.

dirk-thomas:
> Actually it would make sense to also `#ifdef` out the specialization in steady_timer.cpp, so it uses the same implementation as for WallTime (with the time jump detection) on OSX.

I would like to see a PR to implement steady for OS X rather sooner than later. Therefore I don't think we need a patch for the hopefully very short time period in between.

ffurrer:
@flixr I'm using 1.64:
> /usr/local/Cellar/boost/1.64.0_1 (12,628 files, 395.7MB) *
  Poured from bottle on 2017-04-30 at 08:38:38

Issue #946, opened 2016-12-06
Create public Python function for rostopic list in rostopic package


gordonbrander:
Currently, the only function `rostopic` has for listing topics [is private](https://github.com/ros/ros_comm/blob/kinetic-devel/tools/rostopic/src/rostopic/__init__.py#L1181), and it is only useful for the cli usecase because it calls `print`.

`rosgraph.Master("/rostopic").getSystemState()` may be used instead of rostopic, but this seems like an abuse of an API which is meant for graph image generation.

I would like to propose that a public `rostopic.list` function be added to the `rostopic` package, listing all topic names (and perhaps types).

mikepurvis:
I don't disagree with you, but FWIW, what you suggest is exactly how rosbridge handles inspecting the system:

https://github.com/RobotWebTools/rosbridge_suite/blob/a7a2e1a473a1da1acf252d2496fa7db0f211a29f/rosapi/src/rosapi/proxy.py#L49-L56

zbynekwinkler:
Should this be closed since get_topic_list() has been merged?

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #1146, opened 2017-08-18
Created public get_topic_list() function for use in other scripts.

EliteMasterEric:
Resolves issue #946.

The only function in rostopic for listing the available topics is built exclusively for the command line interface; this adds a public function for other applications that want a list of topics.

Usage:
```
import rostopic
publishers, subscribers = rostopic.get_topic_list()
for topic, type, node_count in publishers:
    print("Publisher: %s, [%s] (%d nodes)" % (topic, type, node_count))
for topic, type, node_count in subscribers:
    print("Subscriber: %s, [%s] (%d nodes)" % (topic, type, node_count))
```


dirk-thomas:
After extracting the print-agnostic part of the existing functionality as you did in the current patch please also update the existing code to use this new function (rather than duplicating this logic). Please also consider adding a test for the new function.

EliteMasterEric:
@dirk-thomas Changes have been made, removing redundancy and adding a test to the rostopic test script.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1155,6 +1155,31 @@ def topic_type(t, topic_types):
...
+            return matches[0]
+        return 'unknown type'
+
+    # Return an array of tuples; (<topic>, <type>, <node_count>)
+    master = rosgraph.Master('/rostopic')
+    state = master.getSystemState()
+    topic_types = _master_get_topic_types(master)
+    
+    pubs, subs, _ = state
+    publishers = []
```
dirk-thomas: This line is overwriting the function parameter.

Same below for `subscribers`.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1155,6 +1155,31 @@ def topic_type(t, topic_types):
...
+    if publishers:
+        for topic, nodes in pubs:
+            publishers.append(topic, topic_type(topic, topic_types), len(nodes))
```
dirk-thomas: `list.append()` only accepts a single argument.

Same below.

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1202,14 +1227,7 @@ def _rostopic_list(topic, verbose=False,
...
     try:
-        state = master.getSystemState()
-
-        pubs, subs, _ = state
-        if topic:
-            # filter based on topic
-            topic_ns = rosgraph.names.make_global_ns(topic)        
-            subs = (x for x in subs if x[0] == topic or x[0].startswith(topic_ns))
-            pubs = (x for x in pubs if x[0] == topic or x[0].startswith(topic_ns))
+        pubs, subs = get_topic_list()
```
dirk-thomas, line 1222: The new function returns different information than `getSystemState` before. I doubt that all following code continues to work as is.

dirk-thomas (commented):
Please make sure to run all code parts you have changed in all possible variations to make sure the patch actually works (which is currently not the case).

PR #1147, opened 2017-08-21
Don't need to see the full stack trace for ResourceNotFound

lucasw:
The user typed in something wrong, or is missing a package, or hasn't sourced the right setup script.

The example that prompted this was using $(find package) where that package wasn't on the path:

```
<?xml version="1.0"?>
<launch>
  <param name="robot_description" command="$(find xacro)/xacro.py '$(find bad_package)/urdf/foo.xacro'" />
</launch>
```

The result is a screen full of traceback, followed by a screen full of ROS path output, the line that explains which package wasn't found was sandwiched in-between and is easy to miss.

This eliminates the traceback (but still a lot of ROS path with catkin tools with a workspace with a lot of packages).

Are there other exceptions that are similar that ought to be handled the same?

It would be nice to consolidate the duplicate code in ValueError and RLException above this.

### DIFF
--- a/tools/roslaunch/src/roslaunch/__init__.py
+++ b/tools/roslaunch/src/roslaunch/__init__.py
@@ -36,6 +36,7 @@ from __future__ import print_function
 
 import os
 import logging
+import rospkg
 import sys
 import traceback
 
@@ -315,6 +316,12 @@ def main(argv=sys.argv):
     except ValueError as e:
         # TODO: need to trap better than this high-level trap
         roslaunch_core.printerrlog(str(e))
+        roslaunch_core.printerrlog('The traceback for the exception was written to the log file')
+        if logger:
+            logger.error(traceback.format_exc())
+        sys.exit(1)
+    except rospkg.ResourceNotFound as e:
+        roslaunch_core.printerrlog("Resource not found: " + str(e))
         roslaunch_core.printerrlog('The traceback for the exception was written to the log file')
         if logger:
             logger.error(traceback.format_exc())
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for the patch.

> Are there other exceptions that are similar that ought to be handled the same?

I don't know the answer out of my head. You will need to trace all function calls recursively to find out.

> It would be nice to consolidate the duplicate code in ValueError and RLException above this.

Same goes for the newly added case. Do you want to update the PR for that? I don't mind to merge it as-is or with all three blocks merged.

lucasw:
`except(RLException, ValueError, etc) as e:` would be cleaner in cfcd0d7, but I like the ability to have custom prefix text to the error output (though all the prefixes are basically the name of the exception type, which could be had from `repr` + some formatting, but maybe in the future something more verbose would be desired?).

What is a good way to trigger an RLException or a ValueError?
The first thing I tried was `<param name="foo" command="" />` which generates an IndexError, which could be added here.

(Is there already a discussion about adding the ability to provide the name and line number of a file where it generates an exception or otherwise a problem is noticed?)

lucasw:
Generate an RLException:

```
<param name="foo" command="t" />
```
(or many other variations where param is spelled wrong, or name, or command)

dirk-thomas:
The current patch looks good to me. Do you want to keep iterating on it to cover more cases or should it be merged as is?

> Is there already a discussion about adding the ability to provide the name and line number of a file where it generates an exception or otherwise a problem is noticed?

I am not aware of any effort towards such a feature. One problem for this will be that test coverage is very low and the code based not clean and therefore for any change side effects / regressions are highly likely. That was the main reason why other patches to `roslaunch` haven't been merged in the past.

lucasw:
Let's merge this, I'll make a separate pr for other cases later, and hopefully also generate some tests for them, but it may take a while.

Also I may not have been clear above (though your answer sounds applicable regardless): the files I want names and line numbers of errors out of are the .launch files.

dirk-thomas:
> the files I want names and line numbers of errors out of are the .launch files.

I guessed so :wink: Sounds like a nice feature.

PR #1149, opened 2017-08-21
use not deprecated console_bridge macros and undefine the deprecated ones

### LINKED ISSUE #1145: rospack.h and bag.h cannot both be included in the same compilation unit
A simple program like:
```cpp
#include "rosbag/bag.h"
#include "rospack/rospack.h"

int main(int argc, char** argv) { return 0; }
```

Fails to compile with this error:
```
[100%] Building CXX object rospack_test/CMakeFiles/rospack_test_node.dir/src/main.cpp.o
In file included from /home/jstn/asdf_ws/src/rospack_test/src/main.cpp:2:0:
/opt/ros/indigo/include/rospack/rospack.h:539:1: error: expected identifier before string constant
                  bool append_errno = false);
 ^
/opt/ros/indigo/include/rospack/rospack.h:539:1: error: expected ‘,’ or ‘...’ before string constant
In file included from /opt/ros/indigo/include/rosbag/bag.h:63:0,
                 from /home/jstn/asdf_ws/src/rospack_test/src/main.cpp:1:
/opt/ros/indigo/include/rospack/rospack.h:538:10: error: invalid use of ‘::’
     void logWarn(const std::string& msg,
          ^
In file included from /home/jstn/asdf_ws/src/rospack_test/src/main.cpp:2:0:
/opt/ros/indigo/include/rospack/rospack.h:547:1: error: expected identifier before string constant
                   bool append_errno = false);
 ^
/opt/ros/indigo/include/rospack/rospack.h:547:1: error: expected ‘,’ or ‘...’ before string constant
In file included from /opt/ros/indigo/include/rosbag/bag.h:63:0,
                 from /home/jstn/asdf_ws/src/rospack_test/src/main.cpp:1:
/opt/ros/indigo/include/rospack/rospack.h:546:10: error: invalid use of ‘::’
     void logError(const std::string& msg,
          ^
make[2]: *** [rospack_test/CMakeFiles/rospack_test_node.dir/src/main.cpp.o] Error 1
make[1]: *** [rospack_test/CMakeFiles/rospack_test_node.dir/all] Error 2
make: *** [all] Error 2
```

However, I was able to separate out the rospack functionality into a separate library, and link it with my main executable (which used rosbag) and successfully compile my code.

Tested on Ubuntu 14.04 with ROS Indigo, using catkin and catkin_tools.

dirk-thomas:
Fixes #1145.

This patch avoids the side effect of defining these macro which are often subject to collisions in a ROS header.

### DIFF
--- a/tools/rosbag_storage/include/rosbag/bag.h
+++ b/tools/rosbag_storage/include/rosbag/bag.h
@@ -61,6 +61,18 @@
 #include <boost/iterator/iterator_facade.hpp>
 
 #include "console_bridge/console.h"
+#if defined logDebug
+# undef logDebug
+#endif
+#if defined logInform
+# undef logInform
+#endif
+#if defined logWarn
+# undef logWarn
+#endif
+#if defined logError
+# undef logError
+#endif
 
 namespace rosbag {
 
@@ -569,7 +581,7 @@ void Bag::doWrite(std::string const& topic, ros::Time const& time, T const& msg,
 
         // Check if we want to stop this chunk
         uint32_t chunk_size = getChunkOffset();
-        logDebug("  curr_chunk_size=%d (threshold=%d)", chunk_size, chunk_threshold_);
+        CONSOLE_BRIDGE_logDebug("  curr_chunk_size=%d (threshold=%d)", chunk_size, chunk_threshold_);
         if (chunk_size > chunk_threshold_) {
             // Empty the outgoing chunk
             stopWritingChunk();
@@ -604,7 +616,7 @@ void Bag::writeMessageDataRecord(uint32_t conn_id, ros::Time const& time, T cons
     seek(0, std::ios::end);
     file_size_ = file_.getOffset();
 
-    logDebug("Writing MSG_DATA [%llu:%d]: conn=%d sec=%d nsec=%d data_len=%d",
+    CONSOLE_BRIDGE_logDebug("Writing MSG_DATA [%llu:%d]: conn=%d sec=%d nsec=%d data_len=%d",
               (unsigned long long) file_.getOffset(), getChunkOffset(), conn_id, time.sec, time.nsec, msg_ser_len);
 
     writeHeader(header);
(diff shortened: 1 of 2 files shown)


tools/rosbag_storage/include/rosbag/bag.h
```diff
@@ -569,7 +581,7 @@ void Bag::doWrite(std::string const& topic, ros::Time const& time, T const& msg,
 
         // Check if we want to stop this chunk
         uint32_t chunk_size = getChunkOffset();
-        logDebug("  curr_chunk_size=%d (threshold=%d)", chunk_size, chunk_threshold_);
+        CONSOLE_BRIDGE_logDebug("  curr_chunk_size=%d (threshold=%d)", chunk_size, chunk_threshold_);
```
clalancette, line 584: Unfortunately, this doesn't work on Indigo.  From what I can tell, /usr/include/console_bridge/console.h doesn't have the `CONSOLE_BRIDGE_log*` macros on Indigo.  I have an upcoming patch which I'll push here which defines them for the `rosbag/bag.h` header file.  It's not a generic solution (like having it in `console_bridge/bridge.h` would be), but I'm not sure if we want to introduce that this late into Indigo.  Let me know what you think.

dirk-thomas: This patch is targeting `lunar-devel` and there is no reason to backport this into Indigo. Therefore I don't think the second commit is necessary.

clalancette: Oh, sorry, I missed that.  I got confused because the initial bug report was for Indigo.

Alright, so for Lunar, I agree with your original patch (I'll revert mine off of this branch).  However, that still leaves the original bug report about Indigo.  It sounds like you want to just leave this be on Indigo; is that the case?  If so, once we merge this patch I'll close the original bug report saying that it won't be fixed back in Indigo, but will be fixed going forward.  Sound reasonable?

dirk-thomas: > It sounds like you want to just leave this be on Indigo; is that the case?

Yes, the version of `console_bridge` in Indigo doesn't have the namespaces macros. Therefore I don't think every higher level package needs to work around this upstream problem.

The patch could be backported to Kinetic if the `console_bridge` version on the platforms targeted by that support this.

PR #1155, opened 2017-08-30
add stealth mode for topic_tools/relay

furushchev:
This PR is for adding "stealth mode" to `topic_tools/relay` node.
This is almost the same as a normal relay but it relays only if topic specified by param `~monitor_topic` is being subscribed by any other nodes except itself.

When input topic of relay and monitor topic are pointing to the same topic (this is by default), it looks like the relay node subscribes input topic only when other nodes also subscribe the input topic, which means subscribing without increment subscription count.

This feature is useful for publishing topic for visualization with lazy transport (c.f. https://github.com/ros/nodelet_core/blob/indigo-devel/nodelet_topic_tools/include/nodelet_topic_tools/nodelet_lazy.h ).
For example, lazy-transport-enabled nodes don't subscribe input topics unless other nodes subscribe the output topics of the nodes, but some nodes are expected to always subscribe topics (e.g. `RViZ`) and this breaks "laziness" of the nodes.
By putting `relay` node with stealth mode between `RViZ` and lazy node, this issue will be solved.

### DIFF
--- a/tools/topic_tools/src/relay.cpp
+++ b/tools/topic_tools/src/relay.cpp
@@ -43,9 +43,12 @@ ros::NodeHandle *g_node = NULL;
 bool g_advertised = false;
 string g_input_topic;
 string g_output_topic;
+string g_monitor_topic;
 ros::Publisher g_pub;
 ros::Subscriber* g_sub;
+ros::Timer g_timer;
 bool g_lazy;
+bool g_stealth;
 ros::TransportHints g_th;
 
 void conn_cb(const ros::SingleSubscriberPublisher&);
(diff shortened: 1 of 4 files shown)


mikepurvis:
This seems reasonable to me. :+1:

tools/topic_tools/src/relay.cpp
```diff
@@ -43,24 +43,38 @@ ros::NodeHandle *g_node = NULL;
...
 
 void subscribe()
 {
+  ROS_DEBUG("subscribe");
   g_sub = new ros::Subscriber(g_node->subscribe(g_input_topic, 10, &in_cb, g_th));
 }
 
+void unsubscribe()
+{
+  ROS_DEBUG("unsubscribe");
```
dirk-thomas: Should these debug messages be removed?

dirk-thomas (commented):
The new test doesn't seem to be registered anywhere and therefore does not seem to be run?

furushchev:
@mikepurvis Thanks for a warm comment! :+1: 

@dirk-thomas Thanks for review and nice catch! Registered new test and removed debug logs.

dirk-thomas:
Looks good to me: http://build.ros.org/job/Lpr__ros_comm__ubuntu_xenial_amd64/481/testReport/__main__/TestRelayStealth/

Thank you for the patch. Please add some documentation about this new feature to the wiki page.

furushchev:
@dirk-thomas Sorry for late, updated documentation http://wiki.ros.org/topic_tools/relay :+1:

dirk-thomas:
Thank you for adding the docs. I added a note that this feature is new in Lunar (http://wiki.ros.org/action/diff/topic_tools/relay?action=diff&rev1=18&rev2=19).

furushchev:
@dirk-thomas Thank you too!

PR #1156, opened 2017-08-31
Fix publishing of selected topics from bag file

kartikmohta:
Selecting a subset of topics to publish from a bag file is not working, seems to be a regression from d1d1711.
```
$ rosbag play --topics /tf test.bag
Usage: rosbag play BAGFILE1 [BAGFILE2 BAGFILE3 ...]
rosbag: error: You must specify at least 1 bag file to play back.
```
```
$ rosbag play test.bag --topics /tf
[ INFO] [1504144739.942816979]: Opening --rate-control-max-delay
[FATAL] [1504144739.943025646]: Error opening file: --rate-control-max-delay
```


dirk-thomas:
Thank you for the fix.

Issue #1078, opened 2017-06-12
Topic subscription scalability problem


guillaumeautran:
Looking into understanding the performance of rosbag recorder, I stumble over something which seems to be a performance bottleneck. The actual usecase is to record about >300 topics with varying publication rate. The test setup used for performance measurement and profiling is a single rosbag recorder recording 309 active topics over a period of 2 minutes with no compression. The run is profiled using VTune and profiler data analyzed.

The performance of the existing topic monitoring is shown in the following graph:  
![record_poll_100ms](https://user-images.githubusercontent.com/13353403/27055066-32ec1d5e-4f91-11e7-9db4-756a61035359.png)
Based on those findings, the top CPU usage of the rosbag recorder goes to the ros::PollSet class. This class is polling all file-descriptors which is known to be an inefficient strategy when monitoring over 100 sockets. The poll system call handles descriptors in a linear way. (source: https://daniel.haxx.se/docs/poll-vs-select.html)

In addition to my investigation, I've also implemented a event based implementation (using libev) as a comparison and also profiled the application which yield much better performance as seen in the graph: 
![record_ev](https://user-images.githubusercontent.com/13353403/27055374-37d5320a-4f92-11e7-8770-0f8ffee4e8d3.png)

So, I'd like to ask if anyone has already looked into asynchronous topic socket monitoring and if not, would a modification like the one proposed be of interest?

mikepurvis:
(Guillaume is a colleague at Clearpath)

Looks like the current implementation is here: https://github.com/ros/ros_comm/blob/lunar-devel/clients/roscpp/src/libros/poll_set.cpp

Which is ultimately backed by the `poll_sockets` function here: https://github.com/ros/ros_comm/blob/247459207e20c1da109fc306e58b84d15c4107bd/clients/roscpp/src/libros/io.cpp#L104

There's a pretty extensive Win32-specific implementation already in there, so I'd be nervous about proposals to switch wholesale to a cross-platform solution such as `libev` or `boost::asio`. However, adding a compile-time switch that uses the [epoll functions](https://linux.die.net/man/4/epoll) instead of poll seems like it could be a minimally-invasive option that would improve the situation on epoll-supporting platforms (checking the availability of epoll from CMake is [easy to do](https://github.com/skroll/cmake-modules/blob/master/modules/CheckEpoll.cmake)).

@dirk-thomas Preferences on how to move this forward?

dirk-thomas:
[...]

PR #1162, opened 2017-09-12

guillaumeautran:
Switching to using epoll system calls to improve performance of the topic polling code by a factor of 2. Addresses #1078.

Some previous discussion/review/numbers here: https://github.com/clearpathrobotics/ros_comm/pull/12

### DIFF
--- a/clients/roscpp/CMakeLists.txt
+++ b/clients/roscpp/CMakeLists.txt
@@ -60,11 +60,15 @@ catkin_package(
 
 include(CheckIncludeFiles)
 include(CheckFunctionExists)
+include(CheckCXXSymbolExists)
 
 # Not everybody has <ifaddrs.h> (e.g., embedded arm-linux)
 CHECK_INCLUDE_FILES(ifaddrs.h HAVE_IFADDRS_H)
 # Not everybody has trunc (e.g., Windows, embedded arm-linux)
 CHECK_FUNCTION_EXISTS(trunc HAVE_TRUNC)
+# Not everybody has epoll (e.g., Windows, BSD, embedded arm-linux) 
+CHECK_CXX_SYMBOL_EXISTS(epoll_wait "sys/epoll.h" HAVE_EPOLL)
+
 
 # Output test results to config.h
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/libros/config.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h)
--- a/clients/roscpp/include/ros/io.h
+++ b/clients/roscpp/include/ros/io.h
@@ -138,6 +138,8 @@ namespace ros {
   typedef struct pollfd socket_pollfd;
 #endif
 
+  typedef boost::shared_ptr<std::vector<socket_pollfd> > pollfd_vector_ptr;
+
 /*****************************************************************************
 ** Functions
 *****************************************************************************/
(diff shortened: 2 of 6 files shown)


mikepurvis:
This has been working well for us across several weeks of testing. :+1:

guillaumeautran:
I'm not sure to understand the unit test here. The test add 100 threads who all register the same FD for POLLIN and POLLOUT. The previous implementation behaviour was that a single handler was going to be called (single FD) while the new code behaviour calls ALL of the registered hander on that socket. In this case, all 100 threads will be called granted the ability to write to the socket and all 100 threads will be called to read from the socket. 

I am not sure how the code is expected to behave in the case where multiple handlers are registering for the same socket...

Test code is here: https://github.com/ros/ros_comm/blob/lunar-devel/test/test_roscpp/test/test_poll_set.cpp#L285

guillaumeautran:
After further investigations, it appears that the original code (prior to my change) was behaving differently when multiple handlers are registered for the same socket. Based on the code here: [poll_set.cpp#L76](https://github.com/ros/ros_comm/blob/lunar-devel/clients/roscpp/src/libros/poll_set.cpp#L76) multiple handlers are added to the internal data structure but when registering for events on the socket, [poll_set.cpp#L125](https://github.com/ros/ros_comm/blob/lunar-devel/clients/roscpp/src/libros/poll_set.cpp#L125), only the first handler is marked to receive the event. 
This actually IMHO is a bug of the original implementation because the event handler call that is supposed to receive one type of event may not be able to handle all events correctly (write event on a read-only handler, etc). In addition, deleting the event from the handler will remove the flag from the first handler found in the list, [poll_set.cpp#L144](https://github.com/ros/ros_comm/blob/lunar-devel/clients/roscpp/src/libros/poll_set.cpp#L144), which might not be correct either as that particular handler may still be looking to receive that particular event.

IMHO, the proper behaviour with multiple event handlers should really be such that the handler registering for a particular event should be the one receiving that event. For example, a single handler could be used for socket reads while a separate handler could be used for socket writes. 

Because the code still contains all three versions (Windows, poll, epoll), I suggest we disable this particular unit test and leave it as that.

mikepurvis:
@dirk-thomas Thoughts on the unit test discussion above?

mikepurvis:
Actually, going to close this now so that I can resume use of the `lunar-edge` branch for testing. @guillaumeautran Please make a new PR from a specific feature branch, and we can continue the conversation there.

dirk-thomas:
Since this patch breaks ABI should it target ROS Melodic?

mikepurvis:
Is it mutating any symbols which are linked from user code? I think it's all internal stuff that's being changed, so that should be safe, right?

dirk-thomas:
> Is it mutating any symbols which are linked from user code? I think it's all internal stuff that's being changed, so that should be safe, right?

How can we assume that the changed headers (which are public and nowhere marked as impl / internal) are not used by any userland code?

mikepurvis:
I suppose we cannot be completely assured of that, but the two affected headers (`ros/poll_set.h` and `ros/io.h`) are pretty clearly deep implementation details, and there's not a puff about either of them in the wiki API documentation, nor are the classes involved meant to be extended— `PollSet` has no virtual methods, and there's no other functions which accept a `PollSet` instance.

I think the main risk with `PollSet` is that there could be a member of that type in an API class, which would cause the size of that class to change. However, on a quick grep through the repo, it looks like all instances (TransportTCP, TransportUDP, PollManager), are maintained as pointers or references, so those are safe from size change issues.

Anyway, this is mostly academic— we can actively test/ship this for now, and if it's doesn't merge until `melodic-devel`, that's acceptable to us. But it is also a change which could really benefit a lot of Kinetic/Lunar users, and I wouldn't want to hold it back on an unlikely technicality.

PR #1164, opened 2017-09-13
changed error message for single threaded spinner 

jgueldenstein:
added hint to use MultiThreadedSpinner when starting a single threaded spinner failed because it was on the same queue as a previously started AsyncSpinner.

This behavior was tolerated in kinetic but recommended against.

We would also like to add a similar message to the kinetic-devel branch.
Would you support this too?

### DIFF
--- a/clients/roscpp/src/libros/spinner.cpp
+++ b/clients/roscpp/src/libros/spinner.cpp
@@ -129,8 +129,9 @@ void SingleThreadedSpinner::spin(CallbackQueue* queue)
 
   if (!spinner_monitor.add(queue, true))
   {
-    ROS_FATAL_STREAM("SingleThreadedSpinner: " << DEFAULT_ERROR_MESSAGE << "You might want to use a MultiThreadedSpinner instead.");
-    throw std::runtime_error("SingleThreadedSpinner: " << DEFAULT_ERROR_MESSAGE << "You might want to use a MultiThreadedSpinner instead.");
+    std::string errorMessage = "SingleThreadedSpinner: " + DEFAULT_ERROR_MESSAGE + "You might want to use a MultiThreadedSpinner instead.";
+    ROS_FATAL_STREAM(errorMessage);
+    throw std::runtime_error(errorMessage);
   }
 
   ros::WallDuration timeout(0.1f);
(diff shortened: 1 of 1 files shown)


clients/roscpp/src/libros/spinner.cpp
```diff
@@ -130,7 +130,7 @@ void SingleThreadedSpinner::spin(CallbackQueue* queue)
 
   if (!spinner_monitor.add(queue, true))
   {
-    ROS_FATAL_STREAM("SingleThreadedSpinner: " << DEFAULT_ERROR_MESSAGE);
+    ROS_FATAL_STREAM("SingleThreadedSpinner: " << DEFAULT_ERROR_MESSAGE  << " You might want to use a MultiThreadedSpinner instead.");
```
clalancette: Looking at this critically, I think we probably want to go a bit further and update the language elsewhere.  In particular, the second sentence of `DEFAULT_ERROR_MESSAGE` doesn't seem to make much sense anymore:

```
const std::string DEFAULT_ERROR_MESSAGE =
    "Attempt to spin a callback queue from two spinners, one of them being single-threaded. "
    "This will probably result in callbacks being executed out-of-order.";
```

Since we are throwing a `std::runtime_error` directly afterwards (and thus, nothing will be happening, in-order or out-of-order).  Further, it seems like we should probably use the same string both in the `ROS_FATAL_STREAM` and as the string we use during the throw (though this string would be different depend on whether this is `SingleThreadedSpinner::spin` or `AsyncSpinnerImpl::start`.  Would you mind cleaning up those while we are doing this?  Thanks.

dirk-thomas (changes requested):
The current patch doesn't compile.

Please make sure that patches build locally for you and pass the tests before creating PRs.

v4hn:
Hey dirk. Sorry I pushed @jgueldenstein to reopen this request right away.
He will clean it up shortly. Sorry for the premature ping due to the re-open.

v4hn:
Ping @dirk-thomas @clalancette 

@jgueldenstein fixed the problems with this request and CI passed last week.
We would like to have this upstream to help our students use ROS correctly.
Please re-review and if possible merge.

clients/roscpp/src/libros/spinner.cpp
```diff
@@ -113,8 +113,7 @@ struct SpinnerMonitor
 
 SpinnerMonitor spinner_monitor;
 const std::string DEFAULT_ERROR_MESSAGE =
-    "Attempt to spin a callback queue from two spinners, one of them being single-threaded. "
-    "This will probably result in callbacks being executed out-of-order.";
+    "Attempt to spin a callback queue from two spinners, one of them being single-threaded. ";
```
dirk-thomas: Please remove the trailing space between the dot and the closing quote.

dirk-thomas:

PR #1165, opened 2017-09-13
Timeout can be specified for roscpp service calls

mikepurvis:
Rebase of #738.

FYI @dkargin @efernandez @dirk-thomas

### DIFF
--- a/clients/roscpp/src/libros/service_server_link.cpp
+++ b/clients/roscpp/src/libros/service_server_link.cpp
@@ -327,11 +327,6 @@ void ServiceServerLink::processNextCall()
   }
 }
 
-bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
-{
-  return call(req, resp, -1);
-}
-
 bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp, double timeout)
 {
   CallInfoPtr info(boost::make_shared<CallInfo>());
@@ -405,6 +400,11 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
   }
 
   return info->success_;
+}
+
+bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
+{
+  return call(req, resp, -1);
 }
 
 bool ServiceServerLink::isValid() const
--- a/clients/roscpp/src/libros/service_server_link.cpp
+++ b/clients/roscpp/src/libros/service_server_link.cpp
@@ -369,23 +369,24 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
 
     while (!info->finished_)
     {
-    	if(timeout > 0)
-    	{
-    		boost::chrono::milliseconds duration(static_cast<int>(timeout * 1000));
-    		if(info->finished_condition_.wait_for(lock, duration))
-    		{
-    			ROS_ERROR("Service [%s] call failed: no response for %fsec", service_name_.c_str(), timeout);
-    			interrupted = true;
-    			break;
-    		}
-    	}
-    	else
-    		info->finished_condition_.wait(lock);
+      if(timeout > 0)
+      {
+        boost::chrono::milliseconds duration(static_cast<int>(timeout * 1000));
+        if(info->finished_condition_.wait_for(lock, duration))
+        {
+          ROS_ERROR("Service [%s] call failed: no response for %fsec", service_name_.c_str(), timeout);
+          interrupted = true;
+          break;
+        }
+      }
+      else
+        info->finished_condition_.wait(lock);
     }
+
+    if(interrupted)
+        this->clearCalls();
   }
 
-  if(interrupted)
-	  this->clearCalls();
 
   info->call_finished_ = true;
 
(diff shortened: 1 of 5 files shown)


efernandez:
:+1:

dkargin:
[meta-ros](https://github.com/bmwcarit/meta-ros) still uses indigo-devel and certainly does not reach 'lunar' in near future. There are quite a lot of people using this yocto-ROS distribution. Though we got used to applying external patches here and there, we still hope this pull request be backported to indigo/kinetic

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,14 +363,34 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
     while (!info->finished_)
     {
-      info->finished_condition_.wait(lock);
+      if(timeout > 0)
+      {
+        system_clock::time_point wall_time = now + milliseconds(static_cast<int>(timeout * 1000));
+
+        if(info->finished_condition_.wait_until(lock, wall_time) == boost::cv_status::timeout)
+        {
+          ROS_ERROR("Service [%s] call failed: no response for %fsec", service_name_.c_str(), timeout);
```
dirk-thomas, line 381: I am not sure of outputting an error in these cases is a good choice. Maybe the user uses this feature extensively and is aware that some service calls will time out but doesn't want error messages to appear in those cases?

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -381,6 +402,11 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
   return info->success_;
 }
 
+bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
+{
+  return call(req, resp, -1);
+}
```
dirk-thomas, line 408: To keep the order the same as in the header move this function above the new function signature.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -326,7 +327,7 @@ void ServiceServerLink::processNextCall()
   }
 }
 
-bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
+bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp, double timeout)
```
dirk-thomas, line 330: Depending on the use case it might make sense to specify a timeout not in system time but in simtime?

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,14 +363,34 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
+      else
+        info->finished_condition_.wait(lock);
     }
   }
+  // Clear calls is properly protected
+  if(interrupted)
+    this->clearCalls();
+
 
   info->call_finished_ = true;
```
dirk-thomas, line 395: Is this being set to `true` in the case of a timeout?

dirk-thomas (commented):
Please add tests for this new feature.

dirk-thomas:
> meta-ros still uses indigo-devel and certainly does not reach 'lunar' in near future. There are quite a lot of people using this yocto-ROS distribution. Though we got used to applying external patches here and there, we still hope this pull request be backported to indigo/kinetic

@dkargin The desire to always have the latest features is opposing the desire for stability in a distribution which has been released a long time ago. Therefore any patches are merged into the latest distro first (currently Lunar). If they have shown that they haven't introduced regressions they can be considered for backporting to previous distributions - always with the consideration of effort, chance of regression, etc.

clients/roscpp/src/libros/service_server_link.cpp
dkargin: Though great part of ROS tries to keep 'kinematics' time, some other parts are strongly related to system time. We will have deadlocks when /use_sim_time=true and no /clock is provided or it is frozen.

paulbovbel: Would it make sense to take ros::Duration or ros::WallDuration in the signature rather than double,  to clarify whether the system clock or ROS time is being used?

efernandez:
If you don't mind @dkargin , I'd be happy to contribute the changes requested by @dirk-thomas and move this forward so it can get merged.

Is it possible to get commit access to `pull-738`? @mikepurvis 
Otherwise, I can create a new PR.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,14 +363,34 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
     processNextCall();
   }
 
+  bool interrupted = false;
   {
     boost::mutex::scoped_lock lock(info->finished_mutex_);
 
+    using namespace boost::chrono;
+    system_clock::time_point now = system_clock::now();
```
flixr, line 371: IMHO this should use the `steady_clock` to not be affected by system time changes as in #1250 and #1014

flixr (commented):
To not be affected by system time changes the `steady_clock` should be used.

dirk-thomas:
Waiting for the pending comments to be addressed / replied to.

dkargin:
I do not have access either. Or do I? Though I have no time to deal with it for lunar. We are locked on indigo.

mikepurvis:
Closed in favour of #1271.

PR #1166, opened 2017-09-14
rosnode: Minor fixes for compatibility with both Python 2 & 3

kartikmohta:


tools/rosnode/src/rosnode/__init__.py
```diff
@@ -182,7 +182,6 @@ def get_nodes_by_machine(machine):
     @raise ROSNodeException: if machine name cannot be resolved to an address
     @raise ROSNodeIOException: if unable to communicate with master
     """
-    import urlparse
```
kartikmohta, line 185: Already imported at the top of the file so no need to import here. Moreover there is no `urlparse` module in Python 3.

dirk-thomas:
Please provide a reproducible example.

tools/rosnode/src/rosnode/__init__.py
```diff
@@ -338,7 +337,7 @@ def rosnode_ping(node_name, max_count=None, verbose=False):
                 # 3786: catch ValueError on unpack as socket.error is not always a tuple
                 try:
                     # #3659
-                    errnum, msg = e
+                    errnum, msg = e.args
```
kartikmohta, line 340: This way allows it to work for both Python 2 & 3. `socket.error` is just an alias of `OSError` in Python 3.

dirk-thomas:
Thank you for the patch.

Please include a description in future contributions.

PR #1168, opened 2017-09-20
Fix an error in the comment for NodeHandle::advertise .

csukuangfj:


mikepurvis:
Looks like you added an ampersand and changed some whitespace— was this what you meant to submit?

csukuangfj:
@mikepurvis
Thank you for the reply. The first parameter of a non-static member function should be `this`,
which is a pointer. That is why I added an ampersand before the parameter.

I think only an ampersand was added, no other spaces are introduced.

clalancette:
So, I don't fully understand the `boost::bind` mechanism, but I don't think the change matters.  In a test I did locally, I tried adding a connect callback with `boost::bind` both with and without the ampersand, and it worked fine in either case.  Thus, I'm going to close out this PR since I don't think we need this change.  Feel free to reopen if you disagree.

PR #1172, opened 2017-09-29
Don't try to set unknown socket options

jspricke:
These are not avaible on FreeBSD, for example

### DIFF
--- a/clients/roscpp/src/libros/transport/transport_tcp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_tcp.cpp
@@ -195,7 +195,7 @@ void TransportTCP::setKeepAlive(bool use, uint32_t idle, uint32_t interval, uint
     }
 #endif
 
-#if defined(SOL_TCP)  && defined(TCP_KEEPCNT)
+#if defined(SOL_TCP) && defined(TCP_KEEPCNT)
     val = count;
     if (setsockopt(sock_, SOL_TCP, TCP_KEEPCNT, &val, sizeof(val)) != 0)
     {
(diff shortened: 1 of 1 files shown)


clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -179,7 +179,7 @@ void TransportTCP::setKeepAlive(bool use, uint32_t idle, uint32_t interval, uint
     }
 
 /* cygwin SOL_TCP does not seem to support TCP_KEEPIDLE, TCP_KEEPINTVL, TCP_KEEPCNT */
-#if defined(SOL_TCP) && !defined(__CYGWIN__)
+#if defined(SOL_TCP) && !defined(__CYGWIN__) && defined(TCP_KEEPIDLE) && defined(TCP_KEEPINTVL) && defined(TCP_KEEPCNT)
```
mikepurvis: Is the not-Cygwin check still required now that you're checking all these other defines?

jspricke: I don't know, could be that it's defined but shouldn't be used...

mikepurvis: Makes sense. It's frustrating not having any CI coverage other than the officially-supported platforms.

@dirk-thomas Is there a reason we don't use Travis and Appveyor to get some more coverage of this repo, particularly thinking in this instance of Windows and Python 3?

dirk-thomas: > Is there a reason we don't use Travis and Appveyor to get some more coverage of this repo, particularly thinking in this instance of Windows and Python 3?

The only reason is that nobody has spend the time to setup something like that (and is committed to maintain that setup in the future :wink:).

mikepurvis: Oh, okay! I might be willing, as we're contemplating a Python 3 migration. I just wasn't sure if there was some conviction that we should be using only the buildfarm for testing or something.

dirk-thomas: I am not in which configuration these flags could appear. Would it make sense to move the check for each flag around the block using that flag instead?

dirk-thomas:
Thanks.

PR #1173, opened 2017-09-30
add -d option for 'rosmaster' to output LOG_API

k-okada:
Sometimes, we had seen the CPU Usage of `rosmaster` is quit high (more than 50%). Mostly, this because of calling service call multiple times, without using `persistent` or put `ServiceProxy` within the loop.
To find the node calling these services, enabling LOG_API is very helpful, but currently we do not have to modify source code at this moment ( https://github.com/ros/ros_comm/blob/lunar-devel/tools/rosmaster/src/rosmaster/master_api.py#L137 )
```
rosmaster --core -p 11311 -d __log:=rosmaster.log
tail -f rosmaster.log
```
I'm not sure if we should pass this option from `roscore` -> `roslaunch` -> `rosmaster` like https://github.com/ros/ros_comm/commit/08f6bf06fe4e68efbf6b2c13b558a4d5298d1a41#diff-8517d9bf19c03b642c67e5d937dae851

### DIFF
--- a/tools/rosmaster/src/rosmaster/main.py
+++ b/tools/rosmaster/src/rosmaster/main.py
@@ -71,6 +71,10 @@ def rosmaster_main(argv=sys.argv, stdout=sys.stdout, env=os.environ):
     parser.add_option("-t", "--timeout",
                       dest="timeout",
                       help="override the socket connection timeout (in seconds).", metavar="TIMEOUT")
+    parser.add_option("-d", "--debug",
+                      dest="debug", action="store_true", default=False,
+                      help="print debug information")
+
     options, args = parser.parse_args(argv[1:])
 
     # only arg that zenmaster supports is __log remapping of logfilename
@@ -82,6 +86,10 @@ def rosmaster_main(argv=sys.argv, stdout=sys.stdout, env=os.environ):
     port = rosmaster.master.DEFAULT_MASTER_PORT
     if options.port:
         port = int(options.port)
+
+    if options.debug:
+        rosmaster.master_api.LOG_API = True
+        logging.getLogger("rosmaster.master").setLevel(logging.DEBUG)
 
     if not options.core:
         print("""


clalancette:
The code looks OK to me.  Since this is a deep debugging option, it seems to me like it is fine to just have it on rosmaster for now, and then add it to roscore later.  @dirk-thomas does that seem reasonable to you?

dirk-thomas:
> I'm not sure if we should pass this option from `roscore` -> `roslaunch` -> `rosmaster`

How are you currently using the new option? Is that sufficient for you? If not the argument good be passed along the same way.

tools/rosmaster/src/rosmaster/main.py
```diff
@@ -71,6 +71,10 @@ def rosmaster_main(argv=sys.argv, stdout=sys.stdout, env=os.environ):
     parser.add_option("-t", "--timeout",
                       dest="timeout",
                       help="override the socket connection timeout (in seconds).", metavar="TIMEOUT")
+    parser.add_option("-d", "--debug",
+                      dest="debug", action="store_true", default=False,
+                      help="print debug information")
```
dirk-thomas, line 76: I think the help text (and eventually even the option name) should be more specific to what exactly it is doing, namely enabling logging API calls on the master. Maybe the logging level shouldn't be controlled with the same option. That should be configurable already with the rospy logging config file, no?

k-okada:
@dirk-thomas  > How are you currently using the new option? Is that sufficient for you? If not the argument good be passed along the same way.

I'm currently using this option by `rosmaster --core -p 11311 -d __log:=rosmaster.log`. It's ok for me for now, because this is a deep debugging case as @clalancette noted. But, of course, if many people willing to check this deep debugging message, we should expose this option to roslaunch level.

> namely enabling logging API calls on the master. Maybe the logging level shouldn't be controlled with the same option. That should be configurable already with the rospy logging config file, no?

I tried to enable this message by rospy logging features, but as far as I understand, it turns out that rosmaster is not 'rosnode', it did not run `rospy.init_node()` so we can not configure through rospy tools. Am I correct?

dirk-thomas:
> if many people willing to check this deep debugging message, we should expose this option to roslaunch level.

If you think it would be useful please add this to the current patch. If not I am fine to merge this as-is (after you addressed the pending comment about the option name / help text).

> I tried to enable this message by rospy logging features, but as far as I understand, it turns out that rosmaster is not 'rosnode', it did not run rospy.init_node() so we can not configure through rospy tools. Am I correct?

I haven't tried it but that could indeed be the case.

k-okada:
thank you for comments, I have create new PR #1180 , which passes `--set-master-logger-level` from `roscore` to `rosmaster`. Please review when you have time @wjwwood

dirk-thomas:
Closing in favor of #1180.

PR #1174, opened 2017-09-30
use python defined error codes rather than hardcoded integers

mikaelarguedas:
follow up of https://github.com/ros/genpy/pull/89. Use python defined error codes  everywhere


mikepurvis:
Looks like no functional change, just rationalization. :+1:

PR #1177, opened 2017-10-05
Change rospy.Rate hz type description from int to float

okalachev:
The rospy.Rate constructor can take a float `hz` param, even less that 1 (for rates, that are less frequent than 1 per second). I tested it, and it works OK.

Incorrect type description makes my IDE crazy.


mikepurvis:
LGTM, thanks for the contribution!

PR #1178, opened 2017-10-05
rosnode: Return exit code 1 if there is an error.

### LINKED ISSUE #1171: Inconsistent exit codes of tools in error cases
I just noticed that the console tools (`rosnode`, `rostopic`, ...) are inconsistent in their return code in case of errors.

The case I tested was running them without a running roscore (using kinetic):
* `rostopic list` --> exit code 1
* `rosservice list` --> exit code 2
* `rosnode list` --> exit code 0
* `rosparam list` --> exit code 1

Especially the exit code of `rosnode` is problematic if you expect to get an error code like with the other tools.

luator:
Return with exit code 1 if there is an error in the rosnode tool (e.g. because roscore is not running).
This resolves #1171.

I tested manually by calling `rosnode list` with both roscore running/not running and the result was good. I also ran `rostest rosnode rosnode.test` and `nosetests test_rosnode_command_offline.py`, both passed.
Please let me know in case I should do more testing (I do not know how to provoke other errors than roscore not running, though).

Since I only have a running workstation with Kinetic at the moment, I did the PR for kinetic-devel. What exactly is the policy here? I guess when this is accepted, it should also be merged into lunar-devel.

### DIFF
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -808,9 +808,12 @@ def rosnodemain(argv=None):
             _fullusage()
     except socket.error:
         print("Network communication failed. Most likely failed to communicate with master.", file=sys.stderr)
+        sys.exit(1)
     except rosgraph.MasterError as e:
         print("ERROR: "+str(e), file=sys.stderr)
+        sys.exit(1)
     except ROSNodeException as e:
         print("ERROR: "+str(e), file=sys.stderr)
+        sys.exit(1)
     except KeyboardInterrupt:
         pass


tools/rosnode/src/rosnode/__init__.py
```diff
@@ -808,9 +808,12 @@ def rosnodemain(argv=None):
             _fullusage()
     except socket.error:
         print("Network communication failed. Most likely failed to communicate with master.", file=sys.stderr)
+        sys.exit(1)
```
mikepurvis, line 811: My preferred style would be to `return 1` here, and then `sys.exit(rosnode.rosnodemain())` in the wrapper script, but there are direct sys.exit calls elsewhere in this same function, so I think it makes sense to follow the prevailing style as you have done.

mikepurvis (approved):

mikepurvis:
Looks great, thanks for the contribution!

dirk-thomas:
> Since I only have a running workstation with Kinetic at the moment, I did the PR for kinetic-devel. What exactly is the policy here?

All changes will go into the latest development branch first (currently `lunar-devel`). After they have been released for a bit and no regressions have been identified they will be consider for backporting.

I changed the target branch of this PR. Once CI for Lunar comes back green this can be merged.

Thanks.

luator:
Alright, thanks for the fast review.

PR #1179, opened 2017-10-05
rosconsole: replaced 'while(0)' by 'while(false)' in ros/assert.h and ros/console.h

cwecht:
to avoid clang-tidys modernize-use-bool-literals warnings.


dirk-thomas:
Thank you.

PR #1181, opened 2017-10-06
one more hardcoded python error code

mikaelarguedas:
follow up of https://github.com/ros/ros_comm/pull/1174. looks like I missed one in the previous PR :S


dirk-thomas:
Thanks.

PR #1183, opened 2017-10-11
Test bzfile_ and lz4s_ before reading/writing/closing

jwon02:


dirk-thomas:
Thank you for the patch. The other stream implementation seem to lack the same kind of check. Can you update the PR to also add the checking to those.

jwon02:
Okay, let me take a look.

jwon02:
Updated the PR to test lz4s_ before reading and writing.

dirk-thomas:
Can you please also update `uncompressed_stream.cpp`.

jwon02:
I think there is nothing to check for the uncompressed stream.  Did I miss anything?

dirk-thomas:
Same cases checking `ptr` before trying to use it: e.g. https://github.com/ros/ros_comm/blob/cd7efd4d44d0ec5845df7d59ccdfc204389060f2/tools/rosbag_storage/src/uncompressed_stream.cpp#L62

**Update:** never mind - that would be a different kind of check. The uncompressed does indeed not have a file handle what you are checking for the other types.

dirk-thomas:
Same cases checking `ptr` before trying to use it: e.g. https://github.com/ros/ros_comm/blob/cd7efd4d44d0ec5845df7d59ccdfc204389060f2/tools/rosbag_storage/src/uncompressed_stream.cpp#L62

**Update:** never mind - that would be a different kind of check. The uncompressed does indeed not have a file handle what you are checking for the other types.

dirk-thomas:
Thank you for the patch.

Issue #1176, opened 2017-10-03
Bug in rosbag_storage: can't reuse the same instance or rosbag::Bag


facontidavide:
In my application I used to create a single instance or rosbag::Bag and use it as follows

1. open a file
2. parse using rosbag::View 
3. close() 
4. repeat from point 1.

This is not possible anymore. Running the code with Valgrind, it seems that something odd is happening.

> Invalid read of size 8
>   1: rosbag::View::iterator::increment() in /opt/ros/indigo/lib/librosbag_storage.so
>   2: void boost::iterator_core_access::increment<rosbag::View::iterator>(rosbag::View::iterator&) in /usr/include/boost/iterator/iterator_facade.hpp:520
>   3: boost::iterator_facade<rosbag::View::iterator, rosbag::MessageInstance, boost::forward_traversal_tag, rosbag::MessageInstance&, long>::operator++() in /usr/include/boost/iterator/iterator_facade.hpp:660
>   4: DataLoadROS::readDataFromFile(QString const&, QString&) in /home/dfaconti/ws_plotjuggler/src/PlotJuggler/plugins/ROS/DataLoadROS/dataload_ros.cpp:129
> 
> ....
> 
> Address 0x1baa01b0 is 16 bytes before a block of size 16 alloc'd  1: operator new(unsigned long) in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so
>   2: std::vector<rosbag::ViewIterHelper, std::allocator<rosbag::ViewIterHelper> >::_M_insert_aux(__gnu_cxx::__normal_iterator<rosbag::ViewIterHelper*, std::vector<rosbag::ViewIterHelper, std::allocator<rosbag::ViewIterHelper> > >, rosbag::ViewIterHelper const&) in /opt/ros/indigo/lib/librosbag_storage.so
>   3: rosbag::View::iterator::populate() in /opt/ros/indigo/lib/librosbag_storage.so
>   4: rosbag::View::iterator::iterator(rosbag::View*, bool) in /opt/ros/indigo/lib/librosbag_storage.so
>   5: rosbag::View::begin() in /opt/ros/indigo/lib/librosbag_storage.so
>   6: DataLoadROS::readDataFromFile(QString const&, QString&) in /home/dfaconti/ws_plotjuggler/src/PlotJuggler/plugins/ROS/DataLoadROS/dataload_ros.cpp:129
>   
>

racko:
[...]

PR #1188, opened 2017-10-14
Failing test case for #1176

racko:
See https://github.com/ros/ros_comm/issues/1176


dirk-thomas:
Based in the discussion this should be updated to read different bags.

dirk-thomas:
#1192 demonstrates the failing test in the first commit and the fix in the second.

PR #1189, opened 2017-10-14
Enabled rosbag::Bag move operations if compiler support is available

racko:


tools/rosbag_storage/include/rosbag/bag.h
```diff
@@ -111,6 +112,17 @@ class ROSBAG_DECL Bag
...
 
+#ifndef BOOST_NO_CXX11_RVALUE_REFERENCES
+    Bag(Bag&& other) {
+        init();
+        swap(other);
+    }
+    Bag& operator=(Bag&& other) {
+        swap(other);
+        return *this;
+    }
```
dirk-thomas: Please keep the implementation of these methods in the cpp file even though they are short.

racko:
Updated

dirk-thomas:
Thank you for the patch.

PR #1190, opened 2017-10-14
Added rosbag::Bag::isOpen

racko:


tools/rosbag_storage/include/rosbag/bag.h
```diff
@@ -185,6 +185,8 @@ class ROSBAG_DECL Bag
 
     void swap(Bag&);
 
+    bool isOpen() const { return file_.isOpen(); }
```
dirk-thomas: Please keep the implementation of this method in the cpp file even though it is short.

racko:
Updated

dirk-thomas:
The patch looks good. ~~Since it changes ABI it might be better to target the next ROS distro.~~ Nvm, the new method is non-virtual and should therefore not break ABI.

dirk-thomas:

dirk-thomas:
The patch looks good. ~~Since it changes ABI it might be better to target the next ROS distro.~~ Nvm, the new method is non-virtual and should therefore not break ABI.

PR #1191, opened 2017-10-16
Fixed an out of bounds read in void rosbag::View::iterator::increment()

racko:
- Only triggered if reduce_overlap_ = true
- When iters_.size() == 1 and iters_.pop_back() gets called in the loop,
  the next loop condition check would read from iters_.back(), but
  iters_ would be empty by then.

See https://github.com/ros/ros_comm/issues/1176#issuecomment-337029676


dirk-thomas:
Thank you for the fix.

PR #1192, opened 2017-10-17
Fixed corrupted messages when reopening a rosbag with a different file (#1176)

racko:
* Cleaned up more state in Bag::close() and ChunkedFile::close()
* It was not investigated how exactly the bug worked

I kept #1188 as-is to demonstrate the failing test. (I think I remember that the CI only builds the latest commit in a pull-request, right?)

### DIFF
--- a/tools/rosbag_storage/src/bag.cpp
+++ b/tools/rosbag_storage/src/bag.cpp
@@ -180,6 +180,8 @@ void Bag::close() {
     chunks_.clear();
     connection_indexes_.clear();
     curr_chunk_connection_indexes_.clear();
+
+    init();
 }
 
 void Bag::closeWrite() {
--- a/test/test_rosbag_storage/src/create_and_iterate_bag.cpp
+++ b/test/test_rosbag_storage/src/create_and_iterate_bag.cpp
@@ -1,7 +1,10 @@
 #include "ros/time.h"
 #include "rosbag/bag.h"
 #include "rosbag/view.h"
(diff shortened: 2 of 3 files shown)


tools/rosbag_storage/src/bag.cpp
```diff
@@ -173,6 +173,21 @@ void Bag::close() {
...
+    bag_revision_ = 0;
+    file_size_ = 0;
+    file_header_pos_ = 0;
+    index_data_pos_ = 0;
+    connection_count_ = 0;
+    chunk_count_ = 0;
+    chunk_open_ = false;
+    curr_chunk_data_pos_ = 0;
+    current_buffer_ = 0;
+    decompressed_chunk_ = 0;
```
dirk-thomas: This should be easier now that #1189 has been merged. Please update the patch to call the new `init` method instead.

racko:
Done. I also analyzed the bug a bit more and found the following cause, which I added to the commit message:

    * Bag::decompressed_chunk_ was not reset in close(). decompressed_chunk_
    stores the offset (within the file) of the compressed chunk whose
    decompressed content is currently stored in decompress_buffer_. Since it
    wasn't reset in close(), if the offset is the same as that of the first
    chunk in the file opened next, Bag::decompressChunk() assumed that
    it had already decompressed that chunk in the new file and just returned.
    As a result, some member functions of Bag worked on the decompressed
    chunk of the old bag.

dirk-thomas:
Thank you for the patch.

PR #1193, opened 2017-10-18
[roslaunch-check] Use roslaunch.core.printerrlog for printing error message.

130s:
Errors are printed without much stress, which makes hard to find the specific error occurred.

### DIFF
--- a/tools/roslaunch/scripts/roslaunch-check
+++ b/tools/roslaunch/scripts/roslaunch-check
@@ -38,6 +38,7 @@ import sys
 
 import rospkg
 
+import roslaunch
 import roslaunch.rlutil
 import rosunit
 
@@ -116,7 +117,7 @@ if __name__ == '__main__':
             message = "roslaunch check [%s] failed"%(roslaunch_path)
             f.write(rosunit.junitxml.test_failure_junit_xml(test_name, message, stdout=error_msg,
                     class_name="roslaunch.RoslaunchCheck", testcase_name="%s_%s" % (pkg, outname)))
-        print("wrote test file to [%s]"%test_file)
+        roslaunch.core.printerrlog("wrote test file to [%s]"%test_file)
         sys.exit(1)
     else:
         print("passed")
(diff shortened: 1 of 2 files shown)


tools/roslaunch/scripts/roslaunch-check
```diff
@@ -116,7 +117,7 @@ if __name__ == '__main__':
             message = "roslaunch check [%s] failed"%(roslaunch_path)
             f.write(rosunit.junitxml.test_failure_junit_xml(test_name, message, stdout=error_msg,
                     class_name="roslaunch.RoslaunchCheck", testcase_name="%s_%s" % (pkg, outname)))
-        print("wrote test file to [%s]"%test_file)
+        roslaunch.core.printerrlog("wrote test file to [%s]"%test_file)
```
dirk-thomas, line 120: This doesn't seem to be an error message?

130s: This `if` block is when error occurred. To highlight the file path this change is still useful.

dirk-thomas:
Thank you for the patch.

130s:
Thank you for the review @dirk-thomas.
Just wondering if there will be another batch backport like https://github.com/ros/ros_comm/pull/1008 soon for older supported branches and if this PR will be considered?

dirk-thomas:
There will likely be another Lunar release in the near term and after that a batch backport to Kinetic. All changes will be considered - the usual rational applies if a change should be backported. While this patch has little impact and wouldn't be necessary it also poses very little risk. Will see when doing the backporting.

tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -219,10 +219,10 @@ def check_roslaunch(f, use_test_depends=False):
             print(e, file=sys.stderr)
             miss_all = True
         if miss_all:
-            print("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)), file=sys.stderr)
+            roslaunch.core.printerrlog("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)), file=sys.stderr)
```
dirk-thomas, line 222: The new function doesn't support the `file` keyword argument. See #1317.

PR #1195, opened 2017-10-20
Static-cast unused parameters in TransportTCP.

mikepurvis:
Silences some warnings under macOS.


clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -185,6 +185,8 @@ void TransportTCP::setKeepAlive(bool use, uint32_t idle, uint32_t interval, uint
     {
       ROS_DEBUG("setsockopt failed to set TCP_KEEPIDLE on socket [%d] [%s]", sock_, cached_remote_host_.c_str());
     }
+#else
+    static_cast<void>(idle);
```
dirk-thomas: Please use `(void)idle;` instead.

Similar below.

dirk-thomas:
Thanks.

Issue #1194, opened 2017-10-19
More "agile" demux?


peci1:
Hi, we've found the implementation of demux is quite slow when switching topics. Or at least the topic connections that need to be re-created during every switch are slow. The result is the switching happens quickly, but any subscribers start receiving messages a few seconds later (up to 15 seconds in my measurement).

Steps to reproduce:

    rosrun topic_tools demux a b c

    rostopic echo b

    rostopic echo c

    rostopic pub  -r 10 /a stamped_msgs/Int32 "header:
      seq: 0
      stamp:
        secs: 0
        nsecs: 0
      frame_id: ''
    data: 0"    

    rosservice call /demux/select b # (or c)

Try calling the `select` service a few times, and you'll see the long delay between a topic gets selected and the `rostopic`s start printing the new values.

What about giving `demux` a parameter that would tell it to keep all the publishers running all the time? Or even throwing away the old behavior and having the new one for all? I don't see any drawbacks, since only one of the publishers is doing some real work at a time...

dirk-thomas:
Sounds like a good improvement to me. Please consider to provide a pull request for this enhancement.

peci1:
Here we go...

dirk-thomas:
Addressed by #1196.

PR #1196, opened 2017-10-20
More agile demux.

peci1:
Publishers in demux are no longer destroyed and recreated when switching, which results in much faster switching behavior. The previous version took even 10 seconds to start publishing on the newly selected topic (all on localhost).

Please, comment if you feel the default behavior should stay as the old was, and this new behavior should be triggered by a parameter.

Solves #1194 .


dirk-thomas:

PR #1198, opened 2017-10-23
trace instrumentation for ros_comm

iluetkeb:
What do you guys think about including this in the standard ROS core?


dirk-thomas:
What is the expected behavior on other platforms like Windows with this patch?

iluetkeb:
@dirk-thomas This patch is not platform-specific, because it only makes calls to the "tracetools" package. That package checks for presence of LTTng and if LTTng is not present, it will do nothing, so it should work on Windows. I'm not currently aware of tracing frameworks for Windows, but those could be added to tracetools, if desired.

iluetkeb:
Looking at the failed build report, it seems I need to make "tracetools" available to rosdep first.

dirk-thomas:
Can you share a link to the `tracetools` package.

iluetkeb:
https://github.com/bosch-robotics-cr/tracetools

I'm currently adding it to rosdistro, but for some reason the nosetests are taking ages, so it's gonna be a few minutes before I can push that.

dirk-thomas:
Adding such a new dependency to ROS core is likely something which should happen in a new ROS distro rather than in already released ones. I will mark this with the label "consider-future-release" for now.

iluetkeb:
@dirk-thomas Would it be different if we included the relevant parts of the tracetools package into ros_comm? It's just one header and one cpp file -- the analysis stuff could be kept external, there is no direct dependency on that.

dirk-thomas:
Since the `tracetools` package doesn't have a dependency on `LTTng` nothing would change for binary packages. So this would only be effecting from-source builds when the user has the dependency installed. That part sounds ok to me.

I don't think that integrating it into `ros_comm` is a good idea. This repo is already bloated and contains too much stuff - I would rather not add more to it. Keeping it separate also clarifies the different maintainer role for it. For new ROS distros that will require to do initial releases very early in the release process since it would become a requirement before `ros_comm` can be released (maybe the repo should be moved to the `ros` org unit but that is a detail we can figure out later).

The major reason why I think I should only go into a future ROS distro is that is breaks ABI compatibility. And therefore it should go into existing distros without a very strong reason to break ABI which I don't see.

@clalancette @dhood @mikaelarguedas @tfoote @wjwwood Maybe any of you has feedback on this?

wjwwood:
I would say, release it as a ROS package (the one header and cpp file), then have `ros_comm` depend on it.

+1 to not adding bloat to `ros_comm`.

+1 for not breaking ABI for this.

A compromise might be to make a back port which defaults to a stable ABI, but with a compile definition the tracing could be enabled. That way the binaries stay stable, but someone building from source could turn it on, though maybe not being in the binaries already removes the value of back porting it.

iluetkeb:
@dirk-thomas As far as I can tell, this breaks ABI only in the backward direction -- if you compile against the new version, but at runtime the old is used, that won't work. In the other direction, if you compiled against the old version, but use the one with tracetools at runtime, that should not break ABI. We use this often.

@wjwwood You can already enable a lot of tracing by just dropping our modified version into the workspace. For everything in the workspace, this gives you full information. Stuff that's not in the workspace will lack naming information, which is a big issue on interpretation (you just get function addresses...), but still useful.

All that said, if you prefer putting it into future versions only, I acquiesce. People can always get the instrumented version from your bosch-robotics-cr, if they want it.

dirk-thomas:
Afaik the addition of a virtual function to a class is breaking ABI: https://github.com/ros/ros_comm/pull/1198/files#diff-553a9121209f12c57244b0a30cd246d4R136 (see [REP 9](http://www.ros.org/reps/rep-0009.html#what-s-not-allowed))

Do you have any information why you think this wouldn't be a problem?

README.md
```diff
@@ -0,0 +1,10 @@
+# ROS with Bosch tracing mods
+
+This is a modified version of ros\_comm which adds calls to the tracetools.
+
+## Intended use
+
+The intended use for this is during development, to find timing issues etc. Please see
+the tracetools package for more documentation.
+
+
```
mikepurvis: Please remove this file.

clients/roscpp/CMakeLists.txt
```diff
@@ -1,12 +1,13 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(roscpp)
 
+add_definitions("-std=c++11")
```
mikepurvis: Under ROS Melodic, this is the default; it can be removed.

clients/roscpp/src/libros/service_client_link.cpp
```diff
@@ -154,6 +155,10 @@ bool ServiceClientLink::handleHeader(const Header& header)
     m["callerid"] = this_node::getName();
     connection_->writeHeader(m, boost::bind(&ServiceClientLink::onHeaderWritten, this, _1));
 
+    ros::trace::new_connection(connection_->getTransport()->getAddress(true).c_str(),
+    		connection_->getTransport()->getAddress(false).c_str(),
+			connection_.get(), "ServiceClientLink", service.c_str(), ss->getDataType().c_str());
```
mikepurvis: Spaces over tabs, please (here and elsewhere).

iluetkeb:
Waiting for rosdistro merge (for tracetools release in melodic) before further updates.

iluetkeb:
I'll look into the compile error -- it does not happen on my Ubuntu Xenial, maybe something to do with a newer boost version on Bionic and Stretch.

iluetkeb:
Can we still get this into Melodic, given that it wasn't merged initially, and will thus brake ABI?

dirk-thomas:
> Can we still get this into Melodic, given that it wasn't merged initially, and will thus brake ABI?

The current state of the patch doesn't even compile. Beside that it needs to be rebased in order to be mergable.

After that we can consider to merge it if others think it should go in beside the ABI breakage. @mikepurvis what do you think?

mikepurvis:
:+1: for including this in Melodic. It's a new dependency on `tracetools`, but that package is light and shouldn't cause problems on non-Ubuntu platforms. Should it be moved or forked to the `ros` org?

iluetkeb:
Replaced by https://github.com/ros/ros_comm/pull/1443

PR #1199, opened 2017-10-23
TOPIC TOOLS: add initial_topic param

kev-the-dev:
On a ROS-based robot my lab works on, we use topic_tools/mux to manage potential wrenches from the autonomous controller, an xbox controller, and an rf based controller. We need the option to, on startup, have none of these inputs selected (for safety), so that we have to explicitly select one before anything will move. To do this we mux, we needed to use a dummy topic. 

This adds an optional private parameter to mux, "initial_topic" which will set the input topic the mux will use on startup. Additionally, it can be set to "__none" for no initial topic. If the param is not set, the default behavior or choosing the first input in the arguments is used.

### DIFF
--- a/tools/topic_tools/CMakeLists.txt
+++ b/tools/topic_tools/CMakeLists.txt
@@ -103,6 +103,8 @@ if(CATKIN_ENABLE_TESTING)
   add_rostest(test/relay.test)
   add_rostest(test/relay_stealth.test)
   add_rostest(test/lazy_transport.test)
+  add_rostest(test/mux_initial_none.test)
+  add_rostest(test/mux_initial_other.test)
   ## Latched test disabled until underlying issue in roscpp is resolved,
   ## #3385, #3434.
   #rosbuild_add_rostest(test/relay_latched.test)
--- a/tools/topic_tools/src/mux.cpp
+++ b/tools/topic_tools/src/mux.cpp
@@ -317,9 +317,44 @@ int main(int argc, char **argv)
 
     g_subs.push_back(sub_info);
   }
-  g_selected = g_subs.begin(); // select first topic to start
+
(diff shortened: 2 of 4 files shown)


dirk-thomas:
This looks like a nice enhancements. Can you please add a test covering the new functionality.

kev-the-dev:
Tests added for __none and non-default case

kev-the-dev:
On a side note, it seems like a lot of the tests for topic_tools were disabled... do they still need to be? I think they will pass
```
  ## Latched test disabled until underlying issue in roscpp is resolved,
  ## #3385, #3434.
  #rosbuild_add_rostest(test/relay_latched.test)
  #rosbuild_add_rostest(test/mux.test)
  #rosbuild_add_rostest(test/switch_mux.test)
  #rosbuild_add_rostest(test/switch_mux_leading_slash.test)
  #rosbuild_add_rostest(test/switch_mux_none.test)
  #rosbuild_add_rostest(test/mux_add.test)
```

dirk-thomas:
If they pass it would indeed be great to reenable them.

dirk-thomas:
Thank you for the improvement. Can you please mention the new feature on the [wiki page](http://wiki.ros.org/topic_tools/mux) and link to the change here.

kev-the-dev:
[Wiki updated](http://wiki.ros.org/action/diff/topic_tools/mux?action=diff).

What is the process to get this on kinetic/other versions as well? Send separate PR's for each branch?

dirk-thomas:
A while after a Lunar release all changes will be considered for backporting into Kinetic. I will then create a "big"  PR which backports all the changes desired. That usually includes fixes and sometimes features which have a low chance of regressions (which this one would probably fall into).

Into Indigo usually only significant bug fixes are being backported.

Issue #1209, opened 2017-10-30
 ROSBagException should call the base constructor


lukehansen:
In python 2.7 there is a subtle bug with unpickling custom exceptions; this was fixed in 3.3, see https://bugs.python.org/issue1692335.  As a workaround for 2.7 we can re-define these exceptions as below.

```java
class ROSBagException(Exception):
    """
    Base class for exceptions in rosbag.
    """
    pass
 
class ROSBagFormatException(ROSBagException):
    """
    Exceptions for errors relating to the bag file format.
    """
    pass

class ROSBagUnindexedException(ROSBagException):
    """
    Exception for unindexed bags.
    """
    def __init__(self, *args):
        ROSBagException.__init__(self, 'Unindexed bag')
```

For context, I came across this issue when a `ROSBagUnindexedException` was raised in a `multiprocessing.Pool.map` thread.  It throws the `__init__` exception and hangs, as described here: https://lists.gt.net/python/bugs/1025933

dirk-thomas:
Can you please provide a pull request with the proposed patch against the current default branch (`lunar-devel`). Thanks.

lukehansen:
Sure, though it's not really a ros_comm bug and has been patched in python 3.3.  PR is here: https://github.com/ros/ros_comm/pull/1210

PR #1210, opened 2017-10-30
ROSBagException should call the base constructor

lukehansen:
https://github.com/ros/ros_comm/issues/1209

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -75,24 +75,19 @@ class ROSBagException(Exception):
     """
     Base class for exceptions in rosbag.
     """
-    def __init__(self, value):
-        self.value = value
-
-    def __str__(self):
-        return self.value
+    pass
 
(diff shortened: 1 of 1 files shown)


tools/rosbag/src/rosbag/bag.py
```diff
@@ -75,24 +75,19 @@ class ROSBagException(Exception):
     """
     Base class for exceptions in rosbag.
     """
-    def __init__(self, value):
-        self.value = value
```
dirk-thomas, line 79: Removing `value` will effectively change the public API of this exception. Therefore I don't think this PR can be applied as-is since it would break compatibility with existing code.

Maybe the constructor should be kept and invoke the parent constructor explicitly? If the parent constructor is being called with `value` the custom `__str__` implementation might still be obsolete.

dirk-thomas (changes requested):

dirk-thomas:
Superseded by #1652.

Issue #1207, opened 2017-10-28
non-portable code in rosmaster.util


Yay295:
The last change to `util.py` ( #1104 ) added a function `close_half_closed_sockets()` to fix an issue where some sockets weren't being closed, causing ROS to eventually freeze up. Unfortunately, the call `getsockopt(socket.SOL_TCP, socket.TCP_INFO)` is not portable ([man page](http://man7.org/linux/man-pages/man7/tcp.7.html) "**TCP_INFO** Used to collect information about this socket. The kernel returns a struct tcp_info as defined in the file /usr/include/linux/tcp.h. *This option should not be used in code intended to be portable.*"), and it fails under WSL. There is a [WSL Issue](https://github.com/Microsoft/WSL/issues/1982) tracking this, and it appears they do intend to implement this functionality, but I thought it would be a good idea to bring it up here as well.

dirk-thomas:
Does this also affect other platforms like native Windows?

@otamachan @furushchev Does either of you see a way to avoid this socket option?

otamachan:
It seems that WIndows does not support `TCP_INFO`

```
c:\Python27>python
Python 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:19:30) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import socket
>>> socket.TCP_INFO
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'TCP_INFO'
```

I'm not familiar with Windows sockets, so currently I have no idea how to avoid `TCP_INFO`.

furushchev:
It looks that windows has an api for getting TCP socket states in C++ level but I'm not sure if this functionality is accessible from `python` on windows.

https://msdn.microsoft.com/en-us/library/bb485738(VS.85).aspx

For now, I can only suggest not to use this function on the system that does not support `TCP_INFO` until any solution is found since I have no experience with windows.

smnogar:
I am getting issues with this in macOS High Sierra as well. Errors appear in master.log that look like: 
```
[rosmaster.threadpool][ERROR] 2017-12-21 15:49:02,582: Traceback (most recent call last):
  File "/opt/ros/lunar/lib/python2.7/site-packages/rosmaster/threadpool.py", line 218, in run
    result = cmd(*args)
[...]

PR #1211, opened 2017-10-31
Add TCP_INFO availability check

otamachan:
As using `TCP_INFO` is not a portable code (#1207),  check its availability before using.

### DIFF
--- a/tools/rosmaster/src/rosmaster/util.py
+++ b/tools/rosmaster/src/rosmaster/util.py
@@ -70,18 +70,19 @@ def xmlrpcapi(uri):
 
 
 def close_half_closed_sockets():
-    if hasattr(socket, 'TCP_INFO'):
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the quick fix.

@Yay295 can you please try this patch and comment if it works around the problem you have with WSL.

Yay295:
Unfortunately not. This isn't a Python problem, so `hasattr(socket, 'TCP_INFO') is True`. I might be able to catch it in a try-except though.

dirk-thomas:
@Yay295 Can you please propose a PR with a patch which works for you.

dirk-thomas:
Thank you for the patch. I changed the logic in e5112a77ff76c3c99fda42116b4f124392b9e470 in order to simplify the diff.

PR #1212, opened 2017-10-31
catch exception with `socket.TCP_INFO` on WSL

Yay295:
fixes issue https://github.com/ros/ros_comm/issues/1207
this only catches socket error 92

### DIFF
--- a/tools/rosmaster/src/rosmaster/util.py
+++ b/tools/rosmaster/src/rosmaster/util.py
@@ -50,6 +50,7 @@ monkey_patch()
 del monkey_patch
 
 import socket
+import errno
 
 _proxies = {} #cache ServerProxys
 def xmlrpcapi(uri):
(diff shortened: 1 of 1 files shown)


tools/rosmaster/src/rosmaster/util.py
```diff
@@ -69,12 +69,16 @@ def xmlrpcapi(uri):
...
-            if state == 8:  # CLOSE_WAIT
-                transport.close()
+    try:
+        for proxy in _proxies.values():
+            transport = proxy("transport")
+            if transport._connection and transport._connection[1] is not None and transport._connection[1].sock is not None:
+                state = transport._connection[1].sock.getsockopt(socket.SOL_TCP, socket.TCP_INFO)
+                if state == 8: # CLOSE_WAIT
+                    transport.close()
+    except socket.error as e: # catch [Errno 92] Protocol not available
```
dirk-thomas: Please only wrap the line which might raise in a `try` / `except`.

tools/rosmaster/src/rosmaster/util.py
```diff
@@ -69,12 +69,16 @@ def xmlrpcapi(uri):
...
+        if e.args[0] is not 92:
```
dirk-thomas: Please use the error code in `errno` instead of a hard coded number.

tools/rosmaster/src/rosmaster/util.py
```diff
@@ -69,12 +69,16 @@ def xmlrpcapi(uri):
...
+            raise e
```
dirk-thomas: The `e` is not necessary. Just calling `raise` is sufficient and makes it clearer that the current exception is being raised again.

Yay295:
I've made the requested changes.

tools/rosmaster/src/rosmaster/util.py
```diff
@@ -72,8 +73,14 @@ def close_half_closed_sockets():
...
         transport = proxy("transport")
         if transport._connection and transport._connection[1] is not None and transport._connection[1].sock is not None:
-            state = transport._connection[1].sock.getsockopt(socket.SOL_TCP, socket.TCP_INFO)
-            if state == 8:  # CLOSE_WAIT
+            try:
+                state = transport._connection[1].sock.getsockopt(socket.SOL_TCP, socket.TCP_INFO)
+            except socket.error as e: # catch [Errno 92] Protocol not available
+                if e.args[0] is errno.ENOPROTOOPT:
+                    return
+                else:
```
dirk-thomas: The `else` is not necessary since the `if` block calls `return`.

dirk-thomas:
Just for the reference: this fixes a regression of #1104 which needs to be backported to *kinetic-devel*.

PR #1213, opened 2017-11-01
fix uri in message

dirk-thomas:
Fix regression introduced in #1026.


mathias-luedtke:
I cannot confirm that this resolves our error, because it occurs only sporadically in our tests and I was not able to enforce it egain.
But I can confirm that this patch does not break anything.

Regarding the code: In line 100 `False` gets returned instead of a string.
For consistency reasons I would prefer returning `None` and checking for `is not None`.
However, both versions will work :)

dirk-thomas:
Thank you for checking. In order to keep the patch minimal I will keep it as-is since it will work.

PR #977, opened 2017-02-07
Close socket when connection was not accepted

mikepurvis:
Retry #960, from @mgrrx.

We haven't been able to get a MWE for the problems we saw with this merged, but it came up in our crash reporter originating from unit test runs, workstations, and also robots. An example stack trace:

```
  File "../nptl/sysdeps/unix/sysv/linux/raise.c", line 56, in __GI_raise
  File "abort.c", line 89, in __GI_abort
  File "../sysdeps/posix/libc_fatal.c", line 175, in __libc_message
  File "fortify_fail.c", line 38, in __GI___fortify_fail
  File "chk_fail.c", line 28, in __GI___chk_fail
  File "fdelt_chk.c", line 25, in __fdelt_chk
  File "../../../../../ros-comm-f6da1ec87d6ed4653e65ac2dbae1325f7bc1238d/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp", line 100, in XmlRpc::XmlRpcDispatch::work
  File "../../../../../ros-comm-f6da1ec87d6ed4653e65ac2dbae1325f7bc1238d/utilities/xmlrpcpp/src/XmlRpcServer.cpp", line 133, in XmlRpc::XmlRpcServer::work
  File "../../../../../ros-comm-f6da1ec87d6ed4653e65ac2dbae1325f7bc1238d/clients/roscpp/src/libros/xmlrpc_manager.cpp", line 272, in ros::XMLRPCManager::serverThreadFunc
  File "/usr/lib/x86_64-linux-gnu/libboost_thread.so.1.54.0", in ?? ()
  File "pthread_create.c", line 312, in start_thread
  File "../sysdeps/unix/sysv/linux/x86_64/clone.S", line 111, in clone
```

It's possible that the crashes were coming up during node shutdown, but that still fails affected tests _and_ is a regression from previously.

Issue #222, opened 2013-05-01
Add error message when topic can't connect (roscpp and rospy)


ahendrix:
One of the hard-to-debug problems in ROS is when a publisher is active and registered with the ROS master, but has a bad IP/DNS/ROS_HOSTNAME, so remote clients are not able to connect.

We should consider adding an error message (or making an existing error more visible) to the client libraries whenever the TCP connection to a publisher fails; particularly if it fails during name resolution.

dirk-thomas:
For rospy you can check if changing the two lines from `logdebug()` to `logwarn()` would be what you want:
https://github.com/ros/ros_comm/blob/groovy-devel/clients/rospy/src/rospy/impl/registration.py#L315
https://github.com/ros/ros_comm/blob/groovy-devel/clients/rospy/src/rospy/impl/registration.py#L318

For the cpp part it will be pretty difficult to achieve. Any new connection attempt is scheduled with a PendingConnection. This one tries forever. Adding a warning at that level will be difficult since in a lot of scenarios the connection will intentionally only be established after some time (when the subscriber actually appears).

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #1214, opened 2017-11-02
Unit tests and bug fixes for XmlRpcpp

### LINKED ISSUE #914: xmlrpccpp spins forever when no more file handles

trainman419:
The initial motivation for this was to fix some rare crashes that we were seeing in xmlrpcpp where it would leak file descriptors (as described in https://answers.ros.org/question/250393/rosout-leaks-file-descriptors/ ) and then would crash if given a socket file descriptor with descriptor number above FD_SETSIZE (commonly 1024)

This also fixes #914, where an incoming XmlRpc request would cause the server to use 100% of a core if it could not successfully call `accept()`. (replaces #977, #960). There are two fallback mechanisms here:
Measuring the number of open file descriptors and close new incoming connections if the number of free file descriptors falls below a threshold.
If accept fails completely, stop monitoring the server socket for 1 second and then re-add it, so that we can still accept new connections if the old connections eventually close.

* Fix for file descriptor leaks in XmlRpcClient.
* Reject incoming connections in XmlRpcServer to prevent file descriptor exhaustion.
* Timed retry if accept() fails.
* Fix handling of partial buffer reads and writes in XmlRpcClient.
* Fix use of uninitialized memory in XmlRpcSocket::get_port.
* Break dependency on XmlRpc.h header to make targeted mocking easier.
* Better error reporting when address lookup fails. (#222) 
* Report symbolic state names in XmlRpcClient.
* Make async XmlRpcClient terminate correctly on error.
* Update comments to reflect the fact that an XmlRpcDispatch "exception" actually indicates out-of-band TCP data has arrived.
* Fix XmlRpcDispatch to correctly honor the Exception request again.
* Make XmlRpcUtil::error go to stdout to match log and provide better error reporting. (#364)
* Lots of unit tests.

cc @kmactavish @mikepurvis @mgrrx @ruffsl


dirk-thomas:
The size of the diff is pretty big which makes it rather difficult to review. Can the individual fixes and improvements be split into separate pull requests or at least separate commits?

trainman419:
Ok; I'll try to split this into separate pull requests.

trainman419:
First part of this: #1216 . Many of the other tests depend on that header refactoring, so that will need to merge before I can submit the remainder of my tests and fixes as smaller PRs.

trainman419:
Second part: #1218

trainman419:
Third part: #1221

trainman419:
Fourth part: #1232

trainman419:
Part 5: #1242

trainman419:
Part 6: https://github.com/ros/ros_comm/pull/1243

trainman419:
Part 7: https://github.com/ros/ros_comm/pull/1244

PR #1218, opened 2017-11-03
Unit tests and bug fixes for XmlRpcSocket

trainman419:
* Add system call mocks for socket functions, and use these to test the XmlRpcSocket class.
* Improve the address resolution error messages to include a description the error generated by getaddrinfo.
* Fix an existing TODO where EWOULDBLOCK was not handled as a fatal error on Linux.
* Fix get_port so that it checks the return code from getsockname; this prevents it from using the socket data uninitialized if getsockname fails.
* Make a few of the string arguments to XmlRpcSocket const so that they're compatible with string literals.

### DIFF
--- a/utilities/xmlrpcpp/test/CMakeLists.txt
+++ b/utilities/xmlrpcpp/test/CMakeLists.txt
@@ -3,6 +3,7 @@ if(TARGET xmlrpcvalue_base64)
   target_link_libraries(xmlrpcvalue_base64 xmlrpcpp)
 endif()
 
+if(NOT WIN32)
 catkin_add_gtest(test_socket
   test_socket.cpp
   test_system_mocks.c
@@ -13,6 +14,7 @@ set_target_properties(test_socket PROPERTIES
   LINK_FLAGS
   "-Wl,--wrap=accept -Wl,--wrap=bind -Wl,--wrap=close -Wl,--wrap=connect -Wl,--wrap=getaddrinfo -Wl,--wrap=getsockname -Wl,--wrap=listen -Wl,--wrap=read -Wl,--wrap=setsockopt -Wl,--wrap=select -Wl,--wrap=socket -Wl,--wrap=write -Wl,--wrap=fcntl -Wl,--wrap=freeaddrinfo"
 )
+endif()
 
 catkin_add_gtest(TestValues TestValues.cpp)
 target_link_libraries(TestValues xmlrpcpp)
--- a/utilities/xmlrpcpp/test/CMakeLists.txt
+++ b/utilities/xmlrpcpp/test/CMakeLists.txt
@@ -3,6 +3,17 @@ if(TARGET xmlrpcvalue_base64)
   target_link_libraries(xmlrpcvalue_base64 xmlrpcpp)
 endif()
 
+catkin_add_gtest(test_socket
+  test_socket.cpp
+  test_system_mocks.c
+  ../src/XmlRpcSocket.cpp
+  ../src/XmlRpcUtil.cpp
+)
+set_target_properties(test_socket PROPERTIES
+  LINK_FLAGS
+  "-Wl,--wrap=accept -Wl,--wrap=bind -Wl,--wrap=close -Wl,--wrap=connect -Wl,--wrap=getaddrinfo -Wl,--wrap=getsockname -Wl,--wrap=listen -Wl,--wrap=read -Wl,--wrap=setsockopt -Wl,--wrap=select -Wl,--wrap=socket -Wl,--wrap=write -Wl,--wrap=fcntl -Wl,--wrap=freeaddrinfo"
+)
+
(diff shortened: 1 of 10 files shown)


utilities/xmlrpcpp/test/CMakeLists.txt
```diff
@@ -3,6 +3,17 @@ if(TARGET xmlrpcvalue_base64)
...
 
+catkin_add_gtest(test_socket
+  test_socket.cpp
+  test_system_mocks.c
+  ../src/XmlRpcSocket.cpp
+  ../src/XmlRpcUtil.cpp
+)
+set_target_properties(test_socket PROPERTIES
+  LINK_FLAGS
+  "-Wl,--wrap=accept -Wl,--wrap=bind -Wl,--wrap=close -Wl,--wrap=connect -Wl,--wrap=getaddrinfo -Wl,--wrap=getsockname -Wl,--wrap=listen -Wl,--wrap=read -Wl,--wrap=setsockopt -Wl,--wrap=select -Wl,--wrap=socket -Wl,--wrap=write -Wl,--wrap=fcntl -Wl,--wrap=freeaddrinfo"
```
mikepurvis: Fancy. TIL.

utilities/xmlrpcpp/test/test_socket.cpp
```diff
@@ -0,0 +1,1368 @@
+
```
mikepurvis, line 23: Copyright block.

trainman419: Which license should I use here? I tend towards BSD but the rest of the library is licensed LGPL.

tfoote: Either one would be fine. Embedding BSD inside an LGPL project is not a problem, though you should update the package.xml to declare both. If everything else is LGPL it's simpler/"cleaner" to follow the exisiting projects license. So I guess I'd suggest defaulting to BSD if you're ok with it and it's more liberal.

trainman419: Ok; added an LGPL copyright block to all of the files that I've added in this and the previous PR.

utilities/xmlrpcpp/test/test_system_mocks.c
```diff
@@ -0,0 +1,82 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ *
+ * Author: Austin Hendrix <austin@zoox.com>
+ *
+ */
+
+// The auto-generated mocks here don't use their parameters, so we disable
+// that warning.
+#pragma GCC diagnostic ignored "-Wunused-parameter"
```
dirk-thomas: Does this line need to be wrapped in a conditional in order to not fail on Windows?

utilities/xmlrpcpp/test/CMakeLists.txt
dirk-thomas: I would expect these arguments to fail on Windows?

utilities/xmlrpcpp/test/test_system_mocks.c
trainman419: I'm not sure. If this is built using gcc on windows it's probably fine, but the MSVC compiler may reject it.
I don't have a windows development environment so I can't test this, and REP3 does not specify Windows support or a target compiler or toolchain for Windows either, so I don't think this support should be mandatory.

dirk-thomas: Since some (brave) users do build ROS from source on Windows I would rather not introduce changes which are known ahead of time to break that. (Their might be regressions on Windows due to unforeseen side effects which we can't check without build infrastructure - we can't do anything about that atm.)

trainman419: Ok; I've wrapped both of these in feature guards so that we won't build the mocks on Windows and won't use the GCC diagnostic pragma unless `__GNUC__` is defined.

I don't have a Windows machine and don't know how to set up the development environment on it, so I don't have any way to test this.

dirk-thomas: Since e.g. `clang` also supports this I would suggest using `#ifndef _WIN32` instead.

utilities/xmlrpcpp/test/CMakeLists.txt
```diff
@@ -3,6 +3,19 @@ if(TARGET xmlrpcvalue_base64)
...
+)
```
dirk-thomas: The block within the `if` and `endif` should be indented one level.

trainman419: Done.

dirk-thomas:

dirk-thomas:

PR #1205, opened 2017-10-26
changes between 1.12.7 and 1.13.4 for backporting

dirk-thomas:
The following list of changes has been integrated into ros_comm 1.13.4 (Lunar) since the last Kinetic release (1.12.7).

**Backported:** (these changes are part of this PR)

* fix open mode on Windows #1005

  * bug fix

* fix BZip2 inclusion #1016

  * bug fix

* make roslaunch-check respect if/unless attribute on include #998

  * helpful improvement, backport requested, low chance of regression

* fix rosbag::View::iterator copy assignment operator #1017

  * bug fix, backport requested

* fix rosmsg show from bag #1006

  * bug fix, backport requested, dependent changes in genpy have been backported

* add subscriber to connection log messages #1023

  * helpful improvement

* ensure cwd exists #1031

  * bug fix

* Sleep in rospy wait_for_service even if exceptions raised #1025

  * bug fix

* Avoid deleting XmlRpcClient's while they are being still in use on another thread #1013

  * bug fix, chance for regression?

* abort topic lookup on connection refused #1044

  * helpful improvement

* fix bug in transport_tcp #1050

  * bug fix, backport requested

* fix race condition that lead to miss first message #1058

  * bug fix, backport requested

* ensure pid file is removed on exit #1057

  * bug fix

* changed the check command output to be a little bit more specific #1067

  * helpful improvement

* fix pid file removing condition #1084

  * bug fix for backported change #1057

* add option to reset timer when time moved backwards #1083

  * helpful improvement

* add missing mutex lock for publisher links #1090

  * bug fix

* improve rospy.logXXX_throttle performance #1091

  * helpful improvement

* added logging output when `roslogging` cannot change permissions #1068

  * helpful improvement

* Check /clock publication neatly in publishtest #973

  * helpful improvement

* fix to a critical stack buffer overflow vulnerability #1092

  * bug fix

* only launch core nodes if master was launched by roslaunch #1098

  * bug fix

* made copying rosbag::Bag a compiler error to prevent crashes and added a swap function instead #1000

  * bug fix

* add missing header for writev #1105

  * helpful improvement to support newer toolchains

* add SteadyTimer #1014

  * change behavior, requested for backport

* close CLOSE_WAIT sockets by default #1104

  * helpful improvement

* fix handling connections without indices #1109

  * bug fix

* fix rostopic prining long integers #1110

  * bug fix

* update tests to match stringify changes #1125

  * needs genpy 0.6.6 to be released into Kinetic

[...]

PR #1220, opened 2017-11-07
Revert "Replaced deprecated lz4 function call"

dirk-thomas:
Follow up of #1205.


dirk-thomas:
Since Debian Jessie was never cleanly rebuild this can't be tested by CI :worried:

PR #1221, opened 2017-11-08
Unit tests and bug fixes for XmlRpcClient

trainman419:
Add unit tests for and fix the following bugs in XmlRpcClient:

* Fix error handling in XmlRpcClient so that it closes file descriptors and terminates correctly on errors. This fixes the file descriptor leaks reported in https://answers.ros.org/question/250393/rosout-leaks-file-descriptors/
* Fix handling of partial buffer reads and writes in XmlRpcClient.
* Report symbolic state names in XmlRpcClient.
* Make async XmlRpcClient terminate correctly on error.

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -328,12 +328,7 @@ XmlRpcClient::generateHeader(size_t length) const
   header += buff;
   header += "Content-Type: text/xml\r\nContent-length: ";
 
-  // Windows and glibc have different modifier flags for printing size_t
-#ifdef _MSC_VER
-  snprintf(buff,40,"%Id\r\n\r\n", length);
-#else
   snprintf(buff,40,"%zd\r\n\r\n", length);
-#endif
 
   return header + buff;
 }
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -2,7 +2,8 @@
 #include "xmlrpcpp/XmlRpcClient.h"
 
 #include "xmlrpcpp/XmlRpcSocket.h"
-#include "xmlrpcpp/XmlRpc.h"
+#include "xmlrpcpp/XmlRpcUtil.h"
+#include "xmlrpcpp/XmlRpcValue.h"
 
 #include <stdio.h>
 #include <stdlib.h>
@@ -28,6 +29,24 @@ const char XmlRpcClient::METHODRESPONSE_TAG[] = "<methodResponse>";
 const char XmlRpcClient::FAULT_TAG[] = "<fault>";
 
 
+const char * XmlRpcClient::connectionStateStr(ClientConnectionState state) {
+  switch(state) {
+    case NO_CONNECTION:
+      return "NO_CONNECTION";
+    case CONNECTING:
+      return "CONNECTING";
+    case WRITE_REQUEST:
+      return "WRITE_REQUEST";
+    case READ_HEADER:
+      return "READ_HEADER";
+    case READ_RESPONSE:
+      return "READ_RESPONSE";
+    case IDLE:
+      return "IDLE";
+    default:
+      return "UNKNOWN";
+  }
+}
 
 XmlRpcClient::XmlRpcClient(const char* host, int port, const char* uri/*=0*/)
 {
@@ -79,7 +98,7 @@ struct ClearFlagOnExit {
 bool
 XmlRpcClient::execute(const char* method, XmlRpcValue const& params, XmlRpcValue& result)
 {
-  XmlRpcUtil::log(1, "XmlRpcClient::execute: method %s (_connectionState %d).", method, _connectionState);
+  XmlRpcUtil::log(1, "XmlRpcClient::execute: method %s (_connectionState %s).", method, connectionStateStr(_connectionState));
 
   // This is not a thread-safe operation, if you want to do multithreading, use separate
   // clients for each thread. If you want to protect yourself from multiple threads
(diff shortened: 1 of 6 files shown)


utilities/xmlrpcpp/src/XmlRpcClient.cpp
```diff
@@ -295,19 +323,16 @@ XmlRpcClient::generateHeader(std::string const& body)
...
-  sprintf_s(buff,40,":%d\r\n", _port);
-#else
-  sprintf(buff,":%d\r\n", _port);
-#endif
+  snprintf(buff,40,":%d\r\n", _port);
 
   header += buff;
   header += "Content-Type: text/xml\r\nContent-length: ";
 
+  // Windows and glibc have different modifier flags for printing size_t
```
dirk-thomas: `%zu` should work across all platforms.

utilities/xmlrpcpp/src/XmlRpcClient.cpp
```diff
@@ -322,6 +347,8 @@ XmlRpcClient::writeRequest()
   // Try to write the request
   if ( ! XmlRpcSocket::nbWrite(this->getfd(), _request, &_bytesWritten)) {
     XmlRpcUtil::error("Error in XmlRpcClient::writeRequest: write error (%s).",XmlRpcSocket::getErrorMsg().c_str());
+    // If the write fails, we had an unrecoverable error. Close the socket.
+    close();
```
dirk-thomas, line 346: Instead of calling `close()` in the various locations where a read / write fails would it be possible that the calling code actually does that when receiving `false` instead? Why should this be added here and change the behavior of the API?

utilities/xmlrpcpp/src/XmlRpcClient.cpp
trainman419: No; `z` is a GNU-specific size specifier and is not listed in Microsoft's documentation: https://msdn.microsoft.com/en-us/library/tcxf1dw6.aspx

wjwwood: I think the `z` is defined for C++11: http://en.cppreference.com/w/cpp/io/c/fprintf, and it appears to work on Windows (we use it all the time in ROS 2).

I tried this program out on http://webcompiler.cloudapp.net/:

```c++
#include <cstdio>

int main()
{
   size_t i = 42;
   printf("%zu\n", i);
   return 0;
}
```

The output was:

```
Compiled with  /EHsc /nologo /W4
main.cpp

Compilation successful!

Total compilation time: 328ms

42

Total execution time: 140ms
```

utilities/xmlrpcpp/src/XmlRpcClient.cpp
trainman419: No; there is already a contract between this code and the calling code that determines if the caller should call close on error or not, based on the return value of the setKeepOpen() function: https://github.com/ros/ros_comm/blob/8b741ac9232a9fadefb19e7fdee654225619d2f8/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp#L157-L160

It looks like this is used in the client to keep the socket open after a request is complete so that the socket is available for subsequent requests. I didn't want to change the behavior of that mechanism or interfere with the concept of keeping a client TCP socket open between requests so I opted to keep the logic within the XmlRpcClient class.

I contemplated a few different ways to structure this change:
1. Make the client code call close in the correct scenarios, but this breaks encapsulation and further exposes the client to implementation details of this class.
1. Make the XmlRpcSocket class close the socket on error. This is cleaner, but would require a complete rewrite of XmlRpcSocket to make it own the file descriptor. This was a bigger set of changes than I wanted to make and I thought they would be seen as too invasive to be accepted into ros_comm.
1. Test XmlRpcSocket to make sure that it always returns false when there is an error, and then fix XmlRpcClient so that it always closes the socket when there is an error. I deemed this to be the least invasive and least likely to include changes that could inadvertently change the library behavior.

This is not a breaking change to the API; file descriptors are only closed in places where they were already in an error state, and it's safe to call close multiple times, so client code which calls close again will not fail. Any client which was expecting a persistent connection and which ends up with a broken connection will behave better than the previous implementation. In the previous implementation it would discover the broken socket when attempting to send a new request and would be forced to close the socket and reconnect. In the updated implementation the socket is already closed, so the client immediately knows that it needs to establish a new connection.

utilities/xmlrpcpp/src/XmlRpcClient.cpp
trainman419: We use C++11 internally but it was my understanding that this code still builds in C++03 mode to avoid forcing downstream packages to enable C++11, so I took extra steps to avoid using any C++11 features.

dirk-thomas: The `.cpp` file can be compiled with C++11. We don't want to expose C++11 feature in the headers though since that requires all downstream packages to choose C++11 too.

trainman419: Since these changes are fixing stability issues with XmlRpcpp, it also my hope that they will be backported to the LTS versions of ROS.

dirk-thomas: After they have been released into Lunar and had no regressions for a while the changes can be considered for backporting into Kinetic. But even for Kinetic a C++11 compiler is a requirement already. See [REP 3](http://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021).

wjwwood: That's a fair point, but given that all the modern MSVC compilers use C++11 by default (there is no `-std=c++11` like option), and all other compilers support `zu` without C++11 (that we care about), I think it's safe, in this instance, to just use `%zu`.

trainman419: Updated.

dirk-thomas: `size_t` is *unsigned* so it needs to be `%zu` (not `%zd`).

trainman419:
Bump. Is there anything else I can do to keep this moving?

dirk-thomas:
> Is there anything else I can do to keep this moving?

Please see latest pending comment: https://github.com/ros/ros_comm/pull/1221#discussion_r150038518

utilities/xmlrpcpp/src/XmlRpcClient.cpp
trainman419: Updated.

trainman419:
Sorry; I missed the comment because it was on an outdated diff. Fixed the format string.

dirk-thomas:
Thank you for your effort on this.

PR #1228, opened 2017-11-09
Make master process explicitly 'required' for parent launch

paulbovbel:
In the existing implementation, if `rosmaster` dies, the parent launch will continue running unaffected.

This is particularly unpleasant when the launch is a standalone `roscore`. Adding the required argument here causes any roslaunch process manager that brings up `rosmaster` to also come down if `rosmaster` dies unexpectedly.

### DIFF
--- a/tools/roslaunch/src/roslaunch/nodeprocess.py
+++ b/tools/roslaunch/src/roslaunch/nodeprocess.py
@@ -100,7 +100,7 @@ def create_master_process(run_id, type_, ros_root, port, num_workers=NUM_WORKERS
 
     _logger.info("process[master]: launching with args [%s]"%args)
     log_output = False
-    return LocalProcess(run_id, package, 'master', args, os.environ, log_output, None)
+    return LocalProcess(run_id, package, 'master', args, os.environ, log_output, None, required=True)
 
 def create_node_process(run_id, node, master_uri):
     """


paulbovbel:
Maintainers, any thoughts on this change? This seems like sensible default behaviour, but if it's not I'm happy to put it behind a command line arg.

dirk-thomas:
I am travelling for the next couple weeks so it might be a while before I can take a look at this.

mikepurvis:
I can't think of any scenario where this wouldn't be desirable behaviour— I guess with the one caveat that the master isn't _actually_ required once the system is up and running if there are no new topics coming up after initial setup. But a system in that state is pretty minimally usable, since you won't be able to inspect much of anything.

That said, I'm curious what's crashing your rosmaster. Do you have core dumps from that that are shareable, or a MWE? In parallel to this change, it would be great to figure out the underlying cause of the crash in what should be a relatively simple, stable component.

In any case, 👍  from me.

paulbovbel:
Sadly the errors were transient, so we don't have an MWE. Hopefully with the patched behaviour, it'll be easier to spot a downed roscore and collect artifacts for reproduction.

paulbovbel:
Hello maintainers, any further thoughts on this patch?

paulbovbel:
Ping for review

dirk-thomas:
Thank you the patch. It sounds reasonable to me.

PR #1232, opened 2017-11-13
Unit tests for XmlRpcDispatch

trainman419:
* Fix handling of out-of-band (Exception) request and and event processing to match previous select() implementation.
* Update comments to describe the events that actually trigger and Exception state.
* Unit tests for XmlRpcDispatch to verify that it calls poll() correctly, requests the correct events and calls the correct handler for those events.

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -159,9 +159,8 @@ XmlRpcDispatch::work(double timeout)
 
       if ( ! newMask) {
         _sources.erase(thisIt);  // Stop monitoring this one
-        if ( ! src->getKeepOpen()) {
+        if ( ! src->getKeepOpen())
           src->close();
-        }
       } else if (newMask != (unsigned) -1) {
         thisIt->getMask() = newMask;
       }
@@ -230,4 +229,5 @@ XmlRpcDispatch::getTime()
   return (tv.tv_sec + tv.tv_usec / 1000000.0);
 #endif /* USE_FTIME */
 }
+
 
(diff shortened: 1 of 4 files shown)


trainman419:
Bump. @dirk-thomas @mikepurvis ?

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
```diff
@@ -132,11 +134,12 @@ XmlRpcDispatch::work(double timeout)
...
       bool readable = (pfd.events & POLLIN_REQ) == POLLIN_REQ;
       bool writable = (pfd.events & POLLOUT_REQ) == POLLOUT_REQ;
+      bool oob = (pfd.events & POLLEX_REQ) == POLLEX_REQ;
       if (readable && (pfd.revents & POLLIN_CHK))
         newMask &= src->handleEvent(ReadableEvent);
       if (writable && (pfd.revents & POLLOUT_CHK))
         newMask &= src->handleEvent(WritableEvent);
-      if (pfd.revents & POLLEX_CHK)
+      if (oob && (pfd.revents & POLLEX_CHK))
         newMask &= src->handleEvent(Exception);
```
mikepurvis, line 143: Can you describe more about what is going on here? Is this a behaviour change/fix which arose as a result of the added tests?

mikepurvis:
This all seems reasonable to me, but TBH I've never really had to grapple with these internals before, so my thumbs-up is not the most valuable one.

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
trainman419: I initially wrote this test against the previous version of this, which was using `select()`, since our internal fork of ROS is mostly based on Indigo. When I updated this test to the most recent version of ros_comm to get the `poll()` implementation, it no longer behaved the same way when there was an exception event on the socket. This fixes it so that the `poll()` implementation behaves more like the original `select()` implementation.

Also of note is that for TCP sockets, the exception event actually indicates the reception of out-of-band data over the TCP connection, and doesn't indicate an error event on the socket. I've updated the comments and the implementation to reflect this and restore the original behavior.

(I light of this I think the original behavior might be a little misguided, but my goal here is to preserve all of the existing behavior that isn't clearly a bug.)

dirk-thomas (approved):
LGTM.

I will wait with the merge until CI is green and @trainman419 had a chance to answer the pending question from @mikepurvis.

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
mikepurvis: Fun, okay. Yeah, we're having some similar issues with select/poll/epoll subtleties over in #1217.

dirk-thomas:

PR #1234, opened 2017-11-14
added parameter to omit topics list from rosout logs

trobertson-cpr:
Added a cached parameter to omit topics list from rosout logs. The parameter /rosout/omit_topics stops rosout.cpp from adding the list of all topics to the logs. It does not stop rospy or roscpp from generating the list of topics and sending them over the network stack.


dirk-thomas:
Thank you for the patch. Please consider adding information about this new feature to the wiki page and post a link to the modification here.

PR #1235, opened 2017-11-15
backward compatibility with libconsole-bridge in Jessie (take 3)

mikaelarguedas:
I didnt follow the original discussions regarding console bridge support, so forgive me if this has been discussed before.
Shouldn't all macros be treated the same ?
I'd expect any package including this header to not be able to use logWarn or logInform anymore while they're still able to use logError and logDebug. Is that the expected behavior ?

### DIFF
--- a/tools/rosbag_storage/include/rosbag/bag.h
+++ b/tools/rosbag_storage/include/rosbag/bag.h
@@ -62,25 +62,23 @@
 
 #include "console_bridge/console.h"
 
-#ifdef logDebug
-# ifndef CONSOLE_BRIDGE_logDebug
-#  define CONSOLE_BRIDGE_logDebug logDebug
-# endif
+// Remove this when no longer supporting platforms with libconsole-bridge-dev < 0.3.0,
+// in particular Debian Jessie: https://packages.debian.org/jessie/libconsole-bridge-dev
+#ifndef CONSOLE_BRIDGE_logDebug
+# define CONSOLE_BRIDGE_logDebug(fmt, ...)  \
+  console_bridge::log(__FILE__, __LINE__, console_bridge::CONSOLE_BRIDGE_LOG_DEBUG, fmt, ##__VA_ARGS__)
 #endif
-#ifdef logInform
-# ifndef CONSOLE_BRIDGE_logInform
-#  define CONSOLE_BRIDGE_logInform logInform
-# endif
+#ifndef CONSOLE_BRIDGE_logInform
+# define CONSOLE_BRIDGE_logInform(fmt, ...) \
+  console_bridge::log(__FILE__, __LINE__, console_bridge::CONSOLE_BRIDGE_LOG_INFO,  fmt, ##__VA_ARGS__)
 #endif
-#ifdef logWarn
-# ifndef CONSOLE_BRIDGE_logWarn
-#  define CONSOLE_BRIDGE_logWarn logWarn
-# endif
+#ifndef CONSOLE_BRIDGE_logWarn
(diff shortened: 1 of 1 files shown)


tools/rosbag_storage/include/rosbag/bag.h
```diff
@@ -71,10 +71,22 @@
...
+# endif
 #endif
 #if defined logWarn
-# undef logWarn
+# ifdef CONSOLE_BRIDGE_logWarn
+#  undef logWarn
+# else
+// Remove this when no longer supporting platforms with libconsole-bridge-dev < 0.3.0,
+// in particular Debian Jessie: https://packages.debian.org/jessie/libconsole-bridge-dev
+#  define CONSOLE_BRIDGE_logInform logWarn
```
dirk-thomas: Typo :`Inform`?

dirk-thomas (commented):
Have you tested that this resolves the failing downstream builds on Jessie?

tools/rosbag_storage/include/rosbag/bag.h
mikaelarguedas: Indeed will fix

mikaelarguedas:
> Have you tested that this resolves the failing downstream builds on Jessie?

No I don't think it will. I just noticed the inconsistency when looking into it and figured it deserved fixing as well. Haven't found the source of pcl_ros failure yet

mikaelarguedas:
As I don't have more time to look into the breakage of downstream Jessie packages, I'd suggest merging this as it at least makes the definition of macro the same for all logging macros (but no need to release before the actual problem is fixed).
I suspect the issue comes from class_loader and rosbag acting on these definitions in a different way and don't play nice with each others. If anybody has time to look into it that'd be greatly appreciated
http://build.ros.org/view/Kbin_dj_dJ64/job/Kbin_dj_dJ64__pcl_ros__debian_jessie_amd64__binary/42

mikaelarguedas:
@mikepurvis @dirk-thomas FYI

My conclusion so far is that the current undef logic is very fragile and should be reverted (this PR updated accordingly)
If the main goal is to not exposed the non-namespaced macros I would suggest to request a new version of console_bridge to be released without the deprecated macros (https://github.com/ros/console_bridge/issues/18)

The current logic is not scalable given that if any other header does the same logic it will result in breaking any package including both headers (on a system with a version of console_bridge lower than 0.3.0).

Let's imagine I have a package `foo` depending on `bar` and `baz`:
bar.h and baz.h are identical
```
#include <ros/ros.h>
#ifdef ROS_INFO
# ifndef MYNAMESPACED_ROS_INFO
#  define MYNAMESPACED_ROS_INFO ROS_INFO
# else
#  undef ROS_INFO
# endif
#endif
```

foo.cpp:
```
#include <ros/ros.h>
#include <bar/bar.h>
#include <baz/baz.h>

int main(int argc, char** argv) {
  MYNAMESPACED_ROS_INFO("Hello World");

  return 0;
}
```

:boom: :boom: 
```
error: ‘ROS_INFO’ was not declared in this scope
   MYROS_INFO("Hello World");
                                                 ^
```

mikaelarguedas:
The current state of this PR has been tested on jessie and does fix the failing pcl_ros

dirk-thomas:
Can you please readd the removed comments that the whole block should be removed in the future when all targeted platforms provide a new enough version of console_bridge.

dirk-thomas:
After some long offline discussion with @mikaelarguedas we decide to move this hack into `rospack`.

dirk-thomas:
Confidence, :crossed_fingers: and time pressure makes me merge this ahead of CI :pray:

PR #1236, opened 2017-11-15
Enable Trusty support for lunar-devel

ayrton04:
This PR enables us to build the latest version of ros_comm on Trusty. The LZ4 API changed at version 1.7.2 to use `LZ4_compress_default` instead of `LZ4_compress_limitedOutput`, and this PR defines a version-dependent macro that allows us to build with the older LZ4 API.

### DIFF
--- a/utilities/roslz4/src/lz4s.c
+++ b/utilities/roslz4/src/lz4s.c
@@ -47,11 +47,11 @@
 #define DEBUG(...)
 #endif
 
-// Make sure the LZ library has the function we need
-// Note that the later versions of LZ contain a macro that does this version math, but the earlier
+// Make sure the LZ4 library has the function we need
+// Note that the later versions of LZ4 contain a different macro that does this version math, but the earlier
 // versions lack it.
-#define LZ_VERSION (LZ4_VERSION_MAJOR * 10000 + LZ4_VERSION_MINOR * 100 + LZ4_VERSION_RELEASE)
-#if LZ_VERSION >= 10700
+#define LZ4_VERSION (LZ4_VERSION_MAJOR * 100 * 100 + LZ4_VERSION_MINOR * 100 + LZ4_VERSION_RELEASE)
+#if LZ4_VERSION >= 10702
 #define COMPRESS_DEFAULT LZ4_compress_default
 #else
 #define COMPRESS_DEFAULT LZ4_compress_limitedOutput
--- a/utilities/roslz4/src/lz4s.c
+++ b/utilities/roslz4/src/lz4s.c
@@ -51,10 +51,10 @@
 // Note that the later versions of LZ4 contain a different macro that does this version math, but the earlier
 // versions lack it.
 #define LZ4_VERSION (LZ4_VERSION_MAJOR * 100 * 100 + LZ4_VERSION_MINOR * 100 + LZ4_VERSION_RELEASE)
-#if LZ4_VERSION >= 10702
-#define COMPRESS_DEFAULT LZ4_compress_default
+#if LZ4_VERSION >= 10700
+#define LZ4_COMPRESS_DEFAULT LZ4_compress_default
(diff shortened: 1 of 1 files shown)


dirk-thomas:
[This](https://github.com/lz4/lz4/commit/1b17bf2ab8cf66dd2b740eca376e2d46f7ad7041) commit seems to have added the new function `LZ4_compress_default`. According to the commit the first tag it is part of is `r130`. That tag identifies itself as version 1.7.0 which is also the same for the earlier tag `r129`. I also don't see a tag for version 1.7.2 which you mentioned. What information did you base the choice on?

Beside that ROS Lunar is specifically *not* targeting Ubuntu Trusty. Many other packages will have similar problems. Therefore I am not convinced that it makes sense to add this patch to the code base.

ayrton04:
There wasn't a tag; the version number was incremented in a commit without getting tagged. My original [commit](https://github.com/ros/ros_comm/pull/1236/commits/c333b05f5f3f3e29da146fd544ce771c62b68a24#diff-908b2ea7de4ac66e6e36b226b0199d3cR54) (and the version we had been using internally) was set to use version 1.7.0 as the version in which the switch occurred, but in verifying that I had the version number right before PR-ing it, I managed to make it wrong.

Given that Lunar is not targeting Trusty and that this check would ensure the behavior doesn't change for anyone running on supported platforms, is there danger in merging (once I correct the version number, that is)? We have been using this version of `ros_comm` internally for a while with no problems, though we admittedly use updated versions of many package that were not originally intended for Trusty.

In any case, it's not a big deal if you'd prefer not to merge it. We're just trying to minimize our differences with upstream.

dirk-thomas:
> is there danger in merging (once I correct the version number, that is)?

That is why I am a little hesitant. Basically every change has the potential for regressions. With the recent Kinetic backports we noticed that sadly multiple times that is probably why I am currently extra cautious.

I guess this can be merged since it *shouldn't* effect the targeted platforms. One thing to change beside the exact version is the name of the macro. `COMPRESS_DEFAULT` seems too generic imo. Can you please rename it to `LZ4_COMPRESS_DEFAULT`.

mikepurvis:
I'm supportive of merging. Obviously we can't support old platforms forever, but when it's such a small change, it's certainly nice to provide that migration path forward.

I know at Clearpath we were using the Kinetic+ versions of many core repos long before we actually switched our underlying OS to 16.04, and it sure was nice being able to stage the rollout like that.

ayrton04:
Sounds good. Will fix.

utilities/roslz4/src/lz4s.c
```diff
@@ -47,6 +47,16 @@
 #define DEBUG(...)
 #endif
 
+// Make sure the LZ4 library has the function we need
+// Note that the later versions of LZ4 contain a different macro that does this version math, but the earlier
+// versions lack it.
+#define LZ4_VERSION (LZ4_VERSION_MAJOR * 100 * 100 + LZ4_VERSION_MINOR * 100 + LZ4_VERSION_RELEASE)
+#if LZ4_VERSION >= 10700
```
dirk-thomas: >  That tag identifies itself as version 1.7.0 which is also the same for the earlier tag `r129`.

As mentioned before since an older release (`r129`) identifies itself with this version too I don't think it can be used for the check.

ayrton04: Oh, apologies, I misunderstood your issue. I will get the first post-1.7.0 tag, which I believe is 1.7.3.

dirk-thomas: `r131` identifies as 1.7.1. Would that be more precise?

ayrton04: Yes, it would.

dirk-thomas:
Thank you for iterating on this.

PR #1237, opened 2017-11-16
Fixed warn of --max-splits without --split

v-do:
`rosbag record` always gives this warning even without specifying `--max-splits`:
> [ WARN] [1510806945.020499157]: --max-splits is ignored without --split

The `--max-splits` option has a default value causing the program to always think that `--max-splits` is specified. This fix removes the default command-line option value but still properly initialises the value to 0.


dirk-thomas:
@v-do Thank you for the fix.

@contradict FYI introduced in #866.

PR #1238, opened 2017-11-18
place console_bridge macros definition in header and use it everywher…

mikaelarguedas:
…e console_bridge is included

This is a follow up of https://github.com/ros/ros_comm/pull/1235.

- Add a header to define the namespaces versions of console_bridge macros if they don't exist (for libconsole-bridge-dev < 0.3.0) 
- Use the namespaced versions of the console_bridge macros everywhere
- Include the compatibility header everywhere console_bridge is included

Compared to #1235, this is not necessary to fix downstream packages so doesn't need to be released right now but is a nice to have for both consistency and having deterministic behavior regardless of what was included before.

### DIFF
--- a/tools/rosbag_storage/include/rosbag/bag.h
+++ b/tools/rosbag_storage/include/rosbag/bag.h
@@ -61,24 +61,9 @@
 #include <boost/iterator/iterator_facade.hpp>
 
 #include "console_bridge/console.h"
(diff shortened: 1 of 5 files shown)


dirk-thomas:
Thank you for the patch. Just to highlight: this is intentionally targeting `kinetic-devel` since it is not required for Lunar and newer.

PR #1239, opened 2017-11-18
use namespaced macros instead of deprecated macros

mikaelarguedas:
The rest of rosbag uses the new namespaced macros. 
Given that the old console bridge macros are undefined in:
https://github.com/ros/ros_comm/blob/35d1e0d788890fe1e33c71a4f97dcdfc3c4f074a/tools/rosbag_storage/include/rosbag/bag.h#L65-L76
The rest of the code base needs to use the new ones


mikepurvis:
LGTM.

PR #1243, opened 2017-11-22
Tests and bug fixes for XmlRpcServer

### LINKED ISSUE #914: xmlrpccpp spins forever when no more file handles

trainman419:
Add integration tests for the XmlRpcServer interacting with an XmlRpcClient in the same process, particularly around the behavior when the server process runs out of available file handles.

Update the XmlRpcServer with two different mitigations for file handle exhaustion (Fixes #914, replaces #960 and #977):
 * Measure the number of free file handles and reject incoming connections if the pool of free file descriptors is too small. This actively rejects incoming clients instead of waiting for complete file
handle exhaustion, which would leave clients in a pending state until a file descriptor becomes free.
 * If accept fails due to complete file handle exhaustion, temporarily stop calling accept on this socket. This prevents a busy-loop where poll() believes the listening socket is readable, but accept() fails to
allocate a file descriptor and leaves the socket in a readable state.

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcServer.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcServer.cpp
@@ -9,16 +9,40 @@
 #include "xmlrpcpp/XmlRpcUtil.h"
 #include "xmlrpcpp/XmlRpcException.h"
 
+#include <errno.h>
+#include <string.h>
+#include <sys/resource.h>
 
 using namespace XmlRpc;
 
+const int XmlRpcServer::FREE_FD_BUFFER = 32;
+const double XmlRpcServer::ACCEPT_RETRY_INTERVAL_SEC = 1.0;
 
 XmlRpcServer::XmlRpcServer()
+  : _introspectionEnabled(false),
+    _listMethods(0),
+    _methodHelp(0),
+    _port(0),
+    _accept_error(false),
+    _accept_retry_time_sec(0.0)
 {
-  _introspectionEnabled = false;
-  _listMethods = 0;
-  _methodHelp = 0;
-  _port = 0;
+  struct rlimit limit = { .rlim_cur = 0, .rlim_max = 0 };
+  int max_files = 1024;
+
+  if(getrlimit(RLIMIT_NOFILE, &limit) == 0) {
+    max_files = limit.rlim_max;
+  } else {
+    XmlRpcUtil::error("Could not get open file limit: %s", strerror(errno));
+  }
+  pollfds.resize(max_files);
+  for(int i=0; i<max_files; i++) {
+    // Set up file descriptor query for all events.
+    pollfds[i].fd = i;
+    pollfds[i].events = POLLIN | POLLPRI | POLLOUT;
+  }
+
+  // Ask dispatch not to close this socket if it becomes unreadable.
+  setKeepOpen(true);
 }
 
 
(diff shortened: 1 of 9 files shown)


utilities/xmlrpcpp/test/CMakeLists.txt
```diff
@@ -3,6 +3,23 @@ if(TARGET xmlrpcvalue_base64)
   target_link_libraries(xmlrpcvalue_base64 xmlrpcpp)
 endif()
 
+# Some of the tests that follow use boost threads.
```
mikepurvis: Please add `<test_depend>boost</test_depend>`, since inclusion of this file is guarded by `CATKIN_ENABLE_TESTING`.

utilities/xmlrpcpp/test/test_dispatch_live.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+TEST_F(XmlRpcTest, HelloNonBlock) {
+  XmlRpcClient c("localhost", port);
+  XmlRpcValue noArgs, result;
+
+  // Call the Hello method, non-blocking
+  ASSERT_TRUE(c.executeNonBlock("Hello", noArgs));
+
+  bool done = false;
+  for (int i = 0; i < 30; i++) {
```
mikepurvis: Since this is a new file, ROS brace style, please (here and elsewhere). If you like, run astyle on it:

http://wiki.ros.org/roslint#Tips

utilities/xmlrpcpp/src/XmlRpcServer.cpp
```diff
@@ -32,14 +56,14 @@ XmlRpcServer::~XmlRpcServer()
 
 
 // Add a command to the RPC server
-void 
+void
```
mikepurvis: Please revert these whitespace changes (a follow-on commit which re-adds them is acceptable, as it will get squashed down anyway).

utilities/xmlrpcpp/test/HelloTest.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+  ASSERT_TRUE(c.execute("system.listMethods", noArgs, result));
+
+  XmlRpcValue methods;
+  methods[0] = "Hello";
+  methods[1] = "HelloName";
+  methods[2] = "Sum";
+  methods[3] = "system.listMethods";
+  methods[4] = "system.methodHelp";
+  methods[5] = "system.multicall";
```
mikepurvis, line 159: I'm partial to `boost::assign::list_of` in these instances, but what's here is fine.

mikepurvis:
Looks great. Will wait for @dirk-thomas to weigh in, but this is a terrific set of validations, thank you!

utilities/xmlrpcpp/test/CMakeLists.txt
trainman419: Done.

utilities/xmlrpcpp/src/XmlRpcServer.cpp
trainman419: Reverted.

utilities/xmlrpcpp/test/test_dispatch_live.cpp
trainman419: Done.

dirk-thomas:
The patch itself looks fine to me. It does break ABI though so I am (as always) hesitant to merge this into an already released distro. I guess the argument will be "it doesn't matter for Lunar to break ABI and we want the patch now instead of for Melodic"?

trainman419:
I don't really have a stake in which version this merges into.

I will say that it fixes a fairly substantial bug for our use case, so I think it there is a strong argument for including it in Lunar. Since this isn't an API that is used by most clients, and it will be released in sync with roscpp, I think the impact of the ABI breakage will be low.

dirk-thomas:
Thank you for the patch.

NikolausDemmel:
Looks like this might have broken roscore on macos: https://github.com/ros/ros_comm/issues/1357

PR #1244, opened 2017-11-22
Test and fix uninitialized data in XmlRpcClient

trainman419:
Fix test problems that were exposed by uninitialized memory, properly initialize memory and add tests for proper memory initialization.

Since uninitialized memory was causing the retry logic to be skipped in these tests, the tests have been updated to properly test the retry logic in the client.

### DIFF
--- a/utilities/xmlrpcpp/test/test_client.cpp
+++ b/utilities/xmlrpcpp/test/test_client.cpp
@@ -79,6 +79,10 @@ TEST_F(MockSocketTest, constructor) {
   EXPECT_EQ(false, a._eof);
   EXPECT_EQ(true, a.getKeepOpen());
   EXPECT_EQ(-1, a.getfd());
+  EXPECT_EQ(0, a._sendAttempts);
+  EXPECT_EQ(0, a._bytesWritten);
+  EXPECT_EQ(0, a._contentLength);
+  EXPECT_FALSE(a._isFault);
   EXPECT_FALSE(sourceInList(&a, a._disp._sources));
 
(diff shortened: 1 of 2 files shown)


utilities/xmlrpcpp/test/test_client.cpp
```diff
@@ -829,14 +877,21 @@ TEST_F(MockSocketTest, readHeader_partial_err) {
...
-  EXPECT_FALSE(a.readHeader());
+  // Expect a reconnect attempt
+  Expect_socket(8);
+  Expect_setNonBlocking(8, true);
+  Expect_connect(8, "localhost", 42, true);
+
+  EXPECT_TRUE(a.readHeader());
   // Check that state machine is in the correct state after getting the header.
-  EXPECT_EQ(XmlRpcClient::NO_CONNECTION, a._connectionState);
+  EXPECT_EQ(XmlRpcClient::WRITE_REQUEST, a._connectionState);
```
mikepurvis, line 888: What's the rationale here?

trainman419: The original logic was falling into the not connected state during this test because _sendAttempts was not correctly initialized to zero. This caused the client to immediately fail instead of attempting to retry.

By fixing the initialization, the client now correctly transitions to the write request state if the header read fails on the first attempt. I've updated this test and the others to correctly capture the retry behavior and the failure if readHeader fails again on a second attempt.

dirk-thomas:
Thank you for the patch.

PR #1247, opened 2017-11-24
use SteadyTimer in message_filters

flixr:
To make sure that the message filters also work if system time is set back, use SteadyTimer instead of the normal Timer

### DIFF
--- a/utilities/message_filters/include/message_filters/time_sequencer.h
+++ b/utilities/message_filters/include/message_filters/time_sequencer.h
@@ -84,7 +84,7 @@ public:
    * \param delay The minimum time to hold a message before passing it through.
    * \param update_rate The rate at which to check for messages which have passed "delay"
    * \param queue_size The number of messages to store
-   * \param nh (optional) The NodeHandle to use to create the ros::Timer that runs at update_rate
+   * \param nh (optional) The NodeHandle to use to create the ros::SteadyTimer that runs at update_rate
    */
   template<class F>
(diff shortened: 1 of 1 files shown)


utilities/message_filters/include/message_filters/time_sequencer.h
```diff
@@ -213,22 +213,22 @@ class TimeSequencer : public SimpleFilter<M>
...
-  void update(const ros::TimerEvent&)
+  void update(const ros::SteadyTimerEvent&)
   {
     dispatch();
   }
 
   void init()
   {
-    update_timer_ = nh_.createTimer(update_rate_, &TimeSequencer::update, this);
+    update_timer_ = nh_.createSteadyTimer(ros::WallDuration(update_rate_.toSec()), &TimeSequencer::update, this);
```
mikepurvis, line 223: Can `WallDuration` really not be initialized from a `Duration`? That seems like an API fail.

flixr: IIRC it can't... but I didn't check that again and did it explicitly just to be sure...

mikepurvis:
LGTM. It's a behaviour change, but I'd vote for this go into Lunar. It's extremely unlikely someone would be depending on the current broken behaviour.

That said, there's a potential ABI issue would be if a third party had message_filter class members, those will now change in type (though not in size, so maybe it wouldn't actually be a problem, since there's nothing to link?).

dirk-thomas:
Thank you for the patch.

PR #1249, opened 2017-11-30
fix xmlrpc timeout using monotonic clock

flixr:
in XmlRpcDispatch::work the _endTime check was using system time instead of monotonic time,
so when time was set back it didn't process all events..

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -223,7 +223,17 @@ XmlRpcDispatch::getTime()
 	  ((double) tbuff.timezone * 60));
 #else
   struct timespec ts;
+# ifdef __MACH__ // OS X does not have clock_gettime, use clock_get_time
+  clock_serv_t cclock;
+  mach_timespec_t mts;
+  host_get_clock_service(mach_host_self(), SYSTEM_CLOCK, &cclock);
+  clock_get_time(cclock, &mts);
+  mach_port_deallocate(mach_task_self(), cclock);
+  ts.tv_sec = mts.tv_sec;
+  ts.tv_nsec = mts.tv_nsec;
+#else
   clock_gettime(CLOCK_MONOTONIC, &ts);
+#endif
   return (ts.tv_sec + ts.tv_nsec / 1000000000.0);
 #endif /* USE_FTIME */
 }
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -3,6 +3,8 @@
 #include "xmlrpcpp/XmlRpcSource.h"
 #include "xmlrpcpp/XmlRpcUtil.h"
 
+#include "ros/time.h"
+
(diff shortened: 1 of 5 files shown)


mikepurvis:
LGTM, but this should be confirmed to work correctly OS X and FreeBSD.

flixr:
Good point.
According to https://www.unix.com/man-page/FreeBSD/2/clock_gettime/ it's available on FreeBSD, but apparently not on OSX: https://stackoverflow.com/questions/5167269/clock-gettime-alternative-in-mac-os-x

mikepurvis:
Reference implementation:

https://github.com/wjwwood/serial/blob/827c4a784dd4fdd35dc391f37ef152eab7c9c9b2/src/impl/unix.cc#L81-L97

dirk-thomas:
The same code already exists in [rostime](https://github.com/ros/roscpp_core/blob/64da30f22720ded414c4df37e9e0be41993c277e/rostime/src/time.cpp#L185). It should either be reused or at last be copied with a comment from where the code is coming and the same PP conditions should be used.

dirk-thomas:
Instead of copy-n-pasting a non-trivial code block like this wouldn't it be better to expose `ros_steadytime` in a header in `rostime` and call the function here?

flixr:
I thought it would be better to not add this dependency here, but I could do that if this is preferred...

flixr:
@dirk-thomas is it ok to leave it like this, or should we really add rostime as a dependency to xmlrpcpp?

flixr:
ping @dirk-thomas

dirk-thomas:
I think the risk is pretty high that when the implementation in `rostime` is updated / fixed / modified in the future this copied code isn't. Therefore I think it is worth to expose the function in `rostime` in a header and update this patch to use the new function (and declare a version dependency on `rostime` with the next patch release.

dirk-thomas:
I will have to release `roscpp_core` into Lunar in order for the PR job to pass... (ros/rosdistro#16898)

dirk-thomas:

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
```diff
@@ -222,11 +224,10 @@ XmlRpcDispatch::getTime()
   return ((double) tbuff.time + ((double)tbuff.millitm / 1000.0) +
 	  ((double) tbuff.timezone * 60));
 #else
-  struct timeval	tv;
-  struct timezone	tz;
+  uint32_t sec, nsec;
 
-  gettimeofday(&tv, &tz);
-  return (tv.tv_sec + tv.tv_usec / 1000000.0);
+  ros_steadytime(sec, nsec);
```
dirk-thomas: This might need the `ros::` namespace?

dirk-thomas:
I tried to fix this "on-the-fly" using the GitHub web interface which wasn't really successful. Can you please update the PR to make it pass CI. Thanks.

flixr:
Done.
Sorry about the missing things earlier... I didn't have machine handy to test it and wrongly thought it would be trivial enough to do without tests..

dirk-thomas:
No problem. Thank you for addressing it quickly.

PR #1250, opened 2017-11-30
monotonic clock for callback queue timeouts

flixr:
Fix callback queue timeouts in presence of system time jumps analogous to #1014


mikepurvis:
:+1:

dirk-thomas:
Thank you for the patch.

NickZ:
This introduced a serious regression. See #1343.

PR #1259, opened 2017-12-13
Replaced deprecated syntax (backticks with repr()).

ajeetdsouza:


ajeetdsouza:
Any idea why the build fails?

dirk-thomas:

mikepurvis:
Succeeded now. This is an exactly equivalent syntax except that Py3 doesn't support it— see bullet 2 in this list: https://docs.python.org/3.1/whatsnew/3.0.html#removed-syntax

dirk-thomas:
Thank you for the patch.

PR #1260, opened 2017-12-13
Removed unreachable exceptions.

ajeetdsouza:
`ROSInterruptException` is a subclass of `KeyboardInterrupt` and both exceptions are being handled the same way (`pass`). I removed the `ROSInterruptException` handling since it was placed after the `KeyboardInterrupt` and was thus unreachable.

### DIFF
--- a/tools/roslaunch/src/roslaunch/remoteprocess.py
+++ b/tools/roslaunch/src/roslaunch/remoteprocess.py
@@ -203,7 +203,7 @@ class SSHChildROSLaunchProcess(roslaunch.server.ChildROSLaunchProcess):
                 err_msg = "Unable to establish ssh connection to [%s%s:%s]: %s"%(username_str, address, port, e)
             except socket.error as e:
                 # #1824
-                if e[0] == 111:
+                if e.args[0] == 111:
                     err_msg = "network connection refused by [%s:%s]"%(address, port)
                 else:
                     err_msg = "network error connecting to [%s:%s]: %s"%(address, port, str(e))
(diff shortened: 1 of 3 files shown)


mgrrx:
LGTM

tools/roslaunch/src/roslaunch/remoteprocess.py
```diff
@@ -203,7 +203,7 @@ def _ssh_exec(self, command, address, port, username=None, password=None):
                 err_msg = "Unable to establish ssh connection to [%s%s:%s]: %s"%(username_str, address, port, e)
             except socket.error as e:
                 # #1824
-                if e[0] == 111:
+                if e.args[0] == 111:
```
dirk-thomas, line 206: How is this change related to the issue title?

dirk-thomas (commented):
Can you please clarify why these exceptions are unreachable.

Please also spend a minimum amount of time when filling and issue / pull request to provide the necessary context.

tools/roslaunch/src/roslaunch/remoteprocess.py
ajeetdsouza: I updated the description. Also, changing `e[0]` to `e.args[0]` wasn't really related - I will create separate PRs the next time. Given that I have deleted my fork, is there any way to fix this on my end? Alternatively, you could cherry pick the first commit and ignore the second.

dirk-thomas: Thank you for the additional information.

With the forked repo being delete I am not sure I can cherry-pick individual commits. I will just squash both commits during the merge in this case.

PR #1261, opened 2017-12-13
make xmlrpcpp specific include directory work in devel space

### LINKED ISSUE #1248: Source builds on Kinetic fail with XmlRpcValue.h inclusion problem
As reported in https://github.com/ros/ros_comm/pull/962#issuecomment-347181447:

While trying to build Kinetic `desktop-full` from sources on Ubuntu Xenial (amd64) I'm running into a header inclusion problem with `XmlRpcValue.h` in `diagnostic_aggregator`:

```
In file included from /catkin_ws/src/diagnostics/diagnostic_aggregator/src/analyzer_group.cpp:37:0:
/catkin_ws/src/diagnostics/diagnostic_aggregator/include/diagnostic_aggregator/analyzer_group.h:51:25: fatal error: XmlRpcValue.h: No such file or directory
compilation terminated.
make[2]: *** [CMakeFiles/diagnostic_aggregator.dir/src/analyzer_group.cpp.o] Error 1
```

This seems to be the same issue that is reported in https://github.com/ros/geometry/issues/144#issuecomment-335630911. It should probably be fixed in `ros/diagnostics`, but seeing as the change introduced in #962 was supposed to be bw-compatible, I figured I report it here as well.

Adding `include/xmlrpcpp` to the `catkin_package(..)` `INCLUDE_DIRS` in `ros_com/xmlrpcpp/CMakeLists.txt:12` does seem to make the build continue, but that should not be necessary.

Environment: Ubuntu Xenial (16.04), amd64. Building with `catkin_tools`.

dirk-thomas:
Fixed #1248. Needs to be backported to Kinetic.


mikepurvis:
LGTM 👍

PR #1268, opened 2017-12-15
Fixes #1262 - bad IPv6 initialization order.

### LINKED ISSUE #1262: ROS_IPV6 is checked too late in roscpp
ROS_IPV6 environment variable is checked late in initialization process (in `ros::start`):
```cpp
env_ipv6 = getenv("ROS_IPV6"); //...
bool use_ipv6 = (env_ipv6 && strcmp(env_ipv6,"on") == 0); //...
XmlRpc::XmlRpcSocket::s_use_ipv6_ = use_ipv6;
```

However, first connection to master can be initialized as early as in `ros::init` - failing the whole process and preventing nodes from starting. 

If I change default value of `XmlRpc::XmlRpcSocket::s_use_ipv6_` to `true` manually, these nodes work in IPv6 environment. The correct solution would be to move ROS_IPV6 check to the beginning of `ros::init`.

zielmicha:
ROS_IPV6 environment variable was checked after first connection could be established.
This commit moves this check into early initialization.

### DIFF
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -316,17 +316,6 @@ void start()
     }
   }
 
-  char* env_ipv6 = NULL;
-#ifdef _MSC_VER
-  _dupenv_s(&env_ipv6, NULL, "ROS_IPV6");
-#else
-  env_ipv6 = getenv("ROS_IPV6");
-#endif
-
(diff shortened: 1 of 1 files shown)


zielmicha:
Any chances to get this merged? Or should I improve something with this PR?

dirk-thomas:
Thanks for the patch. It looks as is.

PR #1269, opened 2017-12-17
make xmlrpcpp specific include directory work in devel space [Kinetic backport]

kentsommer:
Backport #1261 to Kinetic


dirk-thomas:
Thank you for the effort to create this patch.

All changes in Lunar will be considered for backports before the next patch release in Kinetic. They will be collected in a single PR for review. This makes it much easier to track what has been backports and what hasn't (and why not).

Therefore I will close this for now.

PR #1271, opened 2017-12-19
Add review changes for #1165

efernandez:
This PR addresses the review changes from @dirk-thomas @paulbovbel @flixr in #1165 

I had to create a new PR because I don't have write access to `pull-738` branch.

FYI @dkargin @mikepurvis

### DIFF
--- a/clients/roscpp/include/ros/service.h
+++ b/clients/roscpp/include/ros/service.h
@@ -58,17 +58,61 @@ namespace service
  * @param service_name The name of the service.
  * @param req The request message.
  * @param[out] res Storage for the response message.
+ * @param timeout The amount of time to wait for, in milliseconds.  If timeout is -1,
+ * waits until the call is complete
+ *
+ * @return true on success, false otherwise.
+ */
+template<class MReq, class MRes>
+bool call(const std::string& service_name, MReq& req, MRes& res, double timeout)
+{
+  namespace st = service_traits;
+  NodeHandle nh;
+  ServiceClientOptions ops(ros::names::resolve(service_name), st::md5sum(req), false, M_string());
+  ServiceClient client = nh.serviceClient(ops);
+  client.setTimeout(timeout);
+  return client.call(req, res);
+}
+
+/** @brief Invoke an RPC service.
+ *
+ * This method invokes an RPC service on a remote server, looking up the
+ * service location first via the master.
+ *
+ * @param service_name The name of the service.
+ * @param req The request message.
+ * @param[out] res Storage for the response message.
  *
  * @return true on success, false otherwise.
  */
 template<class MReq, class MRes>
 bool call(const std::string& service_name, MReq& req, MRes& res)
 {
+  return call(service_name, req, res, -1);
+}
+
+/** @brief Invoke an RPC service.
+ *
+ * This method invokes an RPC service on a remote server, looking up the
+ * service location first via the master.
+ *
+ * @param service_name The name of the service.
+ * @param service The service class that contains the request and response messages
+ * @param timeout The amount of time to wait for, in milliseconds.  If timeout is -1,
+ * waits until the call is complete
+ *
+ * @return true on success, false otherwise.
+ */
+template<class Service>
+bool call(const std::string& service_name, Service& service, double timeout)
+{
   namespace st = service_traits;
+
   NodeHandle nh;
-  ServiceClientOptions ops(ros::names::resolve(service_name), st::md5sum(req), false, M_string());
+  ServiceClientOptions ops(ros::names::resolve(service_name), st::md5sum(service), false, M_string());
   ServiceClient client = nh.serviceClient(ops);
-  return client.call(req, res);
+  client.setTimeout(timeout);
+  return client.call(service.request, service.response);
 }
 
 /** @brief Invoke an RPC service.
@@ -84,12 +128,7 @@ bool call(const std::string& service_name, MReq& req, MRes& res)
 template<class Service>
 bool call(const std::string& service_name, Service& service)
 {
-  namespace st = service_traits;
-
-  NodeHandle nh;
-  ServiceClientOptions ops(ros::names::resolve(service_name), st::md5sum(service), false, M_string());
-  ServiceClient client = nh.serviceClient(ops);
-  return client.call(service.request, service.response);
+  return call(service_name, service, -1);
 }
 
 /**
--- a/clients/roscpp/include/ros/service.h
+++ b/clients/roscpp/include/ros/service.h
@@ -58,7 +58,7 @@ namespace service
  * @param service_name The name of the service.
  * @param req The request message.
  * @param[out] res Storage for the response message.
- * @param timeout The amount of time to wait for.  If timeout is -1, waits until the call is complete
+ * @param timeout The amount of time to wait for.  If timeout is <= 0, waits until the call is complete
  *
  * @return true on success, false otherwise.
  */
@@ -97,7 +97,7 @@ bool call(const std::string& service_name, MReq& req, MRes& res)
  *
  * @param service_name The name of the service.
  * @param service The service class that contains the request and response messages
- * @param timeout The amount of time to wait for.  If timeout is -1, waits until the call is complete
+ * @param timeout The amount of time to wait for.  If timeout is <= 0, waits until the call is complete
  *
  * @return true on success, false otherwise.
  */
--- a/clients/roscpp/include/ros/service.h
+++ b/clients/roscpp/include/ros/service.h
@@ -58,13 +58,12 @@ namespace service
  * @param service_name The name of the service.
  * @param req The request message.
  * @param[out] res Storage for the response message.
- * @param timeout The amount of time to wait for, in milliseconds.  If timeout is -1,
- * waits until the call is complete
+ * @param timeout The amount of time to wait for.  If timeout is -1, waits until the call is complete
  *
  * @return true on success, false otherwise.
  */
 template<class MReq, class MRes>
-bool call(const std::string& service_name, MReq& req, MRes& res, double timeout)
+bool call(const std::string& service_name, MReq& req, MRes& res, const ros::WallDuration& timeout)
 {
   namespace st = service_traits;
   NodeHandle nh;
@@ -88,7 +87,7 @@ bool call(const std::string& service_name, MReq& req, MRes& res, double timeout)
 template<class MReq, class MRes>
 bool call(const std::string& service_name, MReq& req, MRes& res)
 {
-  return call(service_name, req, res, -1);
+  return call(service_name, req, res, ros::WallDuration(-1));
 }
 
 /** @brief Invoke an RPC service.
@@ -98,13 +97,12 @@ bool call(const std::string& service_name, MReq& req, MRes& res)
  *
  * @param service_name The name of the service.
  * @param service The service class that contains the request and response messages
- * @param timeout The amount of time to wait for, in milliseconds.  If timeout is -1,
- * waits until the call is complete
+ * @param timeout The amount of time to wait for.  If timeout is -1, waits until the call is complete
  *
  * @return true on success, false otherwise.
  */
 template<class Service>
-bool call(const std::string& service_name, Service& service, double timeout)
+bool call(const std::string& service_name, Service& service, const ros::WallDuration& timeout)
 {
   namespace st = service_traits;
 
@@ -128,7 +126,7 @@ bool call(const std::string& service_name, Service& service, double timeout)
 template<class Service>
 bool call(const std::string& service_name, Service& service)
 {
-  return call(service_name, service, -1);
+  return call(service_name, service, ros::WallDuration(-1));
 }
 
 /**
--- a/clients/roscpp/include/ros/service_client.h
+++ b/clients/roscpp/include/ros/service_client.h
@@ -165,7 +165,7 @@ public:
    * \param timeout The amount of time to wait for call response. If timeout is -1 (default),
    * waits until call is complete
    */
-  void setTimeout(double timeout);
+  void setTimeout(const ros::WallDuration& timeout);
 
   operator void*() const { return isValid() ? (void*)1 : (void*)0; }
   bool operator<(const ServiceClient& rhs) const
@@ -206,7 +206,7 @@ private:
     M_string header_values_;
     std::string service_md5sum_;
     bool is_shutdown_;
-    double timeout;		///< Timeout for calling service
+    ros::WallDuration timeout;  ///< Timeout for calling service
   };
   typedef boost::shared_ptr<Impl> ImplPtr;
   typedef boost::weak_ptr<Impl> ImplWPtr;
--- a/clients/roscpp/include/ros/service_client.h
+++ b/clients/roscpp/include/ros/service_client.h
@@ -160,6 +160,13 @@ public:
    */
   std::string getService();
 
+  /**
+   * \brief Set timeout for service calls
+   * \param timeout The amount of time to wait for call response. If timeout is -1 (default),
+   * waits until call is complete
+   */
+  void setTimeout(double timeout);
+
   operator void*() const { return isValid() ? (void*)1 : (void*)0; }
   bool operator<(const ServiceClient& rhs) const
   {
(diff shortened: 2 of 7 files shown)


clients/roscpp/include/ros/service_server_link.h
```diff
@@ -106,6 +106,7 @@ class ROSCPP_DECL ServiceServerLink : public boost::enable_shared_from_this<Serv
    * it has finished.
    */
   bool call(const SerializedMessage& req, SerializedMessage& resp);
+  bool call(const SerializedMessage& req, SerializedMessage& resp, double timeout);
```
efernandez: @paulbovbel I haven't changed `double timeout` to `ros::Duration` or `ros::WallDuration` because internally it's used with `boost::chrono::steady_clock` (as suggested by @flixr , instead of `boost::chrono::system_clock`).

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,14 +368,35 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
+        info->finished_condition_.wait(lock);
     }
   }
+  // Clear calls is properly protected
+  if (interrupted)
+  {
+    this->clearCalls();
+  }
 
   info->call_finished_ = true;
```
efernandez, line 398: @dirk-thomas Yes, this is also set to `true` when it times out.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,14 +368,35 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
     while (!info->finished_)
     {
-      info->finished_condition_.wait(lock);
+      if (timeout > 0)
+      {
+        steady_clock::time_point wall_time = now + milliseconds(static_cast<int>(timeout * 1000));
+
+        if (info->finished_condition_.wait_until(lock, wall_time) == boost::cv_status::timeout)
+        {
+          ROSCPP_LOG_DEBUG("Service [%s] call failed: no response for %fsec", service_name_.c_str(), timeout);
```
efernandez: @dirk-thomas I reduce this from `ROS_ERROR` to `ROSCPP_LOG_DEBUG`

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,14 +368,35 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
     processNextCall();
   }
 
+  bool interrupted = false;
   {
     boost::mutex::scoped_lock lock(info->finished_mutex_);
 
+    using namespace boost::chrono;
+    steady_clock::time_point now = steady_clock::now();
```
efernandez: @flixr Using `steady_clock` instead of `system_clock` now

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -327,6 +328,11 @@ void ServiceServerLink::processNextCall()
 }
 
 bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
+{
```
efernandez, line 331: @dirk-thomas Moved before, to respect the header file ordering.

efernandez, line 331: @dirk-thomas Move before, to respect the header file ordering.

efernandez (commented):
I've added some comments to highlight the review changes I did.

Other than that, I did style changes to use `if (` instead of `if(` and use `{ }` even for single line `if`/`else` blocks.

clients/roscpp/include/ros/service_client.h
```diff
@@ -160,6 +160,13 @@ class ROSCPP_DECL ServiceClient
    */
   std::string getService();
 
+  /**
+   * \brief Set timeout for service calls
+   * \param timeout The amount of time to wait for call response. If timeout is -1 (default),
```
flixr: should probably mention that the unit is seconds

clients/roscpp/include/ros/service_client.h
```diff
@@ -199,6 +206,7 @@ class ROSCPP_DECL ServiceClient
     M_string header_values_;
     std::string service_md5sum_;
     bool is_shutdown_;
+    double timeout;		///< Timeout for calling service
```
flixr: should probably mention that unit is seconds

clients/roscpp/include/ros/service_server_link.h
flixr: It could still be a `ros::WallDuration` as you can just call `toSec()` or `toNSec` on it.
Personally I don't have a strong opinion on whether to use a simple double or WallDuration here.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,14 +368,37 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
+    steady_clock::time_point now = steady_clock::now();
+
     while (!info->finished_)
     {
-      info->finished_condition_.wait(lock);
+      if (timeout > 0)
+      {
+        steady_clock::time_point wall_time = now + milliseconds(static_cast<int>(timeout * 1000));
+
+        if (info->finished_condition_.wait_until(lock, wall_time) == boost::cv_status::timeout)
```
flixr: To make this actually work as expected you will need to include the backported boost headers and define `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` in the header file (service_server_link.h) where the condition variable is declared. You can look at #1250 for an example.
Also it would be easier to use `wait_for` in this case...

clients/roscpp/src/libros/service_server_link.cpp
flixr: IMHO this should be a warning...

flixr (changes requested):
For this to really use the steady_clock it needs some change analogous to #1250

clients/roscpp/include/ros/service.h
```diff
@@ -58,40 +58,79 @@ namespace service
  * @param service_name The name of the service.
  * @param req The request message.
  * @param[out] res Storage for the response message.
+ * @param timeout The amount of time to wait for, in milliseconds.  If timeout is -1,
```
efernandez: Just added `ros::service::call` top-level versions with `timeout`.

I copied the doxygen documentation from other functions in the file. That's why it says `milliseconds`. I'm going to fix that together with the changes suggested by @flixr 

I've also added a unit test, which uses this new function (as the other unit test in the same test suite).

clients/roscpp/src/libros/service_server_link.cpp
efernandez: Done.

Thanks for the detailed comments. I wasn't aware of that issue with boost < 1.61.

Re. `wait_until` vs. `wait_for`, I think it was already changed in the original PR, but I think you're right that with `wait_for` it'd be simpler. I'd try it and update this PR.

clients/roscpp/include/ros/service.h
efernandez: Done

clients/roscpp/include/ros/service_client.h
efernandez: Changed it to `ros::WallDuration`, so no need for that anymore.

clients/roscpp/include/ros/service_server_link.h
efernandez: Changed to `ros::WallDuration`.

clients/roscpp/src/libros/service_server_link.cpp
efernandez: Changed to `ROS_WARN`.

@dirk-thomas Do you agree with this?

clients/roscpp/include/ros/service_client.h
efernandez: Changed it to `ros::WallDuration`, so no need for that anymore.

efernandez:
@flixr I think I've addressed all your comments.

I'm also using `wait_for` and `boost::chrono::nanoseconds` in the main logic now.

clients/roscpp/include/ros/service.h
```diff
@@ -58,40 +58,77 @@ namespace service
  * @param service_name The name of the service.
  * @param req The request message.
  * @param[out] res Storage for the response message.
+ * @param timeout The amount of time to wait for.  If timeout is -1, waits until the call is complete
```
flixr: To be precise, it should say "If timeout is <= 0, waits until the call is complete"
Same for other occurrences...

clients/roscpp/include/ros/service_server_link.h
```diff
@@ -37,6 +37,21 @@
 
 #include "ros/common.h"
 
+// check if we might need to include our own backported version boost::condition_variable
```
flixr: This should be added at the very beginning, before the include of "ros/common.h" to make sure that boost thread/condition_variable is not accidentally included before...

flixr (changes requested):
Hm... tests fail...
Include order should be fixed in any case...

clients/roscpp/include/ros/service.h
efernandez: Done

clients/roscpp/include/ros/service_server_link.h
efernandez: Done.

The test is passing locally to me all time. Hopefully it was just this in the buildfarm.
Or maybe it's something related with the `wait_for` + `nanoseconds` changed ?!

There are also the md5sum mismatch errors before (http://build.ros.org/job/Lpr__ros_comm__ubuntu_xenial_amd64/679/consoleFull), but I also have them locally, so it's probably part of some test ?!

UPDATE Yes, the md5sum mismatches are part of the `callSrvWithWrongType` test.

``` bash
~/dev/ws/cpr_ws/build/test_roscpp$ make run_tests_test_roscpp_rostest_test_launch_service_call.xml
-- run_tests.py: execute commands
  /home/eperdomo/dev/ws/cpr_ws/src/ros_comm/tools/rostest/scripts/rostest --pkgdir=/home/eperdomo/dev/ws/cpr_ws/src/ros_comm/test/test_roscpp --package=test_roscpp --results-filename test_launch_service_call.xml --results-base-dir /home/eperdomo/dev/ws/cpr_ws/build/test_roscpp/test_results /home/eperdomo/dev/ws/cpr_ws/src/ros_comm/test/test_roscpp/test/launch/service_call.xml 
... logging to /home/eperdomo/.ros/log/rostest-nb-eperdomo-19190.log
[ROSUNIT] Outputting test results to /home/eperdomo/dev/ws/cpr_ws/build/test_roscpp/test_results/test_roscpp/rostest-test_launch_service_call.xml
[ERROR] [/service_adv] [1513807564.761673776]: client wants service /service_adv to have md5sum 6b193467bc2d41be616d6b0e985e2a63, but it has 671f8e4998eaec79f1c47e339dfd527b. Dropping connection.
[ERROR] [/service_adv] [1513807564.762505964]: client wants service /service_adv to have md5sum 6b193467bc2d41be616d6b0e985e2a63, but it has 671f8e4998eaec79f1c47e339dfd527b. Dropping connection.
[ERROR] [/service_adv] [1513807564.763258836]: client wants service /service_adv to have md5sum 6b193467bc2d41be616d6b0e985e2a63, but it has 671f8e4998eaec79f1c47e339dfd527b. Dropping connection.
[ERROR] [/service_adv] [1513807564.764014530]: client wants service /service_adv to have md5sum 6b193467bc2d41be616d6b0e985e2a63, but it has 671f8e4998eaec79f1c47e339dfd527b. Dropping connection.
[Testcase: testservice_call] ... ok

[ROSTEST]-----------------------------------------------------------------------

[test_roscpp.rosunit-service_call/callSrv][passed]
[test_roscpp.rosunit-service_call/callSrvUnicode][passed]
[test_roscpp.rosunit-service_call/callSrvMultipleTimes][passed]
[test_roscpp.rosunit-service_call/callSrvWithWrongType][passed]
[test_roscpp.rosunit-service_call/callSrvHandle][passed]
[test_roscpp.rosunit-service_call/callSrvPersistentHandle][passed]
[test_roscpp.rosunit-service_call/callSrvLongRunning][passed]
[test_roscpp.rosunit-service_call/callSrvLongRunningTimeout][passed]
[test_roscpp.rosunit-service_call/callSrvWhichUnadvertisesInCallback][passed]
[test_roscpp.rosunit-service_call/handleValid][passed]
[test_roscpp.rosunit-service_call/waitForServiceTimeout][passed]

SUMMARY
 * RESULT: SUCCESS
 * TESTS: 11
 * ERRORS: 0
 * FAILURES: 0

rostest log file is in /home/eperdomo/.ros/log/rostest-nb-eperdomo-19190.log
-- run_tests.py: verify result "/home/eperdomo/dev/ws/cpr_ws/build/test_roscpp/test_results/test_roscpp/rostest-test_launch_service_call.xml"
Built target run_tests_test_roscpp_rostest_test_launch_service_call.xml
```

efernandez:
I can't reproduce the buildfarm failures in the unit test locally. @flixr Could you give it a try locally as well?

Any ideas what's wrong?
The tests are run with `make -j1` so no test is running in parallel, and even if there was a previous called in the service queue, it shouldn't return before `2` seconds. Assuming the `ros::Duration(2.0).sleep()` takes `2` wall (actually steady) time seconds.

flixr:
Test fails locally here... not really sure what's going on...

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,14 +368,32 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
 
     while (!info->finished_)
     {
-      info->finished_condition_.wait(lock);
+      if (timeout > ros::WallDuration(0.0))
+      {
+        if (info->finished_condition_.wait_for(lock, boost::chrono::nanoseconds(timeout.toNSec())) == boost::cv_status::timeout)
+        {
+          ROS_WARN("Service [%s] call failed: no response for %fsec", service_name_.c_str(), timeout.toSec());
+          interrupted = true;
```
flixr, line 382: The `info->success_` that is returned doesn't seem to be set to `false`, or am I missing something here?

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,14 +368,32 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
+      else
+      {
+        info->finished_condition_.wait(lock);
+      }
     }
   }
+  // Clear calls is properly protected
+  if (interrupted)
+  {
+    this->clearCalls();
```
flixr, line 395: It seems this clears/cancels all calls instead of only the current one...?

flixr:
I don't understand how this is supposed to work as I can't see how the actual service request (connection) is aborted/canceled after this timeout.
Am I missing something or does it really not cancel the service request?

Also how did you test this? Invocation of these tests is quite cumbersome (or I just don't get how to run them properly)...
I also don't see any warnings like "Service [xyz] call failed: no response for x sec" that I actually expected.

clients/roscpp/src/libros/service_server_link.cpp
efernandez: Good catch

clients/roscpp/src/libros/service_server_link.cpp
efernandez: If I'm not wrong, it's initialized to `false` here: https://github.com/efernandez/ros_comm/blob/14b168bfbec4c1a2fd74302836dc2247fa9317a1/clients/roscpp/src/libros/service_server_link.cpp#L340

And set to `true/false` if the call is attended here: https://github.com/efernandez/ros_comm/blob/14b168bfbec4c1a2fd74302836dc2247fa9317a1/clients/roscpp/src/libros/service_server_link.cpp#L213

efernandez:
Closing this PR in favour of #1282

> Am I missing something or does it really not cancel the service request?
I guess you already solved that in #1282 by using `wait_until` instead of `wait_for` b/c it's inside a loop.

> Also how did you test this?
I'm building with `catkin build` and then inside the `build/<pkg>` directory I `make <target>` where `<target>` is the test that includes the new tests introduced with this PR.
Re. the warning message, my guess is that you need to use `output="screen"` in the launch/test file used in the rostest. I don't know if you can pass `--screen` somehow, from the `CMakeLists.txt`

PR #1276, opened 2017-12-21
Replaced Thread.setDaemon() using new API.

ajeetdsouza:
https://docs.python.org/2/library/threading.html#threading.Thread.daemon


mikepurvis:
Thanks for the contribution! If you could, please also update the tests which use it:

https://github.com/ros/ros_comm/search?q=setDaemon

ajeetdsouza:
Done.

mikepurvis:
Looks great, thanks!

This is supported/recommended as of Python 2.6, and our target is Python 2.7.

Issue #1254, opened 2017-12-12
Possible documentation error in rosnode


ajeetdsouza:
rosnode documentation seems to have outdated docstrings, making references to XMLRPC handles in places where we now use instances of `rosgraph.Master` (a convenience API for interaction with the ROS Master).

https://github.com/ros/ros_comm/blob/lunar-devel/tools/rosnode/src/rosnode/__init__.py

dirk-thomas:
Addressed by #1278.

PR #1278, opened 2017-12-23
Fixed incorrect docstrings in rosnode/__init__.py

ajeetdsouza:
Relevant issue: #1254 
I closed the previous pull request, since it had conflicts.

### DIFF
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -84,7 +84,7 @@ def _succeed(args):
 _caller_apis = {}
 def get_api_uri(master, caller_id, skip_cache=False):
     """
-    @param master: XMLRPC handle to ROS Master
+    @param master: rosgraph Master instance
     @type  master: rosgraph.Master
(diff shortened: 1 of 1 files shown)


tools/rosbag/src/rosbag/bag.py
```diff
@@ -778,7 +778,7 @@ def __str__(self):
                     else:
                         rows.append(('', s))
         
-        except Exception as ex:
+        except:
```
dirk-thomas: Please do not remove the `Exception` type. This changes behavior and catches other exception not subclasses from `Exception` like `KeyboardInterrupt`.

Same below.

dirk-thomas (commented):
Please do not mix unrelated changes in a single PR. The title of this PR is about fixing docstring.

Please create separate patches for removing the unnecessary variables.

ajeetdsouza:
Sorry! I've reverted the unrelated commits, could you check?

dirk-thomas:
Thank you for the patch.

PR #1281, opened 2017-12-29

guillaumeautran:
Address a long standing issue where the `sockets_changed_` flag would never be reset causing the native poll set to be re-created everytime doing a poll.


guillaumeautran:
FYI: @mikepurvis @dirk-thomas

mikepurvis:
LGTM

guillaumeautran:
Ping for review

dirk-thomas:
Can you please describe how you tested this patch.

guillaumeautran:
We've been running the patch on our robots internally since 2018-01-19. 
This represent ~260km driven autonomously and a total runtime of ~768hours (cumulative on all of our robots).

dirk-thomas:
Thanks for the info. Since @mikepurvis already gave his approval lets merge this.

Thank you for the patch!

PR #1282, opened 2017-12-29
service timeouts: addition to #1271

flixr:
just for review/testing:
This is the same as #1271 from @efernandez with one additional commit that *should* (tm) fix things...

@efernandez feel free to update your PR with that commit

### DIFF
--- a/clients/roscpp/src/libros/service_server_link.cpp
+++ b/clients/roscpp/src/libros/service_server_link.cpp
@@ -329,10 +329,10 @@ void ServiceServerLink::processNextCall()
 
 bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
 {
-  return call(req, resp, -1);
+  return call(req, resp, ros::WallDuration(-1));
 }
 
-bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp, double timeout)
+bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp, const ros::WallDuration& timeout)
 {
   CallInfoPtr info(boost::make_shared<CallInfo>());
   info->req_ = req;
@@ -372,18 +372,13 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
   {
     boost::mutex::scoped_lock lock(info->finished_mutex_);
 
-    using namespace boost::chrono;
-    steady_clock::time_point now = steady_clock::now();
-
     while (!info->finished_)
     {
-      if (timeout > 0)
+      if (timeout > ros::WallDuration(0.0))
       {
-        steady_clock::time_point wall_time = now + milliseconds(static_cast<int>(timeout * 1000));
-
-        if (info->finished_condition_.wait_until(lock, wall_time) == boost::cv_status::timeout)
+        if (info->finished_condition_.wait_for(lock, boost::chrono::nanoseconds(timeout.toNSec())) == boost::cv_status::timeout)
         {
-          ROSCPP_LOG_DEBUG("Service [%s] call failed: no response for %fsec", service_name_.c_str(), timeout);
+          ROSCPP_LOG_DEBUG("Service [%s] call failed: no response for %fsec", service_name_.c_str(), timeout.toSec());
           interrupted = true;
           break;
         }
--- a/clients/roscpp/src/libros/service_server_link.cpp
+++ b/clients/roscpp/src/libros/service_server_link.cpp
@@ -327,6 +327,11 @@ void ServiceServerLink::processNextCall()
   }
 }
 
+bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
+{
+  return call(req, resp, -1);
+}
+
 bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp, double timeout)
 {
   CallInfoPtr info(boost::make_shared<CallInfo>());
@@ -400,11 +405,6 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
   }
 
   return info->success_;
-}
-
-bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
-{
-  return call(req, resp, -1);
 }
 
 bool ServiceServerLink::isValid() const
--- a/clients/roscpp/src/libros/service_server_link.cpp
+++ b/clients/roscpp/src/libros/service_server_link.cpp
@@ -327,11 +327,6 @@ void ServiceServerLink::processNextCall()
   }
 }
 
-bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
-{
-  return call(req, resp, -1);
-}
-
 bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp, double timeout)
 {
   CallInfoPtr info(boost::make_shared<CallInfo>());
@@ -405,6 +400,11 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
   }
 
   return info->success_;
+}
+
+bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& resp)
+{
+  return call(req, resp, -1);
 }
 
 bool ServiceServerLink::isValid() const
--- a/clients/roscpp/src/libros/service_server_link.cpp
+++ b/clients/roscpp/src/libros/service_server_link.cpp
@@ -369,23 +369,24 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
 
     while (!info->finished_)
     {
-    	if(timeout > 0)
-    	{
-    		boost::chrono::milliseconds duration(static_cast<int>(timeout * 1000));
-    		if(info->finished_condition_.wait_for(lock, duration))
-    		{
-    			ROS_ERROR("Service [%s] call failed: no response for %fsec", service_name_.c_str(), timeout);
-    			interrupted = true;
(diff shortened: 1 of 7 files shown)


flixr:
Ahrg, there is still a temporary deadlock (via `finished_mutex_`) and the calls seem to always take at least as long as the timeout...

efernandez:
Good catch. Indeed, the original implementation was using `wait_until`, but it was changed to `wait_for` are requested in some the of the review comments.

Your code LGTM. Could you elaborate a little more on how that deadlock via `finished_mutex_` happens, so I can try to help?

I'd also like to know how you build and run the test, so you can reproduce the failure locally. I can't reproduce it locally, as it always passes for me.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,15 +368,37 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
+        }
+      }
+      else
+      {
+        info->finished_condition_.wait(lock);
+      }
     }
   }
 
+  // Clear calls is properly protected
```
efernandez: I can't really see any deadlock on `info->finished_mutex_`.

However, it seems that on timeout we should do something with the `connection_`, so the read callback is cancelled somehow. Thoughts?

Another minor comment:
Why instead of this `if` block with `cancelCall`, don't we set `info->finished_ = true;` before the `break;` in the loop?
There we own the lock on `info->finished_mutex_`.

efernandez (approved):

test/test_roscpp/test/src/service_call.cpp
```diff
@@ -178,6 +178,45 @@ TEST(SrvCall, callSrvLongRunning)
...
+  ASSERT_FALSE(success)
+      << service_name << " did NOT fail after timeout (res = " << res.str << ")";
+  ASSERT_TRUE(d.toSec() < service_time_seconds)
+      << service_name << " did NOT time out after " << short_timeout.toSec() << "s, took " << d.toSec() << "s instead.";
+
+  start = ros::SteadyTime::now();
+  success = ros::service::call(service_name, req, res, long_timeout);
+  d = ros::SteadyTime::now() - start;
+  ASSERT_TRUE(success)
+      << service_name << " finished successfully after " << d.toSec() << "s (timeout " <<long_timeout.toSec() << "s)";
```
efernandez: Add space after `<<` in `<<long_timeout.toSec()`

clients/roscpp/src/libros/service_server_link.cpp
flixr: > I can't really see any deadlock on info->finished_mutex_.

Sorry, the suspected deadlock was a wrong guess (because the test failed, but that was due to the "server" still processing the previous request, fixed in new tests....)

> Why instead of this if block with cancelCall, don't we set info->finished_ = true; before the break; in the loop?
> There we own the lock on info->finished_mutex_.

Tried that and turned out that is not working reliably (actually same for the cancelCall).

flixr:
> Good catch. Indeed, the original implementation was using wait_until, but it was changed to wait_for are requested in some the of the review comments.

Yeah, that was my suggestion, but I didn't see that it's called in a loop before...

> However, it seems that on timeout we should do something with the `connection_`, so the read callback is cancelled somehow. Thoughts?

Yes, that is what I meant in my previous comments about aborting/cancelling the request in your PR... Not really sure how to best do this, but IMHO this need to be done properly even though the tests pass now.

> I'd also like to know how you build and run the test, so you can reproduce the failure locally. I can't reproduce it locally, as it always passes for me.

I was basically doing it in the same way as you, and then running via `rostest -t ros_comm/test/test_roscpp/test/launch/service_call.xml`.
It also took me a while to notice that the test was (is?) not deterministic and there was (is?) a race condition and it didn't fail every time.

test/test_roscpp/test/src/service_call.cpp
```diff
@@ -201,18 +201,24 @@ TEST(SrvCall, callSrvLongRunningTimeout)
   start = ros::SteadyTime::now();
   success = ros::service::call(service_name, req, res, short_timeout);
   d = ros::SteadyTime::now() - start;
+  printf("call with timeout of %f s took %f s\n", short_timeout.toSec(), d.toSec());
   ASSERT_FALSE(success)
       << service_name << " did NOT fail after timeout (res = " << res.str << ")";
   ASSERT_TRUE(d.toSec() < service_time_seconds)
```
efernandez: Use `ASSERT_LT(d.toSec(), service_time_seconds)` here and something similar below.
That way you can simplify the custom message on error, since gtest will print the expected and actual value for us.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,15 +368,38 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
     processNextCall();
   }
 
+  bool timed_out = false;
   {
     boost::mutex::scoped_lock lock(info->finished_mutex_);
 
+    using namespace boost::chrono;
+    steady_clock::time_point end_time = steady_clock::now() + nanoseconds(timeout.toNSec());
```
dirk-thomas: This variable is only needed when `timeout > 0`. Please avoid calling `steady_clock::now()` if the result isn't used.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,15 +368,38 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
+
     while (!info->finished_)
     {
-      info->finished_condition_.wait(lock);
+      if (timeout > ros::WallDuration(0.0))
+      {
+        if (info->finished_condition_.wait_until(lock, end_time) == boost::cv_status::timeout)
+        {
+          ROS_WARN("Service [%s] call failed: no response for %fsec", service_name_.c_str(), timeout.toSec());
+          timed_out = true;
```
dirk-thomas, line 385: I would suggest to remove the new variable `timed_out` and do the call of `info->finished_condition_.wait(lock);` here instead.

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -362,15 +368,38 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
+      else
+      {
+        info->finished_condition_.wait(lock);
+      }
     }
   }
 
+  if (timed_out)
+  {
+    // TODO: actually properly cancel the request (connection read)
```
dirk-thomas: What is the status of this TODO?

clients/roscpp/src/libros/service_server_link.cpp
flixr: sorry, I'm not quite following...

clients/roscpp/src/libros/service_server_link.cpp
flixr: I'm actually unsure about the internal workings here, but as far as I can see `callFinished` and the likes only remove the callback but don't really "cancel" the actual request.
Not even sure if that is possible.
Notifying the service call provider is probably out of the scope anyway.

flixr:
@efernandez have you ever tested this with persistent service connections?

Some comments/questions:
- In case of a timeout, only the callback for the response is removed on the client side and the request is not really "canceled".
- On the server/provider side the service call still takes the full time. If its single threaded, any subsequent service call will have to wait until the old canceled one is done before the current service call can be processed. This could be unexpected if you call a service repeatedly from the same node/code. 
- Notifying the service call provider is probably out of the scope anyway?

efernandez:
@flixr In the particular code where we had issues with a call not returning ever, it wasn't a persistent service.

I agree with your comments/questions, and I think they're indeed a problem.

For my particular use case, where a node got blocked forever, I don't think it'd matter though, as I know the service should be attended very fast, so I can have a really huge timeout to handle the weird case when it never returns, for whatever reason that is (likely a low level issue where the socket got closed or the response was lost).

For a general case, someone would expect everything to be destroyed/stopped after the timeout, so subsequent calls are attended immediately. Unfortunately this seems difficult to implement.

For me it'd be great to get this initial contribution in, but I'd understand opinions against that as we still don't have the full story clear. Maybe it could be added as an experimental API, just mentioning that in the doxygen and enumarating the missing part and what we think would happen in some scenarios.

mikepurvis:
@flixr, @efernandez Since this is an ABI-breaking change, is it something you're wanting to get in in time for Melodic? What's the status of it?

flixr:
While I would really welcome this functionality, to me this is still unfinished or at least not tested enough yet.
To me it's actually unclear if just "dropping the callback on the client side" (as it's done now) is sufficient/desired, keeping in mind that the service is not actually aborted on the service server/provider side....

flixr:
At the very least it still has to be tested with persistent service connections....

efernandez:
IMHO I think more work needs to happen here. I don't think "dropping the callback on the client side" only will be enough.

Unfortunately, I don't have time to look into it, and it doesn't look trivial to me.

krisklau:
Hi! What is the status of this?

flixr:
No progress.
And IMHO also not easily do-able as it would require a new API to notify the called service that the client ran into a timeout and doesn't require the result anymore.

UnexpectedLobster:
Is this issue addressed in ROS2 ? Or is the only way to address this to only use actions ?

dirk-thomas:

JanezCim:
> Closing due to long time of inactivity and targeting an EOL branch. Please consider opening a new PR targeting the current default branch if you are still interested in getting this patch merged.

So, ironically, this patch timed out before my service call could...

Issue #1283, opened 2017-12-29
xmlrpc_manager tries to reuse closed connections


mgrrx:
I found time again to revisit the CLOSE_WAIT. Although most CLOSE_WAIT problems are already fixed, roscpp nodes still have always one port in CLOSE_WAIT. This can be easily reproduced by just starting a roscore. The rosout process then has one port in CLOSE_WAIT.

As already addressed in #371, the xmlrpc server used in rosmaster does not use HTTP/1.1. Activating it still leads to the same behavior as reported back then.
Surprisingly, the xmlrpc_manager in roscpp tries to reuse the connections even if the server replies with HTTP/1.0 and closes the connection. Adding the keep-alive flag to the request doesn't help - the server will always close the connection.

I see three solutions to this problem:
1. Find out why activating HTTP/1.1 in the xmlrpc servers leads to the behavior described in #371 and fix this issue. 
2. Detect that the server responds with HTTP/1.0 and close the connection accordingly.
3. Do not cache any xmlrpc connection in roscpp.

I have a working solution for solution 2 which I'll upload in a bit.

mgrrx:
1. Is implemented in #1287
2. Is implemented in #1284
3. Is then obsolete

flixr:
I guess this can be closed as with #1287 merged this doesn't happen anymore in melodic.

PR #1284, opened 2017-12-29
[kinetic] Close sockets when server responds with HTTP/1.0

mgrrx:
This PR tries to detect that the server responds with HTTP/1.0 and closes the connection accordingly in the manager.

I moved the clearing part of header_ to the end of parseResponse(). Putting setKeepOpen(false) into readHeader() didn't work for me as the internal state machine is affected by it. The header_ member is public but I couldn't find any usage of it, so I would assume that this doesn't affect anybody.

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -107,6 +107,7 @@ XmlRpcClient::execute(const char* method, XmlRpcValue const& params, XmlRpcValue
     return false;
 
   XmlRpcUtil::log(1, "XmlRpcClient::execute: method %s completed.", method);
+  _header = "";
   _response = "";
   return true;
 }
@@ -403,7 +404,6 @@ XmlRpcClient::readHeader()
 
   // Otherwise copy non-header data to response buffer and set state to read response.
   _response = bp;
-  _header = "";   // should parse out any interesting bits from the header (connection, etc)...
   _connectionState = READ_RESPONSE;
   return true;    // Continue monitoring this source
 }
@@ -447,6 +447,7 @@ XmlRpcClient::parseResponse(XmlRpcValue& result)
   int offset = 0;
   if ( ! XmlRpcUtil::findTag(METHODRESPONSE_TAG,_response,&offset)) {
     XmlRpcUtil::error("Error in XmlRpcClient::parseResponse: Invalid response - no methodResponse. Response:\n%s", _response.c_str());
+    _header = "";
     return false;
   }
 
@@ -458,15 +459,31 @@ XmlRpcClient::parseResponse(XmlRpcValue& result)
     if ( ! result.fromXml(_response, &offset)) {
       XmlRpcUtil::error("Error in XmlRpcClient::parseResponse: Invalid response value. Response:\n%s", _response.c_str());
       _response = "";
+      _header = "";
       return false;
     }
   } else {
     XmlRpcUtil::error("Error in XmlRpcClient::parseResponse: Invalid response - no param or fault tag. Response:\n%s", _response.c_str());
     _response = "";
+    _header = "";
     return false;
   }
       
   _response = "";
+
+  if (_header.size() < 8 || _header.find("HTTP/") != 0)
+    XmlRpcUtil::error("Error XmlRpcClient::parseResponse: Header does not start with protocol");
+  else
+  {
+    const unsigned protocolVersion = atoi(&_header[5]) * 10 + atoi(&_header[7]);
+    if (protocolVersion == 10)
+    {
+      setKeepOpen(false);
+      close();
+    }
+  }
+
+  _header = "";
   return result.valid();
 }
 
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -471,16 +471,10 @@ XmlRpcClient::parseResponse(XmlRpcValue& result)
       
   _response = "";
 
-  if (_header.size() < 8 || _header.find("HTTP/") != 0)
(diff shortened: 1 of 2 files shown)


utilities/xmlrpcpp/src/XmlRpcClient.cpp
```diff
@@ -506,15 +507,31 @@ XmlRpcClient::parseResponse(XmlRpcValue& result)
...
     XmlRpcUtil::error("Error in XmlRpcClient::parseResponse: Invalid response - no param or fault tag. Response:\n%s", _response.c_str());
     _response = "";
+    _header = "";
     return false;
   }
       
   _response = "";
+
+  if (_header.size() < 8 || _header.find("HTTP/") != 0)
+    XmlRpcUtil::error("Error XmlRpcClient::parseResponse: Header does not start with protocol");
```
dirk-thomas: In case a response is received which doesn't contain the header this patch will add a new error. Is this really necessary?

I am also not sure about if it is reasonable to change the current behavior for HTTP 1.0 connections to not-keep-them-open.

Also how does this related to #1287?

dirk-thomas:
If #1287 gets merged should this PR be closed or still considered to be merged?

mgrrx:
Sorry for coming back to this after so many months. Let me summarize why I think this PR still makes sense, especially for older releases:
Without #1287 (which is the case for all releases < Melodic) the rocore always responds with HTTP/1.0 and closes the connections. However, a C++ client would try to keep the connection open which in the end leads to a socket in CLOSE_WAIT as explained in #1283.

utilities/xmlrpcpp/src/XmlRpcClient.cpp
mgrrx: It's not really necessary, I can remove it.

dirk-thomas:
@mgrrx Please consider retargeting a different branch. Lunar is already EOL. Either target kinetic-devel (if it should only be applied to Kinetic) or melodic-devel.

mgrrx:
Updated as requested

utilities/xmlrpcpp/src/XmlRpcClient.cpp
dirk-thomas: @mgrrx This comment seems to be still pending.

mgrrx: sure will do ASAP

mgrrx: I simplified the state ment. I assume it does the job without introducing the new error message.

utilities/xmlrpcpp/src/XmlRpcClient.cpp
```diff
@@ -458,15 +459,25 @@ XmlRpcClient::parseResponse(XmlRpcValue& result)
...
     XmlRpcUtil::error("Error in XmlRpcClient::parseResponse: Invalid response - no param or fault tag. Response:\n%s", _response.c_str());
     _response = "";
+    _header = "";
     return false;
   }
       
   _response = "";
+
+  // Close connection if protocol is HTTP/1.0
+  if (_header.size() > 8 && _header[5] == '1' && _header[7] == '0') {
```
dirk-thomas: Why not continue to check for the full string? Otherwise this would match e.g. `OTHER120`.

Also why `> 8`? Isn't the header size exactly 8 for `HTTP/1.0`?

utilities/xmlrpcpp/src/XmlRpcClient.cpp
dirk-thomas: > I assume it does the job without introducing the new error message.

Please make sure to test the latest change since this is a significant change for a distro which is already stable for 4 years (and the other comment suggests it doesn't work correctly atm).

utilities/xmlrpcpp/src/XmlRpcClient.cpp
```diff
@@ -458,15 +459,25 @@ XmlRpcClient::parseResponse(XmlRpcValue& result)
...
     XmlRpcUtil::error("Error in XmlRpcClient::parseResponse: Invalid response - no param or fault tag. Response:\n%s", _response.c_str());
     _response = "";
+    _header = "";
     return false;
   }
       
   _response = "";
+
+  // Close connection if protocol is HTTP/1.0
+  if (_header.rfind("HTTP/1.0", 0) == 0) {
```
dirk-thomas, line 475: Why use `rfind` here? Existing code uses this: https://github.com/ros/ros_comm/blob/c6ad71e7105f151d912c8c0db3bd18caad771e89/utilities/xmlrpcpp/src/XmlRpcServerConnection.cpp#L135

mgrrx: I was checking the code and actually _header contains much more than you'd expect:
```
HTTP/1.1 200 OK
Server: BaseHTTP/0.3 Python/2.7.17
Date: Mon, 10 Aug 2020 19:48:04 GMT
Content-type: text/xml
Content-length: 274

<?xml version='1.0'?>
<methodResponse>
<params>
<param>
<value><array><data>
<value><int>1</int></value>
<value><string>Unregistered [/talker] as provider of [/rosout]</string></value>
<value><int>1</int></value>
</data></array></value>
</param>
</params>
</methodResponse>
```

Is what _header contains when I run this branch locally on my melodic machine. A plain find would potentially find HTTP/1.0 randomly inside the xml payload. I've no clue why _header contains much more than just the header…

I actually don't have a kinetic machine at hand. Our robots are also on melodic. Is there any chance somebody else could take over here?

dirk-thomas: I see now. So the `rfind` is used to mimic "startswith" and avoid searching through the rest of the string. Sounds good to me. (Can't wait for `starts_with` in C++20...)

dirk-thomas (approved):

Issue #1280, opened 2017-12-29
Provide nicer split times when using rosbag with `--split` and `--duration`


berkowski:
The current behavior splits files at intervals of the provided duration, so given a start time of 01:43:23 and a max duration of 1 hour we'll get files like:

- 2017-12-29-01-43-23_0.bag
- 2017-12-29-02-43-23_1.bag
- 2017-12-29-03-43-23_2.bag

We'd like to be able to generate log splits at "nice" times.  Using the same example:

- 2017-12-29-01-43-23_0.bag
- 2017-12-29-02-00-00_1.bag
- 2017-12-29-03-00-00_2.bag

I am working on a large project transitioning an existing system over to ROS-kinetic and we will implement this kind of behavior, as well as providing the ability to set the timestamp format in the filename.  I am quite willing to share the result if there's interest here.

dirk-thomas:
With the given start time and duration the behavior described with the first bullets makes total sense to me as-is.

~~Please feel free to create a pull request adding an option to the code base to switch to your proposed naming schema.~~ Just saw the referenced PR, thanks.

dirk-thomas:
With the given start time and duration the behavior described with the first bullets makes total sense to me as-is.

~~Please feel free to create a pull request adding an option to the code base to switch to your proposed naming schema.~~ Just saw the referenced PR, thanks.

dirk-thomas:
I close this issue in favor of the pending PR #1286 (to only have one open item). Thank you for working on this.

PR #1286, opened 2017-12-30
rosbag rollovers based on nice wall-clock values with changable datestring format.

berkowski:
Addresses  #1280 for ~~`kinetic-devel`~~ `lunar-devel` (retargeted)

- Adds `--time-format` option to `rosbag record` to change the string formatting used for datestrings in the generated bag files.
- Changes rosbag split behavior when `--duration` is used.

The new behavior now splits at nice wall clock times.  For example, for `--duration=5m` the split times would be at 00:00, 00:05, 00:10, etc.  The first bag file created will have it's duration shortened to shift the interval splitting onto the nice times

### DIFF
--- a/tools/rosbag/src/recorder.cpp
+++ b/tools/rosbag/src/recorder.cpp
@@ -112,7 +112,8 @@ RecorderOptions::RecorderOptions() :
     max_duration(-1.0),
     node(""),
     min_space(1024 * 1024 * 1024),
-    min_space_str("1G")
+    min_space_str("1G"),
+    file_timestamp_format_str("%Y-%m-%d-%H-%M-%S")
 {
 }
 
@@ -269,14 +270,14 @@ bool Recorder::shouldSubscribeToTopic(std::string const& topic, bool from_node) 
 }
 
 template<class T>
-std::string Recorder::timeToStr(T ros_t)
+string Recorder::timeToStr(T ros_t, const std::string &format)
 {
     (void)ros_t;
     std::stringstream msg;
     const boost::posix_time::ptime now=
         boost::posix_time::second_clock::local_time();
     boost::posix_time::time_facet *const f=
-        new boost::posix_time::time_facet("%Y-%m-%d-%H-%M-%S");
+        new boost::posix_time::time_facet(format.c_str());
     msg.imbue(std::locale(msg.getloc(),f));
     msg << now;
     return msg.str();
@@ -345,7 +346,7 @@ void Recorder::updateFilenames() {
     if (prefix.length() > 0)
         parts.push_back(prefix);
     if (options_.append_date)
-        parts.push_back(timeToStr(ros::WallTime::now()));
+        parts.push_back(timeToStr(ros::WallTime::now(), options_.file_timestamp_format_str));
     if (options_.split)
         parts.push_back(boost::lexical_cast<string>(split_count_));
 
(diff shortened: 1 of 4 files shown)


tools/rosbag/src/recorder.cpp
```diff
@@ -167,6 +168,23 @@ int Recorder::run() {
...
+    {
+        // Our last interval number
+        unsigned int prev_interval = (unsigned int) (floor(start_time_.toSec() / options_.max_duration.toSec()));
+
+        // This is the ros::Time that marks the beginning of the interval we've started recording in.
+        ros::Time prev_rollover_time;
+        prev_rollover_time.fromSec(prev_interval * options_.max_duration.toSec());
+
+        // Shift the start time backwards by our 'elapsed' time since the previous period begin
+        start_time_ -= (start_time_ - prev_rollover_time);
```
dirk-thomas, line 186: This seems to be changing the `start_time` unconditionally of using the new option?

berkowski:
Added another command line switch to toggle file splitting time behavior.

tools/rosbag/src/recorder.cpp
```diff
@@ -268,14 +286,14 @@ bool Recorder::shouldSubscribeToTopic(std::string const& topic, bool from_node)
 }
 
 template<class T>
-std::string Recorder::timeToStr(T ros_t)
+string Recorder::timeToStr(T ros_t, const std::string &format)
```
mikepurvis: Be consistent about `string` vs `std::string` please.

tools/rosbag/src/recorder.cpp
mikepurvis: This is fixed now, as the whole block is behind `if (options_.nice_split_times ...`.

tools/rosbag/src/recorder.cpp
berkowski: Fixed.

tools/rosbag/src/recorder.cpp
dirk-thomas: Currently the value of `start_time_` is set to `prev_rollover_time` (because `start_time_ - start_time_` is always `0`). I would suggest to either change the assignment explicitly to that or instead compute a variable which contain the duration of `now` since the last interval start and subtract that variable from `start_time_`.

dirk-thomas:
I tried to retarget `lunar-devel` which didn't work out-of-the-box. Please rebase the patch accordingly and then target the `lunar-devel` branch.

berkowski:
Sure, I can start migrating the patches to `lunar-devel`. Might it be possible to split the PR into two, one for `kinetic` and one for `lunar`?  We're using `kenetic` in a large project due to the LTS badge and would really prefer not to keep a local fork of `ros-comm` with these changes.

dirk-thomas:
New patches are always merged into `lunar-devel` (the current default branch, soon `melodic-devel`) first. After that they can be backported (in a PR tiger with other changes) if necessary / demanded. New features are coming less likely be backported than fixes though.

berkowski:
Makes sense.  I should be able to pick it back up next week.

berkowski:
Rebased the commits onto `lunar-devel`  and retargeted the PR merge branch.  Unfortunately I can't change the name of the topic branch without creating a whole new PR and losing the discussion history, so it still has "kinetic" in the name for now.

I still have a version targeted at `kinetic-devel` that can be used for backporting if desired.

dirk-thomas:
@berkowski Please consider targeting the latest default branch since Lunar is EOL for a while.

dirk-thomas:

Issue #1222, opened 2017-11-08
rospy.log*_throttle (and _once) do not accept (kw)args


gavanderhoorn:
It would appear the `rospy.logX_throttle(..)` and `rospy.logX_once(..)` methods introduced in #812 and #1041 do not accept args or keyword args like the regular `logX(..)` methods do.

This means that instead of `rospy.loginfo("fmt", arg0, arg1, argN)` they have to be passed a bare string (ie: `rospy.loginfo_throttle(1.0, "some string")`).

This can be worked around by using `str.format(..)` itself, but symmetry with the base `logX(..)` methods is lost this way.

Was this intentional, or an oversight?

dirk-thomas:
@wkentaro Maybe you can answer the question and/or provide a follow up PR to align the signatures.

dirk-thomas:
Addressed by #1290.

PR #1289, opened 2018-01-01
Added args and kwargs to rospy.log*

mgrrx:
Implements #1222 

I changed the signature of `_base_logger` so that e.g. a key named `logger_level` can be used for formatting `msg`.

### DIFF
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -159,10 +159,9 @@ def _base_logger(msg, args, kwargs, throttle=None,
                  throttle_identical=False, level=None, once=False):
 
     rospy_logger = logging.getLogger('rosout')
-    name = kwargs.get('logger_name')
+    name = kwargs.pop('logger_name', None)
     if name:
         rospy_logger = rospy_logger.getChild(name)
-        del kwargs['logger_name']
     logfunc = getattr(rospy_logger, level)
 
     if once:
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -155,13 +155,8 @@ def _frame_to_caller_id(frame):
     return pickle.dumps(caller_id)
 
 
-def _base_logger(msg, *args, **kwargs):
-
-    name = kwargs.pop('logger_name', None)
-    throttle = kwargs.pop('logger_throttle', None)
-    level = kwargs.pop('logger_level', None)
-    once = kwargs.pop('logger_once', False)
-    throttle_identical = kwargs.pop('logger_throttle_identical', False)
+def _base_logger(msg, args, kwargs, name=None, throttle=None,
+                 throttle_identical=False, level=None, once=False):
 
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/core.py
```diff
@@ -145,44 +145,42 @@ def rospywarn(msg, *args):
...
-    level = kwargs.pop('logger_level', None)
-    once = kwargs.pop('logger_once', False)
+def _base_logger(msg, args, kwargs, throttle=None,
+                 level=None, once=False):
 
     rospy_logger = logging.getLogger('rosout')
+    name = kwargs.get('logger_name')
     if name:
         rospy_logger = rospy_logger.getChild(name)
+        del kwargs['logger_name']
```
dirk-thomas: Instead of `kwargs.get(...)` and `del kwargs[...]` could this continue to call `kwargs.pop(...)` as it was before?

mgrrx:
Rebased against melodic-devel and changed what was commented.

clients/rospy/src/rospy/core.py
```diff
@@ -155,52 +155,48 @@ def _frame_to_caller_id(frame):
...
 
 
-def _base_logger(msg, *args, **kwargs):
-
-    name = kwargs.pop('logger_name', None)
-    throttle = kwargs.pop('logger_throttle', None)
-    level = kwargs.pop('logger_level', None)
-    once = kwargs.pop('logger_once', False)
-    throttle_identical = kwargs.pop('logger_throttle_identical', False)
+def _base_logger(msg, args, kwargs, throttle=None,
```
dirk-thomas, line 158: I don't see why `args` are `kwargs` can't continue to be passed as `*args` and `**kwargs`. What is the benefit wrapping them into a single parameter?

mgrrx: Since the logger functions (logdebug, loginfo, ...) accept *args and **kwargs and _base_logger has additional optional arguments I wanted to clearly separate them. It's a somewhat artificial case but let's assume one would write the following code:

`rospy.loginfo("foo", once=True, throttle=rospy.Duration(1.0), level='fatal')`
This would make no sense at all but would break the logic in _base_logger.

For me it only makes sense to set change the logger_name when calling rospy.log*.

larslue:
Any chance that this pull request will be merged? I stumbled over the limitations of `rospy.log*` recently and realized that they can be fixed by this PR. I know that it's not urgent to merge it, but one has to continue to build workarounds around `rospy.log*` to support all ways of logging. A common example would be the exc_info, i.e. `rospy.logdebug(msg, exc_info=True)` which is not supported at the moment.

mgrrx:
@dirk-thomas which change is required here? I thought I addressed all issues

dirk-thomas:
> which change is required here? I thought I addressed all issues

That is why I *removed* the label.

dirk-thomas:

dirk-thomas:
Thanks for the patch.

PR #1290, opened 2018-01-01
Simplified internal loggers

mgrrx:
This allows using args and kwargs as in the python logging module

### DIFF
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -132,18 +132,18 @@ _rospy_logger = logging.getLogger("rospy.internal")
 # other sorts of information that scare users but are essential for
 # debugging
 
-def rospydebug(msg, *args):
+def rospydebug(msg, *args, **kwargs):
     """Internal rospy client library debug logging"""
-    _rospy_logger.debug(msg, *args)
-def rospyinfo(msg, *args):
+    _rospy_logger.debug(msg, *args, **kwargs)
+def rospyinfo(msg, *args, **kwargs):
     """Internal rospy client library debug logging"""
-    _rospy_logger.info(msg, *args)
-def rospyerr(msg, *args):
+    _rospy_logger.info(msg, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/core.py
```diff
@@ -131,18 +131,10 @@ def parse_rosrpc_uri(uri):
 # other sorts of information that scare users but are essential for
 # debugging
 
-def rospydebug(msg, *args):
-    """Internal rospy client library debug logging"""
```
dirk-thomas: The information that these functions are "internal" is being lost in the patch. Based on the fact that the functions are intended to be internal I am wondering why you want to allow `kwargs` here? Anyway if it makes still sense to allow them here please update the patch to add them to the existing function definitions and pass them along.

mgrrx:
My intention here was to remove code that basically does nothing and at the same time prevents to pass additional keyword arguments. I stumbled upon it when implementing https://github.com/ros/ros_comm/pull/1289
Rationale: The python logging module by default inspects two keyword arguments: exc_info and extra. I found it quite helpful in the past to pass exc_info and I think it would not hurt to use it e.g. in https://github.com/ros/ros_comm/blob/melodic-devel/clients/rospy/src/rospy/impl/tcpros_base.py#L567

dirk-thomas:
The information about the "internal" nature of the API is still lost from the documentation. Before this note appeared in the docblock of the function.

To enable `kwargs` put keep the docblock you could do something like this:

```
def rospydebug(msg, *args, **kwargs):
    """Internal rospy client library debug logging"""
    _rospy_logger.debug(msg, *args, **kwargs)
```

mgrrx:
changed as suggested. I also fixed the wrong docstring

dirk-thomas:
Please rebase the patch to resolve the conflicts.

mgrrx:
done

dirk-thomas:

dirk-thomas:
Thanks for the patch.

PR #1291, opened 2018-01-01
Use string.Template in RosoutStreamHandler

mgrrx:
This allows to use e.g. $message instead of ${message} in ROSCONSOLE_FORMAT.

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -36,6 +36,7 @@ Library for configuring python logging to standard ROS locations (e.g. ROS_LOG_D
 
 import os
 import sys
+import string
 import time
 import logging
 import logging.config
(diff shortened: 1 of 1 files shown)


tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -232,31 +233,32 @@ def __init__(self, colorize=True, stdout=None, stderr=None):
...
+        msg = self._template.substitute(severity=level,
+                                        message=record_message,
+                                        walltime='%f' % time.time(),
+                                        thread=record.thread,
+                                        logger=record.name,
+                                        file=record.pathname,
+                                        line=record.lineno,
+                                        function=record.funcName,
+                                        node=node_name,
+                                        time=time_str)
```
mikepurvis, line 262: Slight preference here for creating a `replacements` dict and then doing

```
msg = self._template.substitute(**replacements)
```

But I don't think it much matters, and what you have is more succinct.

mikepurvis (approved):
Seems like a good change to me; would be great to have tests that cover these log expansions somehow, but whatever.

dirk-thomas:
I am not sure if that is likely but if an existing format string contains a dollar sign which isn't followed by a variable name in curly braces this patch will make this use case break. While the chances are probably low I am not sure this patch adds a lot of valuable on the other side for this chance of breakage.

mgrrx:
My intention here was to simplify the existing code:
- iteratively replacing substrings by values seems like a bad idea to me. And since string.Template uses regular expressions it is also faster
- I moved the os.environ.get part into the constructor since the environment variables should not change during run time.

dirk-thomas:
> I am not sure if that is likely but if an existing format string contains a dollar sign which isn't followed by a variable name in curly braces this patch will make this use case break. While the chances are probably low I am not sure this patch adds a lot of valuable on the other side for this chance of breakage.

@mgrrx Can you please respond to my earlier concern.

Until then I will mark this for future distros to avoid potential regressions with existing uses.

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -233,31 +234,32 @@ def __init__(self, colorize=True, stdout=None, stderr=None):
         except ImportError:
             self._get_time = None
             self._is_wallclock = None
+        self._template = string.Template(os.environ.get(
+            'ROSCONSOLE_FORMAT', '[${severity}] [${time}]: ${message}') + '\n')
```
dirk-thomas, line 238: In the case where the format string is read from the environment variable I would expect it to not have a trailing newline?

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -233,31 +234,32 @@ def __init__(self, colorize=True, stdout=None, stderr=None):
...
             node_name = get_name()
         except ImportError:
             node_name = '<unknown_node_name>'
-        msg = msg.replace('${node}', node_name)
+
         time_str = '%f' % time.time()
         if self._get_time is not None and not self._is_wallclock():
             time_str += ', %f' % self._get_time()
-        msg = msg.replace('${time}', time_str)
-        msg += '\n'
```
mgrrx, line 260: @dirk-thomas answering your question: because of this \n here ;)

dirk-thomas: But this line is being removed.

mgrrx: yep this line is removed and instead I added the \n in line 238

dirk-thomas: Nvm, the trailing `\n` is *outside* of the get env call.

dirk-thomas:
@mgrrx Please rebase to target `noetic-devel`.

dirk-thomas:
@mgrrx Friendly ping.

dirk-thomas:
> Until then I will mark this for future distros to avoid potential regressions with existing uses.

With Noetic being released and likely being the last ROS 1 distro and this ticket proposing to change behavior which shouldn't target an already released distro I will go ahead and close it.

PR #1292, opened 2018-01-01
Fixed search strategy for python_logging config

mgrrx:
The current strategy tries to first find a file named python_logging.conf and, in case none was found, continues searching the yaml version.
A custom configuration in /etc/ros/python_logging.yaml would be ignored as the file python_logging.conf is always present in the share/conf folder of rosgraph.

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -149,17 +149,18 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
             except OSError as e:
                 sys.stderr.write("INFO: cannot create a symlink to latest log directory: %s\n" % e)
 
-    if 'ROS_PYTHON_LOG_CONFIG_FILE' in os.environ:
-        config_file = os.environ['ROS_PYTHON_LOG_CONFIG_FILE']
-    else:
-        # search for logging config file in /etc/.  If it's not there,
-        # look for it package-relative.
-        config_file = None
+    config_file = os.environ.get('ROS_PYTHON_LOG_CONFIG_FILE')
+    if not config_file:
+        # search for logging config file in ROS_HOME, ROS_ETC_DIR or relative
+        # to the rosgraph package directory.
         rosgraph_d = rospkg.RosPack().get_path('rosgraph')
-        for fname in ['python_logging.conf', 'python_logging.yaml']:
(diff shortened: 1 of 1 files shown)


tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -149,17 +149,16 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
...
         # search for logging config file in /etc/.  If it's not there,
         # look for it package-relative.
-        config_file = None
         rosgraph_d = rospkg.RosPack().get_path('rosgraph')
-        for fname in ['python_logging.conf', 'python_logging.yaml']:
-            for f in [os.path.join(rospkg.get_ros_home(), 'config', fname),
-                      '/etc/ros/%s'%(fname),
-                      os.path.join(rosgraph_d, 'conf', fname)]:
+        for config_dir in [os.path.join(rospkg.get_ros_home(), 'config'),
+                           '/etc/ros/',
```
mikepurvis: Please use `get_etc_ros_dir`, since that accounts for the `ROS_ETC_DIR` envvar:

http://docs.ros.org/independent/api/rospkg/html/rospkg_environment.html#get_etc_ros_dir

tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -149,17 +149,16 @@ def configure_logging(logname, level=logging.INFO, filename=None, env=None):
...
+                           os.path.join(rosgraph_d, 'conf')]:
+            for extension in ['.conf', '.yaml', '.yml']:
```
mikepurvis: Consider `os.extsep`, but it isn't used a lot elsewhere in ROS tooling, and `.` works on all of the supported ROS platforms.

mikepurvis (approved):
Seems reasonable, but can you describe more about how you verified the new and existing behaviour?

mgrrx:
Updated as suggested. I played a bit and placed files here and there. And from the code it was somehow obvious that when a file is found, the search ends.

mikepurvis:
👍

dirk-thomas:
I don't think we need to introduce the option for `yml`. Support `yaml` should be sufficient. Therefore I removed it in 9801667bc53bc9bf835c508987acf5ca80584dde.

Thank you for the patch. This order makes much more sense.

PR #1293, opened 2018-01-02
adding decompress to free(state) before return

aswinthomas:
Calling ``roslz4_compressEnd(&stream)`` and ``roslz4_decompressEnd(&stream)`` to avoid a potential memory leak

### DIFF
--- a/utilities/roslz4/src/lz4s.c
+++ b/utilities/roslz4/src/lz4s.c
@@ -573,7 +573,10 @@ int roslz4_buffToBuffCompress(char *input, unsigned int input_size,
 
   int ret;
   ret = roslz4_compressStart(&stream, block_size_id);
-  if (ret != ROSLZ4_OK) { return ret; }
+  if (ret != ROSLZ4_OK) {
+    roslz4_decompressEnd(&stream);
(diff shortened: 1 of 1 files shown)


utilities/roslz4/src/lz4s.c
```diff
@@ -573,7 +573,10 @@ int roslz4_buffToBuffCompress(char *input, unsigned int input_size,
 
   int ret;
   ret = roslz4_compressStart(&stream, block_size_id);
-  if (ret != ROSLZ4_OK) { return ret; }
+  if (ret != ROSLZ4_OK) {
+    roslz4_decompressEnd(&stream);
```
tfoote: This doesn't look symmetric should it be `roslz4_compressEnd`?

aswinthomas: Ah Yes! Thanks for catching this

mikepurvis:
LGTM

dirk-thomas:
Thank you for the patch. I am closing this in favor of #1313 which uses your patch but targets the latest development branch.

@aswinthomas please see the discussion / comments on the new PR.

PR #1294, opened 2018-01-02
rosbag record modes

rhaschke:
This PR improves `rosbag record` in the following fashion:
- Re-enable `snapshot` mode, which was implemented but not exposed, because `RecorderOptions::snapshot` was always initialized `false`.
- Add new `startstop` mode, which allows to start/stop recording without subscription delay. In this mode, the recorder listens to the topic `start_stop`, which accepts a string as follows:
-- `start` (re)starts recording
-- `start:<filename>` (re)starts recording to new filename
-- `stop` pauses recording

### DIFF
--- a/tools/rosbag/include/rosbag/recorder.h
+++ b/tools/rosbag/include/rosbag/recorder.h
@@ -55,6 +55,7 @@
 #include <ros/time.h>
 
 #include <std_msgs/Empty.h>
+#include <std_msgs/String.h>
 #include <topic_tools/shape_shifter.h>
 
 #include "rosbag/bag.h"
@@ -84,6 +85,7 @@ public:
     ros::Time                    time;
 };
 
+enum RecordingMode { CONTINUOUS, SNAPSHOT, STARTSTOP };
 struct ROSBAG_DECL RecorderOptions
 {
     RecorderOptions();
@@ -94,7 +96,7 @@ struct ROSBAG_DECL RecorderOptions
     bool            do_exclude;
     bool            quiet;
     bool            append_date;
-    bool            snapshot;
+    RecordingMode   mode;
     bool            verbose;
     CompressionType compression;
     std::string     prefix;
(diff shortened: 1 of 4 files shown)


rhaschke:
ping

dirk-thomas:
The current patch doesn't only break ABI but also API. Wouldn't it be feasible to introduce this enhancement in an API compatible fashion instead? I would rather not break existing API.

rhaschke:
@dirk-thomas I guess you are talking about the replacement of `RecorderOptions::snapshot` with `RecorderOptions::mode`, don't you?
I tried to write a clean new interface to offer both the old and new recording modes. Adding yet another boolean to indicate another mode alternative, didn't seem to be appropriate. Neither the python interface nor the C++ binary exposed the snapshot option. Hence, it was only accessible via the library. Doing so, if I remember correctly, wasn't working either. So, I had doubts that this API was actively used.
So far my motivation ;-)

To maintain API compatibility, I could keep the `RecorderOptions::snapshot` flag and internally translate it to the new mode variable, issuing a deprecation message if it was actually used. Does this sound reasonable?

dirk-thomas:
> I guess you are talking about the replacement of `RecorderOptions::snapshot` with `RecorderOptions::mode`.

Yes.

> it was only accessible via the library

And that means that existing code might use exactly that constructor signature. Changing it would break such code.

> if I remember correctly, wasn't working either.

If that is the case then please consider to create a separate ticket about that issue and consider to provide a pull request to address the problem.

> I could keep the RecorderOptions::snapshot flag and internally translate it to the new mode variable, issuing a deprecation message if it was actually used.

Looking closer into the patch I am not sure if replacing the previous bool with an enum to choose between the three modes is the right thing to do. The `snapshot` flag allows to trigger the writing tot disk. The new start/stop mode decides which the time window when messages are collected. Imo these two are orthogonal features.

It makes sense to use the start/stop mode but still write the data immediately. It would also be feasible to use the start/stop mode to collect specific messages and then "snapshot" write them later to disk. Using an enum to choose between these three modes seems to limit the options.

On another note I am not sure of a string message is the right message type to trigger the start/stop logic. If would either suggest a message with a boolean or a custom message which defines the possible values as enums / constants.

dirk-thomas:

PR #1303, opened 2018-01-11
Add add_rostest_gmock function

v-lopez:
This is an extension of the work done here: https://github.com/ros/catkin/pull/897

If/when this is approved, i'll make a pull request for kinetic-devel as well, or feel free to cherry-pick the commit, it shouldn't have any conflicts.


dirk-thomas:
Can you please update the patch by increasing the minimum version of [catkin](https://github.com/ros/ros_comm/blob/6051daa788fc090558d6521fe33f0b7294a9dafe/tools/rostest/package.xml#L13) to 0.7.9 (which is when the new GMock API was introduced.

v-lopez:
Done

dirk-thomas:
Thank you for the patch and the quick update.

PR #1306, opened 2018-01-16
Fixing dangling if-else statement

### LINKED ISSUE #1305: Dangling if-else triggers compilation error | rosbag_storage
I am currently building ROS Kinetic from source using the install generator
on Gentoo Linux (32 bit) --- actually, all this for the Pepper robot, but that is just FYI.

However, [rosbag_storage](https://github.com/ros/ros_comm/tree/1.12.0) fails to compile due to two dangling if else statements (1.12.0) in 

1st)
https://github.com/ros/ros_comm/blob/1.12.10/tools/rosbag_storage/src/lz4_stream.cpp#L180

2nd)
https://github.com/ros/ros_comm/blob/lunar-devel/tools/rosbag_storage/src/bz2_stream.cpp#L131

GCC is 5.4.0.

warp1337:


mikepurvis:
This commit will have to be cherry-picked to the `lunar-devel` branch and then can be merged there. It will be considered for backport when we're next preparing a release into Kinetic.

warp1337:
Actually in the lunar-devel branch, due to 

<pre>
CONSOLE_BRIDGE_logError
</pre>

the error is "mitigated". Thus, the fix as proposed only needs to be applied 
to tags and/or branches that still use the  

<pre>
logError("unused data already available")
</pre>

If I am right...

dirk-thomas:
The fix should still be applied to the `lunar-devel` branch since without using curly braces the code is fragile in case the macros changes to be either empty or expands into two statements.

warp1337:
Agree. I never understood why people (prefer to) omit the curly braces...

dirk-thomas:
Can you please rebase the patch in order to make it mergable.

warp1337:
I was out of office last week. Will do within 48 hours.

warp1337:
I just tried to rebase based on lunar devel, lots of conflicts... Wouldn't it be easier to add those four curly braces on the lunar devel branch and create a new pull request?

dirk-thomas:
That would be fine too.

dirk-thomas:
Addressed by #1637.

PR #1307, opened 2018-01-16

guillaumeautran:
Address a long standing issue where the `sockets_changed_` flag would never be reset causing the native poll set to be re-created everytime doing a poll.


guillaumeautran:
FYI: @mikepurvis 
I'd like to merge this into `lunar-edge` waiting for the other PR#1281 to get merged into `lunar-devel`...

PR #1311, opened 2018-01-19
Raise the correct exception from AnyMsg.serialize

drigz:
Previously, this was causing `TypeError: 'module' object is not
callable`, which was not noticed due to the bare `except:` in the test.

I wasn't able to run the tests, so I'm relying on CI to test that part of the change. I tried running them within a `ros:lunar` docker container with:
```
catkin_make --pkg ros_comm
catkin_make run_tests_test_rospy_rostest
```
but got lots of errors like:
```
  File "/catkin_ws/src/ros_comm/clients/rospy/src/rospy/client.py", line 68, in <module>
    from roscpp.srv import GetLoggers, GetLoggersResponse, SetLoggerLevel, SetLoggerLevelResponse
ImportError: No module named srv
```

### DIFF
--- a/clients/rospy/src/rospy/msg.py
+++ b/clients/rospy/src/rospy/msg.py
@@ -69,7 +69,7 @@ class AnyMsg(genpy.Message):
     def serialize(self, buff):
         """AnyMsg provides an implementation so that a node can forward messages w/o (de)serialization"""
         if self._buff is None:
-            raise rospy.exceptions("AnyMsg is not initialized")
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thank you for the patch.

In the future please see the buildfarm documentation on how to reproduce the same build locally: https://github.com/ros-infrastructure/ros_buildfarm/blob/master/doc/jobs/devel_jobs.rst#run-the-devel-job-locally

PR #1313, opened 2018-01-29
adding decompress to free(state) before return (targeting Lunar)

dirk-thomas:
Replaces #1293 targeting the latest development branch.

### DIFF
--- a/utilities/roslz4/src/lz4s.c
+++ b/utilities/roslz4/src/lz4s.c
@@ -584,7 +584,7 @@ int roslz4_buffToBuffCompress(char *input, unsigned int input_size,
   int ret;
   ret = roslz4_compressStart(&stream, block_size_id);
   if (ret != ROSLZ4_OK) {
-    roslz4_decompressEnd(&stream);
+    roslz4_compressEnd(&stream);
     return ret;
   }
 
--- a/utilities/roslz4/src/lz4s.c
+++ b/utilities/roslz4/src/lz4s.c
@@ -616,10 +616,7 @@ int roslz4_buffToBuffDecompress(char *input, unsigned int input_size,
 
   int ret;
   ret = roslz4_decompressStart(&stream);
-  if (ret != ROSLZ4_OK) {
-    roslz4_decompressEnd(&stream);
-    return ret;
-  }
+  if (ret != ROSLZ4_OK) { return ret; }
 
   while (stream.input_left > 0 && ret != ROSLZ4_STREAM_END) {
     ret = roslz4_decompress(&stream);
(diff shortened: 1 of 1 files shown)


utilities/roslz4/src/lz4s.c
```diff
@@ -583,7 +583,10 @@ int roslz4_buffToBuffCompress(char *input, unsigned int input_size,
 
   int ret;
   ret = roslz4_compressStart(&stream, block_size_id);
-  if (ret != ROSLZ4_OK) { return ret; }
+  if (ret != ROSLZ4_OK) {
+    roslz4_compressEnd(&stream);
```
clalancette, line 587: It seems kind of weird to call `compressEnd` when `compressStart` is the one that is failing.  I think it might make more sense to fix `roslz4_compressStart` so that if the call to `streamResizeBuffer` fails, it calls `streamStateFree` and then returns the error.  What do you think?

utilities/roslz4/src/lz4s.c
```diff
@@ -613,7 +616,10 @@ int roslz4_buffToBuffDecompress(char *input, unsigned int input_size,
 
   int ret;
   ret = roslz4_decompressStart(&stream);
-  if (ret != ROSLZ4_OK) { return ret; }
+  if (ret != ROSLZ4_OK) {
+    roslz4_decompressEnd(&stream);
```
clalancette: Same idea here, though in this case, I don't think that `roslz4_decompressStart` can leak memory on an allocation failure, so we might not want to change this at all.

utilities/roslz4/src/lz4s.c
dirk-thomas: Looking at the current implementation of these functions the patch looks good to me (`compressStart` is doing two things which can individually fail). Maybe you can clarify what you think can go wrong with this change?

utilities/roslz4/src/lz4s.c
dirk-thomas: I agree with you that this shouldn't be necessary since `roslz4_decompressStart` only allocates when returning `OK`. So in case of it failing it doesn't need to deallocate anything.

utilities/roslz4/src/lz4s.c
clalancette: This change will work fine, but I'm suggesting that the contract with `roslz4_compressStart` should be that it either succeeds with some memory allocated, or should fail with no visible side-effects.  That would mean that if something failed internally, it would clean up after itself and not depend on the caller to cleanup for it.  This would also make it easier for any future user of the API to use it, since they wouldn't have to remember to cleanup on failure.

dirk-thomas: I will consider to merge the current patch as-is. Please feel free to refactor the current API / implementation in a separate PR.

clalancette: Sure, that sounds fine.

PR #1314, opened 2018-01-30
[roslaunch-deps] Print warning for package.xml, not manifest.xml.

130s:
Mentioning about manifest.xml could be just confusing with little benefit these days.

```
$ apt-cache policy ros-kinetic-roslaunch
ros-kinetic-roslaunch:
  Installed: 1.12.12-0xenial-20171116-224009-0800
  Candidate: 1.12.12-0xenial-20171116-224009-0800

$ roslaunch-deps -w app.launch
Dependencies:
rviz robot_state_publisher joint_state_publisher tf rgbd_launch openni2_launch nodelet

Missing declarations:
pkg_AA/manifest.xml:
  <depend package="rviz" />
  <depend package="robot_state_publisher" />
  <depend package="joint_state_publisher" />
  <depend package="tf" />
  <depend package="openni2_launch" />
```

### DIFF
--- a/tools/roslaunch/src/roslaunch/depends.py
+++ b/tools/roslaunch/src/roslaunch/depends.py
@@ -365,7 +365,7 @@ def roslaunch_deps_main(argv=sys.argv):
         print('\nMissing declarations:')
         for pkg, miss in missing.items():
             if miss:
-                print("%s/manifest.xml:"%pkg)
+                print("%s/package.xml:"%pkg)
                 for m in miss:
                     print('  <depend package="%s" />'%m)
     


130s:
I just noticed that `  <depend package="rviz" />` is invalid in package.xml format. But with version 1-2-3 of package.xml formats and all variants (`{build, build_export, run, etc.}_depend`) I don't know what to do so just leave it as it is now.

tfoote:
That looks correct for manifest.xml. It would be better to add logic to detect the manifest.xml vs package.xml and change the output of both the name and dependencies accordingly instead of assuming now it's package.xml as they might be in a legacy rosbuild package.

dirk-thomas:
I would be fine with just printing the warning with the current "standard" - so `package.xml` files.

Anyway the following dependency tag is incorrect for `package.xml` files and the corresponding test fails and needs to be update. Please update the patch accordingly.

dirk-thomas:

PR #1317, opened 2018-02-01
printerrlog does not have file keyword

k-okada:
c.f. #1193


dirk-thomas:
Thank you for catching this regression.

PR #1320, opened 2018-02-03
Use std::allocator_traits for forward-compatibility

### LINKED ISSUE #1304: Use std::allocator_traits<ContainerAllocator>::template rebind instead of ContainerAllocator::template rebind
`allocator_traits` is not available till C++11, so this might be in the backlog for a while. Switching conditionally on the compiler version might be reasonable.

Using `::rebind` directly on the type is deprecated as of C++17.

Thankfully, no changes are needed to `ContainerAllocator` to make this work.

eric-wieser:
This requires C++11, but continues to support pre-C++11-style allocators

Fixes #1304.

Also remove the spaces in `> >`, which are no longer required in C++11.

### DIFF
--- a/clients/roscpp/rosbuild/scripts/msg_gen.py
+++ b/clients/roscpp/rosbuild/scripts/msg_gen.py
@@ -58,7 +58,7 @@ MSG_TYPE_TO_CPP = {'byte': 'int8_t', 'char': 'uint8_t',
                    'uint64': 'uint64_t', 'int64': 'int64_t',
                    'float32': 'float',
                    'float64': 'double',
-                   'string': 'std::basic_string<char, std::char_traits<char>, typename ContainerAllocator::template rebind<char>::other > ',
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think this is covered by any tests. I assume you tested it locally somehow? Since this is only effects legacy rosbuild parts I will just go ahead and merge it. Thank you.

eric-wieser:
>  I assume you tested it locally somehow?

No, I was hoping there might be CI on this repo that would test it for me - sorry for not making that clear.

> Since this is only effects legacy rosbuild parts

What do you mean by that? Is this code not used by every cpp ROS node?

dirk-thomas:
> Is this code not used by every cpp ROS node?

No, `rosbuild` was the used build system before the transition to `catkin`.

eric-wieser:
So what piece of code generates the message C++ headers under catkin? That was what I was trying to target here

dirk-thomas:
[gencpp](https://github.com/ros/gencpp)

eric-wieser:
Perhaps it would be wise to revert this, and I should reapply the change there - where hopefully there are existing tests

PR #1324, opened 2018-02-11
Add rosconsole echo

efernandez:
This PR adds a new command to `rosconsole` that allows to print the logger messages from the `/rosout` or `/rosout_agg` topic using only the CLI.

This new command is `echo`, so we can simply call `rosconsole echo` to see all the logger messages. The command takes multiple optional arguments that modify its behaviour, as shown below:
```bash
$ rosconsole echo -h
usage: rosconsole echo [-h] [-1] [--topic TOPIC] [--nocolor] [-d]
                       [FILTER] [LEVEL]

Print logger messages

positional arguments:
  FILTER         regular expression to filter the logger name (default: .*)
  LEVEL          minimum logger level to print (default: warn)

optional arguments:
  -h, --help     show this help message and exit
  -1, --once     prints one logger message and exits
  --topic TOPIC  topic to read the logger messages from (default: /rosout)
  --nocolor      output without color
  -d, --detail   print full logger details
```

This video shows how it works using two test nodes from https://github.com/efernandez/rosconsole_echo_test:
[https://youtu.be/Bn7f26OKBD8](http://www.youtube.com/watch?v=Bn7f26OKBD8)

The two positional arguments supports auto-complete hints thanks to https://github.com/ros/ros/pull/168.

This is slightly inspired on @guillaumeautran 's [`rosdiagnostic` tool](https://github.com/ros/diagnostics/blob/indigo-devel/rosdiagnostic/src/rosdiagnostic/rosdiagnostic.py). I tried to make it closer to how the other `rosconsole` commands are though, and also auto-complete friendly.

### DIFF
--- a/clients/rospy/src/rospy/rosconsole.py
+++ b/clients/rospy/src/rospy/rosconsole.py
@@ -154,6 +154,8 @@ class RosConsoleEcho(object):
         'FATAL': 95,  # Light magenta
     }
 
+    LEVEL_MAX_LENGTH = max([len(level) for level in LEVEL_COLOR.keys()])
+
     def __init__(self, options):
         self._filter = re.compile(options.filter)
         self._level = getattr(Log, options.level.upper())
@@ -166,9 +168,10 @@ class RosConsoleEcho(object):
         rospy.Subscriber(options.topic, Log, callback)
 
     def _stringify(self, level):
-        string = level.ljust(5)
+        string = level.ljust(RosConsoleEcho.LEVEL_MAX_LENGTH)
 
-        return string if self._nocolor else '\033[{}m{}\033[0m'.format(self.LEVEL_COLOR[level], string)
+        return string if self._nocolor else \
+               '\033[{}m{}\033[0m'.format(self.LEVEL_COLOR[level], string)
 
     @staticmethod
     def get_levels():
--- a/clients/rospy/src/rospy/rosconsole.py
+++ b/clients/rospy/src/rospy/rosconsole.py
@@ -144,32 +144,36 @@ def _rosconsole_cmd_get(argv):
 
 
 class RosConsoleEcho(object):
-    LEVEL_STRING_NO_COLOR = {
-        Log.DEBUG: 'DEBUG',
-        Log.INFO : 'INFO ',
-        Log.WARN : 'WARN ',
-        Log.ERROR: 'ERROR',
-        Log.FATAL: 'FATAL',
-    }
-
-    LEVEL_STRING_COLOR = {
-        Log.DEBUG: '\033[92mDEBUG\033[0m',
-        Log.INFO : '\033[97mINFO \033[0m',
-        Log.WARN : '\033[93mWARN \033[0m',
-        Log.ERROR: '\033[91mERROR\033[0m',
-        Log.FATAL: '\033[95mFATAL\033[0m',
+    # See ANSI/VT100 terminal color codes here:
+    # https://misc.flogisoft.com/bash/tip_colors_and_formatting
+    LEVEL_COLOR = {
+        'DEBUG': 92,  # Light green
+        'INFO' : 97,  # White
+        'WARN' : 93,  # Light yellow
+        'ERROR': 91,  # Light red
+        'FATAL': 95,  # Light magenta
     }
 
     def __init__(self, options):
-        self._level_string_map = self.LEVEL_STRING_NO_COLOR if options.nocolor else \
-                                 self.LEVEL_STRING_COLOR
-
         self._filter = re.compile(options.filter)
         self._level = getattr(Log, options.level.upper())
+        self._nocolor = options.nocolor
         self._verbose = options.verbose
+
+        self._level_string_map = {getattr(Log, level): self._stringify(level) for level in self.LEVEL_COLOR.keys()}
 
         callback = self._once_callback if options.once else self._callback
         rospy.Subscriber(options.topic, Log, callback)
+
+    def _stringify(self, level):
+        string = level.ljust(5)
+
+        return string if self._nocolor else '\033[{}m{}\033[0m'.format(self.LEVEL_COLOR[level], string)
+
+    @staticmethod
+    def get_levels():
+        """Get levels sorted by increasing severity."""
+        return sorted(RosConsoleEcho.LEVEL_COLOR.keys(), key=lambda level: getattr(Log, level))
 
     def _print(self, msg):
         print('[ {} ] [\033[1m{}\033[21m]: {}'.format(
@@ -198,7 +202,7 @@ def _get_cmd_echo_argparse(prog):
                         help='regular expression to filter the logger name (default: %(default)s)')
 
     parser.add_argument('level', metavar='LEVEL', type=str, nargs='?', default='warn',
-                        choices=RosConsoleEcho.STRING_LEVEL.keys(),
+                        choices=[level.lower() for level in RosConsoleEcho.get_levels()],
                         help='minimum logger level to print (default: %(default)s)')
 
     parser.add_argument('-1', '--once', action='store_true', dest='once',
--- a/clients/rospy/src/rospy/rosconsole.py
+++ b/clients/rospy/src/rospy/rosconsole.py
@@ -201,7 +201,8 @@ def _get_cmd_echo_argparse(prog):
     parser.add_argument('filter', metavar='FILTER', type=str, nargs='?', default='.*',
                         help='regular expression to filter the logger name (default: %(default)s)')
 
-    parser.add_argument('level', metavar='LEVEL', type=str, nargs='?', default='warn',
+    parser.add_argument('-l', '--level', action='store', metavar='LEVEL',
+                        type=str, default='warn', dest='level',
                         choices=[level.lower() for level in RosConsoleEcho.get_levels()],
                         help='minimum logger level to print (default: %(default)s)')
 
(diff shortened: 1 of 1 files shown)


efernandez:
FYI @mikepurvis @dirk-thomas

clients/rospy/src/rospy/rosconsole.py
```diff
@@ -33,15 +33,23 @@
...
 
 import os
+import re
 import socket
 import sys
 
+from datetime import datetime
+from dateutil.tz import tzlocal
+
+from argparse import ArgumentParser
```
dirk-thomas, line 43: It seems it would make sense to use `argparse` for the whole script. But I wouldn't couple that with the feature you are adding here. It is certainly good to use it for the new `echo` verb.

clients/rospy/src/rospy/rosconsole.py
```diff
@@ -135,13 +143,107 @@ def _rosconsole_cmd_get(argv):
...
+    parser.add_argument('-1', '--once', action='store_true', dest='once',
+                        help='prints one logger message and exits')
+
+    parser.add_argument('--topic', action='store', metavar='TOPIC',
+                        type=str, default='/rosout', dest='topic',
+                        help='topic to read the logger messages from (default: %(default)s)')
+
+    parser.add_argument('--nocolor', action='store_true', help='output without color')
+
+    parser.add_argument('-d', '--detail', action='store_true', help='print full logger details')
```
dirk-thomas: Often such an option is called `-v` / `--verbose`.

clients/rospy/src/rospy/rosconsole.py
```diff
@@ -135,13 +143,107 @@ def _rosconsole_cmd_get(argv):
...
+    parser = _get_cmd_echo_argparse(' '.join([os.path.basename(argv[0]), argv[1]]))
+    args = parser.parse_args(argv[2:])
+
+    rospy.init_node('rosconsole', anonymous=True)
+
+    rosconsole = RosConsoleEcho(args)
+
+    rospy.spin()
+
+    del rosconsole
```
dirk-thomas: Deleting the local variable doesn't do anything. Since the script is exited after this function call anywhere there is nothing which needs to be done. As an alternative - if the goal is to also be able to call this function without exiting right after it - the `RosConsoleEcho` class could have a `deregister` method which stops the subscription.

clients/rospy/src/rospy/rosconsole.py
```diff
@@ -135,13 +143,107 @@ def _rosconsole_cmd_get(argv):
...
+
+def _get_cmd_echo_argparse(prog):
+    parser = ArgumentParser(prog=prog, description='Print logger messages')
+
+    parser.add_argument('filter', metavar='FILTER', type=str, nargs='?', default='.*',
+                        help='regular expression to filter the logger name (default: %(default)s)')
+
+    parser.add_argument('level', metavar='LEVEL', type=str, nargs='?', default='warn',
+                        choices=RosConsoleEcho.STRING_LEVEL.keys(),
+                        help='minimum logger level to print (default: %(default)s)')
```
dirk-thomas: Since both positional arguments `filter` and `level` are optional it is not very clear how to pass e.g. only a level. Maybe consider to change either of them into `--` option.

clients/rospy/src/rospy/rosconsole.py
```diff
@@ -135,13 +143,107 @@ def _rosconsole_cmd_get(argv):
...
+        'error': Log.ERROR,
+        'fatal': Log.FATAL,
+    }
+
+    def __init__(self, options):
+        self._level_string_map = self.LEVEL_STRING_NO_COLOR if options.nocolor else \
+                                 self.LEVEL_STRING_COLOR
+
+        self._filter = re.compile(options.filter)
+        self._level = self.STRING_LEVEL[options.level.lower()]
```
dirk-thomas: You could avoid the definition of `STRING_LEVEL` with the following code:

```
self._level = getattr(Log, options.level.upper())
```

clients/rospy/src/rospy/rosconsole.py
```diff
@@ -135,13 +143,107 @@ def _rosconsole_cmd_get(argv):
...
+        'debug': Log.DEBUG,
+        'info' : Log.INFO,
+        'warn' : Log.WARN,
+        'error': Log.ERROR,
+        'fatal': Log.FATAL,
+    }
+
+    def __init__(self, options):
+        self._level_string_map = self.LEVEL_STRING_NO_COLOR if options.nocolor else \
+                                 self.LEVEL_STRING_COLOR
```
dirk-thomas: In order to reduce the two variables `LEVEL_STRING_NO_COLOR` and `LEVEL_STRING_COLOR` you could use the following more compact logic (just written in the browser):

```
LEVEL_COLOR = {
    'DEBUG': 92,
    'INFO': 97,
    'WARN': 93,
    'ERROR': 91,
    'FATAL': 95,
}
self._level_string_map = {getattr(Log, level): level.ljust(5) for level in LEVEL_COLOR.keys()}
if not options.nocolor:
    [
        self._level_string_map[getattr(Log, level)] = '\033[' + str(color) + 'm' + self._level_string_map[getattr(Log, level)] + '\033[0m'
        for level, color i LEVEL_COLOR.items()
    ]
```

In the (unlikely) event of another level being added in the future the lines to update would be much lower.

dirk-thomas (commented):
The CI failure is unrelated to this patch and has been addressed by #1325. Please rebase this branch to the latest commit of the targeted branch so that CI can pass.

clients/rospy/src/rospy/rosconsole.py
efernandez: Removed.

clients/rospy/src/rospy/rosconsole.py
efernandez: Changed.

clients/rospy/src/rospy/rosconsole.py
efernandez: Cool! Done.

clients/rospy/src/rospy/rosconsole.py
efernandez: Makes sense.

I've done a different implementation though. Let me know if it looks good to you.

And some minor fixes related to the order of the levels when printed by `argparse` when the user introduces a wrong level.

clients/rospy/src/rospy/rosconsole.py
efernandez: The main reason I made them positional arguments is because I didn't know how to provide auto-complete hints with `rosbash` when I have something like `-l` (or `--level`).

I initially had `--filter` and `--level`. If I only have `--level`, do you if it's possible to have auto-complete hints in https://github.com/ros/ros/pull/168?

clients/rospy/src/rospy/rosconsole.py
efernandez: Do you mean I should make it local to the function where it's being used?

I honestly don't think it makes a big difference, but I can do that if that's what you're requesting here.

dirk-thomas: Sorry, for not being clearer. I wasn't referring to the location of the import statement. Instead I was trying to point out that the first command line argument of `rosconsole` is inspected "manually" by looking at `sys.argv` and then `argparse` is only used for the new `echo` verb. It would be nice if `argparse` would be used for all arguments including the decision/selection based on the verb.

But that is clearly outside the scope of this PR.

clients/rospy/src/rospy/rosconsole.py
dirk-thomas: I think it is possible. Manually implemented completion is usually a :zipper_mouth_face: In newer code I usually rely on `argcomplete` but that would be a drastic change.

I will leave it up to you if you want to change the positional arguments to options.

clients/rospy/src/rospy/rosconsole.py
efernandez: Gotcha! Yes, that makes a lot of sense, but definitely not a change for this PR.

clients/rospy/src/rospy/rosconsole.py
efernandez: I've had a quick look at bash completion and it seems it's definitely possible. I don't know how easy and clean it'd be though.

Since I really like the idea of having `--level`, so we can change the level only, I'd give it a try.
I'll come back to you soon, after updating both PRs.

efernandez: Changed to `-l` (`--level`) and update the rosbash auto-complete function to support it in https://github.com/ros/ros/pull/168

clients/rospy/src/rospy/rosconsole.py
```diff
@@ -135,13 +143,102 @@ def _rosconsole_cmd_get(argv):
...
+        self._nocolor = options.nocolor
+        self._verbose = options.verbose
+
+        self._level_string_map = {getattr(Log, level): self._stringify(level) for level in self.LEVEL_COLOR.keys()}
+
+        callback = self._once_callback if options.once else self._callback
+        rospy.Subscriber(options.topic, Log, callback)
+
+    def _stringify(self, level):
+        string = level.ljust(5)
```
efernandez: @dirk-thomas I feel I should change this to `string = level.ljust(max([len(level) for level in LEVEL_COLOR.keys()]))`

Sounds good?

efernandez: Done.

Also rebased onto `lunar-devel`, although the tests were already passing before that.

efernandez:
Let me know if there is something else to do, or otherwise I'll squash all commits into one before you merge this.

dirk-thomas:
Thank you!

PR #1325, opened 2018-02-12
fix various compiler warnings on bionic

dirk-thomas:
http://build.ros.org/view/Mdev/job/Mdev__ros_comm__ubuntu_bionic_amd64/2/warnings22Result/new/

### DIFF
--- a/test/test_roscpp/test/test_transport_tcp.cpp
+++ b/test/test_roscpp/test/test_transport_tcp.cpp
@@ -132,7 +132,7 @@ TEST_F(Synchronous, readWhileWriting)
   for (int i = 0; i < 10; ++i)
   {
     const uint32_t buf_size = 1024*1024;
-    std::auto_ptr<uint8_t> read_buf(new uint8_t[buf_size]);
+    boost::shared_ptr<uint8_t[]> read_buf(new uint8_t[buf_size]);
 
     std::stringstream ss;
     for (int i = 0; i < 100000; ++i)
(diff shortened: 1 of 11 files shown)


test/test_roscpp/test/test_transport_tcp.cpp
```diff
@@ -132,7 +134,7 @@ TEST_F(Synchronous, readWhileWriting)
   for (int i = 0; i < 10; ++i)
   {
     const uint32_t buf_size = 1024*1024;
-    std::auto_ptr<uint8_t> read_buf(new uint8_t[buf_size]);
+    boost::shared_ptr<uint8_t[]> read_buf(new uint8_t[buf_size]);
```
clalancette, line 135: I know that `ros_comm` generally uses boost, but given that this is buried in a test, can we just use `std::shared_ptr`?

utilities/xmlrpcpp/test/TestValues.cpp
```diff
@@ -457,7 +457,7 @@ TEST(XmpRpc, errors) {
   EXPECT_FALSE(v);
 
   // Conversions to other types should now throw an XmlRpcException.
-  EXPECT_THROW((int)value, XmlRpcException);
+  EXPECT_THROW((void)(int)value; , XmlRpcException);
```
clalancette: Is a semicolon supposed to be here?

test/test_roscpp/test/test_transport_tcp.cpp
dirk-thomas: Since the package isn't building with C++11 it can't use `std` smart pointers. And changing the C++ standard version was beyond the scope of this patch.

utilities/xmlrpcpp/test/TestValues.cpp
dirk-thomas: Thanks for catching this. It is a left over of a previously tried workaround. I removed it.

PR #1323, opened 2018-02-09
changes between 1.12.12 and 1.13.6 for backporting

dirk-thomas:
The following list of changes has been integrated into `ros_comm` 1.13.6 (Lunar) since the last Kinetic release (1.12.12).

**Backported:** (these changes are part of this PR)

* fix warn of --max-splits without --split #1237

  * bug fix

* fix documentation for cleanup_master_blacklist() #1253

  * documentation bug fix

* change error message for single threaded spinner #1164

  * helpful improvement, slightly modified patch

* make xmlrpcpp specific include directory work in devel space #1261

  * fix regression introduced in previous backport

* return an error status on error in rosbag #1257

  * bug fix

* replaced deprecated syntax using backticks #1259

  * helpful improvement

* performance improvement for lower/upper bound #1223

  * helpful improvement

* monotonic clock for callback queue timeouts #1250

  * helpful improvement

* topic subscription scalability fix in #1217

  * while risky (?) a significant improvement

* fix bad IPv6 initialization order #1268

  * bug fix

* fix incorrect docstrings in `rosnode/__init__.py` #1278

  * helpful improvement

* add decompress to free(state) before return #1313

  * bug fix

* raise the correct exception from AnyMsg.serialize #1311

  * bug fix

* make master process explicitly 'required' for parent launch #1228

  * helpful improvement, risk (?) of breaking if users were "relying" on launch to continue when the master crashes

* add TCP_INFO availability check #1211

  * bug fix

* add add_rostest_gmock function #1303

  * helpful improvement to be available for Kinetic and Lunar (as the underlying API in catkin)

* topic subscription scalability fix in #1281

  * while risky (?) a significant improvement

* add process listeners to XML RPC server #1319

  * bug fix

* fix xmlrpc timeout using monotonic clock #1249

  * helpful improvement

**Not backported:**

* tests for XmlRpcValue and XML conversion #1216

  * no need to backport

* unit tests and bug fixes for XmlRpcSocket #1218

  * no need to backport

* fix error handling #1226

  * only applicable to code on the Lunar branch

* unit tests and bug fixes for XmlRpcClient #1221

  * no need to backport

* unit tests for XmlRpcDispatch #1232

  * no need to backport

* update lz4 function name to support Trusty #1236

  * only applicable to code on the Lunar branch

* use namespaced macros instead of deprecated macros #1239

[...]

PR #1327, opened 2018-02-13
only launch core nodes if master was launched by roslaunch

mcopejans:
Backports https://github.com/ros/ros_comm/issues/1097 to Indigo.


dirk-thomas:
Thank you for creating the PR. We usually don't backport individual patches in separate PRs but create a single PR which lists all patches since the last release and decide on each one if it should be backported or not (see #1323 for an example). Therefore I will close this PR for now.

PR #1330, opened 2018-02-22
roslaunch: add --log option

k-okada:
add --log option which force all node output to log, Useful when you store all information in log file

### DIFF
--- a/tools/roslaunch/src/roslaunch/__init__.py
+++ b/tools/roslaunch/src/roslaunch/__init__.py
@@ -140,6 +140,9 @@ def _get_optparse():
     parser.add_option("--screen",
                       dest="force_screen", default=False, action="store_true",
                       help="Force output of all local nodes to screen")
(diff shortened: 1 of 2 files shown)


mikepurvis:
FYI @guillaumeautran

mikepurvis:
Looks reasonable and non-harmful. Thanks for the contribution!

mikepurvis:
Please add a note to the following wiki page about the new functionality: http://wiki.ros.org/roslaunch/Commandline%20Tools

k-okada:
@mikepurvis ok, updated the wiki http://wiki.ros.org/roslaunch/Commandline%20Tools#preview

dirk-thomas:
I marked the new option with "New in Melodic": http://wiki.ros.org/action/diff/roslaunch/Commandline%20Tools?action=diff&rev1=50&rev2=51

Issue #1329, opened 2018-02-20
Deadlock when using StatisticsLogger


DieHertz:
There seems to be a possibility for a deadlock in an otherwise single-threaded application when using `StatisticsLogger` due to how it's publisher is initialized:
```cpp
if (!pub_.getTopic().length())
{
  ros::NodeHandle n("~");
  // creating the publisher in the constructor results in a deadlock. so do it here.
  pub_ = n.advertise<rosgraph_msgs::TopicStatistics>("/statistics", 1);
}
```

Pre-requisites:
1. ros-kinetic, likely present in further versions a well
2. Have multiple subscription to the same topic
3. Enable topic statistics logging

Test sample:
```cpp
int main(int argc, char** argv) {
  ros::init(argc, argv, "statistics_logger_deadlock");
  ros::NodeHandle nh{"~"};

  // This call returns properly
  ros::Subscriber sub_one = nh.subscribe<TopicType>("/topic", 1, callback);

  // In case of a deadlock, the following function never returns
  ros::Subscriber sub_another = nh.subscribe<TopicType("/topic", 1, callback);

  ros::spin();
}
```

After successfully subscribing via `sub_one`, a message arriving to `/topic` via `PollManager` in another thread calls `StatisticsLogger::callback`, which attempts to `advertise` it's own topic.
If, at the same time, in the main thread execution happens to be inside the second `subscribe` call, the two may come into a deadlock because of locking mutexes in the opposite.

Internally `roscpp` handles subsequent subscriptions to the same topic by appending them to the same `Subscription` instance, and that is one of precursors to deadlock problem.
Another is `TopicManager::addSubCallback` calling `sub->addCallback` while `subs_mutex_` is locked.

In my application, the issue manifests itself as follows:
```
(gdb) info thread
info thread
  Id   Target Id         Frame
* 1    Thread 0x7f2fd01f79c0 (LWP 28376) "node" __lll_lock_wait ()
  2    Thread 0x7f2fa13d1700 (LWP 28425) "node" 0x00007f2fc0a7374d in poll ()
  3    Thread 0x7f2fa09be700 (LWP 28493) "node" __lll_lock_wait ()
  4    Thread 0x7f2fa01bd700 (LWP 28495) "node" 0x00007f2fc0a7374d in poll ()
  5    Thread 0x7f2f9f9bc700 (LWP 28496) "node" pthread_cond_wait@@GLIBC_2.3.2 ()
[...]

PR #1331, opened 2018-02-22
Fix StatisticsLogger Deadlock

DieHertz:
For a detailed description have a look at #1329.

Review and suggestions are welcome, in case I've missed something.

### DIFF
--- a/clients/roscpp/include/ros/statistics.h
+++ b/clients/roscpp/include/ros/statistics.h
@@ -34,6 +34,7 @@
 #include "publisher.h"
 #include <ros/time.h>
 #include "ros/subscription_callback_helper.h"
+#include <boost/atomic.hpp>
 #include <map>
 
 namespace ros
@@ -56,9 +57,9 @@ public:
   StatisticsLogger();
 
   /**
-   * Actual initialization. Must be called before the first call to callback()
+   * Actual initialization. Must be called early for callback() to not skip first messages.
    */
-  void init(const SubscriptionCallbackHelperPtr& helper);
+  void init(bool hasHeader, const std::string& topic);
 
   /**
    * Callback function. Must be called for every message received.
@@ -66,6 +67,8 @@ public:
   void callback(const boost::shared_ptr<M_string>& connection_header, const std::string& topic, const std::string& callerid, const SerializedMessage& m, const uint64_t& bytes_sent, const ros::Time& received_time, bool dropped);
 
 private:
+  // indicates parameter and publisher initialization for callback()
+  boost::atomic<bool> initialized_;
 
   // these are hard constrains
   int max_window;
--- a/clients/roscpp/include/ros/subscription.h
+++ b/clients/roscpp/include/ros/subscription.h
@@ -99,6 +99,14 @@ public:
   bool addCallback(const SubscriptionCallbackHelperPtr& helper, const std::string& md5sum, CallbackQueueInterface* queue, int32_t queue_size, const VoidConstPtr& tracked_object, bool allow_concurrent_callbacks);
   void removeCallback(const SubscriptionCallbackHelperPtr& helper);
 
+  /**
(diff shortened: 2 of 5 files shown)


DieHertz:
Seems like in this variant it also causes a more obvious, but a deadlock :-(

DieHertz:
When `StatisticsLogger` is initialized from `subscribe` thread, `subs_mutex_` gets locked twice in the sequence `subscribe (topic) -> advertise (statistics)`. `subs_mutex_` could be made a `recursive_mutex` to allow locking it more than once.

Also, `StatisticsLogger` shouldn't be initialized for topics `/clock` and `/statistics` to avoid locking inside `ros::init`.

Another solution (which works in my tests, doesn't reproduce the original issue, but I can't be sure it's entirely correct yet) is to introduce a `advertise_subscribe_mutex_` into `TopicManager` and lock it inside `advertise/unadvertise/subscribe/unsubscribe`.

Which one should I go with? Need your advice.

DieHertz:
Decided to go with another approach, extracting `StatisticLogger::pub_` initialization to `TopicManager::subscribe`, but outside of `Subscription` creation.
That way I can no longer reproduce any of the issues I had nor other associated deadlock that were possible.

clients/roscpp/include/ros/statistics.h
```diff
@@ -84,6 +87,7 @@ class ROSCPP_DECL StatisticsLogger
   double pub_frequency_;
 
   // publisher for statistics data
+  boost::atomic<bool> pub_initialized_;
```
kev-the-dev: Above comment is no longer accurate.

kev-the-dev (commented):
EDIT: See the issue for the backtrace and code insights what show the deadlock scenario

I've been unable to reproduce this on my machine with the following program like the one you mentioned in the issue (I've run dozens of times). Still looking into it... Do you have a  specific program I can see this with/a rostest which produces this?:
```
#include <ros/ros.h>
#include <std_msgs/Int32.h>
#include <iostream>
#include <vector>

void c1(const std_msgs::Int32::ConstPtr& msg)
{
  std::cout << "C1" << std::endl;
}

void c2(const std_msgs::Int32::ConstPtr& msg)
{
  std::cout << "C2" << std::endl;
}

int main(int argc, char** argv) {
  ros::init(argc, argv, "statistics_logger_deadlock");
  ros::NodeHandle nh;
  
  std::vector<ros::Subscriber*> subs;
  // This call returns properly
  for (size_t i = 0; i < 100; ++i)
  {
    ros::Subscriber* sub = new ros::Subscriber();
    *sub = nh.subscribe<std_msgs::Int32>("/topic", 1, c1);
    ros::spinOnce();
    subs.push_back(sub);
  }

  std::cout << "Bout to spin" << std::endl;

  ros::spin();
}
```

DieHertz:
To reproduce it I have run a really CPU intensive computation task on the same machine.
Visual inspection of the code and a captures trace from 2 locked threads shows that the situation is indeed possible, albeit not too frequent.

I'll try to come up with a minimal project to reproduce this weekend though.

clients/roscpp/include/ros/statistics.h
DieHertz: Will swap the lines

kev-the-dev:
No need, I can see it in the code now. I'll checkout your solution

kev-the-dev (approved):
I agree with this solution. Only downside is that some messages could be skipped if the callback is called before initializeStatisticsLogger() runs. Would appreciate comments above the new functions / variables describing what they do.

DieHertz:
I have reworked it a bit in order to simplify, looks like two separate `init*` functions for `StatisticsLogger` aren't necessary.
Also added comments where appropriate.

dirk-thomas:

dirk-thomas:

PR #1333, opened 2018-02-23
Remove trailing spaces.

csukuangfj:


dirk-thomas:
Thank you for spending the effort to create this patch. This repo contains multiple actively maintained branches. If this change would be merged that would significantly increase the burden on any backport in the future. Therefore I am very sorry that I can't accept the contribution (as much as I would prefer the cleaner code style).

A possible time window might be in one year when Indigo and Lunar are about to be EOLed and before N-Turtle is being branches. At that time there are only two active branches (Kinetic and Melodic). Maybe at that time such a patch could be applied to both branches at the same time... :crossed_fingers:

---

For future contributions it might be good to create a ticket for such an idea beforehand to get feedback from the maintainers before spending the effort on it. I would rather not have anyone spend their time on something which doesn't get accepted.

Issue #1339, opened 2018-03-03
Can't pass in a numerical string as arg param to C++ node


lucasw:
If a numerical string like "4235" or "57.23" is set onto the parameter server as a command line argument to a C++ node then it is turned into type int or float on the parameter server and a getParam of type string will not get it.  

```
#include <ros/ros.h>

int main(int argc, char** argv)
{
  ros::init(argc, argv, "string_param_test");
  ros::NodeHandle nh("~");

  // this comes up blank if str_param was passed in as argument, or set
  // rosparam set /string_param_test/str_param "98432"
  // the type is int so a string getParam fails.
  std::string test = "";
  nh.getParam("str_param", test);
  std::cout << test << "\n";

  // this works as desired
  nh.setParam("str_param2", "12345");
  nh.getParam("str_param2", test);
  std::cout << test << "\n";

  return 0;
}
```

```
$ rosrun param_test string_param_test _str_param:="98432"

12345
```
 
```
$ rosparam get /string_param_test/str_param2
'12345'
```

It looks like the root of the issue is that setting a parameter with quotes that is a valid int or float turns into an int or float:

```
$ rosparam set /string_param_test/str_param '8945'
$ rosparam get /string_param_test/str_param 
8945
```

I'm using kinetic on Ubuntu 16.04.  

I also tried this with kinetic-devel, though got a `‘ros_steadytime’ is not a member of ‘ros’` error (even when bringing in the latest roscpp_core) so went back to d4788bc8aa1fc354448d5d7ad45d45451668eade .

(I experimented with adding unit tests here but I'm not sure how to quickly build and test them in a catkin tools build environment - it takes a long time to build and run many tests of test_roscpp, I can't selectively just do a few.  How to build just `./devel/lib/test_roscpp/test_roscpp-params` and not run it or anything else?  Extract the build line out of a verbose build/run of everything?)

lucasw:
Double quotes with single inner quotes works for rosparam set, the outer quotes are stripped of, and the inner quotes clue rosparam that the type is string:

```
rosparam set /test "'98432'"
rosparam get /test 
'98432'
```

But using that as an arg to a C++ node and the string is '98432', with the single quotes as characters in the string, but in a python node the quotes are stripped and the type is correctly determined to be a string.

rosparam get on a string with single quotes yields

```
'''98432'''
```

```
#!/usr/bin/env python

import rospy

rospy.init_node("string_param_test")
val = rospy.get_param("~str_param", None)
[...]

PR #1340, opened 2018-03-04
Strip quotes around strings before putting onto parameter server

lucasw:
The behavior of setting strings through rosparam set and rospy node command line args should all be the same, this allows setting numerical strings through the command line e.g. _foo:="'43'" which gets stored as string 43.

For #1339 

Consider stripping the quotes from roslaunch params for additional consistency.


dirk-thomas:

dirk-thomas:

PR #1345, opened 2018-03-08
Support respawning for machine tag and add some logic to correctly re…

mehditlili:
I had a use case not yet supported by roslaunch, using the machine tag to launch nodes on a remote machine (additional Rpi or Odroid gathering sensor data on a robot).

Unfortunately it happens from time to time that the remote machine loses power and reboots, this can only be recovered by restarting the whole robot's bringup in the current implementation. I added a respawn an respawn_delay attributes the the Machine class in order to first respawn the child roslaunch process on the remote machine, then a function "relaunch_nodes" to the SSHChildROSLaunchProcess class that does basically the same as launch_remote_nodes from ROSRemoteRunner.

There are two points I still want to discuss though:

1. Rebooting takes some time, so I added a while loop in _ssh_exec that keeps retrying to connect whenever a socket.error is thrown (remote machine still not rebooted yet). It will fail after TIMEOUT_SSH_REBOOT = 30 has elapsed. A simpler approach would be to just increase the newly introduced respawn_delay in the machine tag to a time after which the remote machine should have already rebooted, but that is rather heuristic.

2. In my actual pull request, relaunch_nodes and launch_remote_nodes are redundant, I could just move launch_remote_nodes to SSHChildROSLaunchProcess.

Some feedback would be helpful.

### DIFF
--- a/tools/roslaunch/src/roslaunch/core.py
+++ b/tools/roslaunch/src/roslaunch/core.py
@@ -332,10 +332,11 @@ class Machine(object):
     specification.
     """
     __slots__ = ['name', 'address', 'ssh_port', 'user', 'password', 'assignable',
-                 'env_loader', 'timeout']
+                 'env_loader', 'timeout', 'respawn', 'respawn_delay']
     def __init__(self, name, address,
                  env_loader=None, ssh_port=22, user=None, password=None, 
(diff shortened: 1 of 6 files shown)


dirk-thomas:

dirk-thomas:

PR #1348, opened 2018-03-12
Add feature to rostopic pub: repeatedly publish a single message read…

### LINKED ISSUE #1347: rostopic: pub a single message from file
Hello,

when trying to publishing a single message from file, it is published only once, even though a rate is given.

It is certainly desired that, when having multiple messages in the file, they are published one after another at the desired rate. But when providing just one message, it should be repeatedly published.

#### as desired, `argv_publish`:

`rostopic pub -r 1 /test std_msgs/Header '{}'`
-> publishes a header every second

#### misleading, from file via `stdin_publish`
`rostopic pub -r 1 /test std_msgs/Header -f header.yaml`

with header.yaml:

```
  seq: 0
  stamp:
    secs: 0
    nsecs:         0
  frame_id: 'test_frame'
```
-> publishes the message once in latching mode

m-naumann:
… from file, resolves #1347

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -1933,7 +1933,14 @@ def stdin_publish(pub, msg_class, rate, once, filename, verbose):
     else:
         iterator = stdin_yaml_arg
 
-    number_of_messages = len(list(iterator()))
+    # determine whether there is exactly one message
+    exactly_one_message = True
+    number_of_messages = 0
+    for _ in iterator():
+        number_of_messages += 1
+        if number_of_messages > 1:
+            exactly_one_message = False
+            break
 
     r = rospy.Rate(rate) if rate is not None else None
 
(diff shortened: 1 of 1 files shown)


tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1933,6 +1933,8 @@ def stdin_publish(pub, msg_class, rate, once, filename, verbose):
     else:
         iterator = stdin_yaml_arg
 
+    number_of_messages = len(list(iterator()))
```
dirk-thomas: For a very large file evaluating the whole iterator at the beginning (rather than incrementally) doesn't seem like a viable approach.

m-naumann: Hey dirk-thomas, thanks for the feedback! I updated it by iterating only up to element two (if exists)

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1933,6 +1933,15 @@ def stdin_publish(pub, msg_class, rate, once, filename, verbose):
     else:
         iterator = stdin_yaml_arg
 
+    # determine whether there is exactly one message
+    exactly_one_message = True
```
dirk-thomas, line 1937: Shouldn't this value start with `False`? If the iterator has no elements the value of the flag incorrectly reports `True`.

koenlek:
I ran into the same issue. Would be great if this could be merged :+1:

dirk-thomas:
Addressed in #1635.

PR #1350, opened 2018-03-15
Take XmlRpcValue by *const* ref. in operator<<

efernandez:
This allows to use the `operator<<` also with const references to `XmlRpcValue` objects, not only non-const ones.

### DIFF
--- a/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcValue.h
+++ b/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcValue.h
@@ -189,7 +189,7 @@ namespace XmlRpc {
 } // namespace XmlRpc
 
 
-std::ostream& operator<<(std::ostream& os, XmlRpc::XmlRpcValue& v);
+std::ostream& operator<<(std::ostream& os, const XmlRpc::XmlRpcValue& v);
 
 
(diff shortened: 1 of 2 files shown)


efernandez:
@dirk-thomas @tfoote 

Sorry I can't add you directly as reviewers.

mikepurvis:
Would have to update here too, right?

https://github.com/ros/ros_comm/blob/lunar-devel/utilities/xmlrpcpp/src/XmlRpcValue.cpp#L656-L661

efernandez:
Yes, I forgot that. It's fixed in both places now. Thanks.

mikepurvis:
Looks like there's a rosdep problem on Ubuntu causing the CI fail, but two of the jobs are still passing, so this is fine by me to be merged. 👍

mikepurvis (approved):

dirk-thomas:
It looks like the Ubuntu package `python-imaging` has been removed from `bionic. While the binary builds passed before they fail [now](http://build.ros.org/view/Msrc_uB/job/Mbin_uB64__rosbag__ubuntu_bionic_amd64__binary/3/).

dirk-thomas:

dirk-thomas:
Thanks for the patch.

PR #1352, opened 2018-03-20
Fix check for header first in rosbag play for rate control topic

efernandez:
The check doesn't work for `sensor_msgs/Image` (see definition [here](http://docs.ros.org/api/sensor_msgs/html/msg/Image.html)). This PR fixes it.

Example code with the new checks for the `sensor_msgs/Image` definition: https://wandbox.org/permlink/1Oin9UytKs9JFf2h


efernandez:
@dirk-thomas @mikepurvis @tfoote for review

FYI @MasterEric

dirk-thomas:

Issue #1051, opened 2017-05-09
Add substitution args parsing to roslaunch yaml loaded from file


DSsoto:
@jbohren's #218 added resolving args in rosparam if the yaml text is inlined inside of the launch file. 

However I have repeatedly found myself missing this feature when loading parameters from a external yaml file using `rosparam`'s `file` attribute. Here is a minimal reproducible example of the issue.

config.yaml:
```
# This arg will not get substituted corrctly
subbed_pkg_from_file : $(find roscpp)
```

param_loader.launch:
```
<launch>
  <rosparam subst_value="true"> subbed_pkg : $(find roscpp) </rosparam>
  <rosparam subst_value="true" file="/home/santiago/config.yaml" />
</launch>
```

Output of `roslaunch param_loader.launch --dump-params`:
> {/subbed_pkg: /opt/ros/kinetic/share/roscpp, /subbed_pkg_from_file: $(find roscpp)}

It would be very useful to actually have `subbed_pkg_from_file` resolve to the package path when `subst_value` is set to true.

dirk-thomas:
The maintainers will likely not have time to implement this in the foreseeable future. Please consider to contribute a pull request to implement this feature.

dirk-thomas:
Implemented in #1354.

PR #1354, opened 2018-03-22
Add substitution when loading yaml files

nano-meter:
implements #1051

### DIFF
--- a/tools/roslaunch/src/roslaunch/xmlloader.py
+++ b/tools/roslaunch/src/roslaunch/xmlloader.py
@@ -235,9 +235,10 @@ class XmlLoader(loader.Loader):
             # load is the default command            
             cmd = cmd or 'load'
             value = _get_text(tag)
+            subst_function = None
             if subst_value:
-                value = self.resolve_args(value, context)
-            self.load_rosparam(context, ros_config, cmd, param, file, value, verbose=verbose)
+                subst_function = lambda x: self.resolve_args(x, context)
+            self.load_rosparam(context, ros_config, cmd, param, file, value, verbose=verbose, subst_function=subst_function)
 
         except ValueError as e:
             raise loader.LoadException("error loading <rosparam> tag: \n\t"+str(e)+"\nXML is %s"%tag.toxml())
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -365,7 +365,7 @@ class Loader(object):
         else:
             ros_config.add_param(Param(param_name, param_value), verbose=verbose)
         
-    def load_rosparam(self, context, ros_config, cmd, param, file_, text, verbose=True):
+    def load_rosparam(self, context, ros_config, cmd, param, file_, text, verbose=True, subst_function=None):
         """
         Load rosparam setting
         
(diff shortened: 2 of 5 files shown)


dirk-thomas:
Can you please add a test for this new functionality.

nano-meter:
Should I instead of creating a new test yaml file just modify params.yaml?

alextoind:
I was exactly searching for this feature right now!

@dirk-thomas do you think you'll backport it also to kinetic-devel when you are happy with the implementation?

tools/roslaunch/src/roslaunch/xmlloader.py
```diff
@@ -235,9 +235,10 @@ def _rosparam_tag(self, tag, context, ros_config, verbose=True):
             # load is the default command            
             cmd = cmd or 'load'
             value = _get_text(tag)
+            subst_function = None
             if subst_value:
-                value = self.resolve_args(value, context)
-            self.load_rosparam(context, ros_config, cmd, param, file, value, verbose=verbose)
+                subst_function = lambda x: self.resolve_args(x, context)
+            self.load_rosparam(context, ros_config, cmd, param, file, value, verbose=verbose, subst_function=subst_function)
```
dirk-thomas, line 241: Without the patch even a plain `value` (no `file`) was being substituted when `subst_value` was set. After this change substitution seems to only happen on the file content. Is that intentional? This seems to change the existing behavior which would be undesired.

dirk-thomas:
> do you think you'll backport it also to kinetic-devel when you are happy with the implementation?

That depends if the final patch changes any API or behavior and how big the changes / chances for regressions are.

tools/roslaunch/src/roslaunch/xmlloader.py
nano-meter: No, I am not changing existing behavior, the code that does the change is moved to 
https://github.com/nano-meter/ros_comm/blob/a7ed7d8c003f52757b75121ee021b578b4b33a07/tools/roslaunch/src/roslaunch/loader.py#L405

This way both value and file content substitution is handled the same way. 
I also created tests for file and value substitution, so if I introduced undesired changes, those probably would have caught it. If you see a problem, please tell me so I could fix it.

dirk-thomas: Thanks for clarifying. I see now this is only relevant when `cmd` is `load` where it is handles in the function. The other commands don't mind about `value` so it is ok to not substitute it in these cases.

dirk-thomas:
> Should I instead of creating a new test yaml file just modify params.yaml?

The new separate yaml file looks fine to me.

Thanks for the patch. Checking CI results before merging...

dirk-thomas:

dirk-thomas:

dirk-thomas:
Since the failing RosAssert unit tests are unrelated (already on the target branch the case and ticketed separately in #1358) I will go ahead and merge this anyway.

miguelprada:
Now that this has been merged, I'd like to bring back the issue raised by @alextoind of backporting to kinetic-devel. Is there any chance of that happening @dirk-thomas?

dirk-thomas:
> Is there any chance of that happening @dirk-thomas?

That will be considered before the next Kinetic release. In general new features have a lower chance of getting backported.

tlind:
Would also be interested in getting this backported to ROS Kinetic, given that it still has a lifetime of 3 years and this feature looks quite elementary.

PR #1361, opened 2018-04-09
Rework the period standard deviation calculation in Sta…

### LINKED ISSUE #1360: Node using StatisticsLogger crashes when message arrival time is large
If enable_statistics is set to True and a node is subscribed to a topic that receives messages infrequently, it's possible for the arrival times between two messages to be very large (on order of hours).  When this happens it's can trigger a "Duration is out of dual 32-bit range" runtime error when calculating the period standard deviation here https://github.com/ros/ros_comm/blob/lunar-devel/clients/roscpp/src/libros/statistics.cpp#L215 which then causes the node to crash.  The solution is to add a try/except clause around the period standard devision calculation, similar to the stamp age deviation here https://github.com/ros/ros_comm/blob/lunar-devel/clients/roscpp/src/libros/statistics.cpp#L162

deng02:
…tisticsLoggger.

Fixes #1360

### DIFF
--- a/clients/roscpp/src/libros/statistics.cpp
+++ b/clients/roscpp/src/libros/statistics.cpp
@@ -202,7 +202,7 @@ void StatisticsLogger::callback(const boost::shared_ptr<M_string>& connection_he
       msg.period_mean *= 1.0 / (stats.arrival_time_list.size() - 1);
 
       // then, calc the stddev
-      msg.period_stddev = ros::Duration(0);
+      double period_variance = 0.0;
       for(std::list<ros::Time>::iterator it = stats.arrival_time_list.begin(); it != stats.arrival_time_list.end(); it++)
       {
         if (it == stats.arrival_time_list.begin())
@@ -212,11 +212,11 @@ void StatisticsLogger::callback(const boost::shared_ptr<M_string>& connection_he
         }
         ros::Duration period = *it - prev;
         ros::Duration t = msg.period_mean - period;
-        msg.period_stddev += ros::Duration(t.toSec() * t.toSec());
+        period_variance += t.toSec() * t.toSec();
         prev = *it;
       }
-      msg.period_stddev = ros::Duration(sqrt(msg.period_stddev.toSec() / (stats.arrival_time_list.size() - 1)));
-
+      double period_stddev = sqrt(period_variance / (stats.arrival_time_list.size() - 1));
+      msg.period_stddev = ros::Duration(period_stddev);
     }
     else
     {


clients/roscpp/src/libros/statistics.cpp
```diff
@@ -212,7 +212,21 @@ void StatisticsLogger::callback(const boost::shared_ptr<M_string>& connection_he
         }
         ros::Duration period = *it - prev;
         ros::Duration t = msg.period_mean - period;
-        msg.period_stddev += ros::Duration(t.toSec() * t.toSec());
+        try
+        {
+          msg.period_stddev += ros::Duration(t.toSec() * t.toSec());
```
dirk-thomas: The result of this temporary result might clearly exceeding the value range of 32 bit. But in the following lines it is divided and then the `sqrt` is being computed which would bring the value back into a valid range.

Therefore I would suggest to way this value is being computed rather than catching the error and skipping the input.

deng02: Agreed.  I will rework this.

clients/roscpp/src/libros/statistics.cpp
```diff
@@ -212,10 +212,23 @@ void StatisticsLogger::callback(const boost::shared_ptr<M_string>& connection_he
...
+        period_variance += t.toSec() * t.toSec();
         prev = *it;
       }
-      msg.period_stddev = ros::Duration(sqrt(msg.period_stddev.toSec() / (stats.arrival_time_list.size() - 1)));
+      double period_stddev = sqrt(period_variance / (stats.arrival_time_list.size() - 1));
+      try
+      {
+        msg.period_stddev = ros::Duration(period_stddev);
+      }
+      catch(std::runtime_error& e)
```
dirk-thomas: With the updated math in which cases do you expect this to happen?

deng02: Short answer is never :)  The slightly longer mathematial answer that I worked out for my own curiosity is that in a scenario where there are 4 samples where the first 3 arrive within nanoseconds of each other (effectively making the first 2 periods 0) and the 4th arrives much later, the 4th sample would have to arrive 5579326178 seconds or 177 years later. And I won't be around then to deal with the exception.  Thanks for pointing this out, I will remove the guard.

dirk-thomas:
Thank you for the patch.

The CI failure on Melodic is due to a know compiler warning which requires a new pluginlib release. So merging anyway.

Issue #770, opened 2016-03-17
Unsafe Double-checked Locking


git-afsantos:
I realize this is being compiled with C++03, and, thus, the workarounds are not possible, as they require C++11. Here is the faulty code, in `roscpp/src/libros/topic_manager.cpp` (line 56).

``` c++
TopicManagerPtr g_topic_manager;
boost::mutex g_topic_manager_mutex;
const TopicManagerPtr& TopicManager::instance()
{
  if (!g_topic_manager)
  {
    boost::mutex::scoped_lock lock(g_topic_manager_mutex);
    if (!g_topic_manager)
    {
      g_topic_manager = boost::make_shared<TopicManager>();
    }
  }

  return g_topic_manager;
}
```

Double-checked locking as is done here is potentially unsafe; it is possible that a thread may read `g_topic_manager` before it is fully initialized. Of course, it depends on whether there are multiple threads contesting the access (which I am not sure about).

Is lazy initialization really necessary here?

dirk-thomas:
Can you please describe in more detail why you think this code is unsafe.

To me it looks like that the outer condition is checked atomically. And if it is not yet initialized the inner check and creation of the singleton is properly protected by the mutex.

git-afsantos:
I have little experience in C++, however, from my brief research, it is vulnerable in the same way Java code is. Consider the assignment statement.

``` c++
g_topic_manager = boost::make_shared<TopicManager>();
```

There are three things happening here:
1. Allocation of memory;
2. Construction of the object on the allocated memory;
3. Pointing to the allocated memory.

Compilers are not forced to follow this ordering. This means that step 3 can occur before step 2, and here lies the problem.
- Thread A goes through steps 1 and 3 and is scheduled out;
- Thread B enters the method, reads `g_topic_manager` (which is now not null) and returns;
- Thread B is able to use `g_topic_manager` before Thread A goes through step 2.

If this does not make much sense, I will leave you with my references:
- I have been reading on the issue [here](http://www.drdobbs.com/cpp/c-and-the-perils-of-double-checked-locki/184405726);
- I have opened a [question on StackOverflow](http://stackoverflow.com/q/36063617/2088822) to ask exactly this, and the answers confirm that it is unsafe.

This is an issue only if multiple threads try to access the topic manager for the first time, at the same time. I am not familiar enough with the source code to know if this is possible (but I imagine it is, hence the mutex in the first place).

dirk-thomas:
[...]

PR #1365, opened 2018-04-14
Replaced Double-Checked Locking pattern with static variable. Fixes #1364

### LINKED ISSUE #1364: More Uses of Unsafe Double-Checked Locking Pattern
The reasoning is the same as per my previous issue, #770. This code pattern is unsafe, and now that recent releases compile under C++11, there is no reason for it to be used. I just found three more occurrences of this unsafe pattern in the code.

1. [*xmlrpc_manager.cpp*](https://github.com/ros/ros_comm/blob/283818529002c6d37e8b9ec554a17b40c6075c20/clients/roscpp/src/libros/xmlrpc_manager.cpp#L100)
2. [*poll_manager.cpp*](https://github.com/ros/ros_comm/blob/283818529002c6d37e8b9ec554a17b40c6075c20/clients/roscpp/src/libros/poll_manager.cpp#L38)
3. [*connection_manager.cpp*](https://github.com/ros/ros_comm/blob/283818529002c6d37e8b9ec554a17b40c6075c20/clients/roscpp/src/libros/connection_manager.cpp#L45)

git-afsantos:
Copy-pasted the solution for #770. If approved, this fixes #1364.

### DIFF
--- a/clients/roscpp/src/libros/connection_manager.cpp
+++ b/clients/roscpp/src/libros/connection_manager.cpp
@@ -40,20 +40,10 @@
 namespace ros
 {
 
-ConnectionManagerPtr g_connection_manager;
-boost::mutex g_connection_manager_mutex;
 const ConnectionManagerPtr& ConnectionManager::instance()
 {
-  if (!g_connection_manager)
-  {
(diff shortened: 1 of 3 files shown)


mikepurvis:
Looks great, thank you.

PR #1367, opened 2018-04-16
topic_tools: Check that output topic is valid in demux. Fixes #1366

### LINKED ISSUE #1366: topic_tools: demux crashes when using __none
The demux node in topic_tools crashes when it receives a message and the selected topic is the magic "__none" topic

Steps to reproduce

    rosrun topic_tools demux in out_1 out_2
    rosservice call /demux/select "__none"
    rostopic pub -1 /in std_msgs/Empty

demux then crashes with a segmentation fault

This was tested in kinetic on Ubuntu 16.04.

I believe this is due to a null pointer dereference in https://github.com/ros/ros_comm/blob/283818529002c6d37e8b9ec554a17b40c6075c20/tools/topic_tools/src/demux.cpp#L156

lfdebrux:


mikepurvis:
Looks reasonable, thanks!

PR #1368, opened 2018-04-17
fix "pass_all_args" for roslaunch-check

jabbenseth:
the option "pass_all_args" was not recognized by roslaunch checks. This works as expected on "our" launch files which create a mixed tree using the "pass_all_args" option as well as explicit argument forwarding

### DIFF
--- a/tools/roslaunch/resources/example-args.launch
+++ b/tools/roslaunch/resources/example-args.launch
@@ -1,0 +1,3 @@
+<launch>
+  <arg name="test_foo" />
+</launch>
--- a/tools/roslaunch/resources/example-pass_all_args.launch
+++ b/tools/roslaunch/resources/example-pass_all_args.launch
@@ -1,0 +1,4 @@
+<launch>
+  <arg name="test_foo" default="value_foo" />
+  <include file="$(find roslaunch)/resources/example-args.launch" pass_all_args="true" />
+</launch>
--- a/tools/roslaunch/src/roslaunch/depends.py
+++ b/tools/roslaunch/src/roslaunch/depends.py
@@ -170,6 +170,12 @@ def _parse_launch(tags, launch_file, file_deps, verbose, context):
                 else:
                     launch_tag = dom[0]
                     sub_context = _parse_subcontext(tag.childNodes, context)
+                    try:
+                        if tag.attributes['pass_all_args']:
+                            sub_context["arg"] = context["arg"]
+                            sub_context["arg"].update(_parse_subcontext(tag.childNodes, context)["arg"])
+                    except KeyError as e:
+                        pass
                     _parse_launch(launch_tag.childNodes, sub_launch_file, file_deps, verbose, sub_context)
             except IOError as e:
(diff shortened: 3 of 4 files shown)


mikepurvis:
Thanks for this contribution. As the active branch has changed, please retarget this to `melodic-devel`, and it will be considered after merge for backport to Kinetic and Lunar.

Please also add a test-case which exposes the problem and is fixed by your proposed change. I would suggest doing so in this file:

https://github.com/ros/ros_comm/blob/melodic-devel/tools/roslaunch/resources/example.launch

There are [several tests](https://github.com/ros/ros_comm/search?q=%22example.launch%22) which load and process this launch file in different contexts; it shouldn't be hard to make one of them fail.

jabbenseth:
I added a test which exposed the mentioned problem. However if i ran the test manually (`rosrun test_roslaunch test_roslaunch_check.py`) it worked but with `rostest test_roslaunch roslaunch.test` I ran into troubles:
```
'unused args [__name, __log] for include of [/…/src/ros_comm/tools/roslaunch/resources/example-args.launch]' is not None
```
Somehow the implicitly defined arguments "__name" and "__log" for the node specified in the testfile get forwarded and end up in the test environment as well. They are blacklisted in https://github.com/ros/ros_comm/pull/1368/commits/7fb090fa3d2f2cce8fb04553b58401293a902667

Finally my original fix is applied in https://github.com/ros/ros_comm/pull/1368/commits/c8ed560b34e3257ddccf8a14bd7b3e8d46c6d3ee

test/test_roslaunch/test/test_roslaunch_check.py
```diff
@@ -0,0 +1,59 @@
...
+        self.msgs = {}
+
+    def test_roslaunch(self):
+        rospack = rospkg.RosPack()
+        filename = os.path.join(rospack.get_path('roslaunch'), 'resources', 'example-pass_all_args.launch')
+        error_msg = roslaunch.rlutil.check_roslaunch(filename)
+        self.assertIsNone(error_msg)
+
+if __name__ == '__main__':
+    rostest.run(PKG, NAME, TestRoslaunchCheck, sys.argv)
```
mikepurvis: You can filter `argv` using a rospy helper, see: http://wiki.ros.org/rospy/Overview/Initialization%20and%20Shutdown#Accessing_your_command-line_arguments

However, I don't think that's quite the right approach here— AFAICT this doesn't need to be a rostest at all. Rostests are specifically tests which take place in the full roslaunch environment (where a rosmaster and param server are present). What you have here could just be a regular nosetest. See `catkin_add_nosetests` documentation and some examples:

http://docs.ros.org/melodic/api/catkin/html/howto/format2/python_nose_configuration.html
https://github.com/ros/ros_comm/search?q=catkin_add_nosetests

This should allow you to drop the hack in `xmlloader.py` as well.

mikepurvis (changes requested):

mikepurvis (approved):

mikepurvis:
Looks great, thanks for your patience in iterating on this with me.

PR #1358, opened 2018-04-03
debug failing RosAssert tests on Debian Stretch

dirk-thomas:

PR #1373, opened 2018-04-22
Use boost::shared_ptr to fix memory leak in BagPlayer

efernandez:
This fixes two memory leaks in `BagPlayer` by changing the `std::map` of RAW pointers to `BagCallback` to smart pointers `boost::shared_ptr<BagCallback>`"

1. On destruction, b/c the destructor doesn't `delete` the RAW pointers in the `std::map`.
2. In `register_callback` is called with the same `topic` name more than once w/o calling `unregister_callback` for that topic name.

This should also work now if we call `unregister_callback` for a topic not previously registered. Before that would `delete` a null pointer.

### DIFF
--- a/tools/rosbag_storage/src/bag_player.cpp
+++ b/tools/rosbag_storage/src/bag_player.cpp
@@ -43,7 +43,7 @@ ros::Time BagPlayer::real_time(const ros::Time &msg_time) {
 void BagPlayer::start_play() {
 
     std::vector<std::string> topics;
-    std::pair<std::string, BagCallback *> cb;
+    std::pair<std::string, boost::shared_ptr<BagCallback> > cb;
     foreach(cb, cbs_)
         topics.push_back(cb.first);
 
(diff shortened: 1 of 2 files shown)


efernandez:
FYI @dirk-thomas

tools/rosbag_storage/src/bag_player.cpp
```diff
@@ -63,7 +63,6 @@ void BagPlayer::start_play() {
 }
 
 void BagPlayer::unregister_callback(const std::string &topic) {
-    delete cbs_[topic];
```
mikepurvis, line 66: This means it will now not be deleted until the whole object goes out of scope. Perhaps there should still be a `cbs_.erase(topic)` or something here?

efernandez: Of course, that's in the next line. I didn't remove it.

mikepurvis: Oh lol, so I see.

efernandez:
Looks like the tests are currently broken b/c of something else, unrelated with this PR, right?

mikepurvis:
Yes, the debian stretch issue is being investigated in #1358; the other test is flakey.

PR #1372, opened 2018-04-22
keep connection header information in rosbag operations (#679, #685)

ddimarco:
This PR provides a fix for the issue that `rosbag filter` & `rosbag compress` removes latching information from topic connection headers.
See #679 for a nice description. This PR is based on that discussion, but follows a slightly different implementation.

You can see the difference by using [this script](https://gist.github.com/ddimarco/f3e22b27d16f7d3d8f2b9c5dece3a78c) (the script needs the patched code to access the bag headers):
```
# using current kinetic-devel branch
$ rosbag compress test_rosbag.bag --output-dir /tmp/
$ show_bag_headers.py /tmp/test_rosbag.bag           
/chatter_no_latch: {u'topic': '/chatter_no_latch', u'message_definition': 'string data\n', u'md5sum': '992ce8a1687cec8c8bd883ec73ca41d1', u'type': 'std_msgs/String'}
/chatter_latch: {u'topic': '/chatter_latch', u'message_definition': 'string data\n', u'md5sum': '992ce8a1687cec8c8bd883ec73ca41d1', u'type': 'std_msgs/String'}

# this PR
$ rosbag compress test_rosbag.bag --output-dir /tmp/
$ show_bag_headers.py /tmp/test_rosbag.bag           
/chatter_no_latch: {u'message_definition': 'string data\n', u'callerid': '/talker', u'latching': '0', u'md5sum': '992ce8a1687cec8c8bd883ec73ca41d1', u'topic': '/chatter_no_latch', u'type': 'std_msgs/String'}
/chatter_latch: {u'message_definition': 'string data\n', u'callerid': '/talker', u'latching': '1', u'md5sum': '992ce8a1687cec8c8bd883ec73ca41d1', u'topic': '/chatter_latch', u'type': 'std_msgs/String'}
```

PR #1374, opened 2018-04-22
Specialize BagCallbackT for MessageInstance

efernandez:
Specialize `BagCallbackT` for `MessageInstance` so the overloaded `call` method calls the callback directly with the `MessageInstance` object.

This way the client callback can access the connection information, which is useful when processing a bag file to create a new one b/c then we can keep the message time and connection information, e.g. the _latch_ property.


efernandez:
FYI @dirk-thomas

mikepurvis:
Looks reasonable— related to #1372 in terms of overall objective, though that one is for the filter command whereas this is the C++ bag API.

:+1:

mikepurvis:

mikepurvis:
Since this is a header-only change, a CI pass isn't super meaningful. It's certainly not a complicated implementation, but I wonder if there might be some kind of trivial test which could be added to [test_rosbag_storage](https://github.com/ros/ros_comm/tree/melodic-devel/test/test_rosbag_storage)?

efernandez:
@mikepurvis I agree, I'd add a test to `test_rosbag_storage`.
Do you know where the bag files used in those tests are?
For example, I see `/tmp/swap1.bag`, but I can't see where it's on the repo or any cmake rule to download it.

mikepurvis:
@efernandez It gets written as part of the test: https://github.com/ros/ros_comm/blob/e96c407c64e1c17b0dd2bb85b67f388380527097/test/test_rosbag_storage/src/swap_bags.cpp#L9-L12

efernandez:
Thanks.

I've just added a unit test (creating a simple test bag on runtime as the other tests in this package do).

mikepurvis:

mikepurvis:
Remaining failures are known-flaky. Thanks for this contribution!

PR #1376, opened 2018-04-24
Fix compiler warning of test_aes_encryptor

jwon02:
This PR would fix the compiler warning of signed-unsigned mismatch for an EXPECT_EQ call made in test_aes_encryptor.


mikepurvis:
Looks great, thanks for the quick follow-up.

dirk-thomas:
Follow up of #1376.

PR #1379, opened 2018-04-29
warn if ROS_IP contains whitespace

v4hn:
This breaks ROS connections and it does not show with `echo $ROS_IP`.

Also "local IP address" often refers to 192.168. et al,
so I resolved this ambiguity too.

### DIFF
--- a/utilities/roswtf/src/roswtf/network.py
+++ b/utilities/roswtf/src/roswtf/network.py
@@ -88,8 +88,11 @@ def ros_ip_check(ctx):
     # best we can do is compare roslib's routine against socket resolution and make sure they agree
     addrs = rosgraph.network.get_local_addresses()
 
+    if " " in ip:
+        return "ROS_IP [%s] contains whitespace. This is not a valid IP."%ip
+
     if ip not in addrs:
-        return "ROS_IP [%s] does not appear to be a local IP address %s."%(ip, str(addrs))
+        return "ROS_IP [%s] does not appear to be an IP address of a local network interface (one of %s)."%(ip, str(addrs))
     
 # Error/Warning Rules
 


tfoote (approved):

jspricke (approved):

mikepurvis:
Is this added check catching a case not already caught by the `in addrs` check which immediately follows?

v4hn:
> Is this added check catching case not caught by the `in addrs` check which immediately follows?

No, not if you put the question like this.

But it gives the user an explicit hint for this special case which can be very frustrating to debug.
After all this is what `roswtf` is for.
Personally, I debugged this case twice in the last three years and it cost me at least an hour each time to figure it out...

dirk-thomas:
Can you please give an example of a value which you ran into? Are we talking about leading / trailing spaces or spaces in the middle of the IP address?

mikepurvis:
LGTM then.

v4hn:
> Can you please give an example of a value which you ran into?

This was my problem this time:
```
> export ROS_IP=$(ip address show eth0 | sed -n 's:inet \([^ /]*\).*:\1:; T; p')
> echo $ROS_IP
1.2.3.4
> roswtf
...
WARNING ROS_IP may be incorrect: ROS_IP [    1.2.3.4] does not appear to be a local IP address [u'127.0.0.1', ...]
```

There is probably a better way to extract the intended IP address, but a bug in the sed script accidentally added 4 spaces before the IP. In theory this is shown in the warning, but in practice it could also be random padding in the output.

Now this will say
```
WARNING ROS_IP may be incorrect: ROS_IP [    1.2.3.4] contains whitespace. This is not a valid IP.
```

Additionally I disambiguate the previous warning when the IP does not match a local adapter.
It will now print the more comprehensive warning:
```
WARNING ROS_IP may be incorrect: ROS_IP [1.2.3.4] does not appear to be an IP address of a local network interface (one of [u'127.0.0.1', ...]).
```

dirk-thomas:
Thank you for the patch and the explanation.

PR #1386, opened 2018-05-03
Show connection info in rosnode info

efernandez:
This shows the **port**, **URI** (of the other side of the socket) and the **socket** file descriptor for each connection next to the `direction`.

This is useful for debugging the traffic on the port/s of the socket use by a topic, b/c it shows the matching between topic and those ports, which can by used in `tcpdump`.

# Example

Before: `    * direction: inbound`
With this PR: `    * direction: inbound (48768 - 123.0.0.1:41908) [28]`

FYI @mikepurvis @dirk-thomas 

I target `melodic-devel` in this PR, but if possible I'd like to get this in `lunar-devel` as well.

### DIFF
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -45,6 +45,7 @@ import errno
 import sys
 import socket
 import time
+import re
 try:
     from xmlrpc.client import ServerProxy
 except ImportError:
@@ -62,6 +63,7 @@ import rostopic
 
 NAME='rosnode'
 ID = '/rosnode'
+CONNECTION_PATTERN = re.compile(r'\w+ connection on port (\d+) to \[(.*) on socket (\d+)\]')
 
 class ROSNodeException(Exception):
     """
@@ -543,11 +545,16 @@ def get_node_connection_info_description(node_api, master):
                     # older ros publisher implementations don't report a URI
                     buff += "    * to: %s\n"%lookup_uri(master, system_state, topic, dest_id)
                     if direction == 'i':
-                        buff += "    * direction: inbound\n"
+                        buff += "    * direction: inbound"
                     elif direction == 'o':
-                        buff += "    * direction: outbound\n"
+                        buff += "    * direction: outbound"
                     else:
-                        buff += "    * direction: unknown\n"
+                        buff += "    * direction: unknown"
+                    if len(info) > 6:
+                        match = CONNECTION_PATTERN.match(info[6])
+                        if match is not None:
+                            buff += " (%s - %s) [%s]" % match.groups()
+                    buff += "\n"
                     buff += "    * transport: %s\n"%transport
     except socket.error:
         raise ROSNodeIOException("Communication with node[%s] failed!"%(node_api))


tools/rosnode/src/rosnode/__init__.py
```diff
@@ -62,6 +63,7 @@
 
 NAME='rosnode'
 ID = '/rosnode'
+CONNECTION_PATTERN = re.compile(r'\w+ connection on port (\d+) to \[(.*) on socket (\d+)\]')
```
mikepurvis, line 66: It's too bad there doesn't seem to be a more robust way to get this information than tying it to the textual output from here:

https://github.com/ros/ros_comm/blob/e96c407c64e1c17b0dd2bb85b67f388380527097/clients/roscpp/src/libros/transport/transport_tcp.cpp#L719-L724

efernandez: True, but I can't find another way to do it from the master API.

BTW, I think the actual function you wanted to mention is the `rospy` one:

https://github.com/ros/ros_comm/blob/fcf37d289e91931b06e05bfbc395290b793ac2f4/clients/rospy/src/rospy/impl/tcpros_base.py#L470-L476

In this class all the information is available from getters, but it seems none of them are exposed in the API.
Changing that would likely be a big change that could also break other code, so I guess use a regex to parse the string is the only thing that can be done. But maybe I'm missing something from the master API ?!

efernandez:
Rebased on top of `melodic-devel` just to see if the tests pass now.

What are your thoughts about this, @dirk-thomas ?
@mikepurvis Are you ok with the current implementation at the end? I honestly don't see another way to implement this functionality.

FYI @guillaumeautran

tools/rosnode/src/rosnode/__init__.py
dirk-thomas: Please add a comment before this regex referencing the line in `rospy` where the string is defined. A comment in `rospy` where the string is defined referencing this new regex would be good to. Both to avoid one side to be changed without the other.

dirk-thomas:
My earlier comment is still pending: https://github.com/ros/ros_comm/pull/1386#discussion_r191494745

dirk-thomas:
Closing in favor of #1497.

efernandez:
Sorry for the late reply.

I was holding from getting back into this because I noticed some significant degraded performance when this information was being retrieved last time I used it. Unfortunately, I didn't have time to give this a try again, I've been too busy and changed laptops a few times, having to re-build workspaces.

My gut feeling is that the performance issue I experienced was unrelated with this PR, but I just wanted to give you a heads up @dirk-thomas . It'd be great if you could give it a try on your end. From the code I wouldn't expect any major impact on the time `rostopic info` takes to print into the terminal.

Since the commit authorship of my commit was taken by @maximest-pierre , we can blame him if something goes wrong :stuck_out_tongue_winking_eye:

dirk-thomas:
> I was holding from getting back into this because I noticed some significant degraded performance when this information was being retrieved last time I used it.

@efernandez Unfortunately there was no reference about a potential performance impact in this ticket. Since the patch has been merged now please do follow up with some testing to ensure that the change doesn't result in significantly worse performance. I don't have the bandwidth to do additional testing for contributed patches and have to rely on the author to do that. Thanks.

PR #1391, opened 2018-05-07
Add ShapeShifterStamped class & synchronized throttle node

furushchev:
Add `synchronized_throttle` node which is almost the same as `topic_tools/throttle` but accepts multiple topics as inputs and timestamp of all output messages are synchronized.

This feature is useful when throttling both RGB and Depth image or stereo images used for 3dimensional processing.

For making the node work, I also added `topic_tools/ShapeShifterStamped` class which is almost the same as `topic_tools/ShapeShifter` but has `header` field in its instance for comparing timestamps on message synchronization.

synchronized throttle nodelet 1


tools/topic_tools/package.xml
```diff
@@ -18,6 +18,8 @@
   <buildtool_depend version_gte="0.5.78">catkin</buildtool_depend>
 
   <build_depend>cpp_common</build_depend>
+  <build_depend>geometry_msgs</build_depend>
```
dirk-thomas: This would introduce a circular dependendcy between this repo and the `common_msgs` repo. That would make the first time release of these repositories problematic since both need each other.

Maybe moving this functionality into a separate repository would be better.

furushchev: @dirk-thomas Thank you for the feedback! The root problem is there is no message which has Header field in this repo which is need for testing. Would it be nice if I create a new package (let the name be `test_topic_tools`) in [test](https://github.com/ros/ros_comm/tree/melodic-devel/test) directory and move all tests for this package into `test_topic_tools` and also generate a dummy message with header field?

dirk-thomas: Sounds good to me :+1:

dirk-thomas: @furushchev Friendly ping.

furushchev: @dirk-thomas Thanks for pinging. Just now updated this PR according to the discussion above.

furushchev: @dirk-thomas If it would be better to move all existing test codes in `topic_tools` to `test_topic_tools`, I'll do it.

dirk-thomas:
I retargeted the `noetic-devel` branch. Please rebase your branch and resolve conflicts.

dirk-thomas:

Issue #1357, opened 2018-03-30
roscore in 1.13.6 is broken on macos: 


NikolausDemmel:
It throws errors like

```
process[rosout-1]: started with pid [72640]
libc++abi.dylib: terminating with uncaught exception of type std::length_error: vector
[rosout-1] process has died [pid 72640, exit code -6, cmd /opt/ros/lunar/lib/rosout/rosout __name:=rosout __log:=/Users/demmeln/.ros/log/8cd76b6b-345f-11e8-8884-6003088b8618/rosout-1.log].
log file: /Users/demmeln/.ros/log/8cd76b6b-345f-11e8-8884-6003088b8618/rosout-1*.log
[rosout-1] restarting process
```

Seems like it was working fine in 1.13.5. See also discussion at https://github.com/mikepurvis/ros-install-osx/issues/114 . It seems like https://github.com/ros/ros_comm/issues/1243 introduced the offending lines.

NikolausDemmel:
For the record, the immediate reason for the exception is that in

https://github.com/ros/ros_comm/blob/2c8a3f65628d0362d97c29d333d5669b2c7803a6/utilities/xmlrpcpp/src/XmlRpcServer.cpp#L33

for some reason `-1` is returned on macos...

NikolausDemmel:
But even with update of `max_files` commented out, there are further issues, since upon starting roscore it now shows the following errors:

```
[ERROR] [1522447653.950314000]: [registerService] Failed to contact master at [eddard:11311].  Retrying...
[ERROR] [1522447654.003783000]: [registerService] Failed to contact master at [eddard:11311].  Retrying...
[ERROR] [1522447654.062192000]: [registerPublisher] Failed to contact master at [eddard:11311].  Retrying..
```

dirk-thomas:
According to the docs `limit.rlim_max` can be `RLIM_INFINITY` which is defined as `-1`. This case is not being catched in the current code.

@trainman419  @NikolausDemmel Can either of you create a pull request and check that it fixes the problem?

NikolausDemmel:
I'm not sure what the appropriate action is in that case? Sticking with the default of 1024? It would be great to get @trainman419's take on this.

However, even when reverting #1243, just like when leaving `max_files` with the default of 1024, I get the above `Failed to contact master` error. I have no clue, might be an unrelated regression...

[...]

PR #1393, opened 2018-05-09
Change to force a rebuild of the pollset on flag changes

guillaumeautran:
This change fixes issues seen in #1357
Goes with #1281


clients/roscpp/src/libros/io.cpp
```diff
@@ -119,6 +119,8 @@ void close_socket_watcher(int fd) {
 void add_socket_to_watcher(int epfd, int fd) {
 #if defined(HAVE_EPOLL)
   struct epoll_event ev;
+  bzero(&ev, sizeof(ev));
```
guillaumeautran, line 122: Fixing valgrind warnings.

clients/roscpp/src/libros/io.cpp
```diff
@@ -147,6 +149,8 @@ void del_socket_from_watcher(int epfd, int fd) {
 void set_events_on_socket(int epfd, int fd, int events) {
 #if defined(HAVE_EPOLL)
   struct epoll_event ev;
+  bzero(&ev, sizeof(ev));
```
guillaumeautran, line 152: Fixing valgrind warnings.

test/test_roscpp/test/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 catkin_add_gtest(${PROJECT_NAME}-test_version test_version.cpp)
 if(TARGET ${PROJECT_NAME}-test_version)
-  target_link_libraries(${PROJECT_NAME}-test_version)
+  target_link_libraries(${PROJECT_NAME}-test_version ${catkin_LIBRARIES})
```
guillaumeautran, line 3: I could not build and run the unit tests without the following changes.

mikepurvis: Did you end up just running the tests on Linux but with a build where `HAVE_EPOLL` was forced to 0?

guillaumeautran: That's correct yes.

mikepurvis: And prior to this fix the tests failed to complete successfully?

guillaumeautran: Yes, sorry, I should have mentioned that. 

Also, I am looking at the current build failure.

mikepurvis: My impression is that that test has been flakey for a while— it definitely has a timing sensitivity and does _sometimes_ pass on the buildfarm. However, if you have further insights on it, that would be great.

guillaumeautran:
I cannot figure out why the test fails on the build farm. They obviously pass just fine on my laptop which makes it hard to troubleshoot...

mikepurvis:
LGTM, and has the thumbs-up from @mpflanzer in https://github.com/ros/ros_comm/issues/1357#issuecomment-388000450.

IMO this should merge now and we can treat the test regression as separate since it predates this change.

PR #1405, opened 2018-05-21
remove rosconsole

dirk-thomas:
See ros/rosdistro#17919.


dirk-thomas:
CI is expected to fail since at this point `rosconsole` is just missing.

Issue #1408, opened 2018-05-23
ros_comm 1.4.1 breaks socketcan_bridge tests


mathias-luedtke:
I just noticed that one test started to fail for melodic shadow-fixed (1.4.1), but works for main (1.3.6).
https://travis-ci.org/ipa-mdl/ros_canopen/builds/381469509

The [failing test](https://github.com/ros-industrial/ros_canopen/blob/melodic-devel/socketcan_bridge/test/to_socketcan_test.cpp#L157), runs `ros::spinOnce()` to fetch one message.

Digging through the commits https://github.com/ros/ros_comm/pull/1365 raised my attention.
I will run further tests, but this seems to change the behaviour at least for this test.

mathias-luedtke:
I was not able to fix this test with an additional wait. It might be that the message got lost.
So it looks more like https://github.com/ros/ros_comm/pull/1393 is the cause.

The only external depedencies are
```
executing command [apt-get install -y ros-melodic-roscpp]
executing command [apt-get install -y ros-melodic-class-loader]
executing command [apt-get install -y ros-melodic-message-runtime]
executing command [apt-get install -y ros-melodic-std-msgs]
executing command [apt-get install -y ros-melodic-roslint]
executing command [apt-get install -y ros-melodic-rostest]
executing command [apt-get install -y ros-melodic-rosunit]
executing command [apt-get install -y ros-melodic-message-generation]
executing command [apt-get install -y libboost-all-dev]
```

dirk-thomas:
@guillaumeautran Can you please take a closer look at this.

mathias-luedtke:
I was not able to dermine the cause of this problem, perhaps it's the combination of different commits.
However, I was able to come up with a test case.
For some reason the `/rosout` handling seems to be involved.

mikepurvis:
Assuming you're on Linux, I'd be doubtful that #1393 could directly be the issue, since that was a fix for non-epoll systems such as macOS.

Does the test still fail if you build ros_comm from source but override `HAVE_EPOLL=0` in this spot?

https://github.com/ros/ros_comm/pull/1217/files#diff-af96143090deaa643970030f962c964dR70

That would kick it back to the legacy poll implementation for handling IO.

guillaumeautran:
I am taking a look right now but the specific tests passes for me at this time. Still doing some more digging.

mathias-luedtke:
> Assuming you're on Linux

[...]

PR #1409, opened 2018-05-23
add test case for #1408

mathias-luedtke:
This is a down-stripped test case for #1408.
Until now I don't have a better name for it.

This test case passes for ros_comm <=1.13.6, but fails for 1.14.1.

If any of `ROS_INFO` , `publisher` or `subscriber` gets removed from the `Disturber` case, the `Disturbed` case passes.


mathias-luedtke:
@mikepurvis: setting `HAVE_EPOLL` explicitly does not have any effect on the test case result

mathias-luedtke:
Wasn't meant to be merged.

PR #1410, opened 2018-05-24
Fix some errors in some probably not frequented code paths

awesomebytes:
These were detected while cythonizing recursively some files.

[Cython](http://cython.org/) (via `cythonize`) tries to create equivalent .c source code files and it acts kind of like a static code analyser.


mikepurvis:
Nice work, these all look like legit fixes to me! 👍

dirk-thomas:
I just changed the target branch to `melodic-devel`. Please rebase the patch to apply cleanly to that branch. Thanks.

Issue #1380, opened 2018-05-01
rosout should allow file logging to be disabled through parameters


yli-cpr:
A MR will be submitted shortly

PR #1241, opened 2017-11-20
added parameter to stop client libraries from generating topics list

trobertson-cpr:
The cached parameter /rosout_disable_topics_generation stops rospy and roscpp from generating the list of topics and sending them over the network stack as part of their log messages. It does not stop rosout from printing an empty topics list.

PR #1413, opened 2018-05-28
Allow disabling rosout file logging (to rosout.log) (#1381)

yli-cpr:
See issue #1380.

Cherry-pick from melodic-devel


mikepurvis:
This is fine to put on `lunar-edge` for the purposes of Clearpath's internal release management, particularly since we already tested it as-is, but do note that it may be affected by some of the concerns which have now been raised in #1241.

Issue #1399, opened 2018-05-15
[rosbag] Restore snapshot functionality


kev-the-dev:
# Background
The deprecated ```rosrecord``` command (replaced with ```rosbag record```) had an [experimental snapshot feature](http://wiki.ros.org/rosrecord#rosrecord-1) which kept the buffer in memory until triggered with a separate command to write to disk. This feature is useful in live testing scenarios where the importance of recording data is only obvious _after_ something interesting happens and is found in proprietary robotics middleware. AFAIK this feature still exists in the source, but is inaccessible from the CLI.

# Proposal
* Restore the snapshot functionality to ```rosbag record``` by exposing the option to the argument parser
* Add an option for snapshotting to buffer a fixed time duration in the past. It's more convenient for users to know they have the last 5 seconds of data like a dashcamera than thinking in terms of buffer memory usage
* Instead of triggering write with a ```std_msgs/Empty```, use an actionlib goal. This goal would allow the user to specify which topics to save, how far back to go, and name to give the file (ex: save the last 15 seconds of pointcloud data to ```cyclist_at_night.bag```). The action component would allow the user to see the progress (feedback) of writing large multigigabyte files to disk.
* Add service to pause recording, so users can have some time to think about what to save before it is overwritten

I would like to take on this project after receiving feedback on the design.

kev-the-dev:
I found #1294 and saw that we don't want to break the existing API. Perhaps this is appropriate for a new subcommand, ```rosbag snapshot```?

ironmig:
Another option is:
* Preserve original trigger mode but also add the custom action I described
* Using --duration flag with -s flag will implement the fixed duration buffer I described
* If ```rosbag record -t``` is run, it will use the old trigger mode. However, additional flags such as ```rosbag record -t  -o filename --start -3.0 --stop -1.0 topic1 topic2 topic3...``` will use the action with the custom time, filename, and time cropping

The only API change here would be giving new purpose to options that previously did nothing

jian-dong:
@kev-the-dev  Any update about this MR?

kev-the-dev:
As discussed in the above linked PR, @DLu volunteered to make this an independent package

jian-dong:
@kev-the-dev Thanks for you reply,  this package link  [rosbag_snapshot](https://github.com/ros/rosbag_snapshot) not reachable , maybe it is private.

DLu:
Still working on it...

jian-dong:
[...]

PR #1414, opened 2018-05-29
ROSBAG: add snapshot subcommand

kev-the-dev:
# Description
Addresses #1399 by adding a new subcommand, ```rosbag snapshot``` which acts similarly to the deprecated ```rosrecord -s``` command but with additional features. It seemed easier to me to implement this as a new subcommand rather than trying to add this to ```rosbag record``` without breaking API/ABI (I can still try that approach if it is preferred). It subscribes to topics and maintains a buffer of recent messages like a dash cam. This is useful in live testing where unexpected events can occur which would be useful to have data on but the opportunity is missed if ```rosbag record``` was not running (disk space limits make always running ```rosbag record``` impracticable). Instead, users may run snapshot in the background and save data from the recent past to disk as needed.

## Usage
```rosbag snapshot``` can be configured through command line flags and with ROS params for more granular control. By default, the command will run in server mode (buffering data). When certain flags are used, program will act as a client by requesting that the server write data to disk or freezing the buffer to preserve interesting data until a user can decide what to write.

### CLI usage
```
$ rosrun rosbag snapshot -h
Usage: rosbag snapshot [options] [topic1 topic2 ...]

Buffer recent messages until triggered to write or trigger an already running instance.

Options:
  -h [ --help ]                produce help message
  -t [ --trigger-write ]       Write buffer of selected topcis to a bag file
  -p [ --pause ]               Stop buffering new messages until resumed or 
                               write is triggered
  -r [ --resume ]              Resume buffering new messages, writing over 
                               older messages as needed
  -s [ --size ] arg (=-1)      Maximum memory per topic to use in buffering in 
                               MB. Default: no limit
  -d [ --duration ] arg (=30)  Maximum difference between newest and oldest 
                               buffered message per topic in seconds. Default: 
                               30
  -o [ --output-prefix ] arg   When in trigger write mode, prepend PREFIX to 
                               name of writting bag file
  -O [ --output-filename ] arg When in trigger write mode, exact name of 
                               written bag file
  --topic arg                  Topic to buffer. If triggering write, write only
                               these topics instead of all buffered topics.

```
**Hold a buffer of the last 30 seconds of data from selected topics until triggered to write**
```rosbag snapshot -d 30 /tf /odom /camera/image_color /camera/camera_info /velodyne_points```
**Buffer the most recent gigabyte of the following topics in the camera namespace**
```ROS_NAMESPACE=camera rosbag snapshot -s 1000 image_rect_color camera_info```
### Example launch file
```
<launch>
  <node name="snapshot" pkg="rosbag" type="snapshot" args="">
    <rosparam>
        default_duration_limit: 1  # Maximum time difference between newest and oldest message, seconds, overrides -d flag
        default_memory_limit: 0.1  # Maximum memory used by messages in each topic's buffer, MB, override -s flag
        topics:
            - test1                # Inherit defaults
            - test2:               # Override duration limit, inherit memory limit
                duration: 2
            - test3:               # Override both limits
                duration: -1       # Negative means no duration limit
                memory: 0.00
    </rosparam>
  </node>
</launch>
```
## Client examples

**Write all buffered data to ```<datatime>.bag```**
```rosbag snapshot -t```

**Write buffered data from selected topics to ```new_lighting<datatime>.bag```**
```rosbag snapshot -t -o new_lighting /camera/image_raw /camera/camera_info```

**Write all buffered data to ```/home/user/crashed_into_wall.bag```**
 ```rosbag snapshot -t -O /home/user/crashed_into_wall``` 

**Pause buffering of new data, holding current buffer in memory until -t or -r is used**
```rosbag snapshot -p```

**Resume buffering new data**
```rosbag snapshot -r```

**Call trigger service manually, specifying absolute window start and stop time for written data**
```
$ rosservice call /trigger_snapshot "filename: 'short_control_instability.bag'
topics:
- '/tf'
- '/odom'
- '/wrench'
- '/motors/cmd'
start_time:
  secs: 62516
  nsecs: 0
stop_time:
  secs: 62528
  nsecs: 0"
```
**View status of buffered data (useful for future tools/GUI)**
```
$ rostopic echo /snapshot_status 
topics: 
  - 
    topic: "/test"
    node_pub: ''
    node_sub: "/snapshot_1527185221120605085"
    window_start: 
      secs: 62516
      nsecs: 761000000
    window_stop: 
      secs: 62521
      nsecs: 984000000
    delivered_msgs: 524
    dropped_msgs: 0
    traffic: 2096
    period_mean: 
      secs: 0
      nsecs:         0
    period_stddev: 
      secs: 0
      nsecs:         0
    period_max: 
      secs: 0
      nsecs:         0
    stamp_age_mean: 
      secs: 0
      nsecs:         0
    stamp_age_stddev: 
      secs: 0
      nsecs:         0
    stamp_age_max: 
      secs: 0
      nsecs:         0
enabled: True
---
```

### DIFF
--- a/test/test_rosbag/test/test_snapshot.py
+++ b/test/test_rosbag/test/test_snapshot.py
@@ -1,0 +1,250 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2018, Open Source Robotics Foundation, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of Open Source Robotics Foundation, Inc. nor the
+#    names of its contributors may be used to endorse or promote products
+#    derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import os
+import sys
+import tempfile
+import unittest
+import rospy
+from rosbag import Bag
+from std_msgs.msg import String
+from std_srvs.srv import SetBool
+from rosbag_msgs.msg import SnapshotStatus
+from rosbag_msgs.srv import TriggerSnapshot, TriggerSnapshotRequest, TriggerSnapshotResponse
+
+
+class TestRosbagSnapshot(unittest.TestCase):
+    '''
+    Tests the "rosbag snapshot" command.
+    Relies on the nodes launched in snapshot.test
+    '''
+    def __init__(self, *args):
+        self.params = rospy.get_param("snapshot")
+        self._parse_params(self.params)
+        self.last_status = None
+        self.status_sub = rospy.Subscriber("snapshot_status", SnapshotStatus, self._status_cb, queue_size=5)
+        self.trigger = rospy.ServiceProxy("trigger_snapshot", TriggerSnapshot)
+        self.enable = rospy.ServiceProxy("enable_snapshot", SetBool)
+        super(TestRosbagSnapshot, self).__init__(*args)
+
+    def _parse_params(self, params):
+        '''
+        Parse launch parameters of snapshot to cache the topic limits in a map
+        so it is easier to check compliance in later tests.
+        '''
+        self.topic_limits = {}
+        self.default_duration_limit = params['default_duration_limit']
+        self.default_memory_limit = params['default_memory_limit']
+        for topic_obj in self.params['topics']:
+            duration = self.default_duration_limit
+            memory = self.default_memory_limit
+            if type(topic_obj) == dict:
+                topic = topic_obj.keys()[0]
+                duration = topic_obj[topic].get('duration', duration)
+                memory = topic_obj[topic].get('memory', memory)
+            else:
+                topic = topic_obj
+            topic = rospy.resolve_name(topic)
+            duration = rospy.Duration(duration)
+            memory = 1E6 * memory
+            self.topic_limits[topic] = (duration, memory)
+
+    def _status_cb(self, msg):
+        self.last_status = msg
+
+    def _assert_no_data(self, topics=[]):
+        '''
+        Asserts that calling TriggerWrite service with
+        specifed parameters responds non-success and did not create
+        a bag file.
+        '''
+        filename = tempfile.mktemp()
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: test/test_rosbag/test/test_snapshot.py [hunk 1]
```diff
@@ -0,0 +1,250 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2008, Open Source Robotics Foundation, Inc.
```
scpeters: 2018

scpeters:
The spelling of `snapshoter.h` and `snapshoter.cpp` seems odd to me; it would feel more natural with a double-t, as in `snapshotter.h` and `snapshotter.cpp`.

Though, I would actually just use `snapshot.h` and `snapshot.cpp` for the library code and use `snapshot_main.cpp` for the executable.

scpeters:
> Though, I would actually just use `snapshot.h` and `snapshot.cpp` for the library code and use `snapshot_main.cpp` for the executable.

@ironmig do you have any thoughts about this comment?

kev-the-dev:
That makes more sense to me, but it may be preferred to follow the convention established for that package (play.cpp player.cpp record.cpp recorder.cpp)

scpeters:
Ah, that makes sense. I had seen the python file [rosbag_main.py](https://github.com/ros/ros_comm/blob/melodic-devel/tools/rosbag/src/rosbag/rosbag_main.py), but there is more of a pattern in the c++ code.

paulbovbel:
Thanks for this feature @ironmig, it's very useful!

Would the ros_comm maintainers consider merging it into melodic?

dirk-thomas:
This needs to be rebased to resolve merge conflicts.

DLu:
@kev-the-dev I'd be happy to if you don't have time.

kev-the-dev:
I'm happy to see this through assuming it doesn't require a significant redesign. I rebased and squashed it. I'll look into the CI failures to see if they're trivial

dirk-thomas:
You can ignore the Noetic related CI results for now. Since not all dependencies of this repo have been released yet they are expected to fail.

dirk-thomas:

tools/rosbag_msgs/package.xml
```diff
@@ -0,0 +1,14 @@
+<package>
+  <name>rosbag_msgs</name>
+  <version>1.0.0</version>
+  <description>Service and message definitions for rosbag</description>
+  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
```
dirk-thomas, line 5: Your name should be listed here.

tools/rosbag_msgs/package.xml
```diff
@@ -0,0 +1,14 @@
...
+  <license>BSD</license>
+  <author>Kevin Allen</author>
```
dirk-thomas, line 7: Same here.

dirk-thomas (commented):
I am not sure that this 1.5k lines contribution should be merged into this repository. It adds another message (which will implicitly be part of `ros_core`) as well as a new command which will implicitly be maintained by me (since you won't be easily able to watch for incoming issues related to it, create patches, and do releases on demand).

Therefore I am leaning towards suggesting to move the proposed implementation into a separate repository.

kev-the-dev:
That's fair, it is a lot to ask to maintain this much new code in ros_comm. Another package is probably the best option. It might be a few weeks/months, but I'll start on that when I get the time

paulbovbel:
We've been using this command for ~1 year at Locus, thanks for the contribution @kev-the-dev.

arneboe:
There is a memory leak in `Snapshotter::subscribe`.
You are binding `topicCB` using `boost::bind` with `queue` as additional parameter.
Therefore the subscriber holds a shared_ptr to the `queue`.
Storing the subscriber inside the queue (`queue->setSubscriber(sub);`) essentially makes the queue hold a shared_ptr to itself and thus it will never be deleted.
Never deleting the subscriber will in turn stop a whole bunch of ros internals from getting shutdown properly.

A possible fix (albeit not a very nice one) is.

```
Snapshotter::~Snapshotter()
{
    for (std::pair<const std::string, boost::shared_ptr<MessageQueue>>& buffer : buffers_) 
    {
      buffer.second->sub_->shutdown();
    }
}
```

DLu:
I am working on a refactor of this contribution to work as an independent package. Would it be possible to get [https://github.com/ros/rosbag_snapshot](https://github.com/ros/rosbag_snapshot) willed into existence?

tfoote:
@DLu I'd suggest you get it going locally and then we can transfer it into the organization.

dirk-thomas:
Assuming this feature will land in its own package I will go ahead and close this ticket.

DLu:
@tfoote After much ado, the package is now available: https://github.com/locusrobotics/rosbag_snapshot

DLu:
@tfoote Can you get it forked over at ros/rosbag_snapshot? Also, is there a proper organization for the release repository for something in `ros/`?

tfoote:
We can fork it, but usually we prefer to transfer it so that we can keep the history of issues and PRs etc. This one doesn't have much history so that's not a big problem. One other thing about a fork is that apparently github search only searches the parent fork so if development is going to diverge it's better to transfer, or you have to request to "break the fork" from github customer service. 

If you transfer it to me I can then transfer it into the organization. 

For the gbp I can create one for you at in https://github.com/ros-gbp: https://github.com/ros-gbp/rosbag_snapshot-release

DLu:
I have requested the transfer. (You can only transfer from an organization to yourself, so I had to transfer it to DLu first, then to you)

tfoote:
It's now here: https://github.com/ros/rosbag_snapshot

ros-discourse:

https://discourse.ros.org/t/rosbag-snapshot/15324/1

PR #1415, opened 2018-05-30
Fix some errors in some probably not frequented code paths melodic

awesomebytes:
Update to https://github.com/ros/ros_comm/pull/1410

Are you gonna backport to Kinetic also?

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_pubsub.py
+++ b/clients/rospy/src/rospy/impl/tcpros_pubsub.py
@@ -242,7 +242,7 @@ class TCPROSHandler(rospy.impl.transport.ProtocolHandler):
         if type(protocol_params) != list or len(protocol_params) != 3:
             return 0, "ERROR: invalid TCPROS parameters", 0
         if protocol_params[0] != TCPROS:
-            return 0, "INTERNAL ERROR: protocol id is not TCPROS: %s"%id, 0
+            return 0, "INTERNAL ERROR: protocol id is not TCPROS: %s"%protocol_params[0], 0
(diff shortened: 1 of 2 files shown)


dirk-thomas:
#1410 had some additional changes. Are they not applicable to the Melodic branch? Should they be pull requested against an older branch only? If yes, which one would be newest distro which needs those?

> Are you gonna backport to Kinetic also?

All patches are considered for backporting. Since this fixes obvious bugs it will very likely be backported for the next patch release.

awesomebytes:
That are not applicable as the error was fixed already (variable "uri" not
existing in the scope).

So if the file gets backported as it is (which looks more correct) it
should be all fine.

dirk-thomas:
> So if the file gets backported as it is (which looks more correct) it
should be all fine.

The "file" doesn't get backported only the applied patches get cherry-picked to the older branches.

So if older distros need additional fixes please create a separate PR for that targeting the branch with the most recent distro which needs the fix.

awesomebytes:
Ok, I'll need a bit of time to apply the same change to previous
branches/distros (other than kinetic and melodic).
Can you reopen the previous PR and re-target it for kinetic again maybe?

I'll take a look at other branches/distros soon.

2018-05-31 12:19 GMT+10:00 Dirk Thomas <notifications@github.com>:

awesomebytes:
I opened PR on all in-development distros.

Indigo and Kinetic have the same changes.

Lunar and Melodic have the same changes.

I don't know what's the process to apply the same change to all the branches, sorry if this is not in your expected format.

dirk-thomas:
Thanks for the patch!

Issue #1346, opened 2018-03-09
rosout: logging + topic subscription: not thread safe


krixkrix:
If a node subscribes to a topic in one thread and logs to rosout in another, a conflict can occur, since the nodes topic list may be altered while the rosout logging traverses it.

The following error is reported:

> Unable to report rosout: Set changed size during iteration

As well as tcpros protocol synchronization is reported lost:

> [ERROR] [1520594758.451498758]
> [/rosout]: a message of over a
> gigabyte was predicted in tcpros. that
> seems highly unlikely, so I'll assume
> protocol synchronization is lost.

> [ERROR] [1520594758.451553409]
> [/rosout]: Exception thrown when
> deserializing message of length [0]
> from [/rosout_test]: Buffer Overrun

> [ERROR] [1520594758.451622886]
> [/rosout]: Exception thrown when
> deserializing message of length [39]
> from [/rosout_test]: Buffer Overrun

This python ROS node reproduces the problem:

    #!/usr/bin/env python
    
    import rospy
    from std_msgs.msg import Float64
    from threading import Thread
    
    def topic_callback(msg):
        pass
    
    def my_topic_thread():
        """endless subscribing and unsubscribing to a list of topics"""
    
        topic_names = ["/some/topic{}".format(i) for i in range(0, 100)]
    
        while not rospy.is_shutdown():
            rospy.loginfo("subscribing")
            subs = []
            for topic in topic_names:
                subs.append(rospy.Subscriber(topic, Float64, topic_callback))
                rospy.sleep(0.001)
    
            rospy.loginfo("unsubscribing")
            for sub in subs:
                sub.unregister()
    
    if __name__ == '__main__':
    
        rospy.init_node('demo')
    
        topic_thread = Thread(target=my_topic_thread)
        topic_thread.start()
    
        while not rospy.is_shutdown():
            # This logging fails intermittently, because the nodes
            # topic list is changed during logging
            # Error: protocol synchronization is lost
            rospy.loginfo("dummy message")
            rospy.sleep(0.001)
    
        topic_thread.join()

This issue was originally raised on ros answers: https://answers.ros.org/question/284911

kinetic, ubuntu 16.04

gavanderhoorn:
From ROS Answers Q&A: [#q284911](https://answers.ros.org/question/284911).

krixkrix:
The problem can be solved by copying the topics list before creating the Log object, though I don't know if this is the proper way.
Patch:

```
[...]

PR #1416, opened 2018-05-31
Fix Issue 1346 - Make a copy of the topic list

pbaughman:
Proposed fix from https://github.com/ros/ros_comm/issues/1346

#### Description
get_topics() returns a set, which is iterated by the Log constructor.  Make a copy of the set (atomic in cpython!) so nobody adds or removes something while the Log constructor iterates over it

A more optimized fix might be to use an immutable data structure inside the topic manager.  I imagine that in the general case the topic list changes a lot less often than than calls are made to _rosout.  It might be better to eat the overhead on topic add/remove instead of adding a copy operation to every _rosout call.

### DIFF
--- a/clients/rospy/src/rospy/topics.py
+++ b/clients/rospy/src/rospy/topics.py
@@ -1348,7 +1348,7 @@ class _TopicManager(object):
         @return: list of topic names this node subscribes to/publishes
         @rtype: [str]
         """                
-        return self.topics
+        return self.topics.copy()
     
     def _get_list(self, rmap):
         return [[k, v.type] for k, v in rmap.items()]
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Moving the `copy` call into the [get_topics()](https://github.com/ros/ros_comm/blob/e1d4a1602b14f47e6543646924520b0e4879d056/clients/rospy/src/rospy/topics.py#L1351) function seems to be a "better" option (applies to all callers, doesn't need to change any other code). Can you update the patch accordingly.

pbaughman:
Aye - if you're comfortable with a slightly bigger footprint, that's probably a more robust fix

clients/rospy/src/rospy/topics.py
```diff
@@ -1348,7 +1348,7 @@ def get_topics(self):
         @return: list of topic names this node subscribes to/publishes
         @rtype: [str]
```
pbaughman, line 1349: This return type is a bit of a fib, I think, unless I don't understand the documentation syntax.  Problem?

tfoote:
This would be a good candidate for backporting. It's known to cause issues all the way back to indigo and is a pure bug fix.

dirk-thomas:
Thank you for the updated patch.

PR #1421, opened 2018-06-06
Fix some errors in some probably not frequented code paths kinetic

awesomebytes:
As https://github.com/ros/ros_comm/pull/1415


dirk-thomas:
Thanks for the patch!

Ignoring CI since it is due to a flaky test.

PR #1424, opened 2018-06-06
import rosbag locally instead of at module level

mikaelarguedas:
This is addressing https://github.com/ros2/ros1_bridge/issues/118#issuecomment-394909837 reducing the dependency overhead for downstream projects wanting to import `rosmsg`

### DIFF
--- a/tools/rosmsg/src/rosmsg/__init__.py
+++ b/tools/rosmsg/src/rosmsg/__init__.py
@@ -51,7 +51,6 @@ import genmsg
 from genpy.dynamic import generate_dynamic
 
 import roslib.message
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@ros/ros_comm-maintainers: It looks like the test failure is unrelated and that this test is flaky (as it fails sometimes on the dev job and other PR jobs), please let me know if it's not the case

mikepurvis:
Looks reasonable to me, thanks!

dirk-thomas:

dirk-thomas:
Thanks for the patch.

PR #1430, opened 2018-06-10
[roswtf] 1) Switch from OptParse to ArgParse. 2) arg parsing portion as an option, not a requirement

130s:
2 changes committed separately.

- Replace with `argparse` - `OptParse` is deprecated since Python 2.7 (ref: https://docs.python.org/2.7/library/optparse.html).
- In Python API there's no equivalent to sh' `roswtf` command that serves as a single point of triggering `roswtf` via API, not via executable.

  sh script `roswtf` calls `roswtf_main()` that internally calls [_roswtf_main()](https://github.com/ros/ros_comm/blob/24e45419bdd4b0d588321e3b376650c7a51bf11c/utilities/roswtf/src/roswtf/__init__.py#L93), which takes in stdin. This still works ok when you have a control over the arguments to pass to roswtf, but doesn't work when you don't. E.g. `catkin run_tests` seems to pass some args while internally running `nosetests`, which the parse module takes in and thus passed to roswtf and ends up causing an error.

### DIFF
--- a/utilities/roswtf/src/roswtf/__init__.py
+++ b/utilities/roswtf/src/roswtf/__init__.py
@@ -93,7 +93,7 @@ def roswtf_main():
 def _roswtf_main():
     # performance optimization
     rospack = rospkg.RosPack()
-    all_pkgs = rospack.list()
+    all_packages = rospack.list()
 
     import argparse
     parser = argparse.ArgumentParser(usage="usage: roswtf [launch file]", description="roswtf is a tool for verifying a ROS installation and running system. Checks provided launchfile if provided, else current stack or package.")
@@ -117,16 +117,19 @@ def _roswtf_main():
                         help="launch files to check. 0...n files can be passed")
     #TODO: --all-pkgs option
     args = parser.parse_args()
-    roswtf_node(args)
+    roswtf_node(args.all_packages, args.disable_plugins, args.launch_files, args.offline)
 
 
-def roswtf_node(args):
+def roswtf_node(all_packages=False, disable_plugins=False, launch_files=[], offline=False):
     """
-    @type args: argparse.Namespace
+    @type all_packages: 
+    @type disable_plugins:
+    @type launch_files: 
+    @type offline: 
     """
     launch_files = names = None
-    if args.launch_files:
-        launch_files = args.launch_files
+    if launch_files:
+        launch_files = launch_files
         if 0:
             # disable names for now as don't have any rules yet
             launch_files = [a for a in args if os.path.isfile(a)]
@@ -143,7 +146,7 @@ def roswtf_node(args):
     import roswtf.roslaunchwtf
     import roswtf.stacks    
     import roswtf.plugins
-    if not args.disable_plugins:
+    if not disable_plugins:
         static_plugins, online_plugins = roswtf.plugins.load_plugins()
     else:
         static_plugins, online_plugins = [], []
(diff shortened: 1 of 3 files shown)


130s:
Tests seem to be unstable, some jobs pass while others go unstable, but I can't tell what part of this PR causes it. I'd appreciate if someone can help.

Also, having seen some CI jobs passing, please review.

130s:

utilities/roswtf/src/roswtf/__init__.py
```diff
@@ -91,33 +91,42 @@ def roswtf_main():
...
+                        help="launch files to check. 0...n files can be passed")
     #TODO: --all-pkgs option
-    options, args = parser.parse_args()
-    if args:
-        launch_files = args
+    args = parser.parse_args()
+    roswtf_node(args)
+
+
+def roswtf_node(args):
```
dirk-thomas: Why should this be a public API? A caller will need to know internal knowledge (what arguments need to be set on the passed `args`) which makes it hard to use. Therefore I think this should either be private or expose a signature with specific keyword arguments.

dirk-thomas: @130s Friendly ping.

130s: > Therefore I think this should either be private or expose a signature with specific keyword arguments.

Thanks, I did the latter of your suggestion.

utilities/roswtf/src/roswtf/__init__.py
```diff
@@ -91,33 +91,45 @@ def roswtf_main():
...
+                        help="launch files to check. 0...n files can be passed")
     #TODO: --all-pkgs option
-    options, args = parser.parse_args()
-    if args:
-        launch_files = args
+    args = parser.parse_args()
+    roswtf_node(args.all_packages, args.disable_plugins, args.launch_files, args.offline)
+
+
+def roswtf_node(all_packages=False, disable_plugins=False, launch_files=[], offline=False):
```
dirk-thomas, line 123: Why is this function named "_node"?

utilities/roswtf/src/roswtf/__init__.py
```diff
@@ -91,33 +91,45 @@ def roswtf_main():
...
+    """
+    @type all_packages: 
+    @type disable_plugins:
+    @type launch_files: 
+    @type offline: 
+    """
+    launch_files = names = None
```
dirk-thomas, line 130: Isn't this overwriting the function parameter?

utilities/roswtf/src/roswtf/__init__.py
```diff
@@ -91,33 +91,45 @@ def roswtf_main():
...
+    if launch_files:
+        launch_files = launch_files
```
dirk-thomas, line 132: That looks like a no-op?

utilities/roswtf/src/roswtf/__init__.py
```diff
@@ -91,33 +91,45 @@ def roswtf_main():
...
                       help="only run offline tests")
 
+    parser.add_argument("launch_files",  nargs='*',
+                        help="launch files to check. 0...n files can be passed")
     #TODO: --all-pkgs option
-    options, args = parser.parse_args()
-    if args:
-        launch_files = args
+    args = parser.parse_args()
+    roswtf_node(args.all_packages, args.disable_plugins, args.launch_files, args.offline)
```
dirk-thomas, line 120: Only `args.all_packages` is passed to the function but not `all_packages`.

dirk-thomas (commented):
The amount of significant issues I noticed on a first look makes me believe that you haven't tried to run this code.

Please thoroughly test your changes to ensure this doesn't cause regressions.

dirk-thomas:

PR #1434, opened 2018-06-15
message_filters: call Subscriber::unsubscribe() in destructor

cwecht:
This makes it obvious, which function is actually called.

In Addition, this fix a clang stytic analyser warning.


dirk-thomas:
Thank you for the patch.

PR #1438, opened 2018-06-19
remove messages that are newer than the newly added message

christian-rauch:
In some circumstances, when replaying a log in loop mode (-l) the synchronisation blocks because messages jump back in time when restarting the log. This commit removes messages that are newer than the newly added message, enabling to synchronise looped log files. This could optionally enabled via a parameter that gives the maximum tolerated backward jump in time.

### DIFF
--- a/utilities/message_filters/src/message_filters/__init__.py
+++ b/utilities/message_filters/src/message_filters/__init__.py
@@ -57,7 +57,7 @@ class SimpleFilter(object):
             cb(*(msg + args))
 
 class Subscriber(SimpleFilter):
-
+    
     """
     ROS subscription filter.  Identical arguments as :class:`rospy.Subscriber`.
 
@@ -167,7 +167,7 @@ class Cache(SimpleFilter):
         if not self.cache_times:
             return None
         return self.cache_times[0]
-
+        
     def getLast(self):
         if self.getLastestTime() is None:
             return None
@@ -258,7 +258,7 @@ class ApproximateTimeSynchronizer(TimeSynchronizer):
         # clear all buffers if jump backwards in time is detected
         now = rospy.Time.now()
         if now < self.last_added:
-            rospy.loginfo("ApproximateTimeSynchronizer: Detected jump back in time. Clearing  buffer.")
+            rospy.loginfo("ApproximateTimeSynchronizer: Detected jump back in time. Clearing buffer.")
             for q in self.queues:
                 q.clear()
         self.last_added = now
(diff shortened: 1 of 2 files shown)


tfoote (changes requested):
Logic to clear the filter when looping bag playback might make sense but I don't think that this is the right approach. It assumes all messages arrive in order except for a loop which is not guarenteed. There could easily be multiple publishers and publishers with higher latency.

There is logic in tf2_ros to reset it's buffers if the clock jumps backwards. I would suggest looking at that sort of logic as that is much more clearly a time discontinuity and behavior should change. Rather than possibly a delayed message arriving into a synchronizer.

christian-rauch:
I agree that `/clock` is a more sensible reference for the time than the timestamped messages. I implemented the buffer clearing logic like in https://github.com/ros/geometry2/pull/68 (I think this is the logic you were referring to).

tfoote:
That looks better to me. Two comments though.

With the additional use of `now()` the [mock initialization](https://github.com/ros/ros_comm/blob/melodic-devel/utilities/message_filters/test/test_approxsync.py#L101) of the time in the test needs to be moved earlier.

And the logic should provide a warning about the reset. Otherwise this could be triggered regularly and all data would potentially be silently dropped.

christian-rauch:
The message about the reset is printed as info, not as warning. From my point of view, clearing the buffer on jumps backward in time is an expected behaviour while the alternative of blocking the synchronisation (by not clearing the buffers) is unexpected.

tfoote:
info output is fine too. It looks like this is being triggered in the current tests and causing a failure. And related to that we should add a unit test to explicitly test that this is working and won't regress in the future.

christian-rauch:
The test fails because it not only scrambles the sequence of messages, but also the clock time that is supposed to be continuous: `rospy.rostime._set_rostime(rospy.Time(i+0.05))` where `i` is randomly sampled. Given that we assume that clock time is continuous (but header stamps can be unordered), does this test of synchronising header-less messages sense? I.e. by definition, messages without header do not have an order.

christian-rauch:
@tfoote Does this look all right to you now?

tfoote (approved):

dirk-thomas:

dirk-thomas:

dirk-thomas:
The C++ equivalent will likely only be part of the next ROS distro (see https://github.com/ros/ros_comm/pull/1518#issuecomment-459550703) I am not sure if you still want to wait with this patch to release them in sync or merge this into Melodic now?

christian-rauch:
We can merge this now. Then at least people which are using the python interface will benefit from this.

dirk-thomas:
Sounds good to me. Thanks for the patch.

PR #1439, opened 2018-06-20
Fixed issue occuring during alternating calls of getParamCached and setParam

cwecht:
fff8901 shows an issue which is likely a race condition in the interaction of the nodes parameter cache and the parameter server.


cwecht:
884152027e175d9ab912ebe7aef2ca5f23b47e0b   provides a fixed for the test introduced in fff89012ca1b6a77edf98880fecc048433d47f07.

I'm not sure if I understand `Registratons`  correctly. `__get_item__` returns a list of pairs of strings. If I understand the documentation correctly this list is either empty or

cwecht:
To get this clear. The race, which this PR fixes, is this:

1. The Node A writes parameter 'param' to **1** calling `NodeHandle::setParam`: 'param' is set in the nodes cache and is transmitted to the parameter server.
2. A reads 'param' with `NodeHandle::getParamCached` and gets the value **1** out of the cache.
3. A writes 'param' again, but this time the value **2**. The value is written to the cache and transmitted to the server.
4. The parameter server notifys all nodes, which have been using 'param' , about the parameter change from step 1. In A's cache the value of 'param' is set to **1**.
5. A read 'param' from the cache and gets value **1** (but expected **2**).

The example chosen in fff8901 might be contrived, but it is a simple way to reproduce this issue.

cwecht:
Ok, as far as i can tell, the test for the rosmaster fails now, because the tests expects `compute_param_updates` to have a pretty wide contract. The documentation of `Registrations` says that its `map`-members is a map from keys to lists of pairs of the form `(caller_id, caller_api)` ([L162](https://github.com/ros/ros_comm/blob/89ca9a7bf288b09d033a1ab6966ef8cfe39e1002/tools/rosmaster/src/rosmaster/registrations.py#L162)).

But test_rosmaster_paramserver.py tests `compute_param_updates` passing a map from keys to caller_ids.

I find this rather confusing. Also I do not see any purpose of this rather than making the test easier.

cwecht:
Maybe there should be new tests for `compute_param_updates` but I'd like to get some feedback about this patch in advance.

dirk-thomas:

cwecht:
Well, two changes, which are pretty unrelated to the original purpose of this PR, fixed the builds now.
@dirk-thomas I'm willing to create separate PRs for this unrelated changes if it makes it easier for you.

cwecht:
@dirk-thomas As far as I am concerned this can be merged now.

dirk-thomas:
Thanks for the patch.

PR #1440, opened 2018-06-20
Throw BagException when disk is full and return from run() with error…

VictorLamoine:
Currently when the disk is full a `ROS_ERROR` message is sent, writing is disabled (see https://github.com/ros/ros_comm/blob/melodic-devel/tools/rosbag/src/recorder.cpp#L649-L651). The problem is that the recording thread does not exit in that case.

This changes the behavior:
- If disk is below the minimum an exception is thrown
- The exception is catched in the recording thread
- The recording thread exits with an error code (`1`)
- The  `run` member functions joins the recording thread and return the exit code

From a user perspective now you can run:
```bash
if (!recorder->run())
  // Handle the fact that we cannot record anymore
```

Note: in `doRecordSnapshotter` no disk space checking is done, I did not add it.

### DIFF
--- a/tools/rosbag/scripts/trim.py
+++ b/tools/rosbag/scripts/trim.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2018, Victor Lamoine - Institut Maupertuis
+# All rights reserved.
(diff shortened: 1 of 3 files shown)


VictorLamoine:
Not sure why the test failed.

PR #1443, opened 2018-06-25
trace instrumentation for ros_comm

iluetkeb:
This is a replacement for the previous PR on kinetic-devel (https://github.com/ros/ros_comm/pull/1198)


iluetkeb:
Note that building on bionic requires an update of the tracetools package, which is already pushed but might not yet be in.

iluetkeb:
As expected, failed by https://github.com/ros/rosdistro/pull/18261

dirk-thomas:

iluetkeb:
Missed a spot that's only built during testing.

iluetkeb:
The Bionic error seems unrelated.

Debian Stretch uses boost version 1.62, which apparently already introduced the split with members. I wasn't aware of that and will modify the preprocessor macro.

iluetkeb:
tracetools 0.2.1 now implements the version number check from 1.62 onwards (I checked on boost.org, that's the first version which has the change), which should fix the Debian stretch build. See https://github.com/ros/rosdistro/pull/18275 and https://github.com/ros/rosdistro/pull/18274

iluetkeb:
Okay, the tracetools build on stretch passed (http://build.ros.org/job/Mbin_ds_dS64__tracetools__debian_stretch_amd64__binary/), so we could try this again.

iluetkeb:
I honestly don't understand the Jenkins log. It builds fine and the error messages during testing seem normal (at least, they are not classified as errors in the test report). 

At the end, there is a problem with publishing the test results, but there's so many other errors related to publishing the status and stuff, that I don't know whether that is serious -- and it doesn't seem related to my changes anyway.

iluetkeb:
@dirk-thomas Could have a look at what's the problem? I don't see any issues that could explain why the checks failed. The builds compile, and while CI declares them "unstable", this is not due to failing tests.

iluetkeb:
@mikepurvis Do you have an idea?

dirk-thomas:

iluetkeb:
Uh... Now we get lots of test failures? When last time we didn't get *any*?

Also, the test results are partially odd and partially ridiculous. Odd things are the timeouts that occur. This could be investigated further.

Things like a Python module complaining that it cannot import "core" from "."? This patch has no Python code whatsoever, how could it break things there?

I can't look at this in more detail before Wednesday, but if anybody has any idea where these issues could come from, it would be much appreciated.

dirk-thomas:
> Now we get lots of test failures? When last time we didn't get any?

The failing tests were unrelated to this patch and should be fixed in the next CI build.

iluetkeb:
Hmm, at the end there is about a 2 minute gap, and then "FATAL: command execution failed". No error message as to why, but maybe some timeout.

Is there an issue with your CI server, or is this more related to other changes on melodic?

kveretennicov:

...on the off-chance that these were transient CI issues.

iluetkeb:
It works for me on my box and on my colleague's boxes. The errors here are puzzling.

If upstream has no interest in investigating, that's okay, let's leave it then.

dirk-thomas:
> If upstream has no interest in investigating, that's okay, let's leave it then.

I will reopen the ticket for now since it is not a lack of interest but simply a lack of resources that the patch hasn't received any attention.

sloretz:
Thank you for the PR!

christophebedard:
For anyone looking at this in the future, the spiritual successor to this in ROS 2 is `ros2_tracing`: https://github.com/ros2/ros2_tracing

PR #1444, opened 2018-06-26
Fix rosbag's internal _read_message API

### LINKED ISSUE #1432: rosbag - _read_message in bag.py throws exception
since melodic, the function "seek_and_read_message_data_record" has different numbers of parameters for the two classes "_BagReader102_Indexed" and "_BagReader200". Call "_read_message" directly will throw an exception when the reader is _BagReader200.

lalten:
This PR fixes the internal rosbag API used by rqt_bag, which was broken with PR https://github.com/ros/ros_comm/pull/1372

This closes https://github.com/ros/ros_comm/issues/1432 and https://github.com/ros-visualization/rqt_bag/issues/27

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -1291,12 +1291,12 @@ class Bag(object):
 
         return sum((h.uncompressed_size for h in self._chunk_headers.values()))
 
-    def _read_message(self, position, raw=False):
+    def _read_message(self, position, raw=False, return_connection_header=None):
         """
         Read the message from the given position in the file.
         """
         self.flush()
-        return self._reader.seek_and_read_message_data_record(position, raw)
+        return self._reader.seek_and_read_message_data_record(position, raw, return_connection_header)
 
     # Index accessing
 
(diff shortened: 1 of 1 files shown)


lalten:
@dirk-thomas I don't understand why the checks failed. It seems like it doesn't have to do anything with the code. Is the Github API is broken?

VictorLamoine:
@dirk-thomas Sorry to ping but this issue is quite annoying, it forces to compile `ros_comm` from sources on every machine where you need `rqt_bag`

Either this merge request or https://github.com/ros/ros_comm/pull/1445 does the job.

VictorLamoine:
@dirk-thomas @tfoote See previous message
I guess you guys are at the beach :beach_umbrella: :sun_with_face:

dirk-thomas:
After reviewing the original (already merged) patch (#1372) as well as both proposed fixes (#1445 and this one) I actually thing both should be merged. Additional the API compatibility which was broken in the merged patch should be restored.

I created #1473 with all fixes combined.

dirk-thomas:
Closing in favor of #1473.

dirk-thomas:
Thank you for contributing the patch - the commits are part of the combined PR.

PR #1457, opened 2018-07-13
Add a default __hash__ function for the Edge class

kartikmohta:
In Python3, using rqt_graph with some compressed image topics present gave the following error:
```python
PluginHandlerDirect._restore_settings() plugin "rqt_graph/RosGraph#0" raised an exception:
  File "/opt/ros/melodic/lib/python3.6/site-packages/qt_gui/plugin_handler_direct.py", line 116, in _restore_settings
    self._plugin.restore_settings(plugin_settings_plugin, instance_settings_plugin)
  File "/opt/ros/melodic/lib/python3.6/site-packages/rqt_graph/ros_graph.py", line 217, in restore_settings
    self._refresh_rosgraph()
  File "/opt/ros/melodic/lib/python3.6/site-packages/rqt_graph/ros_graph.py", line 246, in _refresh_rosgraph
    self._update_graph_view(self._generate_dotcode())
  File "/opt/ros/melodic/lib/python3.6/site-packages/rqt_graph/ros_graph.py", line 280, in _generate_dotcode
    hide_dynamic_reconfigure=hide_dynamic_reconfigure)
  File "/opt/ros/melodic/lib/python3.6/site-packages/rqt_graph/dotcode.py", line 876, in generate_dotcode
    hide_dynamic_reconfigure=hide_dynamic_reconfigure)
  File "/opt/ros/melodic/lib/python3.6/site-packages/rqt_graph/dotcode.py", line 698, in generate_dotgraph
    nt_nodes, edges, image_nodes = self._accumulate_image_topics(nt_nodes, edges, node_connections)
  File "/opt/ros/melodic/lib/python3.6/site-packages/rqt_graph/dotcode.py", line 553, in _accumulate_image_topics
    image_topic_edges_in.update(node_connections[n2].incoming)
TypeError: unhashable type: 'Edge'
```

### DIFF
--- a/tools/rosgraph/src/rosgraph/impl/graph.py
+++ b/tools/rosgraph/src/rosgraph/impl/graph.py
@@ -213,6 +213,7 @@ class Edge(object):
         return "%s->%s"%(self.start, self.end)
     def __eq__(self, other):
         return self.start == other.start and self.end == other.end
+    __hash__ = object.__hash__
 
 def edge_args(start, dest, direction, label):
     """


dirk-thomas:
Can you please provide a reproducible example? Maybe this should be addressed in `rqt_graph` instead.

I am not sure if an object id based hash makes sense for the `Edge` class. Often users expect the hash to be at least "related" to the `__eq__` comparison which wouldn't be the case.

kartikmohta:
Anything that publishes using `image_transport` causes the issue. The simplest example is:
```cpp
#include <ros/ros.h>
#include <image_transport/image_transport.h>

int main(int argc, char *argv[])
{
  ros::init(argc, argv, "image_transport_example");
  ros::NodeHandle nh("~");

  image_transport::ImageTransport it(nh);
  image_transport::Publisher pub = it.advertise("image", 2);

  ros::spin();
  return 0;
}
```

I suppose handling it in `rqt_graph` is a better approach though that probably requires much more work.

Should I change the hash function it to make it "match" the `__eq__` function?

dirk-thomas:
> Should I change the hash function it to make it "match" the `__eq__` function?

That is not recommended either. Basically when you add an edge into a dictionary the hash is computed to determine where the key should be inserted. If you afterwards change any of the fields of the edge the hash would change but the key has already been inserted under the "old" hash. To avoid that kind of mismatch the attributes contributing to the hash should be considered "const" which wouldn't necessarily be the case for the fields of an edge.

Therefore I would lean towards addressing this in `rqt_graph` which "knows" how it is using the edges and what it can ensure.

kartikmohta:
BTW, this change matches what Python 2 does by default, so maybe we need to make the Edge class "unhashable" in Python 2 as well?

dirk-thomas:
> maybe we need to make the Edge class "unhashable" in Python 2 as well?

Maybe - chances are it might break existing code out there though...

kartikmohta:
Yes I agree, so can we pretty please unbreak it for Python3 (with a note/warning)? :slightly_smiling_face:

dirk-thomas:
> so can we pretty please unbreak it for Python3 (with a note/warning)?

Do you mean as-is - using `object.__hash__`? I am just worried that is not what users expect to happen when they have two edges with the same field values...

Is there a reason not to address it in `rqt_graph` (which knows how it using the class and how it handles edges with the same values but different instances)?

kartikmohta:
No particular reason than me being lazy I guess. I'll close this and try to fix it in `rqt_graph`.
Thanks for taking the time to discuss! :slightly_smiling_face:

Issue #72, opened 2013-01-08
roslaunch does not check for localhost ROS_MASTER_URI when doing remote launches


130s:
Migrated from https://code.ros.org/trac/ros/ticket/3801
Reported by:    kwc     Owned by:   kwc
Priority:   major 
Attachment available: No
Last modified ago: Oct 2011

---

log files from a ros user confirm that a remote roslaunch can be passed a localhost URI, which is invalid. The parent roslaunch should never send this value.

Timple:
I experienced this issue today and it takes quite some debugging. I ended up on this [answer](https://answers.ros.org/question/12434/using-roslaunch-to-run-distributed-systems-over-ssh/?answer=18423#post-id-18423). 

Would simply adding a warning would be do-able to close this issue?

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #1463, opened 2018-07-24
Error when hostnames do not match on remote launch

Timple:
Solves #72

### DIFF
--- a/tools/roslaunch/src/roslaunch/remote.py
+++ b/tools/roslaunch/src/roslaunch/remote.py
@@ -39,6 +39,7 @@ Integrates roslaunch remote process launching capabilities.
 import logging
 import socket
 import time
+from urlparse import urlparse
 
 import rosgraph.network as network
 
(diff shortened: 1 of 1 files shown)


mikepurvis:
This looks reasonable to me; can't think of a situation where this would result in a false positive. 👍

Timple:
One thing that popped up in my mind: Would this scenario still work?:

computer1: roscore
computer2: subscribed to computer1-core and roslaunch with remote nodes on computer 3

tfoote:
I think that for #72 localhost has specific meaning that it won't talk externally. Otherwise you may have multihoned computers with different hostname on different interfaces etc and I'm not sure that the generic check is as simple as this proposed change.

Timple:
In that case, would comparing against just 'localhost' help?
Other cases like 127.0.0.1 might be giving identical issues, but checking against localhost only might make sense because this is the default of ROS_MASTER_URI. And therefore the most likely to cause issues with users.

Timple:
A colleague ran into the same issue again and it takes quite some debugging to find that your own ROS_MASTER_URI should not contain localhost (the default).

Is anything wrong with my implementation? Or do you like me to rewrite to throw a warning if localhost is used in combination with remote roslaunch?

130s:

130s:
I don't yet have a deep understanding of the issue #72 (its author was me but I just copied from an older repo back then). Especially these days in a containerized world where each container might be designed as self-contained ROS image, I don't know how this change affects or not. That said in a non-containerized, pure multi-host single ROS Master environment this seems reasonable to me.

dirk-thomas:
Please retarget the latest default branch (currently `melodic-devel`) in order to be considered for review and merge.

dirk-thomas:
@Timple @130s Closing due to no response. Please consider to open a new pull request targeting `noetic-devel`.

PR #1464, opened 2018-07-24
Add hasStarted() const to Timer API

bmagyar:
To assert if a timer is running at the moment, one can hack their way around by using `isValid` and `setDuration` to invalid ones.

I'm proposing the addition of a simple const accessor to the internal `started_` field. 

Should I add a simple test file as well?
Can this be cherry-picked to kinetic-devel too?

### DIFF
--- a/clients/roscpp/include/ros/timer.h
+++ b/clients/roscpp/include/ros/timer.h
@@ -71,6 +71,7 @@ public:
    */
   void setPeriod(const Duration& period, bool reset=true);
 
+  bool hasStarted() const { return impl_->hasStarted(); }
   bool isValid() { return impl_ && impl_->isValid(); }
   operator void*() { return isValid() ? (void*)1 : (void*)0; }
 
@@ -98,6 +99,7 @@ private:
     Impl();
     ~Impl();
 
+    bool hasStarted() const;
     bool isValid();
     bool hasPending();
     void setPeriod(const Duration& period, bool reset=true);
--- a/clients/roscpp/src/libros/timer.cpp
+++ b/clients/roscpp/src/libros/timer.cpp
@@ -42,6 +42,11 @@ Timer::Impl::~Impl()
   stop();
 }
 
+bool Timer::Impl::hasStarted() const
+{
+  return started_;
+}
+
 bool Timer::Impl::isValid()
 {
   return !period_.isZero();


mikepurvis:
LGTM— have wanted this in the past. Slight preference for `isStarted` over `hasStarted`, as there are many more "bool isXYZ()" methods in roscpp. However, the timer already has `hasPending` anyway, so it doesn't really matter.

Either way, 👍

bmagyar:
I am open to changing that. I was a little reluctant to add the const declaration to the function. Although that feels like the correct thing to do, I saw that the current API doesn't have any. Changing it would be breaking API (in the strict sense) so I didn't want to touch anything else.

bmagyar:
@clalancette could we get a rebuild//rerun of tests on Debian Stretch?

clalancette:

mikepurvis:
Adding const should always be safe, shouldn't it?

REP9 does list "changing the const/volatile qualifiers of a member function" as a thing that causes ABI breakage, but that is surprising to me— my impression is that constness is enforced at compilation time, and is basically invisible to the linker. Apparently not.

bmagyar:
The `const` modifier changes the signature of a function, which even causes API breaks AFAIK.

mikepurvis:
It might be an ABI break, but it's definitely not an API break— it's making the method _more_ permissive than previously, since it was previously only callable on non-const instances of the class, but is now callable on const instances as well. So there's no existing usage which will break at compile time by making a method const.

bmagyar:
I based my guess on how gdb often appends const to the end of function
names in the ASM but you are absolutely right in your reasoning.

bmagyar:
@clalancette could you please point us to someone who could review/merge this?

dirk-thomas:

bmagyar:
@tfoote can we cherry-pick this to Kinetic too?

dirk-thomas:
The new API is currently not selected for backporting into Lunar (see #1477). As such it wouldn't be backported to Kinetic (since that happens from the Lunar branch).

Commonly we don't backport new features. If you think this would be very important (and can't be worked around in user code) please comment on the referenced ticket to request it being backported to Lunar.

dirk-thomas:
FYI #1541.

Issue #1461, opened 2018-07-23
SROS Docker and Ubuntu install fails due to missing dependencies etc.


TomL92:
Hello,

While trying to follow the official SROS Tutorials ([http://wiki.ros.org/SROS/Tutorials](http://wiki.ros.org/SROS/Tutorials)), I had some issues when trying to start the `sroskeyserver`. The same errors also occur when running the prebuilt Docker container, so i suppose, that the problem is due to some inconsistencies in the SROS repository or its dependencies.

At first i've pulled the docker and tried to run the command from the install page:
<details> 
  <summary>command (click to expand)</summary>

  ```bash
  docker run --rm -it \
    osrf/sros \
    bash -c "source /ros_entrypoint.sh && \
        sroskeyserver & \
        sleep 3 && \
        sroslaunch rospy_tutorials talker_listener.launch"
  ```
</details>

Somehow this runs into the following problem:

<details>
  <summary> Error (click to expand) </summary>

  ```
  Traceback (most recent call last):
  File "/root/sros_catkin_ws/install/bin/sroskeyserver", line 2, in <module>
    import sros
  File "/root/sros_catkin_ws/install/lib/python2.7/dist-packages/sros/__init__.py", line 7, in <module>
    import rosgraph.sros_consts as sros_consts
  ImportError: No module named rosgraph.sros_consts
  Traceback (most recent call last):
  File "/root/sros_catkin_ws/install/bin/sroslaunch", line 2, in <module>
    import sros
  File "/root/sros_catkin_ws/install/lib/python2.7/dist-packages/sros/__init__.py", line 7, in <module>
    import rosgraph.sros_consts as sros_consts
  ImportError: No module named rosgraph.sros_consts
```

</details>

Therefore I checked the ros_comm package, which is not available in the Docker container as it is ignored by using an `AMENT_IGNORE` file.
<details>
  <summary>see here (click to expand)</summary>

  ```bash
  ./src/ros_comm/ros_comm/AMENT_IGNORE
  ./src/ros_comm/clients/AMENT_IGNORE
  ./src/ros_comm/tools/rostopic/AMENT_IGNORE
  ./src/ros_comm/tools/rosmsg/AMENT_IGNORE
  ./src/ros_comm/tools/roslaunch/AMENT_IGNORE
  ./src/ros_comm/tools/rosbag/AMENT_IGNORE
  ./src/ros_comm/tools/rosparam/AMENT_IGNORE
[...]

PR #1465, opened 2018-07-24
Remove old AMENT_IGNORE

ruffsl:
Fixes https://github.com/ros/ros_comm/issues/1461 and resolves [SROS Docker and Ubuntu install fails due to missing dependencies etc.](https://answers.ros.org/question/298385/sros-docker-and-ubuntu-install-fails-due-to-missing-dependencies-etc/)


dirk-thomas:
This the `sros` branch is considered "in progress" / "unreviewed" (see #858) I will just go ahead and merge the proposed patch.

PR #1466, opened 2018-07-24
Fix issues when built or run on Windows

johnsonshih:


tools/rosgraph/src/rosgraph/roslogging.py
```diff
@@ -59,7 +59,12 @@ def findCaller(self, dummy=False): # Dummy second arg to match Python3 function
         f = inspect.currentframe()
         if f is not None:
             f = f.f_back
+        else:
+            return file_name, lineno, func_name
```
dirk-thomas: Can you please provide some context why this change was necesary? The same for the `break` a few lines below.

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
```diff
@@ -100,9 +105,9 @@ XmlRpcDispatch::work(double timeout)
   while (_sources.size() > 0) {
 
     // Construct the sets of descriptors we are interested in
-    const unsigned source_cnt = _sources.size();
-    pollfd fds[source_cnt];
-    XmlRpcSource * sources[source_cnt];
+    unsigned source_cnt = _sources.size();
```
dirk-thomas: Why can't this variable be `const` anymore?

tools/rosgraph/src/rosgraph/roslogging.py
johnsonshih: without this, roscore won't run on Windows.  I need to investigate more about this change, will revert it from this PR for now.

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
johnsonshih: on MSVC compiler will complain that _sources.size() is not an constant, so I need to remove the const.  I'll investigate to see if we can add it back

johnsonshih: after switching to use std::vector, I'm able to declare source_cnt as a const variable, will add it back.

dirk-thomas:
Thank you for the patch. The failing test is known to be flaky so this could be merged.

Do you want to keep iterating on this for the logging (if necessary?) or address that in a separate PR?

johnsonshih:
I will do a separate PR for the change in logging.

PR #1468, opened 2018-07-28
add option to hide summary from roslaunch output.

newten82:
On large projects the summary output can slow down the launch of a system. The --no-summary option hides the summary to get things running faster.

### DIFF
--- a/tools/roslaunch/src/roslaunch/__init__.py
+++ b/tools/roslaunch/src/roslaunch/__init__.py
@@ -161,6 +161,9 @@ def _get_optparse():
     parser.add_option("-v", action="store_true",
                       dest="verbose", default=False,
                       help="verbose printing")
+    parser.add_option("--no-summary", action="store_true",
+                      dest="no_summary", default=None,
+                      help="hide summary printing")
     # 2685 - Dump parameters of launch files
     parser.add_option("--dump-params", default=False, action="store_true",
                       dest="dump_params",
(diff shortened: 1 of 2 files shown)


tools/roslaunch/src/roslaunch/__init__.py
```diff
@@ -161,6 +161,9 @@ def _get_optparse():
     parser.add_option("-v", action="store_true",
                       dest="verbose", default=False,
                       help="verbose printing")
+    parser.add_option("--no-summary", action="store_true",
+                      dest="no_summary", default=None,
```
dirk-thomas, line 165: Please use `False` as the default value for a boolean parameter.

tools/roslaunch/src/roslaunch/parent.py
```diff
@@ -73,7 +73,7 @@ class ROSLaunchParent(object):
     """
 
     def __init__(self, run_id, roslaunch_files, is_core=False, port=None, local_only=False, process_listeners=None,
-            verbose=False, force_screen=False, is_rostest=False, roslaunch_strs=None, num_workers=NUM_WORKERS, timeout=None):
+            verbose=False, no_summary=None, force_screen=False, is_rostest=False, roslaunch_strs=None, num_workers=NUM_WORKERS, timeout=None):
```
dirk-thomas, line 76: To avoid double negative maybe name the new keyword argument `show_summary` with a default value of `True`?

cwecht:
@dirk-thomas  As this PR has been replaced by #1468, this PR can be closed now.

dirk-thomas:
> As this PR has been replaced by #1468, this PR can be closed now.

I assume you intended to refer to #1655.

Issue #1427, opened 2018-06-08
topic_tools transform doesn't use node namespace for input topic, requires absolute topic name


lucasw:
```
rostopic pub /foo/bar std_msgs/Float64 "data: 1.0" -r 1
rosrun topic_tools transform bar bar2 std_msgs/Float64 "m.data" __ns:=foo
ERROR: Wrong input topic (or topic field): bar
```
instead have to do this:
```
rosrun topic_tools transform /foo/bar bar2 std_msgs/Float64 "m.data"
```

The node namespace is used for the output topic though.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #1470, opened 2018-08-01
Use node namespace when looking up topic

lucasw:
For #1427 

I can add a test for this, though hztest is failing

```
catkin run_tests --no-deps --this
```

### DIFF
--- a/tools/topic_tools/scripts/transform
+++ b/tools/topic_tools/scripts/transform
@@ -75,7 +75,11 @@ class TopicOp:
 
         self.expression = args.expression
 
-        input_topic_in_ns = rospy.get_namespace() + args.input
+        if len(args.input) > 0 and args.input[0] != '/':
+            input_topic_in_ns = rospy.get_namespace() + args.input
+        else:
+            input_topic_in_ns = args.input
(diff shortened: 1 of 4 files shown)


lucasw:
Whoops this needs to check for a leading slash and not add the namespace then.

tools/topic_tools/scripts/transform
```diff
@@ -75,10 +75,15 @@ class TopicOp:
 
         self.expression = args.expression
 
+        if len(args.input) > 0 and args.input[0] != '/':
+            input_topic_in_ns = rospy.get_namespace() + args.input
+        else:
+            input_topic_in_ns = args.input
```
dirk-thomas, line 81: This could be simplified to the following (assuming `args.topic` can't be empty):

```
input_topic = args.input
if not input_topic.startswith('/'):
    input_topic = rospy.get_namespace() + input_topic

dirk-thomas (commented):
Please retarget the latest development branch (currently `melodic-devel`).

dirk-thomas:
Closing in favor of #1663.

PR #1472, opened 2018-08-03
On-error type conformance in the Master_API 

KKostya:
- Changed the default return value from `0` to `[]` in the `@apivalidate` arguments for the `registerSubscriber` and `registerPublisher` methods.
- Similar typing tweak for the `lookupService` method.

The ROS Master XMLRPC API (http://wiki.ros.org/ROS/Master_API) defines `registerSubscriber` and `registerPublisher` return types to be `(int, str, [str])`. Yet, on an error, instead of a list of strings, it returns an integer. 

This stricter typing is needed for our work on OCaml client project in ROS:
https://github.com/AestheticIntegration/imandra-ros/tree/master-api/rosocaml


dirk-thomas:

dirk-thomas:

dirk-thomas:
Thanks for the patch.

PR #1473, opened 2018-08-03
fix regressions from 1372

### LINKED ISSUE #1432: rosbag - _read_message in bag.py throws exception

dirk-thomas:
This patch addresses multiple problems introduced by #1372 and closes #1432:
* The first commit makes the new argument optional to restore API compatibility.
* The second and third are from #1444.
* The fourth commit is from #1445.


dirk-thomas:

PR #1450, opened 2018-07-06
Rename python message_filters.Cache.getLastestTime to getLatestTime

madebyollin:
Per instructions from dirk-thomas on this issue: https://github.com/ros/ros_comm/issues/1449#issuecomment-403118429

PR #1442, opened 2018-06-22
Fix for issue 1441

pbaughman:
Suggested fix for https://github.com/ros/ros_comm/issues/1441

Use the logger.Handler's format method so that by default stack traces
are included in the log message if exc_info is set on the LogRecord

PR #1428, opened 2018-06-08
fix compiler warnings about unused variables

dirk-thomas:
Which appeared in the recent devel job build: http://build.ros.org/view/Mdev/job/Mdev__ros_comm__ubuntu_bionic_amd64/42/

PR #1407, opened 2018-05-23
Fix the count of subscribers.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@gmail.com>

PR #1477, opened 2018-08-09
changes between 1.14.3 and 1.13.6 for backporting

dirk-thomas:
The following list of changes has been integrated into `ros_comm` 1.14.3 (Melodic) since the last Lunar release (1.13.6).

**Commits before branching:** some commits were made after the 1.13.6 release before the branch for Melodic was created. Those are therefore the same on the Lunar and Melodic branch. They won't be enumerated here.

**Backported:** (these changes are part of this PR)

* rosbag record: Add TransportHint options --tcpnodelay and --udp (#1295)

  * Requested backport

* topic_tools: Check that output topic is valid in demux (#1367)

  * Bug fix

* Replaced Double-Checked Locking pattern with static variable (#1365)

  * Bug fix

* Fix "pass_all_args" for roslaunch-check, add nosetest (#1368)

  * Bug fix

* Convert the period standard deviation in StatisticsLogger to Duration at the end. (#1361)

  * Bug fix

* Fix integer overflow for oneshot timers. (#1382)

  * Bug fix

* Change to force a rebuild of the pollset on flag changes (#1393)

  * Bug fix

* XmlRpcpp fixes for OSX (#1402)

  * Bug fix

* Fix the count of subscribers. (#1407)

  * Bug fix

* Fix Issue 1346 - Make a copy of the topic list (#1416)

  * Bug fix

* Fix some errors in some probably not frequented code paths (#1415)

  * Bug fix

* Rename python message_filters.Cache.getLastestTime to getLatestTime (#1450)

  * Fix naming of API

* Add hasStarted()const to Timer API (#1464)

  * Add getter API to avoid the need to maintain this private state separately

**Not backported:**

* added parameter to stop client libraries from generating topics list (#1241)

  * ABI breakage

* Adds latch functionality (#1341)

  * New option

 * Use boost::shared_ptr to fix memory leak in BagPlayer (#1373)

   * ABI breakage

* Use HTTP/1.1 in XMLRPC Server (#1287)

  * Significant change

* fix sphinx warning (#1371)

  * Not necessary

* roslaunch: add --log option (#1330)

  * New option

* add time when timer expired to timer events (#1130)

   * ABI breakage

* Implement Bag encryption/decryption (#1206)

  * New feature, new dependencies

* Keep connection header info in rosbag filter/compress (#1372)

  * New feature

* Fix compiler warning of test_aes_encryptor (#1376)

  * Not applicable

* Specialize BagCallbackT for MessageInstance (#1374)

  * New feature

* warn if ROS_IP contains whitespace (#1379)

  * Improvement

* Changed invocation to `add` to conform template syntax (#1388)

  * Not necessary

* topic_tools: throttling when rostime jump backward (#1397)

  * Improvement

* Allow disabling rosout file logging (to rosout.log) (#1381)

  * New option

* suppress sequential identical messages (#1309)

  * New feature

* remove rosconsole (#1405)

  * Not applicable

* Add in -D_FILE_OFFSET_BITS=64 on machines less than 64-bits. (#1406)

  * Not applicable

* import rosbag locally instead of at module level (#1424)

  * Not necessary

* fix compiler warnings about unused variables (#1428)

  * Not necessary

* Throw BagException when disk is full (#1451)

  * Improvement

* Fix for issue 1441 (#1442)

  * Improvement

* message_filters: call Subscriber::unsubscribe() in destructor (#1434)

  * Not necessary

* fix regressions from 1372 (#1473)

  * Not applicable

@ros/ros_team @ros/ros_comm-maintainers RFC.


mikepurvis:
Looks like good picks to me.

bmagyar:
I would like to request #1464 to be added to the backported list. 

I wouldn't really call it a new feature, it is only a minor getter that allows to avoid hacking around this issue in user code. This is especially desired as Kinetic is likely to stay around for a while within companies using ROS: many haven't even left Indigo yet, let alone planning to move to Melodic.

dirk-thomas:
> This is especially desired as Kinetic is likely to stay around for a while within companies using ROS: many haven't even left Indigo yet, let alone planning to move to Melodic.

This on it's own is not a real argument because at the same time these users demand stability. That is ultimately the reason why we have distributions and changes to existing distros are limitted to bug fixes (since any change has the potential to introduce regressions).

> I wouldn't really call it a new feature, it is only a minor getter that allows to avoid hacking around this issue in user code.

That being said I agree with you that #1464 is a very small API addition which should be safe to backport to provide the convenience not to pass the information in parallel. :+1:

I cherry-picked the corresponding commit in d80aa6d4ab9a28705149ae766ab329051b8d22ad and updated the above description.

dirk-thomas:
The following change wasn't backported as part of this (or any future) backport PR:

> fix compiler warnings about unused variables (#1428)
> > Not necessary

The change has been backported separately to Lunar (#1610) and to Kinetic (#1578.)

PR #1479, opened 2018-08-10
Fix test code build issues on Windows

johnsonshih:


dirk-thomas:
Thank you for this big patch!

PR #1480, opened 2018-08-10
[roswtf] Add an alias script.

130s:
`roswtf` provides federated FAQ analysis. Its plugin mechanism allows customization. In product setting we found it convenient to implement system checkup that the users, not just developers, can run with a fewer steps.

I hate to say but its script name is often concerning when asking customers to run checkup command. Here I'm suggesting to add an "alias" script.

I simply duplicated `roswtf` script and renamed it, so it's not really an alias, but I did not know how I can alias a Python script.

I named it `rosfaq` but I'm all ears for the naming as well.

### DIFF
--- a/utilities/roswtf/scripts/rosfaq
+++ b/utilities/roswtf/scripts/rosfaq
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2018, Open Source Robotics Foundation
+# All rights reserved.
+#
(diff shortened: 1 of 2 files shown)


dirk-thomas:
I don't think creating an alias script is necessary.

You can also refer to it with the "official acronym" mentioned in this [answer](https://answers.ros.org/question/277428/what-does-roswtf-mean/) if that makes it less "concerning".

If you prefer to have a different name you should be able to create an actual alias in your shell.

PR #1484, opened 2018-08-13
add dependency on rosbash to topic_tools

mintar:
Without this, `source /opt/ros/$ROS_DISTRO/setup.bash` fails when `topic_tools` is installed (and `rosbash` is not) because the following variable is empty in `20.transform.bash`:

    _sav_transform_roscomplete_rosrun=

This problem is demonstrated by the following `.gitlab-ci.yml`:

https://gitlab.com/mintar/debug_topic_tools_rosbash/blob/master/.gitlab-ci.yml

The failure is shown in the log here:

https://gitlab.com/mintar/debug_topic_tools_rosbash/-/jobs/88877496

Without rosbash installed, the command `$(complete | grep -w rosrun | awk '{print $3}')` returns nothing, so the `source` command fails.

Installing `rosbash` fixes the problem, which is why I believe `topic_tools` should have a dependency on `rosbash`.

### DIFF
--- a/tools/topic_tools/package.xml
+++ b/tools/topic_tools/package.xml
@@ -19,6 +19,7 @@
 
   <build_depend>cpp_common</build_depend>
   <build_depend>message_generation</build_depend>
+  <build_depend>rosbash</build_depend>
   <build_depend>rosconsole</build_depend>
   <build_depend>roscpp</build_depend>
   <build_depend>rostest</build_depend>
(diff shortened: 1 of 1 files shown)


mintar:
Test failures are unrelated.

dirk-thomas:
I assume the problem exists in the latest development branch too? If yes, please edit the PR and target the `melodic-devel` branch instead.

mintar:
I've found a much better solution: #1486 .

PR #1497, opened 2018-08-28
show connection info on rosnode info

maximeconnolly:
Continuation of #1386 
The only changes are to address the reviews

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -473,6 +473,7 @@ class TCPROSTransport(Transport):
         Similar to getTransportInfo() in 'libros/transport/transport_tcp.cpp'
         e.g. TCPROS connection on port 41374 to [127.0.0.1:40623 on socket 6]
         """
+        # Pattern matching this output in tools/rosnode/src/rosnode/__init__.py CONNECTION_PATTERN
         return "%s connection on port %s to [%s:%s on socket %s]" % (self.transport_type, self.local_endpoint[1], self.remote_endpoint[0], self.remote_endpoint[1], self._fileno)
 
     def fileno(self):
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -45,6 +45,7 @@ import errno
 import sys
 import socket
 import time
+import re
 try:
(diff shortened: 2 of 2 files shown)


tools/rosnode/src/rosnode/__init__.py
```diff
@@ -62,6 +63,8 @@
 
 NAME='rosnode'
 ID = '/rosnode'
+# Line is defined at clients/rospy/src/rospy/impl/tcpros_base.py at line 477
```
dirk-thomas: Please do not refer to a line number since that is going to change if the file is being edited in the future. Instead refer to the file and a symbol, e.g. `The string is defined in clients/rospy/src/rospy/impl/tcpros_base.py TCPROSTransport.get_transport_info`.

clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -473,6 +473,7 @@ def get_transport_info(self):
         Similar to getTransportInfo() in 'libros/transport/transport_tcp.cpp'
         e.g. TCPROS connection on port 41374 to [127.0.0.1:40623 on socket 6]
         """
+	# Changing this line while break tools/rosnode/src/rosnode/__init__.py at line 67
```
dirk-thomas: Please do not refer to a line number since that is going to change if the file is being edited in the future. Instead refer to the file and a symbol, e.g. `Pattern matching this output in tools/rosnode/src/rosnode/__init__.py CONNECTION_PATTERN`.

Also this line currently uses a tab which can't be mixed with space in Python 3. Please change it to use the spaces (as the rest of the file).

maximeconnolly:
@dirk-thomas It should be good now

dirk-thomas:
Great, thank you for picking this up.

For future PRs it would be good to keep the original commit in the author name.

dirk-thomas:

dirk-thomas:
For future PRs it would be good to keep the original commit in the author name.

efernandez:
Thanks @maximest-pierre . Did you notice the `rostopic info` command being significantly slower after this change?

maximeconnolly:
@efernandez  Not really. I just compared it with ```time``` both version and they were almost identical.

efernandez:
Then it should be fine @maximest-pierre . I probably have a laggy connection when I tried, or an overloaded system. :+1:

PR #1499, opened 2018-09-06
fix #1474 move bag encryption plugins into separate library

### LINKED ISSUE #1474: Declaring a rosbag::Bag member in a nodelet class leads to failure to launch the nodelet
Having a class member of type rosbag::Bag in a nodelet leads to failure to launch the nodelet in ROS melodic.
Here is a minimum example:
```cpp
#include <nodelet/nodelet.h>
#include <rosbag/bag.h>

class MyNodelet : public nodelet::Nodelet
{
 public:
  void onInit() {}

 private:
  rosbag::Bag bag_;
};

#include <pluginlib/class_list_macros.h>
PLUGINLIB_EXPORT_CLASS(MyNodelet, nodelet::Nodelet);
```
(Here is the full package: [my_package.tar.gz](https://github.com/ros/ros_comm/files/2264378/my_package.tar.gz))

Compiling and running
```
rosrun nodelet nodelet standalone my_package/MyNodelet
```
leads to the following error:
```
[ERROR] [1533595561.186295932]: Failed to load nodelet [/my_package_MyNodelet] of type [my_package/MyNodelet] even after refreshing the cache: MultiLibraryClassLoader: Could not create object of class type rosbag::NoEncryptor as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
[ERROR] [1533595561.187418380]: The error before refreshing the cache was: MultiLibraryClassLoader: Could not create object of class type rosbag::NoEncryptor as no factory exists for it. Make sure that the library exists and was explicitly loaded through MultiLibraryClassLoader::loadLibrary()
```
This used to work fine in lunar. It seems to be related to some recent changes to add encryption support in rosbag (https://github.com/ros/ros_comm/pull/1206).

Changing the member to be a pointer which is initialized in the `onInit` function works fine.

jalkino:
Moves the bag encryption plugins into a separate library, i.e. keep it out of `librosbag_storage.so`
This setup is required by pluginlib's `class_loader`, see http://wiki.ros.org/class_loader#Caution_of_Linking_Directly_Against_Plugin_Libraries

This fixes #1474 which was caused by `librosbag_storage.so` containing directly linked code (like `rosbag:bag`) _and_ the encryption plugins. This setup is unsupported by class_loader. As described in the link, the `class_loader` tries to make it work anyway for backward compatibility by guessing where the factory object for the plugin was loaded from. 
When using nodelets (which are plugins itself), the `class_loader` was guessing wrong and assigned the factory objects for the encryption plugins to the nodelet's `.so` library. This caused `createInstance` to fail because, according to `encryptor_plugins.xml`, the factory object should come from `librosbag_storage.so`

### DIFF
--- a/tools/rosbag_storage/CMakeLists.txt
+++ b/tools/rosbag_storage/CMakeLists.txt
@@ -28,7 +28,7 @@ add_definitions(${BZIP2_DEFINITIONS})
 set(AES_ENCRYPT_SOURCE "")
 set(AES_ENCRYPT_LIBRARIES "")
 if(NOT WIN32)
-  set(AES_ENCRYPT_SOURCE "src/aes_encryptor.cpp")
+  set(AES_ENCRYPT_SOURCE "src/aes_encryptor.cpp" "src/gpgme_utils.cpp")
   set(AES_ENCRYPT_LIBRARIES "crypto" "gpgme")
 endif()
 
@@ -79,10 +79,10 @@ if(NOT WIN32)
   if(CATKIN_ENABLE_TESTING)
     find_package(rostest)
 
-    catkin_add_gtest(test_aes_encryptor test/test_aes_encryptor.cpp
+    catkin_add_gtest(test_aes_encryptor test/test_aes_encryptor.cpp src/gpgme_utils.cpp
       WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/test)
     if(TARGET test_aes_encryptor)
-      target_link_libraries(test_aes_encryptor rosbag_storage ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+      target_link_libraries(test_aes_encryptor rosbag_storage ${BZIP2_LIBRARIES}  ${AES_ENCRYPT_LIBRARIES} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
     endif()
   endif()
 endif()
--- a/tools/rosbag_storage/CMakeLists.txt
+++ b/tools/rosbag_storage/CMakeLists.txt
@@ -33,14 +33,12 @@ if(NOT WIN32)
 endif()
 
 add_library(rosbag_storage
-  ${AES_ENCRYPT_SOURCE}
   src/bag.cpp
   src/bag_player.cpp
   src/buffer.cpp
   src/bz2_stream.cpp
   src/lz4_stream.cpp
   src/chunked_file.cpp
-  src/encryptor.cpp
   src/message_instance.cpp
   src/query.cpp
   src/stream.cpp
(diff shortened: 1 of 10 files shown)


jalkino:
One check failed because the unit test `__main__.RandomPlay.test_random_play` failed. To me this seem unrelated to this PR, as this error occurred before in other PR's. It seems to be a random error / unstable test.
Can someone please re-trigger the build to check that? Thank you!

C-NR:
For us at the autonomos group on the Dahlem Center for Machine Learning and Robotics this proposed fix works well. With it we are able to record and play back our bag files using ROS Melodic.

sprickett:
I'm also using this fix at Createc to enable our recording and playback nodelet's under Melodic. I've not discovered any issues.

dirk-thomas:

dirk-thomas:

dirk-thomas:

dseifert:
@dirk-thomas, any chance this PR could be merged and integrated for Melodic? It's been open for 3 months and multiple people confirmed that it works.

By the way, the "enhancement" label is a misclassification. This is not an enhancement, but a fix for a regression / bug in rosbag-storage. In my case the bug prevents multiple projects from actually working in Melodic, and was the biggest headache for our switch from Kinetic to Melodic.

Please let us know if there's anything we can do to assist in merging the pull request.

dirk-thomas:
Unfortunately I won't be able to look at this (and other ROS 1 tickets) before January.

dseifert:
Time flies, happy new year, @dirk-thomas :-)

dseifert:
Hi @dirk-thomas. Just a friendly reminder that this bug is still bothering (apparently quite a few) people that are using the rosbag API in nodelets. Please let me know if there's anything that I could do to assist in merging the pull request.

dseifert:
Just to let everybody know that we are still using this pull-request on a regular basis as otherwise we can not use the rosbag API in a nodelet :-( It would be appreciated if somebody could accept the PR so this bug can finally be resolved.

mikepurvis:
This seems pretty harmless to me, and seems to work in our setup as well. I'll merge in 24h if there are no objections.

dirk-thomas:
The reason I hesitated to merge this earlier is that the patch does break API since it removed public API from a header. Though I guess it is not likely that external code uses the moved classes and if it does it should be fairly easy to address by changing an include.

Thanks for the patch.

dseifert:
Thank you!!

dirk-thomas:
Please see #1687 which fixes a regression introduces with this patch.

berndpfrommer:
How and when will this fix find its way into the standard Ubuntu Melodic packages?
I just upgraded all my melodic packages, but alas the ros_comm package is built on tag 1.14.3, which does not have the fix:
http://repositories.ros.org/status_page/ros_melodic_default.html
Could somebody please clarify?
Thanks!

jonfink:
Just a ping to find out what the plan is for this PR to hit updated packages on apt server?

Encountered this same issue integrating rosbag reading into an RVIZ panel today. Can confirm that compiling this version of rosbag_storage from source fixes the issue.

dirk-thomas:
> How and when will this fix find its way into the standard Ubuntu Melodic packages?

The change is already on the `melodic-devel` branch so whenever a new release of this repo is being made it will be included. When that is going to happen is unclear and solely depends on having the necessary time to do so.

PR #1502, opened 2018-09-07
roslaunch: Allow passing _TIMEOUT_SIGINT and _TIMEOUT_SIGTERM as parameters

### LINKED ISSUE #1493: roslaunch: Allow passing _TIMEOUT_SIGINT and _TIMEOUT_SIGTERM as parameters
There is demand for having [these parameters](https://github.com/ros/ros_comm/blob/6b9efd56d6882d1c017152ba11a5780a1496b30b/tools/roslaunch/src/roslaunch/nodeprocess.py#L58) configurable:

* https://bitbucket.org/osrf/ariac/wiki/2018/developer_tips
* https://gist.github.com/plusk01/7b50443138a8d4b0ff742fa86b75c1e5
* https://answers.ros.org/question/11353/how-to-delay-escalation-to-sig-term-after-sending-sig-int-to-roslaunch/
* my own use-case, where I'm re-processing some bagfiles having a `rosbag record --bz2` node in launch file, which doesn't finish saving the output bagfile in the 15 seconds time-frame after the source bag (with `required=true`) stops playing

Is it possible to submit a patch/MR that would add a roslaunch command-line option for this? I don't see any backwards-compatibility problems if we use the current values as defaults...

peci1:
Fixes #1493 by adding roslaunch CLI options `--sigint-timeout` and `--sigterm-timeout`. The timeouts are common to all nodes launched by this command. It would be possible to customize them per-node, but that would require changing the XML tags, and I'm not sure if it'd be useful at all.

It needed an unexpectedly big change to the codebase, but mainly it was just adding and passing keyword arguments of methods, so the API should be backwards compatible (every added kw argument has a default value).

I added unit tests for the local nodes, however, I haven't figured out any other way of testing this than really launching and stopping nodes, so the newly added tests need about 20 seconds to run... I hope it's not a big problem...

I also haven't figured out how to do automatic testing of the remote launch, but it seems that it's not tested at all. So I at least added a script you can run manually in case you have a second machine set up for remote launching. This test passed on my machine.

Theoretically, it should work also on Windows, but I haven't tested that.

### DIFF
--- a/tools/roslaunch/resources/timeouts.launch
+++ b/tools/roslaunch/resources/timeouts.launch
@@ -1,0 +1,5 @@
+<launch>
+
+  <node name="signals" pkg="roslaunch" type="signal_logger.py" />
+
+</launch>
--- a/tools/roslaunch/scripts/roscore
+++ b/tools/roslaunch/scripts/roscore
@@ -59,6 +59,16 @@ def _get_optparse():
     parser.add_option("--master-logger-level",
                       dest="master_logger_level", default=False, type=str,
                       help="set rosmaster.master logger level ('debug', 'info', 'warn', 'error', 'fatal')")
+    parser.add_option("--sigint-timeout",
+                      dest="sigint_timeout",
+                      default=15, type=int,
+                      help="the SIGINT timeout used when killing the core (in seconds).",
+                      metavar="SIGINT_TIMEOUT")
+    parser.add_option("--sigterm-timeout",
+                      dest="sigterm_timeout",
+                      default=2, type=int,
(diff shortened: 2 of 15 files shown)


peci1:
Hmm, the test failures seem to be unrelated...

UnexpectedLobster:
+1, that would be very appreciated and time-saving :)

dirk-thomas:
The patch removes the previously (private) constants and duplicates the values `2` and `15` in numerous places. Please keep the constants (without a leading underscore) and use them everywhere instead of the integers.

peci1:
Closing in favor of #1670.

PR #1505, opened 2018-09-16
[roswtf] Better msg replacement for 'No package or stack in context'.

130s:
When there's no ROS pkg is found on the current directory when `roswtf` is run, a message `No package or stack in context` appears. This message is not useful IMO because what "context" means is unclear and is not something clearly explained anywhere in ROS documents.

This PR suggests clearer message.

**How to reproduce the behavior**
```
$ cd ~
$ roswtf
:
No package or stack in context
:

$ rosversion roswtf
1.12.14
```


dirk-thomas:
Thanks for the improvement.

PR #1507, opened 2018-09-21
Fix memory error due to missing rosout_disable_topics_generation para…

hdino:
Getting the parameter `rosout_disable_topics_generation` was not secured by a default value and thus could cause a memory error if it is not set (cf. [https://lists.debian.org/debian-science/2018/09/msg00046.html](https://lists.debian.org/debian-science/2018/09/msg00046.html).

### DIFF
--- a/clients/roscpp/src/libros/rosout_appender.cpp
+++ b/clients/roscpp/src/libros/rosout_appender.cpp
@@ -107,10 +107,7 @@ void ROSOutAppender::log(::ros::console::Level level, const char* str, const cha
   
   // check parameter server/cache for omit_topics flag
   // the same parameter is checked in rosout.py for the same purpose
-  if (!ros::param::getCached("/rosout_disable_topics_generation", disable_topics_))
-  {
-    disable_topics_ = false;
-  }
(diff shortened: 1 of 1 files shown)


mgrrx (commented):
I suggest to also initialize the variable in the initializer of the class.

clients/roscpp/src/libros/rosout_appender.cpp
```diff
@@ -107,7 +108,9 @@ void ROSOutAppender::log(::ros::console::Level level, const char* str, const cha
   // check parameter server/cache for omit_topics flag
   // the same parameter is checked in rosout.py for the same purpose
   if (!ros::param::getCached("/rosout_disable_topics_generation", disable_topics_))
-      disable_topics_ = false;
+  {
+    disable_topics_ = false;
```
jspricke: I guess we don't need this if the default is set in the initializer, right?

hdino:
Generally not, but I left the check of the return value of `ros::param::getCached` in case a future version alters its argument. However, this might be a bit paranoid...

jspricke (approved):

mgrrx (approved):

dirk-thomas:
Thanks for the patch.

PR #1510, opened 2018-09-25
Unix Domain Socket Support

fujitatomoya:
Hi,

I'm Tomoya Fujita from Sony.

this is the Unix Domain Socket feature extension for ROS message transmission layer. if the system internal message use only, TCP/IP, UDP/IP protocal stack does not do any good to the system, just a pure overhead. so then Unix Domain Socket (UDS) is reasonable to get rid of the overhead with the system internal use for ROS Inter-Process Communication with good affinity of current code.

we understand the current code is not suitable for merge into the mainline, but we'd like to know if anyone is interested on this feature. and actually we are new to make contribution for ROS, so could you please tell us the proper procedure for the contribution? if any request, we can adjust the code for sure and we are willing to do that.

BTW, i will explain the overview of this feature ROSCon2018@Madrid.

thanks,


dirk-thomas:
Since this change is massive in size please provide a very detailed description of this contribution (high level overview, implementation details, changes in behavior, how this is supposed to be used etc.

Also in order to be considered it needs to target the latest development branch (currently `melodic-devel`), must pass all tests and should not introduce regressions for other platforms currently working (e.g. Windows - maybe gracefully fall back on that platform).

fujitatomoya:
this feature is introduced in ROSCon2018@Madrid, https://roscon.ros.org/2018/presentations/ROSCon2018_Aibo.pdf

there is doc/README is ready for this feature and if necessary, we can provide more detailed information for sure.

we understand the current code is for kinetic-devel, so we sill start working on melodic-devel.

fujitatomoya:
since im new to contribute for ros_comm, what do we need to do for the regression test? is there any specific test cases are ready for that. could you please tell us what to do?

dirk-thomas:
> so we sill start working on melodic-devel.

Please retarget this patch to the default branch once it is ready.

> what do we need to do for the regression test?

First all existing test should pass. But beyond that a good amount of manual testing needs to happen too.

fujitatomoya:
understood, thanks.

fujitatomoya:
since this needed to rebase on default branch, closing this thread and then will take care of https://github.com/ros/ros_comm/pull/1549.

PR #1511, opened 2018-09-26
roslaunch: add command to not override params already defined

KamilaLis:
Allow to decide if we want to override already defined params with new value or not (in launch file).
```<rosparam command="load_no_override" file="xxx" />```

### DIFF
--- a/tools/roslaunch/src/roslaunch/config.py
+++ b/tools/roslaunch/src/roslaunch/config.py
@@ -291,7 +291,7 @@ class ROSLaunchConfig(object):
         """
         self.clear_params.append(param)
 
-    def add_param(self, p, filename=None, verbose=True):
+    def add_param(self, p, filename=None, verbose=True, override_params=True):
         """
         Declare a parameter to be set on the param server before launching nodes
         @param p: parameter instance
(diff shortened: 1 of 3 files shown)


mgrrx:
I don't get the use case of this new feature. You have two yaml files, both loaded within the same roslaunch context and you want to avoid that the second file overrides parameters defined in the first one?
I also fear that a user would assume that `load_no_override` also checks if the parameter is already defined on the parameter server and then skips overriding existing parameters entirely. I haven't tested this branch but by looking at the code I assume this is not the case.
Can you please target the current development branch `melodic-devel`.

KamilaLis:
> You have two yaml files, both loaded within the same roslaunch context and you want to avoid that the second file overrides parameters defined in the first one?

The main idea was to allow the modification of certain parameters only in all the launch files in a workspace. I want to have a file in which a person who was not involved directly in a creation of a robot is able to change only some of the parameters that are to be sent to parameter server. The objective of this implementation was to load all the parameters of this file before starting all of the nodes.

> I also fear that a user would assume that `load_no_override` also checks if the parameter is already defined on the parameter server and then skips overriding existing parameters entirely.

Can you suggest any other name for this functionality? The only other idea I have come up with was labeling current `load` command as `load_override`.

mikepurvis:
I see the use-case here— we do something similar with an after-the-fact "configuration" yaml that's basically just param'd in at the end of the main launchfile and can override anything. However, if you're genuinely concerned about locking out a user in this way due to safety or security, this doesn't feel like a good way to get there.

Two other possibilities for how to achieve what you want without changing the codebase:

- Load the "user" parameters upfront, so that any defaults are applied on top of the user-specified options, rather than the other way around.
- Wrap the param loading in a script that applies specific whitelisting rules which affect what the user is allowed to specify. You can achieve this right within roslaunch using `<param command="$(find fancy_pkg safe_param_filter path/to/user/params.yaml" type="yaml">`, just have the script parse the yaml and then dump it to stdout. You can either filter the disallowed options if they exist, or just return a non-zero exit code to fail the whole roslaunch instance.

KamilaLis:
@mikepurvis you are right. Thank you for your help! :)

PR #1515, opened 2018-09-28
rospy: added get_param_cached

mgrrx:
rospy had an incomplete version of a local parameter cache in the code base. The code was never called due to https://github.com/ros/ros_comm/blob/melodic-devel/clients/rospy/src/rospy/msproxy.py#L119 .

I borrowed some of the code in rosmaster and added the rospy toplevel function `get_param_cached` (in accordance to `roscpp::getParamCached`). It was surprisingly easy since the subscribeParam functionality was already there.

### DIFF
--- a/clients/rospy/src/rospy/__init__.py
+++ b/clients/rospy/src/rospy/__init__.py
@@ -51,7 +51,7 @@ from .client import spin, myargv, init_node, \
     wait_for_message, \
     get_master, \
     on_shutdown, \
-    get_param, get_param_names, set_param, delete_param, has_param, search_param,\
+    get_param, get_param_cached, get_param_names, set_param, delete_param, has_param, search_param,\
     DEBUG, INFO, WARN, ERROR, FATAL
 from .timer import sleep, Rate, Timer
 from .core import is_shutdown, signal_shutdown, \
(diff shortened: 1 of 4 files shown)


NickZ:
Is there anything blocking merging on this? This brings the rospy client in-line with the behavior of roscpp client, and will solve a number of performance issues that I am seeing with rospy client.

mgrrx:
This PR was labeled as an enhancement and did not receive further feedback. It would be great if you could test this feature and report feedback.
I guess there should also be some discussion on wether the caching should happen by default (as it was thought to be in the original code) or if the extra method get_param_cached should be introduced. I personally decided for the latter since it is the same in roscpp.

paulbovbel:
We've been running this patch w/ https://github.com/locusrobotics/ros_comm/pull/10 at Locus since February with success. Thanks @mgrrx!

furushchev:
@mgrrx Hi, thank you very much for the patch!
I'm struggling with using `rospy` in environment with latency between rosmaster and rospy client node, and the change in this pull request completely solves the issue I have.
It would be nice if the feature is merged. :+1: 

( I tested the patch in https://github.com/ros/ros_comm/issues/1855 )

IMO, It would be better to append `get_param_cached` instead of changing `get_param`, since  behavior of existing nodes are changed.

flixr:

mgrrx:
any chance to get this one in melodic?

dirk-thomas:

dirk-thomas:
Thanks for the improvement.

Issue #1516, opened 2018-10-03
rosnode ping -a takes close to 30 seconds to finish


msadowski:
I'm running some stress testing of my setup that uses [node_alive], which in turn uses rosnode ping.

My setup:
1. Laptop running master
2. 3 single board PCs 
3. Network switch to link all of the above

I noticed that if I forcibly disconnect network on the single board computer by unplugging the ethernet the command `rosnode ping -a` can take close to 30 seconds to finish with the following output:

```
Will ping the following nodes: 
 * /node1
 * /node2
 * /rosout
 * /node3

pinging /node1 with a timeout of 3.0s
xmlrpc reply from http://10.42.0.121:59371/     time=2.152920ms
pinging /node2 with a timeout of 3.0s
xmlrpc reply from http://10.42.0.134:46530/     time=2.267838ms
pinging /rosout with a timeout of 3.0s
xmlrpc reply from http://10.42.0.1:33253/       time=0.365973ms
pinging /node3 with a timeout of 3.0s
unknown network error contacting node: timed out
unknown network error contacting node: timed out
unknown network error contacting node: timed out
unknown network error contacting node: timed out
unknown network error contacting node: timed out
unknown network error contacting node: timed out
unknown network error contacting node: timed out
unknown network error contacting node: timed out
unknown network error contacting node: timed out
unknown network error contacting node: timed out
unknown network error contacting node: timed out
connection to [/node3] timed out
```

It looks as if the timeout is not respected? 

# Update

Current source code:
```
    try:
        while True:
            try:
                start = time.time()
                pid = _succeed(node.getPid(ID))
                end = time.time()

                dur = (end-start)*1000.
                acc += dur
                count += 1

                if verbose:
                    print("xmlrpc reply from %s\ttime=%fms"%(node_api, dur))
                # 1s between pings
            except socket.error as e:
                # 3786: catch ValueError on unpack as socket.error is not always a tuple
                try:
                    # #3659
                    errnum, msg = e.args
                    if errnum == -2: #name/service unknown
                        p = urlparse.urlparse(node_api)
                        print("ERROR: Unknown host [%s] for node [%s]"%(p.hostname, node_name), file=sys.stderr)
                    elif errnum == errno.ECONNREFUSED:
                        # check if node url has changed
                        new_node_api = get_api_uri(master,node_name, skip_cache=True)
                        if not new_node_api:
                            print("cannot ping [%s]: unknown node"%node_name, file=sys.stderr)
                            return False
[...]

PR #1517, opened 2018-10-04
rosnode: Explicitly handle socket.timeout in rosnode ping

msadowski:
This pull request is in response to #1516. There are some cases where rosnode ping would take over 30 seconds to scan all the nodes due to ValueError happening when trying to unpack socket.error. 

This pull request adds handling of socket.timeout explicitly.

Please let me know if there is anything else you would like me to change.

### DIFF
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -361,7 +361,7 @@ def rosnode_ping(node_name, max_count=None, verbose=False):
                             continue
                         print("ERROR: connection refused to [%s]"%(node_api), file=sys.stderr)
                     else:
-                        print("connection to [%s] timed out"%node_name, file=sys.stderr)
+                        print("connection to [%s] failed"%node_name, file=sys.stderr)
                     return False
                 except ValueError:
                     print("unknown network error contacting node: %s"%(str(e)))
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -336,6 +336,9 @@ def rosnode_ping(node_name, max_count=None, verbose=False):
                 if verbose:
                     print("xmlrpc reply from %s\ttime=%fms"%(node_api, dur))
                 # 1s between pings
+            except socket.timeout:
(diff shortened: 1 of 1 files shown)


tools/rosnode/src/rosnode/__init__.py
```diff
@@ -336,6 +336,9 @@ def rosnode_ping(node_name, max_count=None, verbose=False):
                 if verbose:
                     print("xmlrpc reply from %s\ttime=%fms"%(node_api, dur))
                 # 1s between pings
+            except socket.timeout:
+                print("connection to [%s] timed out"%node_name, file=sys.stderr)
+                return False
```
dirk-thomas, line 341: With this extra block added can line [364](https://github.com/ros/ros_comm/pull/1517/files#diff-548698d13d27ce06b14e7ba0e99f6f76R364) still happen?

msadowski: Looking into documentation of [socket](https://docs.python.org/2/library/socket.html) I believe it should not happen when we handle timeout error explicitly.

Socket has 4 exceptions:
- socket.error - can return either a string or (errno, string)
- socket.herror - (h_errno, string)
- socket.gaierror - (error, string)
- socket.timeout - string

since the timeout is the only subclass of socket.error that returns a single string I think the line 364 should never happen (unless I'm missing anything else that could raise a ValueError that is not related to unpacking the error). 

Would you like me to remove this ValueError check?

dirk-thomas: > since the timeout is the only subclass of socket.error that returns a single string I think the line 364 should never happen

The the following message should be changed to state that something "unknown" has happened, e.g.:

```
                     else:
                         print("connection to [%s] timed out"%node_name, file=sys.stderr)
```

replaced by:

```
                     else:
                         print("connection to [%s] failed"%node_name, file=sys.stderr)
```

> Would you like me to remove this ValueError check?

I would rather keep that - there is no reason related to the issue here (the timeout) that it should be removed.

msadowski: Thanks for clarification. I just pushed the change.

dirk-thomas:
Thanks for the patch.

PR #1520, opened 2018-10-16
Fix #889 : Exclude unused args check if pass_all_args is set

### LINKED ISSUE #889: roslaunch: pass_all_args exception if there are extra args defined
I just tried using this attribute on a fairly complex launch file that has multiple `<arg>`s declared, and it leads to an exception because not all of the `<arg>`s are defined in the `<include>`d file.

For example, in the `drcsim_gazebo` package, [atlas.launch](https://bitbucket.org/osrf/drcsim/src/194be8500fef81593f79607a21ee2badd9700a0e/drcsim_gazebo/launch/atlas.launch?at=default&fileviewer=file-view-default#atlas.launch-14) includes [atlas_no_controllers.launch](https://bitbucket.org/osrf/drcsim/src/194be8500fef81593f79607a21ee2badd9700a0e/drcsim_gazebo/launch/atlas_no_controllers.launch?at=default&fileviewer=file-view-default) and passes three arguments. I removed some of the duplicate argument definitions in [this commit](https://bitbucket.org/osrf/drcsim/commits/d5c93d7db649ca4df6c07d6fcba6f9e77953913a), but roslaunch gives an exception since there are extra `<arg>`s in `atlas.launch` that aren't part of `atlas_no_controllers.launch` (for example, [inertia_args](https://bitbucket.org/osrf/drcsim/src/194be8500fef81593f79607a21ee2badd9700a0e/drcsim_gazebo/launch/atlas.launch?at=default&fileviewer=file-view-default#atlas.launch-8)).

I'm guessing this will be close as a "won't fix", but I wanted to mention it since I was excited when @gerkey added this feature in #710, but now there's quite a few instances when I won't be able to use it.

alextoind:
This follows #889.

The dummy argument added to the test launch file should be fine to verify that there will be no exception even if there is an extra unused argument.

### DIFF
--- a/tools/roslaunch/src/roslaunch/xmlloader.py
+++ b/tools/roslaunch/src/roslaunch/xmlloader.py
@@ -632,8 +632,9 @@ class XmlLoader(loader.Loader):
                 self._recurse_load(ros_config, launch.childNodes, child_ns, \
                                        default_machine, is_core, verbose)
 
-            # check for unused args
-            loader.post_process_include_args(child_ns)
+            if not pass_all_args:
+                # check for unused args
+                loader.post_process_include_args(child_ns)
 
         except ArgException as e:
             raise XmlParseException("included file [%s] requires the '%s' arg to be set"%(inc_filename, str(e)))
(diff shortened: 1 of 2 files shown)


dirk-thomas:

alextoind:
@dirk-thomas should I change anything in the patch?

dirk-thomas:
> should I change anything in the patch?

No, the failing tests were unrelated to this patch and should be fixed in the next CI build.

dirk-thomas:
Thanks for the patch.

v-lopez:
@dirk-thomas Would you consider backporting this to kinetic?

saikishor:
+1 for backporting this to kinetic.

PR #1522, opened 2018-10-23
Add new option to publish when a bag write begin

AnthonyBirot:
I use the split mode of the bag recorder to save multiples hours bags.  And I need to know when the rosbag record node begin to write a new bag, because I want to publish some topics just at the begin.

I don't wan't to publish in continuous topics to be sure it is in bags.
I did the PR in the kinetic branch because in my project i use it and i can't change.

Note: I've seen the build didn't pass, I'll fix it.

### DIFF
--- a/tools/rosbag/include/rosbag/recorder.h
+++ b/tools/rosbag/include/rosbag/recorder.h
@@ -55,6 +55,7 @@
 #include <ros/time.h>
 
 #include <std_msgs/Empty.h>
+#include <std_msgs/String.h>
 #include <topic_tools/shape_shifter.h>
 
 #include "rosbag/bag.h"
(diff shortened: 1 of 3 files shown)


AnthonyBirot:
It's the get_compression_info test which failed. The value of compression is random and some time the unit test work. When it doesn't work, I had 898, 895 ... But I found this comment in the unit test code:
```
# the value varies each run, I suspect based on rand, but seems
# to generally be around 960 to 980 on my comp
```
So I don't think it's my PR which change that.

dirk-thomas:
Please target the latest development branch (currently `melodic-devel`). Patches are only accepted there and afterwards can be considered for backporting (for features rather than bug fixes that is less likely though).

AnthonyBirot:
Ok so i made a new pull request on melodic-devel : #1527.
So i close this pull reqest.

Issue #1523, opened 2018-10-26
rosmaster publisherUpdate fails with: [Errno 9] Bad file descriptor


ymikhailov:
It seems that rosmaster is not thread safe which leads to rare errors with rate of about 2 fails on 700 runs.

```
[rosmaster.master][INFO] 2018-09-18 07:33:15,764: publisherUpdate[/topic_1] -> http://localhost:37219/ ['http://localhost:39893/']: sec=0.00, exception=[Errno 9] Bad file descriptor
[rosmaster.master][INFO] 2018-09-18 07:33:15,766: publisherUpdate[/topic_2] -> http://localhost:40187/ ['http://localhost:35773/', 'http://localhost:39893/']: sec=0.00, result=[1, '', 0]
[rosmaster.master][INFO] 2018-09-18 07:33:15,768: publisherUpdate[/topic_3] -> http://localhost:38589/ ['http://localhost:39893/']: sec=0.01, result=[1, '', 0]
[rosmaster.master][INFO] 2018-09-18 07:33:15,768: publisherUpdate[/topic_2] -> http://localhost:44039/ ['http://localhost:35773/', 'http://localhost:39893/']
[rosmaster.master][INFO] 2018-09-18 07:33:15,771: publisherUpdate[/topic_4] -> http://localhost:38589/ ['http://localhost:38117/', 'http://localhost:39893/']
[rosmaster.master][INFO] 2018-09-18 07:33:15,772: publisherUpdate[/topic_2] -> http://localhost:44039/ ['http://localhost:35773/', 'http://localhost:39893/']: sec=0.00, result=[1, '', 0]
[rosmaster.master][INFO] 2018-09-18 07:33:15,773: publisherUpdate[/topic_2] -> http://localhost:42669/ ['http://localhost:35773/', 'http://localhost:39893/']
[rosmaster.master][INFO] 2018-09-18 07:33:15,775: publisherUpdate[/topic_4] -> http://localhost:38589/ ['http://localhost:38117/', 'http://localhost:39893/']: sec=0.00, result=[1, '', 0]
[rosmaster.threadpool][ERROR] 2018-09-18 07:33:15,775: Traceback (most recent call last):
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rosmaster/threadpool.py", line 218, in run
    result = cmd(*args)
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/rosmaster/master_api.py", line 210, in publisher_update_task
[...]

PR #1471, opened 2018-08-02
Reuse xmlrpc connections everywhere

mgrrx:
I noticed that xmlrpc connections are still not always reused since, due to circular import dependencies, the proxy class in rospy.msproxy can't be imported into these modules:
- rospy.impl.registration
- rospy.impl.tcpros_base
- rospy.impl.masterslave
- rospy.impl.simtime

### Current behavior

The current behavior is that a bunch of connections to the master are created and closed when running e.g. rospy.init_node:

(sudo  tshark -i lo -f "dst port 11311 or src port 11311")
```
    1 0.000000000          ::1 → ::1          TCP 94 36840 → 11311 [SYN] Seq=0 Win=43690 Len=0 MSS=65476 SACK_PERM=1 TSval=2354951144 TSecr=0 WS=128
    2 0.000007516          ::1 → ::1          TCP 74 11311 → 36840 [RST, ACK] Seq=1 Ack=1 Win=0 Len=0
    3 0.000094255    127.0.0.1 → 127.0.0.1    TCP 74 36862 → 11311 [SYN] Seq=0 Win=43690 Len=0 MSS=65495 SACK_PERM=1 TSval=764413847 TSecr=0 WS=128
    4 0.000105717    127.0.0.1 → 127.0.0.1    TCP 74 11311 → 36862 [SYN, ACK] Seq=0 Ack=1 Win=43690 Len=0 MSS=65495 SACK_PERM=1 TSval=764413847 TSecr=764413847 WS=128
    5 0.000116362    127.0.0.1 → 127.0.0.1    TCP 66 36862 → 11311 [ACK] Seq=1 Ack=1 Win=43776 Len=0 TSval=764413847 TSecr=764413847
    6 0.000168308    127.0.0.1 → 127.0.0.1    HTTP/XML 606 POST /RPC2 HTTP/1.1
    7 0.000173955    127.0.0.1 → 127.0.0.1    TCP 66 11311 → 36862 [ACK] Seq=1 Ack=541 Win=44800 Len=0 TSval=764413847 TSecr=764413847
    8 0.001303588    127.0.0.1 → 127.0.0.1    HTTP/XML 491 HTTP/1.1 200 OK
    9 0.001396096    127.0.0.1 → 127.0.0.1    TCP 66 36862 → 11311 [ACK] Seq=541 Ack=426 Win=44800 Len=0 TSval=764413848 TSecr=764413848
[...]

PR #1524, opened 2018-10-26
Fix concurrency error on publisherUpdate task

ymikhailov:
Fix for issue #1523


dirk-thomas:
This looks like a duplicate of #1471.

PR #1525, opened 2018-10-26
[C++14] remove explicit -std=c++11, default to 14

moriarty:
From REP-0003
> Melodic
> As of ROS Melodic, we are using the C++14 (ISO/IEC 14882:2014) standard.

Related to discussion on ros-planning/moveit#1146

### DIFF
--- a/clients/roscpp/CMakeLists.txt
+++ b/clients/roscpp/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(roscpp)
 
 if(NOT WIN32)
-  set_directory_properties(PROPERTIES COMPILE_OPTIONS "-std=c++11;-Wall;-Wextra")
+  set_directory_properties(PROPERTIES COMPILE_OPTIONS "-Wall;-Wextra")
 endif()
 
 find_package(catkin REQUIRED COMPONENTS
--- a/utilities/xmlrpcpp/test/CMakeLists.txt
+++ b/utilities/xmlrpcpp/test/CMakeLists.txt
@@ -8,7 +8,6 @@ if(TARGET test_base64)
   target_link_libraries(test_base64 xmlrpcpp)
 endif()
 
-set_target_properties(test_base64 PROPERTIES COMPILE_FLAGS -std=c++11)
 if(WIN32)
   # On Windows, gtest can be built as shared (dll) or static (lib),
   # to simplify the problem, here we require to match BUILD_SHARED_LIBS we used


dirk-thomas:
What is the purpose of removing the C++11 option? Just because Melodic *could* leverage C++14 doesn't mean it has to.

moriarty:
For clarity and consistency. The lack of triggered discussions on the moveit issue which I linked to above. But @clalancette has recently answered on that issue.

I opened https://github.com/ros/pluginlib/pull/131, https://github.com/ros/roscpp_core/pull/94, and this issue to create consistency throughout the ros_comm stack. 

The REP reads:
> Melodic
> As of ROS Melodic, we are using the C++14 (ISO/IEC 14882:2014) standard.

And in downstream packages when we bump to C++14 (and actually want to take advantage of the features) there is discussion if we can safely use C++14 while some things were still being compiled as  C++11.

To summarize the other issue

--------------------------------------

@gavanderhoorn wrote:
> @davetcoleman wrote:
> 
> > Melodic is C++14
> > http://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023
> 
> Bit late, but I don't think you can say this.
> 
> The language used in the REP is:
> 
> > Targeted Languages:
> > 
> > * C++14
> 
> _but_, if we look at how `roscpp` is compiled (for instance, there are many other packages that do similar things):
> 
> ```
> [ 33%] Building CXX object CMakeFiles/roscpp.dir/src/libros/master.cpp.o
> /usr/bin/c++  -DROSCONSOLE_BACKEND_LOG4CXX -DROS_PACKAGE_NAME=\"roscpp\" -Droscpp_EXPORTS -I/tmp/binarydeb/ros-melodic-roscpp-1.14.3/obj-x86_64-linux-gnu/devel/include -I/tmp/binarydeb/ros-melodic-roscpp-1.14.3/include -I/tmp/binarydeb/ros-melodic-roscpp-1.14.3/obj-x86_64-linux-gnu/devel/include/ros -I/opt/ros/melodic/include -I/opt/ros/melodic/share/xmlrpcpp/cmake/../../../include/xmlrpcpp -I/tmp/binarydeb/ros-melodic-roscpp-1.14.3/obj-x86_64-linux-gnu  -g -O2 -fdebug-prefix-map=/tmp/binarydeb/ros-melodic-roscpp-1.14.3=. -fstack-protector-strong -Wformat -Werror=format-security -DNDEBUG -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC   -std=c++11 -Wall -Wextra -o CMakeFiles/roscpp.dir/src/libros/master.cpp.o -c /tmp/binarydeb/ros-melodic-roscpp-1.14.3/src/libros/master.cpp
> ```
> Note the `-std=c++11` in there.
> 
> Edit: C++11 explicitly set [here](https://github.com/ros/ros_comm/blob/b5ee2db6524d5335e65abca1cc190d948e80a02d/clients/roscpp/CMakeLists.txt#L5).
- https://github.com/ros-planning/moveit/pull/1146#issuecomment-433364616

----------------------------------


@clalancette wrote:
> So, I originally wrote the update to REP-0003 for Melodic. And looking at it again, I agree that the wording is somewhat ambiguous. @gavanderhoorn interpretation is correct; packages are _free_ to use C++14 in Melodic, but are not required to do so. I'd happily review and merge any changes to REP-0003 that makes this less ambiguous.
- (https://github.com/ros-planning/moveit/pull/1146#issuecomment-433945005)

----------------------------------

I think the REP-0003 is clear:
> As of ROS Melodic, we are using the C++14 (ISO/IEC 14882:2014) standard.

These PRs (https://github.com/ros/pluginlib/pull/131, https://github.com/ros/roscpp_core/pull/94, and this #1525) are not actually using any C++14 features, they are just removing any doubts or concerns about actually using C++14 when there are 4 packages within the ros_comm stack which are compiled with **-std=c++11**
```
./ros_comm/xmlrpcpp/test/CMakeLists.txt:set_target_properties(test_base64 PROPERTIES COMPILE_FLAGS -std=c++11)
./ros_comm/roscpp/CMakeLists.txt:  set_directory_properties(PROPERTIES COMPILE_OPTIONS "-std=c++11;-Wall;-Wextra")
./pluginlib/CMakeLists.txt:  check_cxx_compiler_flag("-std=c++11" COMPILER_SUPPORTS_CXX11)
./pluginlib/CMakeLists.txt:      set_target_properties(${PROJECT_NAME}_unique_ptr_test PROPERTIES COMPILE_FLAGS -std=c++11 LINK_FLAGS -std=c++11)
./roscpp_core/rostime/CMakeLists.txt:    set_property(TARGET ${PROJECT_NAME}-test_time APPEND_STRING PROPERTY COMPILE_FLAGS "-std=c++11")
```

This type of clarity isn't too important with minor bumps to C++ standards, C++11->C++14 but C++11->C++17 and C++20 will be large bumps to the C++ standards... And for those future bumps I think it is important that ros_comm sets clear precedent and removes ambiguity by compiling with the standard which is being used by that release.

Edit: I linked to the wrong PR https://github.com/ros/roscpp_core/pull/9 -> https://github.com/ros/roscpp_core/pull/94

dirk-thomas:
Is there any technical reason why "consistency" is needed? Are you aware of any incompatibilities between C++11 and C++14 libraries?

moriarty:
No, there is no technical reason and I'm not aware of any incompatibilities between C++11 -> C++14.

Internally, with regards to compatibility issues, we were going to switch to C++14 and it turned out to be as simple as changing the compile option.
So we soon after went straight to C++17 for some internal components. We did come across one interesting edge cases, but it was caught during the unit/integration tests. The code compiled under all, but only passed the tests when compiled with C++11 & C++14... It was a few months ago, but if I remember it involved some the new std:basic_string_view and type deduction + lambda capture + gcc optimization... it's not something I would be worried about here.

dirk-thomas:

dirk-thomas (approved):
While you mentioned in https://github.com/ros/ros_comm/pull/1525#issuecomment-434442230 that is no technical reason to change this it should be fine to drop the explicit `-std=c++11` from the `melodic-devel` branch.

moriarty:
Thanks for re-triggering the tests, is there a list of known unstable tests?

dirk-thomas:
> is there a list of known unstable tests?

No, there is not. The recent non-green devel and PR builds for the package give an idea which tests are flaky though.

PR #1526, opened 2018-10-30
Fix stamp_age_mean overflow when stamp age very big

KindDragon:
Similar to #1361


dirk-thomas:
Please target the latest development branch - currently `melodic-devel`.

dirk-thomas:
Thanks for the patch.

PR #1529, opened 2018-11-03
Fix template syntax bug

jerry73204:
This patch is a continuation of PR #1388 and issue #1383. The previous PR missed an occurance of `this->add<i>(evt);`. This PR aims to complete the fix,

### DIFF
--- a/utilities/message_filters/include/message_filters/synchronizer.h
+++ b/utilities/message_filters/include/message_filters/synchronizer.h
@@ -355,7 +355,7 @@ private:
   template<int i>
   void cb(const typename mpl::at_c<Events, i>::type& evt)
   {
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please target the latest development branch (currently `melodic-devel`).

jerry73204:
I see. It's done as a backport when I'm porting old ROS to Arch Linux. I can put a patch in the packege instead.

jerry73204:
`melodic-devel` seems have no such issue. Feel free to close the PR if merging is not planned.

dirk-thomas:
Ok, I will close it then since we usually don't "forward support" newer toolchain for released distros - especially since Indigo is 4.5 years old.

PR #1531, opened 2018-11-10
Fix negative numbers in ros statistics

### LINKED ISSUE #1509: traffic is negative when using ros statistics, if a new connection is created between the same nodes
1. the `last_bytes_sent` is stored and will be used when the same `callid` is encountered

Solution:
1. Check out `connection`, a fresh new `bytes_sent` should be used if a new connection is created

foodtooth:
Fix ros/ros_comm#1509

### DIFF
--- a/clients/roscpp/include/ros/statistics.h
+++ b/clients/roscpp/include/ros/statistics.h
@@ -63,13 +63,14 @@ public:
   /**
    * Callback function. Must be called for every message received.
    */
-  void callback(const boost::shared_ptr<M_string>& connection_header, const std::string& topic, const std::string& callerid, const SerializedMessage& m, const uint64_t& bytes_sent, const ros::Time& received_time, bool dropped);
+  void callback(const boost::shared_ptr<M_string>& connection_header, const std::string& topic, const std::string& callerid, const SerializedMessage& m, const uint64_t& bytes_sent, const ros::Time& received_time, bool dropped, int connection_id);
 
 private:
 
   // these are hard constrains
   int max_window;
   int min_window;
+  int previous_connection_id;
 
   // these are soft constrains
   int max_elements;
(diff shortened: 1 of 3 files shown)


dirk-thomas:

foodtooth:
have no idea why ci test failed

dirk-thomas:
> have no idea why ci test failed

The failing tests were unrelated to this patch and should be fixed in the next CI build.

dirk-thomas:

dirk-thomas:
The current patch [breaks ABI](http://www.ros.org/reps/rep-0009.html#what-s-not-allowed) so it shouldn't be merged into Melodic as is but only into future distros. The alternative would be to add a new function with the new argument (in addition to the existing one) to avoid breaking ABI. Then it could be considered for Melodic (and potential backports).

foodtooth:
> The current patch [breaks ABI](http://www.ros.org/reps/rep-0009.html#what-s-not-allowed) so it shouldn't be merged into Melodic as is but only into future distros. The alternative would be to add a new function with the new argument (in addition to the existing one) to avoid breaking ABI. Then it could be considered for Melodic (and potential backports).

Hi, thanks for the link, great content to check out. As to the comments about adding new function, does it mean that I need to overload the existing function, which seems like still forbidden according to the link.

dirk-thomas:
> As to the comments about adding new function, does it mean that I need to overload the existing function, which seems like still forbidden according to the link.

You are right. I missed that adding an overload if there is one yet isn't allowed either.

The only option for staying ABI compatible would be to introduce a new function, e.g. `callback_with_connection_id(...)`. and call that one instead.

dirk-thomas:

dirk-thomas:

dirk-thomas:
Thanks for the patch.

PR #1534, opened 2018-11-20
Fixed typos: awhile -> a while

daniel-s-ingram:


dirk-thomas:
Thanks for the patch.

PR #1536, opened 2018-11-20
Update wiki URLs

VictorLamoine:
From http://ros.org/wiki to http://wiki.ros.org/

```
find . -type f -exec sed -i 's/http\:\/\/www\.ros\.org\/wiki/http\:\/\/wiki\.ros\.org/g' {} +
grep -rn "http://www.ros.org/wiki"
```


mikepurvis:
Looks great, nice simple change.

Failing test is flaky and unrelated.

PR #1537, opened 2018-11-20
Publish last message from latch topics when start time > 0

efernandez:
When we pass `-s TIME` to `rosbag play` the last message before `TIME` from each latch topics (selected)  doesn't get published. This PR fixes that.


dirk-thomas:

PR #1539, opened 2018-11-21
roscpp: Add const specifier to NodeHandle::param(param_name, default_val)

achim-k:
Otherwise the compiler will not allow the following:
```cpp
void foo(const ros::NodeHandle& nh) {
  const bool val = nh.param("bool_param", false);
}
```


dirk-thomas:
Thanks for the patch.

PR #1521, opened 2018-10-22
roslaunch: add an option in XmlLoader to only load arg tags

christophebedard:
This fixes #1300. It's off/false by default, but it's used by `arg_dump` to stop `roslaunch-complete` from failing if any arg has no default value.

PR #1540, opened 2018-11-21
roslaunch/xmlloader: use continue instead of pass for args_only

christophebedard:
As discussed in #1521, it works either way.

However, using `continue` is slightly more appropriate in this current situation.


dirk-thomas:
Thank you for following up on this.

dirk-thomas:

PR #1543, opened 2018-11-22
[rosbag] Fix waitForSubscribers hanging with simtime

AlexReimann:
Use Walltime for sleep as clock is not running yet

### DIFF
--- a/tools/rosbag/src/player.cpp
+++ b/tools/rosbag/src/player.cpp
@@ -405,7 +405,7 @@ void Player::waitForSubscribers() const
         foreach(const PublisherMap::value_type& pub, publishers_) {
             all_topics_subscribed &= pub.second.getNumSubscribers() > 0;
         }
-        ros::Duration(0.1).sleep();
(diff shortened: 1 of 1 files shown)


tools/rosbag/src/player.cpp
```diff
@@ -405,7 +405,7 @@ void Player::waitForSubscribers() const
         foreach(const PublisherMap::value_type& pub, publishers_) {
             all_topics_subscribed &= pub.second.getNumSubscribers() > 0;
         }
-        ros::Duration(0.1).sleep();
+        ros::WallTime::sleepUntil(ros::WallTime::now() +  ros::WallDuration(0.1));
```
dirk-thomas: Please simplify the logic to: `ros::WallDuration(0.1).sleep();`

theg4sh:
```suggestion
        ros::WallDuration(0.1).sleep();
```

dirk-thomas:
Thanks for the patch.

PR #1546, opened 2018-11-26
Avoid calling memcpy on NULL pointer with size 0.

hannob:
This fixes a warning with undefined behavior sanitizer (UBSAN). UBSAN is a compiler tool to check for code that is undefined according to the C standard and can be activated with -fsanitize=undefined.

The problem here is that this code is sometimes called with a NULL pointer of size 0. It's not allowed to call memcpy with a NULL pointer, even if the buffer has a size of zero. Avoid it by not calling memcpy if len is zero.

Stack trace and warning from UBSAN:
```
/var/tmp/portage/dev-ros/xmlrpcpp-1.14.3/work/ros_comm-1.14.3/utilities/xmlrpcpp/test/test_socket.cpp:726:13: runtime error: null pointer passed as argument 2, which is declared to never be null
    #0 0x555b1d591ade in test_getsockname(int, sockaddr*, unsigned int*) /var/tmp/portage/dev-ros/xmlrpcpp-1.14.3/work/ros_comm-1.14.3/utilities/xmlrpcpp/test/test_socket.cpp:726
    #1 0x555b1d612a10 in XmlRpc::XmlRpcSocket::get_port(int) /var/tmp/portage/dev-ros/xmlrpcpp-1.14.3/work/ros_comm-1.14.3/utilities/xmlrpcpp/src/XmlRpcSocket.cpp:386
    #2 0x555b1d59233d in XmlRpcSocketTest_get_port_Test::TestBody() /var/tmp/portage/dev-ros/xmlrpcpp-1.14.3/work/ros_comm-1.14.3/utilities/xmlrpcpp/test/test_socket.cpp:775
    #3 0x7f6f8bc145d9 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (//usr/lib64/libgtest.so+0x3f5d9)
    #4 0x7f6f8bc0a289 in testing::Test::Run() (//usr/lib64/libgtest.so+0x35289)
    #5 0x7f6f8bc0a3e7 in testing::TestInfo::Run() (//usr/lib64/libgtest.so+0x353e7)
    #6 0x7f6f8bc0a4b4 in testing::TestCase::Run() (//usr/lib64/libgtest.so+0x354b4)
    #7 0x7f6f8bc0a997 in testing::internal::UnitTestImpl::RunAllTests() (//usr/lib64/libgtest.so+0x35997)
[...]

PR #1547, opened 2018-11-26
Fix invalid zero index

hannob:
At various places the code uses &variable[0] to reference the start of a vector buffer. However in case the vector is of size zero this is invalid, because there is no index [0] in a zero-sized vector.

This avoids 4 such cases, separated in two commits, one for XmlRpcValue.cpp and one for test_base64.cpp.

Together with #1546 this makes the test suite run without warnings from UBSAN.

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcValue.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcValue.cpp
@@ -497,8 +497,11 @@ namespace XmlRpc {
     xml.resize(xml.size() + base64EncodedSize(_value.asBinary->size()));
 
     base64::encoder encoder;
-    offset += encoder.encode(&(*_value.asBinary)[0], _value.asBinary->size(), &xml[offset]);
-    offset += encoder.encode_end(&xml[offset]);
+    if (_value.asBinary->size() > 0)
+    {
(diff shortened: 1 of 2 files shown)


hannob:
Looking closer into it there may be a more elegant way: Using var.data() instead of &var[0], it would avoid the branching with the if's.

I'll wait for your feedback, but I can rework the PR to use .data() instead if you prefer.

dirk-thomas:
Thank you for the patch.

> I can rework the PR to use .data() instead if you prefer.

I think it would be good to do so to keep the logic simpler. Thanks.

hannob:
vector .data() functions are c++11, and this [1] seems to indicate ROS wants to stay compatible with c++03, so maybe that's not really an option?

[1] https://answers.ros.org/question/216842/ros-using-c-11-how-to-use-with-catkin/

dirk-thomas:
Melodic requires a C++14 compiler. So yes, we can use that function.

dirk-thomas:
Closing in favor of #1631.

gavanderhoorn:
The work that resulted in this PR was supported by ROSIN - ROS-Industrial Quality-Assured Robot Software Components (European Union’s Horizon 2020 research and innovation programme under grant agreement no. 732287).

More information: [rosin-project.eu](http://rosin-project.eu).

PR #1549, opened 2018-11-27
Topic melodic devel uds support

fujitatomoya:
Hi

this is the Unix Domain Socket feature extension for ROS message transmission layer. if the system internal message use only, TCP/IP, UDP/IP protocal stack does not do any good to the system, just a pure overhead. so then Unix Domain Socket (UDS) is reasonable to get rid of the overhead with the system internal use for ROS Inter-Process Communication with good affinity of current code.

rebase on default-branch, this requires from https://github.com/ros/ros_comm/pull/1510,
doc/README.md is provide for overview,

if anything else required, please let me know.


fujitatomoya:
@gavanderhoorn 
> Should the files in the debian sub directory be included in this PR?

really, is that true?
i do not see anything like that on source code, it has to be done build time.

gavanderhoorn:
> @gavanderhoorn
> 
> > Should the files in the debian sub directory be included in this PR?
> 
> really, is that true?
> i do not see anything like that on source code, it has to be done build time.

No, I was mistaken. I either mixed up two PRs, or saw it in one of your previous PRs.

In any case, that was the reason why I deleted my comment.

dirk-thomas:
I marked the PR with "consider-future-release" since it clearly breaks ABI compatibility in various places.

For a potential contribution of that significance there is by far not enough information to review the contribution. A few items where additional context is required: 
* High level description on how this feature integrates into the existing system.
  * Does it change the current behavior / how is the new transport being selected?
* Interoperability with other language clients?
* Cross platform considerations, what happens on e.g. Windows?
* A large number of unit tests is failing with this patch.

This is the same request as on the previous PR: https://github.com/ros/ros_comm/pull/1510#issuecomment-430725562

fujitatomoya:
@dirk-thomas 

http://build.ros.org/job/Mpr_ds__ros_comm__debian_stretch_amd64/379/
seems like Jenkins does not work as expected.
the other jobs also fails as following.
http://build.ros.org/job/Mpr_ds__ros_comm__debian_stretch_amd64/375
http://build.ros.org/job/Mpr_ds__ros_comm__debian_stretch_amd64/377

the rest of your comments, i will add more description for each of them.

thanks,

fujitatomoya:
@dirk-thomas 

thanks for your comment and taking care of this thread,
here is the current behavior and high level description.

* High level description on how this feature integrates into the existing system.
  * if the system  internal message use only, TCP/IP, UDP/IP protocal stack does not do
    any good to the system, just a pure overhead. so then Unix Domain Socket (UDS) is 
    reasonable to get rid of the overhead with the system  internal use for ROS
    Inter-Process Communication.
  * The trasport class is selected via ip address, if it's the same ip address,
    it can detect the connection is in the same host, then UDSROS (Unix Domain Socket)
    is selected. otherwise, it will use TCPROS/UDPROS as before.
  * XMLRPC is extended to take care for unix domain socket path name. but it depends on
    the client if this field is used or not. if the client is UDSROS aware, it can use
    unix domain socket path to connect. but if it is not, just use IP and port as before.
  * this feature does not change current behavior, it just detects if it's localhost or not
    to connect, and if unix domain socket is available, use the UDSROS to connect.
* Interoperability with other language clients?
  * roscpp and rospy supported.
  * it can keep the interoperability with other languages. (confirmed rosjava)
    note that the new xmlrpc handler "requestTopicUds" is not suppoted for other languages,
    so it detects "no such handler" exception but connection established via TCPROS/UDPROS.
* Cross platform considerations, what happens on e.g. Windows?
  * this feature is only effective for linux platform.

thanks,
Tomoya

dirk-thomas:
This needs to be rebased to resolve conflicts.

dirk-thomas:
@fujitatomoya Closing due to no further reply. Please feel free to open a new PR against `noetic-devel`.

PR #1557, opened 2018-12-06
Fix nodes spins rather than blocking and waiting, using 100% CPU [kinetic]

### LINKED ISSUE #1343: Regression in 1.12.13: rosout node spins rather than blocking and waiting, using 100% CPU
Introduced in faab4cbc63d6c750c6dc6d050f6268dd19a2bc68. This is a bug in the roscpp client.

In `callAvailable()` in callback_queue.cpp, if no callback is available, `wait_for()` (located in boost_161_pthread_condition_variable_fwd.h) is called on the condition variable, and is told to wait for a specified amount of time (usually 0.1 seconds). However, instead of waiting for the specified amount of time, it returns immediately, turning this into a spin wait rather than a blocking wait. This leads to the rosout node chewing up 100% of the CPU core.

meyerj:
Fixes #1343, and other potential issues. ~~Most likely `SteadyTimer` never has been steady.~~

As [commented](https://github.com/ros/ros_comm/issues/1343#issuecomment-435545310) in the original issue, there is no guarantee that the compiler will inline the calls to `boost::condition_variable` constructor, `boost::condition_variable::timed_wait()` and `boost::condition_variable::wait_for()` and it might or might not use the backported version from Boost 1.61 introduced in https://github.com/ros/ros_comm/pull/1014. The namespace has not been changed and even within roscpp some compilation units used the patched and some used the unpatched implementation (by including or not including `boost_161_condition_variable.h`). If those methods are mixed, because one gets inlined and the other not, or because the same instance is used in different compilation units with different includes and preprocessor macros defined, the underlying pthread condition variable gets initialized with `CLOCK_MONOTONIC` but is called with system clock timestamps or vice-versa. This situation ultimately leads to the 100% busy-wait problem in `CallbackQueue::callAvailable()` as reported in #1343, if the `CallbackQueue` constructor for whatever reason calls the `boost::condition_variable` constructor from a compilation unit where `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` was not defined.

The first patch resolves this link ambiguity by changing the namespace of the included `condition_variable` and `condition_variable_any` implementations to `boost_161`. New preprocessor macros `ROSCPP_BOOST_CONDITION_VARIABLE_HEADER` and `ROSCPP_BOOST_CONDITION_VARIABLE` have been introduced and exposed in `ros/common.h` to select the implementation when `roscpp` is compiled. Because the memory layout does not change, I would carefully claim that the patch is fully ABI compatible. I copied `libroscpp.so` compiled from a patched workspace to `/opt/ros/kinetic/lib` and used it as a drop-in replacement over the past days, without having to recompile any other package installed from debians.

The second patch, 56ecdfa, fixes an issue with ROS timers that popped up in combination with the first. The `boost::condition_variable::timed_wait()` methods are considered deprecated and uses `boost::get_system_time()` (system clock) unconditionally to calculate the absolute timestamp [here](https://github.com/ros/ros_comm/blob/fb6a6bd2839206ab99994a0d0beb89a6ffab1223/clients/roscpp/include/boost_161_pthread_condition_variable_fwd.h#L169) - as called by [TimerManager::threadFunc()](https://github.com/ros/ros_comm/blob/fb6a6bd2839206ab99994a0d0beb89a6ffab1223/clients/roscpp/include/ros/timer_manager.h#L580). If the condition variable would have been constructed as expected, with the implementation from Boost 1.61 or above and `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` defined, this call blocks the thread almost indefinitely and breaks timers (failing unit tests in `roscpp_tests`) - the inverse problem. It has already been reported [here](https://svn.boost.org/trac10/ticket/12728) and marked as wontfix. The solution is to switch to the Boost Chrono interface as already done in https://github.com/ros/ros_comm/pull/1250 for [ros::CallbackQueue](https://github.com/ros/ros_comm/pull/1250/files#diff-5f391de72dc7aa5abcad51904f99a29a). I have no good explanation why this problem has not been triggered before. It only makes sense if the backported constructor of `boost::condition_variable` would have never been called because of the order of source files in the `roscpp` library and hence `SteadyTimer` never has been fully steady (#1014). Any better ideas?

~~Last but not least, fdca164 applies the patch from #1464 to the other timer types and adds `WallTimer::hasStarted()` and `SteadyTimer::hasStarted()`, just for completeness and to match the three interfaces except for the time, duration and event types.~~ (replaced by #1565)

To be absolutely certain that `TimerManager<T, D, E>` will never be instantiated outside of a compilation unit that defines `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` (see [timer_manager.h:215](https://github.com/ros/ros_comm/pull/1557/files#diff-3decb5e6adcb333e676de1a4812e1378R215)), I suggest to move the implementation of the constructor and member functions to a source file `timer_manager.cpp` with [explicit template instantiation](https://en.cppreference.com/w/cpp/language/class_template#Explicit_instantiation) for the three timer types. I tested this approach to exclude that `TimerManager<T, D, E>` and the internal condition variable is constructed elsewhere, without the definition of BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC. But at the end I reverted that patch because this was not the case. An alternative, but more intrusive solution would be to move the `timer_manager.h` header to the `src` folder and to not expose it at all.

Sorry for targeting `kinetic-devel` although it should probably be `melodic-devel`, but this is the version I have been working on. In general, the patch applies to all versions since kinetic (not tested yet).


fujitatomoya:
confirmed https://github.com/ros/ros_comm/issues/1343 is fixed with this patch.

flixr:
Thanks a lot for digging into this!
Looks like a much better/cleaner solution to me, will try to also test this asap.

clients/roscpp/include/ros/steady_timer.h
```diff
@@ -67,9 +67,11 @@ class ROSCPP_DECL SteadyTimer
 
   /**
    * \brief Set the period of this timer
+   * \param reset Whether to reset the timer. If true, timer ignores elapsed time and next cb occurs at now()+period
    */
   void setPeriod(const WallDuration& period, bool reset=true);
 
+  bool hasStarted() const { return impl_->hasStarted(); }
```
flixr: The corresponding `Impl::hasStarted()` seem to be still missing in steady_timer.cpp

flixr:
> Fixes #1343, and other potential issues. Most likely `SteadyTimer` never has been steady.

Maybe I never ran into any issues like this, because I'm always using Release builds and then the SteadyTimer works as expected.

> Sorry for targeting `kinetic-devel` although it should probably be `melodic-devel`

Melodic (or rather Ubuntu bionic) already has boost 1.65, so at least the problem with the backported headers should not arise there.

clients/roscpp/include/ros/steady_timer.h
meyerj: Thanks, fixed in https://github.com/ros/ros_comm/pull/1557/commits/c37a75bedac9945b640c10c83e5609625e746cbc.

meyerj:
> Maybe I never ran into any issues like this, because I'm always using Release builds and then the SteadyTimer works as expected.

I overlooked the specialization of `TimerManager<T,D,E>::threadFunc()` in [steady_timer.cpp](https://github.com/ros/ros_comm/blob/c37a75bedac9945b640c10c83e5609625e746cbc/clients/roscpp/src/libros/steady_timer.cpp#L34-L100). This one already used the Boost Chrono interface and was working as expected.

However, the code duplication is not required anymore now that the general definition of `threadFunc()` in [timer_manager.h](https://github.com/ros/ros_comm/blob/bb1c24f884a48948283d6992eec65d4de4adffc3/clients/roscpp/include/ros/timer_manager.h#L494-L598) also uses Boost Chrono. It has some additional checks whether time jumped backwards which are not required for the steady timer, but they also should not hurt. I merged the two implementations in https://github.com/ros/ros_comm/pull/1557/commits/bb1c24f884a48948283d6992eec65d4de4adffc3.

flixr:
I guess it would actually make sense to create a separate PR for the added `hasStarted` methods...
They have nothing to do with the rest and can then more easily be reviewed/merged separately.

meyerj:
> I guess it would actually make sense to create a separate PR for the added `hasStarted` methods...
> They have nothing to do with the rest and can then more easily be reviewed/merged separately.

Done: https://github.com/ros/ros_comm/pull/1565

meyerj:
Result of running [abi-compliance-checker](https://lvc.github.io/abi-compliance-checker/) on `libroscpp.so` compiled from [1.12.14](https://github.com/ros/ros_comm/tree/1.12.14) and [fix-1343](https://github.com/meyerj/ros_comm/tree/fix-1343) (this PR) on Ubuntu Xenial, compiled in `RelWithDebInfo` mode:
https://gistpreview.github.io/?e7db43ad2f12f9c44a1df0336983d131/compat_report.html

As expected, the type of two condition variable members has been changed, but because the memory layout did not, this should not be a problem.

meyerj:
Regarding the failing unit test:
```
13:19:44 [ RUN      ] CallbackQueue.raceConditionCallback
13:19:45 /tmp/ws/src/ros_comm/test/test_roscpp/test/test_callback_queue.cpp:466: Failure
13:19:45 Value of: condition_object.condition_one.load()
13:19:45   Actual: true
13:19:45 Expected: false
13:19:45 [  FAILED  ] CallbackQueue.raceConditionCallback (61 ms)
```
(http://build.ros.org/job/Kpr__ros_comm__ubuntu_xenial_amd64/464)

I could reproduce the same failure locally when running with `kinetic-devel` (fb6a6bd2839206ab99994a0d0beb89a6ffab1223). The test case fails approx. 1 out of 20 times. The test case was introduced in https://github.com/ros/ros_comm/pull/1054. From the description, I would assume that it is not related to #1343 or this patch.

Ah, kinetic is missing the patch from 507299f8b7bc4a9e888f59ed1e3edeb85c1845bb. Are there plans to backport bug fixes from lunar or melodic to kinetic or even indigo?

meyerj:
> Ah, kinetic is missing the patch from 507299f. Are there plans to backport bug fixes from lunar or melodic to kinetic or even indigo?

I opened another PR to backport this one: https://github.com/ros/ros_comm/pull/1578

dirk-thomas:
> Sorry for targeting kinetic-devel although it should probably be melodic-devel

Please retarget the `melodic-devel` branch with this PR. In general patches are only accepted to the default branch (except if a patch is only necessary for older distros).

@meyerj Can you please check #1608 and comment how this patch compares to the other one.

dirk-thomas:
@meyerj Friendly ping.

ahoarau:
I want to try this PR. Do I need to recompile ros-base from source and link everything ? Or can I add an overlay suffices ?

dirk-thomas:
@ahoarau An overlay should be sufficient.

ahoarau:
Does not seem to like the PR with a c++14 library (unique_lock from boost/std)
```
install/include/boost_161_pthread_condition_variable_fwd.h:52:13: error: reference to ‘unique_lock’ is ambiguous
             unique_lock<mutex>& lock,
```

peci1:
This PR is against ROS Kinetic, which does not officially support C++14, if I'm not mistaken

ahoarau:
c++11 causes the same problems. Very identical compatibility

ahoarau:
I removed all the using namespace, and now it builds fine.
[fix-fix-1343.tar.gz](https://github.com/ros/ros_comm/files/2901784/fix-fix-1343.tar.gz)

dirk-thomas:
@meyerj Another friendly ping. If this should be included in an upcoming release please consider rebasing this against `melodic-devel`.

meyerj:
I merged the patch from @ahoarau in https://github.com/meyerj/ros_comm/pull/1, which seems to solve the C++11/C++14 compatibility issues according to https://github.com/ros/ros_comm/pull/1557#issuecomment-467109009. Thanks for the patch!

Unfortunately I do not have the time at the moment to test this myself or to compare with the solution proposed in https://github.com/ros/ros_comm/pull/1608.


> Please retarget the melodic-devel branch with this PR. In general patches are only accepted to the default branch (except if a patch is only necessary for older distros).

I opened another PR against `melodic-devel`: https://github.com/ros/ros_comm/pull/1651

However, if melodic normally demands at least Boost 1.62 according to [REP-3](http://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023), it might not be necessary to apply the patch there and we could rather remove the Boost 1.61 backports only required for older versions completely. See my comment in the new PR.

flixr:
We can close this in favor of #1651

meyerj:
> We can close this in favor of #1651

The backport of #1651 to `kinetic-devel` is non-trivial. I am fine with closing this PR, but I would be happy if the patch could be backported and released in ROS kinetic, too. In that case the diff here could serve as a reference.

meyerj:
See https://github.com/ros/ros_comm/pull/2011 for a newer patch targeting `kinetic-devel` that is supposed to fix the same issue.

PR #1565, opened 2018-12-14
Add hasStarted() const to WallTimer and SteadyTimer API

meyerj:
Completes https://github.com/ros/ros_comm/pull/1464 (_Add hasStarted() const to Timer API_):
Also add `hasStarted()` method to `ros::WallTimer` and `ros::SteadyTimer`, such that the timer APIs are identical.

Whitespace changes are the result of merging [steady_timer.cpp](https://github.com/ros/ros_comm/blob/d0021d822f7b3d56ec488ef6fa45943c4fbc64c3/clients/roscpp/src/libros/steady_timer.cpp) with [timer.cpp](https://github.com/ros/ros_comm/blob/d0021d822f7b3d56ec488ef6fa45943c4fbc64c3/clients/roscpp/src/libros/timer.cpp) and [wall_timer.cpp](https://github.com/ros/ros_comm/blob/d0021d822f7b3d56ec488ef6fa45943c4fbc64c3/clients/roscpp/src/libros/wall_timer.cpp). In a future release it would be possible to define a template class `TimerBase<T,D,E>` and three `typedef`s for `Timer`, `WallTimer` and `SteadyTimer` to avoid code duplication.

The patch is fully ABI-compatible because it only adds two new non-virtual member functions.

Addresses https://github.com/ros/ros_comm/pull/1557#issuecomment-447349467:
> I guess it would actually make sense to create a separate PR for the added hasStarted methods...
They have nothing to do with the rest and can then more easily be reviewed/merged separately.

### DIFF
--- a/clients/roscpp/include/ros/steady_timer.h
+++ b/clients/roscpp/include/ros/steady_timer.h
@@ -67,9 +67,11 @@ public:
 
   /**
    * \brief Set the period of this timer
+   * \param reset Whether to reset the timer. If true, timer ignores elapsed time and next cb occurs at now()+period
    */
   void setPeriod(const WallDuration& period, bool reset=true);
 
+  bool hasStarted() const { return impl_->hasStarted(); }
   bool isValid() { return impl_ && impl_->isValid(); }
   operator void*() { return isValid() ? (void *) 1 : (void *) 0; }
 
@@ -97,6 +99,7 @@ private:
     Impl();
     ~Impl();
 
+    bool hasStarted() const;
     bool isValid();
     bool hasPending();
     void setPeriod(const WallDuration &period, bool reset=true);
--- a/clients/roscpp/include/ros/steady_timer.h
+++ b/clients/roscpp/include/ros/steady_timer.h
@@ -71,7 +71,7 @@ public:
    */
   void setPeriod(const WallDuration& period, bool reset=true);
 
-  bool hasStarted() const { return impl_->hasStarted(); }
+  bool hasStarted() const { return impl_ && impl_->hasStarted(); }
   bool isValid() { return impl_ && impl_->isValid(); }
   operator void*() { return isValid() ? (void *) 1 : (void *) 0; }
 
--- a/clients/roscpp/include/ros/wall_timer.h
+++ b/clients/roscpp/include/ros/wall_timer.h
@@ -71,6 +71,7 @@ public:
    */
   void setPeriod(const WallDuration& period, bool reset=true);
 
+  bool hasStarted() const { return impl_->hasStarted(); }
   bool isValid() { return impl_ && impl_->isValid(); }
   operator void*() { return isValid() ? (void*)1 : (void*)0; }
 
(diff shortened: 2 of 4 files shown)


clients/roscpp/include/ros/steady_timer.h
```diff
@@ -67,9 +67,11 @@ class ROSCPP_DECL SteadyTimer
 
   /**
    * \brief Set the period of this timer
+   * \param reset Whether to reset the timer. If true, timer ignores elapsed time and next cb occurs at now()+period
    */
   void setPeriod(const WallDuration& period, bool reset=true);
 
+  bool hasStarted() const { return impl_->hasStarted(); }
```
flixr: should probably be
```suggestion
  bool hasStarted() const { return impl_ && impl_->hasStarted(); }
```
see also #1541

meyerj: I currently do not see how a `SteadyTimer`, `WallTimer` or `Timer` instance can exist with an invalid `impl_` pointer, because it is always initialized in the constructor and never reset until destruction. But you and @kunaltyagi are right that either all or none of the methods must have a nullptr check.

I will update this PR after #1541 gets merged.

flixr: #1541 was merged...

meyerj: Updated.

kunaltyagi: https://github.com/ros/ros_comm/blob/melodic-devel/clients/roscpp/include/ros/timer.h#L49

This line doesn't create an impl. I've yet to see the header file of other timers, but I except similar code decisions.

meyerj: You are right. I did not notice the definition in the header.

The three headers are completely identical with this patch, apart from type names, includes and copyright headers. In combination with #1557 the sources will be identical, too.

dirk-thomas:
Please avoid unrelated white space changes as well as moving code unnecessarily. Both just make backporting changes across branches more difficult.

meyerj:
> Please avoid unrelated white space changes as well as moving code unnecessarily. Both just make backporting changes across branches more difficult.

The goal was to unify whitespace between [steady_timer.cpp](https://github.com/ros/ros_comm/blob/d0021d822f7b3d56ec488ef6fa45943c4fbc64c3/clients/roscpp/src/libros/steady_timer.cpp) with [timer.cpp](https://github.com/ros/ros_comm/blob/d0021d822f7b3d56ec488ef6fa45943c4fbc64c3/clients/roscpp/src/libros/timer.cpp) and [wall_timer.cpp](https://github.com/ros/ros_comm/blob/d0021d822f7b3d56ec488ef6fa45943c4fbc64c3/clients/roscpp/src/libros/wall_timer.cpp), which are almost identical and only have been copied over the years. Some patches, like https://github.com/ros/ros_comm/pull/1464, have only been applied to one of them, although they actually apply to all three implementations. The unified whitespace helps to compare those three on the same branch.

I will minimize the diff as requested.

dirk-thomas:

The CI failures are due to tests known to be flaky :disappointed:

PR #1568, opened 2018-12-24
Change error message. Fixes #1532

### LINKED ISSUE #1532: roslaunch gives deceptive error if chmod +x isn't set
Trying to `roslaunch pack_name launcher.launch` fails with the following error if your source file exists in the appropriate spot, but does not have `+x` permission set. 

```
ERROR: cannot launch node of type [pack_name/node_name.py]: 
can't locate node [node_name.py] in package [pack_name]
```

Looks like it is from: 

https://github.com/ros/ros_comm/blob/eb7e686a9a58109087c956eed3cfa6f2ed3eec3d/tools/roslaunch/src/roslaunch/node_args.py#L256

I spent a lot of time mucking around with path, ENV, packages.xml, namespaces, and the launch file, thinking it is a problem with a path or naming scheme. Once you know what you are looking for, it is obvious from questions on the wiki that this is a [commonly](https://answers.ros.org/question/244463/errorcannot-launch-node-of-typetesttestcannot-locate-node-test-in-package-test/) [encountered](https://answers.ros.org/question/209426/i-cant-launch-a-node/) [problem](https://answers.ros.org/question/241690/roslaunch-error-error-cannot-launch-node-of-type/). 

If the file is `+x` but lacks a shebang, the more helpful error occurs:

```
Unable to launch [foo/image_nexus_node-2]. 
If it is a script, you may be missing a '#!' declaration at the top.
```
Could someone add a more useful helper text here to indicate that this might be a permission error and not a path error? Perhaps: 

```
raise NodeParamsException("can't locate node [%s] in package [%s]. 
Make sure file exists in package path and permission is set to executable (chmod +x)"
%(node.type, node.package))
```

betab0t:
As suggested by @xkortex in #1532, added more useful helper text here to indicate that this might be a permission error.

### DIFF
--- a/tools/roslaunch/src/roslaunch/node_args.py
+++ b/tools/roslaunch/src/roslaunch/node_args.py
@@ -253,7 +253,7 @@ def create_local_process_args(node, machine, env=None):
         # multiple nodes, invalid package
         raise NodeParamsException(str(e))
     if not matches:
-        raise NodeParamsException("can't locate node [%s] in package [%s]"%(node.type, node.package))
+        raise NodeParamsException("Cannot locate node of type [%s] in package [%s]. Make sure file exists in package path and permission is set to executable (chmod +x)"%(node.type, node.package))
     else:
         # old behavior was to take first, do we want to change this in Fuerte-style?
         cmd = matches[0]


dirk-thomas:

dirk-thomas:
Thanks for the improvement.

PR #1570, opened 2019-01-04
Solved static linking problems against rosbag_storage

ivanpauno:
BZip2 wasn't exported in rosbag_storage DEPENDS list, so static linking against it failed.
This solves the issue.


dirk-thomas:
I tried to also update the case in the `find_package()` call but that doesn't work. The CMake config file seems to be case sensitive and the name which it needs to be found with is different than the case of the variables it defines.

Since the exported `DEPENDS BZIP2` will result in `BZIP2` to be `find_package()`-ed when `rosbag_storage` is being `find_package()`-ed I don't think this is working as expected.

dirk-thomas:

PR #1572, opened 2019-01-08
Fix rostopic pub from file and stdin.

LeroyR:
I dont know wtf this was supposed to do but it did not work.

publishing from file only send each message once and then stop.
It will now loop over the messages and honor the --once option.


DavidPL1:
Cherry-pick 5d2ad73 from melodic branch. This should keep the test from failing

PR #1573, opened 2019-01-08
Added dependencies and Docker container alerts to manual test. Fixes #1567

### LINKED ISSUE #1567: sudo usage in test_udp_with_dropped_packets.sh
`ros_comm/test/test_roscpp/test/scripts/test_udp_with_dropped_packets.sh` fails to run on a ROS container in docker.
`sudo` is unavailable in the docker container (the default user is already super user).

betab0t:
Avoid #1567 by checking all dependencies are installed and alert if not. also, added a check if script is running in docker container to make sure --privileged flag is used(required for iptables to work properly)...


dirk-thomas:

PR #1576, opened 2019-01-10
fix compiler warnings about unused variables (backport to kinetic-devel)

meyerj:
Cherry-pick of https://github.com/ros/ros_comm/pull/1428 to fix PR builds that target `kinetic-devel`, in particular:
- https://github.com/ros/ros_comm/pull/1557 (http://build.ros.org/job/Kpr__ros_comm__ubuntu_xenial_amd64/461/)

### DIFF
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -604,8 +604,6 @@ void shutdown()
     XMLRPCManager::instance()->shutdown();
   }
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
The patch was not backported to Lunar as part of #1477 since the compiler warning doesn't appear on Lunar. As such it was never consider for backporting from Lunar to Kinetic.

Thanks for the patch. See also #1610.

PR #1577, opened 2019-01-10
Fix dynamic windowing for Topic Statistics

### LINKED ISSUE #1562: Topic Statistics reports 0 for topics under 1Hz
When `/enable_statistics` parameter is `true`, Subscribers publish to the `/statistics` topic. For a topic that has an expected 0.2Hz publishing frequency, this feature always reports `msg.mean_period` as 0, no matter how much time is given.

ghost:
Fix dynamic windowing for Topic Statistics to accurately adapt to slow or fast topics.

fixes #1562 (Topic statistics report 0Hz for topics under 1Hz)


ghost:
The test that failed in above build on stretch (topic_tools.rosunit-throttle_hztest_explicit/test_hz) seems to have some probability of failing already, and from looking through it, seems like it couldn't have been affected by my change, since that test does not enable topic statistics.

dirk-thomas:
Instead of breaking API to invert the semantic of the member variable and update all cases where it is being used it should be much less intrusive to invert the actual condition in this line: https://github.com/ros/ros_comm/pull/1577/files#diff-fafa872049c30ea0ac4fa62e11cd5ddeR262 ?

emersonknapp:
(merged eknapp-az account with this my main account)

Thanks for taking a look at this! Sorry for my slow update - just moved states and this fell by the wayside.

The linked `publish?` check isn't the only problematic part - the variable `pub_frequency` was used in all cases in the implementation as a period, not a frequency.
* it is added to a time as a duration
* it is compared directly to parameters called `window`, measured in seconds

In the interest of keeping the API untouched like you mentioned, I have uploaded a new version that corrects the semantics only in the implementation. There are two commits there - one is a failing test case for both rospy and roscpp for topic frequencies, the second is the fix to the clients.

This version is a little less intrusive, but still required fixing more than just the check since the semantics were globally wrong.

emersonknapp:
@dirk-thomas do you have any further thoughts or concerns about this?

dirk-thomas:
This needs a rebase after #1625 was merged.

artivis:
Maybe this PR can be closed (cleanup) since #1695 replaces it.

PR #1582, opened 2019-01-21
Fix string error on windows

randoms:
Check if name is empty before read it, or it will cause 'cannot decrement string iterator before begin' error on Windows in debug mode.


dirk-thomas:
Thank you for the patch.

PR #1583, opened 2019-01-22
add ${PYTHON_EXECUTABLE} to ROSTEST_EXE

kejxu:
this is another issue related to the shebang line on Linux, `rostest` (without extension), as a python script, cannot be executed directly on Windows. To work around it, we add the `${PYTHON_EXECUTABLE}` prefix to make the command executable


dirk-thomas:
Thanks for the patch.

PR #1584, opened 2019-01-23
use resource _retriever if available

kejxu:
use `resource _retriever` in `roslaunch` if the module is available, so that `package://` can be used in launch files

### DIFF
--- a/tools/roslaunch/src/roslaunch/config.py
+++ b/tools/roslaunch/src/roslaunch/config.py
@@ -45,6 +45,7 @@ import rospkg
 import rospkg.distro
 import rosgraph.names
 import rosgraph.network
+import resource_retriever
 
 from .core import Master, local_machine, is_machine_local, RLException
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Please describe why you think another approach to reference package specific files is necessary. [$(find pkg)](http://wiki.ros.org/roslaunch/XML#substitution_args) is an already documented and working way to achieve this.

Also the package doesn't have a dependency on `resource_retriever`. So it can't rely rely on it being present - the conditional import is only a hacky way to workaround this.

kejxu:
checked with our team about the original motivation of this change, and confirmed that this is no longer necessary. To qualify for the "enhancement" tag, this change very likely needs changes in the `resource_retriever` package, which have been identified as unnecessary. There was initially a problem with `find` on Windows, but that has been fixed in other changes.

Our apologies for the confusion, closing this pull request

PR #1585, opened 2019-01-23
update filename list in add_message_files

kejxu:
not all the files under `test/test_rosmaster/msg` are included in [CMakeList.txt](https://github.com/ros/ros_comm/blob/melodic-devel/test/test_rosmaster/CMakeLists.txt#L8). to fix this, add all `.msg` files into the list for `add_message_files()`

### DIFF
--- a/test/test_rosmaster/CMakeLists.txt
+++ b/test/test_rosmaster/CMakeLists.txt
@@ -5,21 +5,7 @@ project(test_rosmaster)
 find_package(catkin REQUIRED COMPONENTS genmsg rostest std_msgs)
 
 if(CATKIN_ENABLE_TESTING)
-  add_message_files(DIRECTORY msg
-    FILES
-    String.msg
-    Arrays.msg
-    CompositeA.msg CompositeB.msg Composite.msg
-    Embed.msg Floats.msg Simple.msg
-    RosmsgA.msg
-    RosmsgB.msg
-    RosmsgC.msg
-    TestArrays.msg
-    TestHeader.msg
-    TestPrimitives.msg
-    TestString.msg
-    TVals.msg
-  )
+  add_message_files(DIRECTORY msg)
   add_service_files(DIRECTORY srv
     FILES
     AddTwoInts.srv
--- a/test/test_rosmaster/CMakeLists.txt
+++ b/test/test_rosmaster/CMakeLists.txt
@@ -5,7 +5,26 @@ project(test_rosmaster)
 find_package(catkin REQUIRED COMPONENTS genmsg rostest std_msgs)
 
 if(CATKIN_ENABLE_TESTING)
-  add_message_files(DIRECTORY msg)
+  add_message_files(DIRECTORY msg
+    FILES
+    Arrays.msg
+    Composite.msg
+    CompositeA.msg
+    CompositeB.msg
+    Embed.msg
+    Empty.msg
+    Floats.msg
+    RosmsgA.msg
+    RosmsgB.msg
+    RosmsgC.msg
+    Simple.msg
+    String.msg
+    TestArrays.msg
+    TestHeader.msg
+    TestPrimitives.msg
+    TestString.msg
+    TVals.msg
+  )
   add_service_files(DIRECTORY srv
     FILES
     AddTwoInts.srv


test/test_rosmaster/CMakeLists.txt
```diff
@@ -5,21 +5,7 @@ project(test_rosmaster)
...
-    RosmsgA.msg
-    RosmsgB.msg
-    RosmsgC.msg
-    TestArrays.msg
-    TestHeader.msg
-    TestPrimitives.msg
-    TestString.msg
-    TVals.msg
-  )
+  add_message_files(DIRECTORY msg)
```
dirk-thomas: > remove the file names list to let `add_message_files` automatically pick up all `.msg` files under the msg directory

Globbing (source) files in CMake is not recommended.

dirk-thomas (commented):
> `Empty.msg` under `test_rosmaster\msg` is not included in the `FILES` list in `CMakeLists.txt`; however, it is referenced in `tools\rosmsg\test\msg\RosmsgB.msg`

`rosmsg` doesn't depend on `test_rosmaster` so it can't use its message anyway. Also `rosmsg` doesn't seem to process the `RosmsgB.msg` file.

Can you please describe what problem you are trying to address with the patch?

kejxu:
> `rosmsg` doesn't depend on `test_rosmaster` so it can't use its message anyway. Also `rosmsg` doesn't seem to process the `RosmsgB.msg` file.

this sounds like a separate issue, maybe a `<test_depend>` is needed in `rosmsg`? because the test cases in `rosmsg` seems to be [using messages from `test_rosmaster`](https://github.com/ros/ros_comm/blob/melodic-devel/tools/rosmsg/test/test_rosmsg.py#L72)

> Can you please describe what problem you are trying to address with the patch?

Just tried `catkin_make_isolated --pkg rosmsg --make-args run_tests` in Ubuntu environment, there are failures saying this:
```
Cannot locate message [Empty] in package [test_rosmaster] with paths [['...']]
```

test/test_rosmaster/CMakeLists.txt
kejxu: will change to list all messages

dirk-thomas:
> maybe a `<test_depend>` is needed in `rosmsg`

The `test_*` do not install any of their files. They are even never being released. So `rosmsg` can't depend on it`. The test probably needs to be changed to use a different message from another package it can depend on.

kejxu:
> The test probably needs to be changed to use a different message from another package it can depend on.

Feels like this should be tackled in another issue. Only adding `Empty.msg` to the `FILES` list so all the `.msg` files are included (and minor styling update) in this change.

dirk-thomas:
> Only adding `Empty.msg` to the `FILES` list so all the `.msg` files are included (and minor styling update) in this change.

You haven't answered my previous question:

> Can you please describe what problem you are trying to address with the patch?

kejxu:
> Can you please describe what problem you are trying to address with the patch?

Thanks for the prompt response! My apologies for the confusion, have been working on other pull requests too, and I thought we covered this earlier. The motivation for this change is because of the following test error in the `rosmsg` package:
```
ERROR: test_get_msg_text (test_rosmsg.TestRosmsg)
  File "/home/kejxu/ros_dev/src/ros_comm/tools/rosmsg/test/test_rosmsg.py", line 82, in test_get_msg_text
    self.assertEquals(text, rosmsg.get_msg_text(type_, raw=False))
  File "/home/kejxu/ros_dev/install_isolated/lib/python2.7/dist-packages/rosmsg/__init__.py", line 434, in get_msg_text
    raise ROSMsgException("Unable to load msg [%s]: %s"%(type_, e))
ROSMsgException: Unable to load msg [test_rosmaster/RosmsgB]: Cannot locate message [Empty] in package [test_rosmaster] with paths [['/home/kejxu/ros_dev/install_isolated/share/test_rosmaster/msg']]
```

Is this the best fix? Given we've mentioned that the test should probably use message types from another package, I'm not sure. Yet we think it wouldn't hurt to include all `.msg` files under `test/test_rosmaster/msg` for the `add_message_files` call.

Ubuntu output attached:
before `CMakelist.txt` change: [rosmsg.run_tests.txt](https://github.com/ros/ros_comm/files/2830034/rosmsg.run_tests.txt)
after `CMakelist.txt` change: [rosmsg.run_tests.test_rosmaster-update.txt](https://github.com/ros/ros_comm/files/2830033/rosmsg.run_tests.test_rosmaster-update.txt)

dirk-thomas:
> The motivation for this change is because of the following test error in the rosmsg package:

When do you see this test failure? The CI for this repo reports that all tests are passing. What are the exact steps you are following?

kejxu:
> When do you see this test failure? The CI for this repo reports that all tests are passing. What are the exact steps you are following?

you're right, this test from `rosmsg` has always been passing for CI tests. I am seeing this from a clean local `ros_base` source build on Ubuntu 18.04, but I'm not sure why there is this difference between CI build and my local build. Will create another local clean build and check if this error happens again

yet I still think it would make sense to add `Empty.msg` into the `FILES` list =)

dirk-thomas:
Maybe compare your exact build steps with what the buildfarm is doing to determine if there is any difference.

> I still think it would make sense to add `Empty.msg` into the `FILES` list =)

Only when there is a reason to do so :wink:

dirk-thomas:

PR #1586, opened 2019-01-23
normalize paths before comparison in rosmsg

kejxu:
this is a minor update to the existing `results[0] != path:` comparison:
normalize all paths before comparison or adding them into `paths` to make sure all elements in `paths` are really unique
* `_get_package_paths` will return 2 paths for the same package; for example: `C:\\ros-win\\install_isolated\\share\\test_rosmaster` and `C:/ros-win/install_isolated\\share\\test_rosmaster` will both be returned for `test_rosmaster`, resulting in duplicated output

### DIFF
--- a/tools/rosmsg/src/rosmsg/__init__.py
+++ b/tools/rosmsg/src/rosmsg/__init__.py
@@ -551,7 +551,7 @@ def _get_package_paths(pkgname, rospack):
     path = rospack.get_path(pkgname)
     paths.append(path)
     results = find_in_workspaces(search_dirs=['share'], project=pkgname, first_match_only=True, workspace_to_source_spaces=_catkin_workspace_to_source_spaces, source_path_to_packages=_catkin_source_path_to_packages)
-    if results and path.replace(os.path.sep, '/') != results[0].replace(os.path.sep, '/'):
+    if results and results[0].replace(os.path.sep, '/') != path.replace(os.path.sep, '/'):
         paths.append(results[0])
     return paths
     
--- a/tools/rosmsg/src/rosmsg/__init__.py
+++ b/tools/rosmsg/src/rosmsg/__init__.py
@@ -549,14 +549,15 @@ _catkin_source_path_to_packages = {}
 def _get_package_paths(pkgname, rospack):
     paths = []
     path = rospack.get_path(pkgname)
-    path = os.path.normcase(os.path.normpath(path))
     paths.append(path)
     results = find_in_workspaces(search_dirs=['share'], project=pkgname, first_match_only=True, workspace_to_source_spaces=_catkin_workspace_to_source_spaces, source_path_to_packages=_catkin_source_path_to_packages)
     if results:
-        path_in_workspaces = results[0]
-        path_in_workspaces = os.path.normcase(os.path.normpath(path_in_workspaces))
-        if path_in_workspaces != path:
-            paths.append(results[0])
+        if os.path.sep != '/':
+            if path.replace(os.path.sep, '/') != results[0].replace(os.path.sep, '/'):
+                paths.append(results[0])
+        else:
+            if path_in_workspaces != path:
+                paths.append(results[0])
     return paths
     
 def rosmsg_search(rospack, mode, base_type):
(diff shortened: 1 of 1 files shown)


tools/rosmsg/src/rosmsg/__init__.py
```diff
@@ -549,10 +549,14 @@ def iterate_packages(rospack, mode):
 def _get_package_paths(pkgname, rospack):
     paths = []
     path = rospack.get_path(pkgname)
+    path = os.path.normcase(os.path.normpath(path))
```
dirk-thomas: As mentioned in another of your PRs the usage of `normpath` is problematic since if symlinks are present it might change the semantic of the path. Therefore please remove the usage of it.

tools/rosmsg/src/rosmsg/__init__.py
```diff
@@ -549,10 +549,14 @@ def iterate_packages(rospack, mode):
...
     paths.append(path)
     results = find_in_workspaces(search_dirs=['share'], project=pkgname, first_match_only=True, workspace_to_source_spaces=_catkin_workspace_to_source_spaces, source_path_to_packages=_catkin_source_path_to_packages)
-    if results and results[0] != path:
-        paths.append(results[0])
+    if results:
+        path_in_workspaces = results[0]
+        path_in_workspaces = os.path.normcase(os.path.normpath(path_in_workspaces))
```
dirk-thomas: Please try to avoid too many single use temporary variables.

tools/rosmsg/src/rosmsg/__init__.py
kejxu: will remove `os.path.normpath`

tools/rosmsg/src/rosmsg/__init__.py
kejxu: will remove extra local variable

tools/rosmsg/src/rosmsg/__init__.py
```diff
@@ -551,8 +551,13 @@ def _get_package_paths(pkgname, rospack):
...
     paths.append(path)
     results = find_in_workspaces(search_dirs=['share'], project=pkgname, first_match_only=True, workspace_to_source_spaces=_catkin_workspace_to_source_spaces, source_path_to_packages=_catkin_source_path_to_packages)
-    if results and results[0] != path:
-        paths.append(results[0])
+    if results:
+        if os.path.sep != '/':
+            if path.replace(os.path.sep, '/') != results[0].replace(os.path.sep, '/'):
+                paths.append(results[0])
+        else:
+            if path_in_workspaces != path:
```
dirk-thomas: Undefined variable `path_in_workspaces`?

tools/rosmsg/src/rosmsg/__init__.py
```diff
@@ -551,8 +551,13 @@ def _get_package_paths(pkgname, rospack):
...
+                paths.append(results[0])
```
dirk-thomas: The duplication of control seems unnecessary. This could simple stay a single condition since on platforms where `os.path.sep` is `/` the `replace()` calls are no-ops:

```
if results and results[0].replace(os.path.sep, '/') != path.replace(os.path.sep, '/'):
    paths.append(results[0])
````

File: tools/rosmsg/src/rosmsg/__init__.py (hunk 3 above)
Comment (kejxu) [reply tools/rosmsg/src/rosmsg/__init__.py]: embarrassed, change to use `results[0]`

**Comment by dirk-thomas** [comment]
Thanks for the patch.

**Comment by kejxu** [comment]
@dirk-thomas, thanks for merging this change! While not related to this change, here is something I'd love to share and ask for opinions:

the reason for the test failure behind this change is essentially because 1 path is normalized, and the other one is not. this normalization happens during https://github.com/ros/ros_comm/blob/melodic-devel/tools/rosmsg/src/rosmsg/__init__.py#L551
```
path = rospack.get_path(pkgname)
```

and inside `rospack.get_path`, `abspath` is [called](https://github.com/ros-infrastructure/rospkg/blob/f187910cd7af7e34bbe1d6f386bbc27a9f2cbf01/src/rospkg/rospack.py#L59)
```
path = os.path.abspath(path)
```

from [python documentation](https://docs.python.org/2/library/os.path.html#os.path.abspath):
```
os.path.abspath(path)
Return a normalized absolutized version of the pathname path. On most platforms, this is equivalent to calling the function normpath() as follows: normpath(join(os.getcwd(), path)).
```

I haven't had time to dig in further on the behavior of `abspath` on Linux, but given our experience working with symlinks, I feel like this might be worth looking into. ROS should probably consider to
- normalize all paths (if this would also make sense with the presence of symlinks)
- or, avoid any normalization (which might make sanitizing paths very painful)

dirk-thomas:
Due to the size of the code base I can't suggest either of the two approaches. Just blindly applying normalization doesn't sounds appealing to me. Removing it from that specific instance might have unintended side effects. I guess it would need to be tried...

PR #1587, opened 2019-01-23
include cctype for std::tolower

kejxu:
`rosout.cpp` uses [`std::tolower`](https://github.com/ros/ros_comm/blob/melodic-devel/tools/rosout/rosout.cpp#L91) from `cctype`, adding this header to fix build break on Windows


dirk-thomas:
Thanks for the patch.

PR #1588, opened 2019-01-23
move winsock2.h into .cpp and add macro guard for pollfds

kejxu:
- `winsock2.h` is not used in the header file, move into the `.cpp` file that actually uses it
- the use of `pollfds` is [guarded by macros](https://github.com/ros/ros_comm/blob/melodic-devel/utilities/xmlrpcpp/src/XmlRpcServer.cpp#L31), add `_WINDOWS` macro guard to fix build failure on Windows


dirk-thomas:
Thank you for the patch.

PR #1589, opened 2019-01-23
add python prefix for python scripts when there is no .py extension

kejxu:
similar to https://github.com/ros/ros/pull/197, a lot of python scripts don't have the `.py` extension. To work around this on Windows, add `python` prefix in order to execute/launch these python scripts

### DIFF
--- a/tools/roslaunch/src/roslaunch/node_args.py
+++ b/tools/roslaunch/src/roslaunch/node_args.py
@@ -261,8 +261,9 @@ def create_local_process_args(node, machine, env=None):
         raise NodeParamsException("Cannot locate node of type [%s] in package [%s]"%(node.type, node.package))
     cmd = [cmd]
 
-    # ROS nodes as python scripts without .py extension could run on Linux thanks to the shebang line
-    # add python prefix to execute such kind of scrips on Windows
+    # Python scripts in ROS tend to omit .py extension since they could become executable
+    # by adding a shebang line (#!/usr/bin/env python) in Linux environments
+    # special handle this case by executing the script with the Python executable in Windows environment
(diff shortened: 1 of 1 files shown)


kejxu:
this change is only scoped to Windows platform, not related to the test fail on ubuntu 18.04

tools/roslaunch/src/roslaunch/node_args.py
```diff
@@ -260,7 +260,10 @@ def create_local_process_args(node, machine, env=None):
...
         raise NodeParamsException("Cannot locate node of type [%s] in package [%s]"%(node.type, node.package))
     cmd = [cmd]
-    if sys.platform in ['win32']:
-        if os.path.splitext(cmd[0])[1] == '.py':
-            cmd = ['python'] + cmd
+
+    # Python scripts in ROS tend to omit .py extension since they could become executable
+    # by adding a shebang line (#!/usr/bin/env python) in Linux environments
+    # special handle this case by executing the script with the Python executable in Windows environment
+    if os.name == 'nt' and os.path.splitext(cmd[0])[1].lower() in ['.py', '']:
```
dirk-thomas: Why does it change from `sys.platform` to `os.name`?

kejxu: change back to `if sys.platform in ['win32']`

this was changed to align with other Windows porting changes we made, so that most platform-specific checks would use the same condition (`os.name == 'nt'`). however, `os.name` and `sys.platform` are not that different, and are both determined at compile time, so this might be unnecessary. We thought about using `platform.sys()` instead, since that's determined at runtime, but there are lots of places where `platform.sys()` could/should be used; changing all of them might just lead to too many random pull requests

dirk-thomas:
Thanks for the patch.

PR #1590, opened 2019-01-23
respawn if process died while checking should_respawn()

kejxu:
If process died while checking `should_respawn()`, treat it as over-due and respawn by returning a negative floating point number

### DIFF
--- a/tools/roslaunch/src/roslaunch/pmon.py
+++ b/tools/roslaunch/src/roslaunch/pmon.py
@@ -246,7 +246,9 @@ class Process(object):
         if self.time_of_death is None:
             if self.is_alive():
                 return False
-        return (self.time_of_death + self.respawn_delay) - time.time()
+        due_interval = (self.time_of_death + self.respawn_delay) - time.time()
+        #if due at this moment, treat it as over-due, return a negative value
+        return -0.001 if due_interval == 0.0 else due_interval
 
     def stop(self, errors=None):
         """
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This change doesn't look good to me. Returning 0.0 seems to be the correct thing to do here (which the code already does).

What is the purpose of this change? Did you run into any issues you could describe? Maybe on of the callers isn't interpreting the returned value correctly?

kejxu:
> What is the purpose of this change? Did you run into any issues you could describe? Maybe on of the callers isn't interpreting the returned value correctly?

Checked with our team:
> Yes, we would run into problem when the due_interval is 0 (which mean the process is due at the moment performing the check).  If we return 0.0 then it will be a bool (false), no respawn happens; return a negative value will trigger the respawn.
> This happens on Windows quite often, our time.time() tick resolution is not high enough, in process monitor test case, it fails quite often due to this issue.

dirk-thomas:
> If we return 0.0 then it will be a bool (false), no respawn happens; return a negative value will trigger the respawn.

This describes the symptom but the current patch is not the right way to address the problem. Instead wherever the function is being called its return value should be checked more precisely in order to distinguish `0.0` from `false`.

kejxu:
> This describes the symptom but the current patch is not the right way to address the problem. INstead wherever the function is being called its return value should be checked more precisely in order to distinguish `0.0` from `false`.

thanks for the guidance. we'll revisit the failing test case and make necessary change

dirk-thomas:
Thanks for the patch and for updating it.

dirk-thomas:

PR #1591, opened 2019-01-23
DLL import/export macros update

kejxu:
DLL import/export macros update


dirk-thomas:
Thank you for the patch.

The CI failures on Debian Stretch are only due to two known to be flaky tests.

PR #1592, opened 2019-01-24
fix paths (and regex for paths) comparison issues

kejxu:
- [both filenames](https://github.com/ros/ros_comm/blob/melodic-devel/tools/rosgraph/src/rosgraph/roslogging.py#L65) need to be normalized before comparison
- for platforms that use `\` as path separator (Windows, Mac OS), need to escape `os.path.sep` for regular expression comparison

### DIFF
--- a/test/test_rospy/test/unit/test_rospy_core.py
+++ b/test/test_rospy/test/unit/test_rospy_core.py
@@ -116,6 +116,8 @@ class TestRospyCore(unittest.TestCase):
             base, ext = os.path.splitext(this_file)
             if ext == '.pyc':
                 this_file = base + '.py'
+            if os.sep == '\\':
+                this_file = this_file.replace(os.sep, r'\\')
 
             try:
                 # hack to replace the stream handler with a debug version
--- a/test/test_rospy/test/unit/test_rospy_core.py
+++ b/test/test_rospy/test/unit/test_rospy_core.py
@@ -116,7 +116,6 @@ class TestRospyCore(unittest.TestCase):
             base, ext = os.path.splitext(this_file)
             if ext == '.pyc':
                 this_file = base + '.py'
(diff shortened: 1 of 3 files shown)


test/test_rospy/test/unit/test_rospy_core.py
```diff
@@ -116,6 +116,8 @@ def test_loggers(self):
             base, ext = os.path.splitext(this_file)
             if ext == '.pyc':
                 this_file = base + '.py'
+            if os.path.sep == '\\':
+                this_file = this_file.replace(os.path.sep, r'\\')
```
dirk-thomas: Wouldn't it make more sense to `re.escape(this_file)`?

Same below.

kejxu: thanks for sharing this with us! will change to use `re.escape`

test/test_rospy/test/unit/test_rospy_core.py
```diff
@@ -116,6 +116,7 @@ def test_loggers(self):
             base, ext = os.path.splitext(this_file)
             if ext == '.pyc':
                 this_file = base + '.py'
+            this_file = re.escape(this_file)
```
dirk-thomas: Please apply the escaping where it is being used - in this case directly in line 182.

tools/rosgraph/test/test_roslogging.py
```diff
@@ -92,6 +93,7 @@ def test_rosconsole__logging_format():
         base, ext = os.path.splitext(this_file)
         if ext == '.pyc':
             this_file = base + '.py'
+        this_file = re.escape(this_file)
```
dirk-thomas: Please apply the escaping where it is being used - in this case directly in line 114.

dirk-thomas:

PR #1593, opened 2019-01-24
normalize strings to utf-8 before setting as environment variable

kejxu:
normalize strings to utf-8 before setting as environment variable


dirk-thomas:
Thank you for the patch.

PR #1594, opened 2019-01-24
load binary files with Binary() (only)

kejxu:
open binary files in `r` mode and convert to binary format using `Binary` class

### DIFF
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -479,7 +479,7 @@ class Loader(object):
                 from xmlrpc.client import Binary
             except ImportError:
                 from xmlrpclib import Binary
-            with open(binfile, 'rb') as f:
+            with open(binfile, 'r') as f:
                 return Binary(f.read())
         elif command is not None:
             try:


tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -479,7 +479,7 @@ def param_value(self, verbose, name, ptype, value, textfile, binfile, command):
                 from xmlrpc.client import Binary
             except ImportError:
                 from xmlrpclib import Binary
-            with open(binfile, 'rb') as f:
+            with open(binfile, 'r') as f:
```
dirk-thomas, line 482: The change looks wrong to me. The `Binary` constructor expects `bytes` and therefore the file is opened with `rb`. When opening the file with only `r` the read data will be of type `str` and not `bytes`.

dirk-thomas:
Also the ticket does not contain any information what motivates the change in the first place.

tools/roslaunch/src/roslaunch/loader.py
kejxu: thanks for sharing the expertise! revisiting this change to check why it was made in the first place

kejxu:
> Also the ticket does not contain any information what motivates the change in the first place.

the previously mentioned test case passes without problem. my apologies for the confusion, closing the pull request

PR #1595, opened 2019-01-24
Fix race due unprotected access to callbacks_ in roscpp client

igor-semenov:
Access to **callbacks_** in **Publication::peerConnect()** and **Publication::peerDisconnect()** is not protected with mutex, so if other thread removes callback while call to one of these functions, we may fall into SEGFAULT, trying to access via invalid iterator.


dirk-thomas:
Thank you for the fix.

PR #1596, opened 2019-01-24
query ipv6 only if specified

kejxu:
query ipv6 only if `rosgraph.network.use_ipv6()` returns `true`; otherwise, scope the query to ipv4 only using `socket.AF_INET`


dirk-thomas:
Thank you for the patch.

PR #1597, opened 2019-01-27
move check parameter for /rosout_disable_topics_generation only for initialization time

k-okada:
PR #1241 outputs a lot of warning message with a debug log level.
c.f. https://github.com/ros/ros_comm/pull/1241#issuecomment-457920032

other workaround is to check hasParam before retriving a cached parameter.

```
diff --git a/clients/roscpp/src/libros/rosout_appender.cpp b/clients/roscpp/src/libros/rosout_appender.cpp
index f22cd9596..c7adaca73 100644
--- a/clients/roscpp/src/libros/rosout_appender.cpp
+++ b/clients/roscpp/src/libros/rosout_appender.cpp
@@ -106,7 +106,8 @@ void ROSOutAppender::log(::ros::console::Level level, const char* str, const cha
   
   // check parameter server/cache for omit_topics flag
   // the same parameter is checked in rosout.py for the same purpose
-  ros::param::getCached("/rosout_disable_topics_generation", disable_topics_);
+  if (ros::param::has("/rosout_disable_topics_generation"))
+    ros::param::getCached("/rosout_disable_topics_generation", disable_topics_);
 
   if (!disable_topics_){
     this_node::getAdvertisedTopics(msg->topics);
```

### DIFF
--- a/clients/roscpp/src/libros/rosout_appender.cpp
+++ b/clients/roscpp/src/libros/rosout_appender.cpp
@@ -47,6 +47,7 @@ namespace ros
 
 ROSOutAppender::ROSOutAppender()
 : shutting_down_(false)
+, disable_topics_(false)
 , publish_thread_(boost::bind(&ROSOutAppender::logThread, this))
 {
   AdvertiseOptions ops;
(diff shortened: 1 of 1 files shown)


dirk-thomas (commented):
Moving the logic from the `log` function to the constructor changes behavior. Without the patch it is possible to set the parameter afterwards to influence the behavior at runtime. The patch removes that possibility. Therefore I think this shouldn't be applied as is.

paulbovbel:
~If https://github.com/ros/ros_comm/pull/1515 was pulled in, this check could run using the cached parameter API~ please disregard, got mixed up with rospy issues.

flixr:
While this patch changes the behavior, I would still be in favor or merging this.
It is the lesser of two evils...

dirk-thomas:
Resolved via #1823 instead.

PR #1601, opened 2019-01-30
fix various test problems

dirk-thomas:
Basically all cases tried to use resources they don't depend on and there are not guaranteed to be accessible when building / testing only with the declared dependencies.

See https://github.com/ros/ros_comm/pull/863#discussion_r252368448 where one of the cases was introduced. In that case this patch only fixes the testing but the missing dependency at runtime is still a problem.


dirk-thomas:
The CI failure is due to a test known to be flaky.

Issue #1545, opened 2018-11-24
roscpp multithreaded spinners eat up CPU when callbacks take too long


peci1:
When debugging a performance problem with my node, I created an MWE in C++ which shows interesting behavior with serious performance hit. In case you use MultiThreadedSpinner or AsyncSpinner with >1 threads and one of your callbacks takes too long (messages arrive quicker), the spinner eats up a lot of CPU.

I'll post the MWE on Monday, so for now just a textual description. I have a normal node with the MT/Async spinner, I subscribe to one 50 Hz PointCloud2 topic with a single subscriber. The callback contains just `ros::Duration::sleep(time)`.

When running the code with a single thread (either with SingleThreadedSpinner, or even MT spinner with 1 thread), the node takes up about 5% of CPU and just throws away the messages it did not have time to process.

As soon as there are at least 2 threads, CPU consumption goes high (40-400% of CPU) and scales (lineraly?) with the duration I wait in the callback and the number of threads.

My research ended in [`spinner.cpp`](https://github.com/ros/ros_comm/blob/669fbd32d2f92cc295f4b024fcb2f982fddec0f0/clients/roscpp/src/libros/spinner.cpp#L253), where I see that in case of 1 thread, `callAvailable()` is used, whereas with more threads, `callOne()` is used. However, I can't see through all the multi-threading magic in there. I ran a profiler on the node, and most of the time is spent with mutexes and other sync primitives.

I haven't tried yet with `SubscribeOptions.allow_concurrent_callbacks` .

Possibly related:
- https://github.com/ros/ros_comm/issues/1343
- https://answers.ros.org/question/298750/getmtnodehandlegetmtprivatenodehandle-doesnt-result-in-additional-callbacks-but-does-take-more-cpu/?answer=309229#post-id-309229

peci1:
The MWE:

```
#include <ros/ros.h>
#include <sensor_msgs/PointCloud2.h>

class TestClass {
    ros::NodeHandle pubNh;
    ros::Subscriber s;

public:
    TestClass();
    void callback(sensor_msgs::PointCloud2ConstPtr msg);
};

TestClass::TestClass() :
      s(pubNh.subscribe("/scan_point_cloud", 10, &TestClass::callback, this)) {

}

void TestClass::callback(sensor_msgs::PointCloud2ConstPtr msg) {
  ros::Duration(0.01).sleep();
}

int main(int argc, char **argv) {
[...]

PR #1602, opened 2019-01-30
[WIP] Fix subscription busy wait

peci1:
PR showing a solution for bug #1545 .

This is not yet meant to be merged - treat it like I'm offering a solution open for comments.

This PR consists of two commits.

The first one solves the issue in a backwards-compatible way, however I've noticed that it's performance when calling `CallbackQueue::addCallback` rapidly drops compared to upstream (to thousands/sec compared to hundreds of thousands/sec).

The second one takes it further, breaking ABI (but not API, although it adds to it). This solution doesn't suffer from the performance issues, or at least not as much (there is actually some mutex locking added, so it is probably a bit slower, but taking into account that the upstream now fully throttles the CPU...)

What should the tests show: calling `callOne()` too often (as in upstream) is quite expensive, and since upstream enters a busy-wait-loop around `callOne`, it throttles 10 cores to maximum. The fixed versions should be much more friendly to your CPU, which is shown by the fact that it does not call excessive numbers of `callOne` (roughly 2-3 times the number of subscription callbacks pushed to the queue).

Another interesting metric showed to be the number of subscription callbacks that it's possible to push in the queue in the given time-frame. As I stated earlier, the version from first commit has a very strong drop in this metric. It might be interesting to test it with some real pusblisher/subscriber system, but I haven't yet had time to test it...

### DIFF
--- a/clients/roscpp/include/ros/callback_queue_interface.h
+++ b/clients/roscpp/include/ros/callback_queue_interface.h
@@ -36,6 +36,7 @@
 #define ROSCPP_CALLBACK_QUEUE_INTERFACE_H
 
 #include <boost/shared_ptr.hpp>
+#include <boost/thread/condition_variable.hpp>
 #include "common.h"
 #include "ros/types.h"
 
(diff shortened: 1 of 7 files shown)


dirk-thomas:
Please target the current default branch with your patch, currently `melodic-devel`.

peci1:
@dirk-thomas retargeted, #1608 .

dirk-thomas:
Thanks, closing in favor of #1608.

PR #1604, opened 2019-01-31
Use cryto only if cryptodome is not found

CdeMills:
This is a bit safer as cryptodome is favored over crypto. Yet if crypto is the only installed lib, it will be used.


dirk-thomas:

There is no need to support two different dependencies. Closing in favor of one of the many other PRs opened for the same issue.

PR #1605, opened 2019-01-31
Safest approch: use cryptodome over crypto, and ensure os.urandom() i…

CdeMills:
This one is a bit more extensive, and should be the preferred way to go.


dirk-thomas:

Closing as duplicate of #1609.

PR #1606, opened 2019-01-31
Another approach: use python-cryptography instead

CdeMills:
This introduces a dependency w.r.t. https://cryptography.io/en/latest/

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -50,8 +50,14 @@ import threading
 import time
 import yaml
 
-from Crypto import Random
-from Crypto.Cipher import AES
+#from Crypto import Random
+# from Crypto.Cipher import AES
+# taken from Crypto.Cipher.AES.block_size
+AES_BLOCK_SIZE = 16
+
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from cryptography.hazmat.backends import default_backend
+
 import gnupg
 
 try:
@@ -221,15 +227,21 @@ class _ROSBagAesCbcEncryptor(_ROSBagEncryptor):
         f.seek(chunk_data_pos)
         chunk = _read(f, chunk_size)
         # Encrypt chunk
-        iv = Random.new().read(AES.block_size)
+        # iv = Random.new().read(AES.block_size)
+        backend = default_backend()
+        iv = os.urandom(AES_BLOCK_SIZE)
         f.seek(chunk_data_pos)
         f.write(iv)
-        cipher = AES.new(self._symmetric_key, AES.MODE_CBC, iv)
-        encrypted_chunk = cipher.encrypt(_add_padding(chunk))
+        # cipher = AES.new(self._symmetric_key, AES.MODE_CBC, iv)
+        cipher = Cipher(algorithms.AES(self._symmetric_key), modes.CBC(iv), backend=backend)
+        encryptor = cipher.encryptor()
+        # encrypted_chunk = cipher.encrypt(_add_padding(chunk))
+        encrypted_chunk = encryptor.update(_add_padding(chunk)) + encryptor.finalize()
         # Write encrypted chunk
         f.write(encrypted_chunk)
         f.truncate(f.tell())
-        return AES.block_size + len(encrypted_chunk)
+        # return AES.block_size + len(encrypted_chunk)
+        return AES_BLOCK_SIZE + len(encrypted_chunk)
 
     def decrypt_chunk(self, encrypted_chunk):
         """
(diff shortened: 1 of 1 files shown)


mikepurvis:
Is this fully compatible with the original approach? Either way, it doesn't seem like we're gaining much in brevity with this new dependency.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -221,15 +227,21 @@ def encrypt_chunk(self, chunk_size, chunk_data_pos, f):
...
+        # iv = Random.new().read(AES.block_size)
+        backend = default_backend()
+        iv = os.urandom(AES_BLOCK_SIZE)
         f.seek(chunk_data_pos)
         f.write(iv)
-        cipher = AES.new(self._symmetric_key, AES.MODE_CBC, iv)
-        encrypted_chunk = cipher.encrypt(_add_padding(chunk))
+        # cipher = AES.new(self._symmetric_key, AES.MODE_CBC, iv)
+        cipher = Cipher(algorithms.AES(self._symmetric_key), modes.CBC(iv), backend=backend)
+        encryptor = cipher.encryptor()
```
dirk-thomas, line 237: What is the benefit of using this different API?

tools/rosbag/src/rosbag/bag.py
```diff
@@ -221,15 +227,21 @@ def encrypt_chunk(self, chunk_size, chunk_data_pos, f):
         f.seek(chunk_data_pos)
         chunk = _read(f, chunk_size)
         # Encrypt chunk
-        iv = Random.new().read(AES.block_size)
+        # iv = Random.new().read(AES.block_size)
+        backend = default_backend()
+        iv = os.urandom(AES_BLOCK_SIZE)
```
dirk-thomas, line 232: What is the difference of `os.urandom` over `Random`?

dirk-thomas (commented):

The new dependency proposed in this patch (which still needs to be added to the package manifest) is available as of Ubuntu Bionic: [python-pycryptodome](https://packages.ubuntu.com/bionic/python-pycryptodome), [python3-pycryptodome](https://packages.ubuntu.com/bionic/python3-pycryptodome)

Please describe the pros/cons over the proposed patch in #1609.

CdeMills:
Hello,
pro and cons are difficult to tell. Those are two crypto lib aiming at implementing the same functionalities. pyca/cryptography is linked with OpenSSL, while cryptodome isn't. I think the choice should also rely on timing both approaches.
Regards

Pascal

dirk-thomas:
Comparing the statistics of the two GitHub repos I would think [cryptography](https://github.com/pyca/cryptography/) is the more mature and active project compared to [pycryptodome](https://github.com/Legrandin/pycryptodome/).

Therefore I closed the other PR.

@CdeMills Please answer the pending questions when you have a chance and also update the package manifest to add the new dependency (and replace the old one).

tools/rosbag/src/rosbag/bag.py
```diff
@@ -240,14 +252,21 @@ def decrypt_chunk(self, encrypted_chunk):
         @rtype:  str
         @raise ROSBagFormatException: if size of input chunk is not multiple of AES block size
         """
-        if len(encrypted_chunk) % AES.block_size != 0:
+        #        if len(encrypted_chunk) % AES.block_size != 0:
+        if len(encrypted_chunk) % AES_BLOCK_SIZE != 0:
             raise ROSBagFormatException('Error in encrypted chunk size: {}'.format(len(encrypted_chunk)))
-        if len(encrypted_chunk) < AES.block_size:
+        # if len(encrypted_chunk) < AES.block_size:
+        if len(encrypted_chunk) < AES.BLOCK_SIZE:
```
dirk-thomas, line 259: `AES.BLOCK_SIZE` doesn't exist. Either `AES_BLOCK_SIZE` or `AES.block_size`.

tools/rosbag/src/rosbag/bag.py
```diff
@@ -345,7 +368,8 @@ def get_info_str(self):
     def _build_symmetric_key(self):
         if not self._gpg_key_user:
             return
-        self._symmetric_key = Random.new().read(AES.block_size)
+        # self._symmetric_key = Random.new().read(AES.block_size)
+        self._symmetric_key = os.urandom(AES.block_size)
```
dirk-thomas, line 372: Why `AES.block_size` here instead of `AES_BLOCK_SIZE`?

tools/rosbag/src/rosbag/bag.py
```diff
@@ -354,21 +378,30 @@ def _decrypt_encrypted_header(self, f):
...
+        # size -= AES.block_size
+        size -= AES_BLOCK_SIZE
         encrypted_header = _read(f, size)
-        cipher = AES.new(self._symmetric_key, AES.MODE_CBC, iv)
-        header = cipher.decrypt(encrypted_header)
+        # cipher = AES.new(self._symmetric_key, AES.MODE_CBC, iv)
+        cipher = Cipher(algorithms.AES(self._symmetric_key), modes.CBC(iv), backend=backend)
+        decryptor = cipher.decryptor()
+        # header = cipher.decrypt(encrypted_header)
+        header = decryptor_update(encrypted_header) + decryptor_finalize()
```
dirk-thomas, line 398: Should be `decryptor.update(`...

tools/rosbag/src/rosbag/bag.py
```diff
@@ -50,8 +50,14 @@
 import time
 import yaml
 
-from Crypto import Random
-from Crypto.Cipher import AES
+#from Crypto import Random
+# from Crypto.Cipher import AES
+# taken from Crypto.Cipher.AES.block_size
+AES_BLOCK_SIZE = 16
```
dirk-thomas, line 56: Why use this constant over `AES.block_size` (which is `128`)?

dirk-thomas:
The passing CI clearly indicates that non of the touched code is covered by unit tests. Otherwise they would have failed due to the bugs in the current patch.

dirk-thomas:
Closing in favor of #1793 which redos this patch.

PR #1607, opened 2019-01-31
Ensure minimal c++ standard is c++11

CdeMills:
On older platforms (CentOS7 f.i.), by default the C++ compiler standard is c++98, which doesn't know about nullptr. This patch ensures to raise minimum compatibility with c++11


dirk-thomas:
Melodic targets a C++ 14 compiler. If your platform uses a compiler which doesn't enable that standard by default you might either want to update to a newer compiler version or set this compiler flag globally using an environment variable.

Since this change would be necessary on a lot of packages we made the decision on relying on the default standard chosen by the compiler.

PR #1609, opened 2019-01-31
Add pycryptodome as default on melodic

### LINKED ISSUE #1599: roscomm/ros_bag crypto implementation

maximeconnolly:
Fix #1599 

This patch will only work on Ubuntu 18.04 and up (Melodic and up). The package pycryptodome is not available on earlier version.

### DIFF
--- a/tools/rosbag/package.xml
+++ b/tools/rosbag/package.xml
@@ -30,7 +30,7 @@
   <run_depend>boost</run_depend>
   <run_depend>genmsg</run_depend>
   <run_depend>genpy</run_depend>
-  <run_depend>python-crypto</run_depend>
+  <run_depend>python-pycryptodome</run_depend>
   <run_depend>python-gnupg</run_depend>
   <run_depend>python-rospkg</run_depend>
   <run_depend>rosbag_storage</run_depend>
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -50,8 +50,9 @@ import threading
 import time
 import yaml
 
-from Crypto import Random
-from Crypto.Cipher import AES
+from Cryptodome.Cipher import AES
+from Cryptodome.Random import random
+
 import gnupg
 
 try:
@@ -221,7 +222,7 @@ class _ROSBagAesCbcEncryptor(_ROSBagEncryptor):
         f.seek(chunk_data_pos)
         chunk = _read(f, chunk_size)
         # Encrypt chunk
-        iv = Random.new().read(AES.block_size)
+        iv = random.getrandbits(AES.block_size)
         f.seek(chunk_data_pos)
         f.write(iv)
         cipher = AES.new(self._symmetric_key, AES.MODE_CBC, iv)
@@ -299,7 +300,7 @@ class _ROSBagAesCbcEncryptor(_ROSBagEncryptor):
                 v = v.encode()
             header_str += _pack_uint32(len(k) + 1 + len(v)) + k + equal + v
 
-        iv = Random.new().read(AES.block_size)
+        iv = random.getrandbits(AES.block_size)
         enc_str = iv
         cipher = AES.new(self._symmetric_key, AES.MODE_CBC, iv)
         enc_str += cipher.encrypt(_add_padding(header_str))
@@ -345,7 +346,7 @@ class _ROSBagAesCbcEncryptor(_ROSBagEncryptor):
     def _build_symmetric_key(self):
         if not self._gpg_key_user:
             return
-        self._symmetric_key = Random.new().read(AES.block_size)
+        self._symmetric_key = random.getrandbits(AES.block_size)
         self._encrypted_symmetric_key = _encrypt_string_gpg(self._gpg_key_user, self._symmetric_key)
 
     def _decrypt_encrypted_header(self, f):


tools/rosbag/package.xml
```diff
@@ -31,6 +31,7 @@
   <run_depend>genmsg</run_depend>
   <run_depend>genpy</run_depend>
   <run_depend>python-crypto</run_depend>
```
dirk-thomas: I assume this dependency can be removed when only one API is used?

tools/rosbag/package.xml
```diff
@@ -31,6 +31,7 @@
...
+  <run_depend>python-pycryptodome</run_depend>
```
dirk-thomas, line 33: There needs to be a `rosdep` key created for this: https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml

tools/rosbag/src/rosbag/bag.py
```diff
@@ -50,8 +50,14 @@
...
 
-from Crypto import Random
-from Crypto.Cipher import AES
+try:
+    from Cryptodome.Cipher import AES
+    from Cryptodome.Random import random
+
+except ImportError:
+    from Crypto.Cipher import AES
+    from Crypto.Random import random
```
dirk-thomas: There is no need to support two different dependencies. Please remove the usage of `Crypto` in favor of the new dependency.

dirk-thomas (commented):
Please describe the pros/cons over the proposed patch in #1606.

tools/rosbag/package.xml
maximeconnolly: Yes it can, I only need to add the keys to the rosdistro.

maximeconnolly:
I am going to add pycryptodome to the rosdistro and address the reviews.

maximeconnolly:
This propose patch has the advantage of keeping compatibility with pycrypto. One of the cons of taking this route instead of #1606 is that we need to add pycryptodome to the rosdistro.

tools/rosbag/package.xml
maximeconnolly: https://github.com/ros/rosdistro/pull/20128

CdeMills:
Hello,
pro and cons are difficult to tell. Those are two crypto lib aiming at implementing the same functionalities. pyca/cryptography is linked with OpenSSL, while cryptodome isn't. I think the choice should also rely on timing both approaches.
Regards

Pascal

dirk-thomas:
Comparing the statistics of the two GitHub repos I would think [cryptography](https://github.com/pyca/cryptography/) is the more mature and active project compared to [pycryptodome](https://github.com/Legrandin/pycryptodome/).

Therefore will close this PR in favor of #1606.

Thank you for your contribution - even if the patch wasn't merged.

tfoote:
cryptodome is also only available in bionic and newer: https://packages.ubuntu.com/search?keywords=python-pycryptodome

vs cryptography has been packaged since at least xenial: https://packages.ubuntu.com/search?keywords=python-cryptography

This would be a blocker for backporting.

dirk-thomas:
With the uncertainty around the `python-cryptography` PR #1793 I will reopen this since it is a much smaller patch with less chance of regression.

dirk-thomas:
> This would be a blocker for backporting.

Backporting this change to Kinetic is not even on the table so this is not a problem.

cottsay (approved):
LGTM with passing CI

dirk-thomas:

dirk-thomas:

dirk-thomas:

tools/rosbag/src/rosbag/bag.py
```diff
@@ -221,7 +222,7 @@ def encrypt_chunk(self, chunk_size, chunk_data_pos, f):
         f.seek(chunk_data_pos)
         chunk = _read(f, chunk_size)
         # Encrypt chunk
-        iv = Random.new().read(AES.block_size)
+        iv = random.getrandbits(AES.block_size)
```
mikepurvis, line 225: This should have been `get_random_bytes`, not `getrandbits`; PR incoming.

mikepurvis: See #1814.

PR #1614, opened 2019-02-18
RFC: multiple launch files for a single test executable with add_rostest_g{test,mock}

theg4sh:
The base change in usage is allow to run a test executable with multiple launch files, e.g.:
```
	add_rostest_gtest(noname_test
		LAUNCH_FILES
			test/noname_test.launch
			test/noname_bad_arguments_test.launch
		SOURCES test/noname_test.cpp)
```

An extended usage allow to link and build dependencies for a test target, like that:
```
	add_rostest_gtest(noname_test
		LAUNCH_FILES
			test/noname_test.launch
			test/noname_bad_arguments_test.launch
		SOURCES test/noname_test.cpp
		LIBRARIES ${catkin_LIBRARIES} ${GTEST_BOTH_LIBRARIES}
		DEPENDENCIES
			noname_node
			std_msgs_generate_messages_cpp
			mavros_msgs_generate_messages_cpp
			noname_msgs_generate_messages_cpp)
```

There is also added notification when add_rostest_g{test,mock} skips a target due to unsetted GTEST_FOUND or GMOCK_FOUND variables.

### DIFF
--- a/tools/rostest/cmake/rostest-extras.cmake.em
+++ b/tools/rostest/cmake/rostest-extras.cmake.em
@@ -60,29 +60,51 @@ endfunction()
 # This is an internal function, use add_rostest_gtest or
 # add_rostest_gmock instead
 #
-# :param type: "gtest" or "gmock"
-# The remaining arguments are the same as for add_rostest_gtest
+# The function arguments are the same as for add_rostest_gtest
 # and add_rostest_gmock
 #
-function(_add_rostest_google_test type target launch_file)
-  if (NOT "${type}" STREQUAL "gtest" AND NOT "${type}" STREQUAL "gmock")
+function(_add_rostest_google_test target)
+  set(multiValueArgs LAUNCH_FILES SOURCES LIBRARIES DEPENDENCIES)
+  cmake_parse_arguments(_rostest_google_test "DEPRECATED" "TYPE" "${multiValueArgs}" ${ARGN})
+
+  if(NOT "${_rostest_google_test_TYPE}" STREQUAL "gtest" AND NOT "${_rostest_google_test_TYPE}" STREQUAL "gmock")
     message(FATAL_ERROR
-      "Invalid use of _add_rostest_google_test function, "
-      "first argument must be 'gtest' or 'gmock'")
+      "Invalid use of _rostest_google_test function, "
+      "TYPE argument must be 'gtest' or 'gmock'")
     return()
   endif()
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't see why this functionality should be provided by the `add_rostest_gtest` function. It seems much easier to use a `foreach` loop over all launch files in the calling code and invoke one function within the loop. That would also ensure that each test has its own name and can easily be invoked separately. Since `add_rostest_gtest` builds the `gtest` executable it can't be used repeatedly but this could be refactored to provide a single function to be called in the `foreach` loop.

In general please avoid unrelated changes in a single PR. E.g. the "added notification" should not be mixed in the patch but separated in another PR.

Also as far as I understand the patch currently used signatures of the `add_rostest_gtest` / `add_rostest_gmock` functions are being marked as deprecated? I don't see any reason to do so (forcing existing usage to be updated for no apparent reason). So any patch should maintain backward compatibility (without deprecations).

Please also include tests exercising the newly added functionality.

theg4sh:
@dirk-thomas I see.

> It seems much easier to use a foreach loop over all launch files in the calling code and invoke one function within the loop. 

If you mean to call `add_rostest_gtest` inside `foreach`'s body, so, in that case you'll get an build error due to attempting to call `add_executable` for existed target inside of `_add_rosgtest_google_test`.

Different launch files looks useful when testable node (blackbox testing) required to be run with different arguments or yaml configurations, as it happened in my case.

Btw, it's my bad. Deep investigation shows that multiple launch files could be executed with one test target much easier than I thought. The following code do the thing I want:
 ```
add_rostest_gtest(testname_test some.launch testname.cpp)
add_rostest(testname_test someanother.launch)
add_rostest(testname_test someelse.launch)
```

So, due to that I close the issue. Sorry for inconvenience.

jfmi, @dirk-thomas, could you please show an example with a `foreach` inside of `CMakeLists.txt`? :)

dirk-thomas:
> If you mean to call `add_rostest_gtest` inside `foreach`'s body, so, in that case you'll get an build error due to attempting to call `add_executable` for existed target inside of `_add_rosgtest_google_test`.

I mentioned that specific problem in my previous comment. What I was suggesting was to first create the `gtest` executable using [catkin_add_executable_with_gtest](https://github.com/ros/catkin/blob/6157b75b19780278d6892364c98f8f6b4951d50f/cmake/test/gtest.cmake#L93-L109) and the use a `foreach` loop to add multiple tests using `add_rostest`.

PR #1615, opened 2019-02-19
test_rospy: added queue_size arguments to Publishers to avoid warnings

cwecht:


dirk-thomas:
Thanks for the patch.

PR #1616, opened 2019-02-20
add windows.h usage explicitly

kejxu:
add `windows.h` usage explicitly to enable build on Windows


dirk-thomas:
Thanks for the patch.

PR #1617, opened 2019-02-20
fix IOError during Python file operation

kejxu:
fix `IOError` during Python file operation

this error is Windows-specific and hard to catch, adding justification as comments so it's easier to understand


dirk-thomas:
Thanks for the patch.

PR #1618, opened 2019-02-20
update CMakeLists.txt in rosbag_storage

kejxu:
- update `RUNTIME_OUTPUT_DIRECTORY ` for devel space so `.dll` gets binplaced into `rosbag_storage\lib\` instead of `rosbag_storage\bin\`. this might not be necessary if https://github.com/ros/pluginlib/pull/145 is merged (?)
- update `RUNTIME DESTINATION` for `rosbag_storage` according to catkin guidelines
- always export `encryptor_plugins.xml`


dirk-thomas:
Thanks for the patch.

PR #1619, opened 2019-02-20
add POSIX flag for shlex.split()

kejxu:
add POSIX flag for [shlex.split()](https://docs.python.org/2/library/shlex.html) so it operates in non-POSIX mode correctly when on Windows


dirk-thomas:
Thanks for the patch.

PR #1620, opened 2019-02-20
update install destination for roslz4

kejxu:
update install destination for roslz4 according to catkin guidelines


dirk-thomas:
Thanks for the patch.

PR #1622, opened 2019-02-20
fix infinite loop in rosbag buffer resize

ipa-fez:
If you open enough bags and the buffer fills enough, this was looping infinitely because capacity_ would overflow and never be >= capacity

Now it fails with an BagIOException instead because it doesn't read enough bytes.


dirk-thomas:
I will close this in favor of #1623 since the backport will happen from the `melodic-devel` branch together with other commits.

PR #1623, opened 2019-02-20
fix infinite loop in rosbag buffer resize

ipa-fez:
Same as #1622 but for melodic

### DIFF
--- a/tools/rosbag_storage/src/buffer.cpp
+++ b/tools/rosbag_storage/src/buffer.cpp
@@ -35,6 +35,7 @@
 #include <stdlib.h>
 #include <assert.h>
 #include <utility>
+#include <limits>
 
 #include "rosbag/buffer.h"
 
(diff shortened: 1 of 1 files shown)


tools/rosbag_storage/src/buffer.cpp
```diff
@@ -65,7 +66,12 @@ void Buffer::ensureCapacity(uint32_t capacity) {
         capacity_ = capacity;
     else {
         while (capacity_ < capacity)
+        {
+          if (static_cast<uint64_t>(capacity) * 2 >=  std::numeric_limits<uint32_t>::max())
```
dirk-thomas: Nitpick: please remove the second consecutive space after the operator. Also using `>` should be sufficient here.

tools/rosbag_storage/src/buffer.cpp
```diff
@@ -65,7 +66,12 @@ void Buffer::ensureCapacity(uint32_t capacity) {
...
+            capacity_ = capacity;
```
dirk-thomas: Why should this be set to `capacity` and not to `std::numeric_limits<uint32_t>::max()` since the general policy is to grow by a factor of 2? Otherwise this would reallocate every time the capacity is increases by e.g. 1 if it is already greater than `max / 2`.

ipa-fez: Good idea.

dirk-thomas:

Issue #1621, opened 2019-02-20
ros_comm: Bug in statistics.cpp causing Topic Statistics rate under 0.016 Hz for some connections


wentz89:
I have a plot-tool to visualize the statistics datas from rosbags and it seems that for some topic-connections there are only 7 messages over 420 sec on the /statistics topic, but with a pretty large amount of evaluated datas. I know that the statistics calculation uses a dynamic window if to less datas are available. This happened with more than one Topic. 

![_localization_odometry _localization--- _path_control](https://user-images.githubusercontent.com/19651717/53089299-82b09100-350c-11e9-86e7-a21ce2da2e0b.png)

In the picture you can see the data plots i have taken.  In my opinion, however, this behavior is unintentional and unwanted. Thats why i open this issue.

The Problem in the source code is described in one of my comments at the bottom.

wentz89:
Note: The /localization/odometry Topic has 2 other connections. These plots show a similar behavior, with a few more data-points (but not many).

dirk-thomas:
From the description of this ticket it is unclear what exactly your problem is, there is not enough information about what you are doing in order to reproduce what you are seeing and it is unclear if this is actually a bug in the sources of this repo or just related to what / how you are doing.

Therefore I would suggest to ask you question on [answers.ros.org](http://answers.ros.org/) instead. If you have asked your question there please add a link to the question to this ticket so that future readers can find the related question.

If that lead to enough evidence that there is actually a bug please feel free to comment here with additional information and the ticket can be reopened.

tfoote:
Note that default reporting periods are between once every 4 seconds and once every 64 seconds with the goal of 10 to 100 samples per window. https://github.com/ros/ros_comm/blob/acd831359ea245c523d1d12449f52bdac41de229/clients/rospy/src/rospy/impl/statistics.py#L72-L79 https://github.com/ros/ros_comm/blob/acd831359ea245c523d1d12449f52bdac41de229/clients/roscpp/src/libros/statistics.cpp#L46-L50 and they are adjustable by parameter if you'd like to change them.

wentz89:
Thanks for that hint.

But anyway the window show there are more than 6300 evaluated elements per statistics-msg (see Delivered in the Plot).
I just took a look on my other data plots and there are connections that only evaluate 1 or 2 messages on one connection per statistics message (the topic frequency was about 1 Hz).
Also the update frequency for the most publisher-subscriber - connections is 1 Hz.

[...]

PR #1625, opened 2019-02-21
fixed bug in statistics decision making if one should publish

wentz89:


dirk-thomas:
@wentz89 Please provide at least a minimal description when creating a ticket. In this case the PR aims to fix #1621.

kejxu:
great catch! `time lapsed = 1 / frequency`

> I recognized that pub_frequency_ is a double value, which doesn't make sense

why not, frequency could be a floating point number. and it (implicitly) makes sure `1 / frequency` yields the correct value

wentz89:
@kejxu well yes, but actually no ;). I should have written "it doesn't make sense in this case" cause:
1. the program evaluats a time + Duration(frequenzy), when it should be time + Duration(rate)
2. (but this is my opinion) One usually say this publisher should have a frequency of 100 Hz or 1 Hz or 10 Hz. One never hear 10,5 Hz or 100,7 Hz should be the frequency, therefore it could be an integer value

dirk-thomas:
Thanks for patch.

dirk-thomas:
@wentz89 @kejxu Please see #1577 for a more complete patch for this problem. It would be great if either of you could try that patch and comment on it.

Issue #1626, opened 2019-02-22
node topic statistics broken


kikass13:
Hello,

after we tried testing the contents of the /statistics topic via 'rosparam set enable_statistics true' in our robot, our hosts cpu usage went nuts.

After that i tried to debug the roscpp statistics class ... 

1. Original kinetic statistics analysis and bugfix
- i put some chrono timers in the callback function of the statistics class, so that i can visualize them later 
- i started a cpp node with 20 publishers (200Hz TX frequency for each one)
- i started a cpp node with 20 subcribers
- x axis in the graph is the element index, y axis is the processing time of the callback (in us)
- the red circles define the point, where the callback function actually calculates stuff and publishes it's contents onto the /statistics topic

![image](https://user-images.githubusercontent.com/6985609/53243379-efac5e00-36a7-11e9-9a1c-3f00c925d936.png)

As can be seen in the image, the spikes occur when the class calculates and publishes it's contents. **IMPORTANT: the time window is of the publishes is doubled over time (halfing the frequency each time)**

I found the problem here (kinetic and melodic): 
https://github.com/ros/ros_comm/blob/194c776737134a429619f6851f1850734843e210/clients/roscpp/src/libros/statistics.cpp#L116

This line has to be changed to this:
` if (stats_it->second.last_publish + ros::Duration(1/pub_frequency_) < received_time)`

By doing that, we get a much more stable (and not so resource heavy) frequency in the timing analysis:

![image](https://user-images.githubusercontent.com/6985609/53243836-66962680-36a9-11e9-99c5-19e9ab5f837d.png)

2. Performance improvements 

I didn't like the performance and the way this `  std::map<std::string, struct StatData> map_` object was handled inside the callback ... 
so i tried to tweak some things (by only using the original iterator and directly changing the data structures instead of doing multiple (slow, binary tree) map lookups of the topic names. (i guess a `<unordered_map>` is not suitable because there are some code parts, where the order of the elements inside the map matters.

The results, after messing around with the code and changing each access to the map to it's original iterator, can be seen here:

![image](https://user-images.githubusercontent.com/6985609/53244143-4450d880-36aa-11e9-9958-a2b3308b1cbc.png)

> medianTime = 3 us
> which means my code performs around 12 times faster on my machine (intel i5, 2.2GHz, publishers and subsribers on localhost [loopback])

_________________________________________

i can provide a merge request if needed, but i first want to get some feedback on this issue ... 
[...]

PR #1627, opened 2019-02-22
bugfix: statistics related to #1626 #1621

kikass13:
+ (Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz-- x12 times) performance optimization included

see analysis in #1626 and #1621

From original:
ros_statistics_test_result_original_kinetic_roscpp

Over bugfix:
ros_statistics_test_result_bugfixed_kinetic_roscpp

To Optimization:
ros_statistics_test_result_optimized_kinetic_roscpp

### DIFF
--- a/clients/roscpp/include/ros/statistics.h
+++ b/clients/roscpp/include/ros/statistics.h
@@ -82,6 +82,8 @@ private:
 
   // frequency to publish statistics
   double pub_frequency_;
+  // time to publish statistics
+  double pub_time_;
 
   // publisher for statistics data
   ros::Publisher pub_;
@@ -100,6 +102,8 @@ private:
     // latest total traffic volume observed
     uint64_t stat_bytes_last;
   };
+
+  bool initialized = false;
 
   // storage for statistics data
   std::map<std::string, struct StatData> map_;
--- a/clients/roscpp/src/libros/statistics.cpp
+++ b/clients/roscpp/src/libros/statistics.cpp
@@ -33,29 +33,40 @@
 #include "std_msgs/Header.h"
 #include "ros/param.h"
 
+#include <chrono>
+
+
+//#define DEBUG
+#undef DEBUG
+
 namespace ros
 {
 
 StatisticsLogger::StatisticsLogger()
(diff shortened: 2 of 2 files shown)


dirk-thomas:
Please target the latest development branch - currently melodic-devel. Also the patch seems to break ABI and as such can only be considered for future ROS distros.

dirk-thomas:
Also please avoid any unrelated changes (e.g. to whitespaces) in the patch as well as don't include statements you used for local debugging.

kikass13:
thanks for your reply, this code is just here so that I can visualize and discuss changes that would be suitable. 

I can obviously remove all the stuff, which is breaking the abi policies (define, added member variables and provide that instead. 
(I will probably do that on Monday)

would such a change be feasible?  what is gonna happen with the kinetic devel branch ... I mean this class is obviously broken since it was created 5 years ago ...

We use kinetic, would these changes be merged back into kinetic as well?  

I mean I would probably have to implement a fixed ros-kinetic roscpp fork inside our code library and jump through the hoops of wrapping the publisher/subscriber classes into a separate namespace ... Just to have access to this statistics feature (which is obviously a must have for every serious system monitoring for autonomous robots).

So my points are: 
 * I need this in kinetic
 * this feature is broken (publish fails after sole time and cpu resources just go nuts for each node * publisher * message_callback 
 * this has to be fast 

thanks for your time :)

dirk-thomas:
> I can obviously remove all the stuff, which is breaking the abi policies (define, added member variables and provide that instead. would such a change be feasible?

Probably yes, but for sure I can only say after reviewing the actual patch.

> what is gonna happen with the kinetic devel branch

A bug fix which is API/ABI compatible can be backported to older distros after it has been applied to the current one.

dirk-thomas:
@kikass13 Friendly ping.

kikass13:
@dirk-thomas this was allready done via #1625 in melodic-devel


https://github.com/ros/ros_comm/blob/3131ac6b1cf78f997653ced9b2f2487075a6d188/clients/roscpp/src/libros/statistics.cpp#L116

PR #1629, opened 2019-02-23
/rosbag_storage/CMakeLists.txt: require C++14

### LINKED ISSUE #1600: rosbag: allow compilation on CentOS
Hello,

CentOS 7 platform uses gcc-4.8.5; by default the c++ standard is 98. rosbag_storage and rosbag expect compilers using natively c++14. 

I produced a patch w.r.t files ros_comm/rosbag/CMakeLists.txt and 
ros_comm/rosbag_storage/CMakeLists.txt; unfortunately I can't add them to this issue. Basically it's about adding a few lines like 
target_compile_features(rosbag_storage PUBLIC cxx_std_11)
requesting that C++ standard is at least 11

Regards

Pascal

cwecht:
A first step to fix #1600. The C++-version is now set to C++14 explicitly.

### DIFF
--- a/tools/rosbag_storage/CMakeLists.txt
+++ b/tools/rosbag_storage/CMakeLists.txt
@@ -1,4 +1,7 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.1)
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 project(rosbag_storage)
 


dirk-thomas:
> A first step to fix #1600.

Do you have an estimate where else similar patches will be necessary. I just want to make sure that this is actually feasible. In the past we recommended users trying to build Melodic on platforms which don't default to C++14 to set this via an environment variable instead.

cwecht:
After thinking a bit more about it, this PR is not necessarily an improvement. The CMAKE_CXX_STANDARD must be set in every project, which exports an compiled target. For header-only libraries it does not work this way. It might arguably be not a good idea to have packages with differing c++ standards linked together. Currently this works, but in the future, ABI-breaking changes in the C++-Standard are still possible (like from C++03 to C++11).

Nevertheless the current solution is sub-optimal, since there is no documentation, what needs to be done to make the build work on platforms with older compilers, as far as I know.

dirk-thomas:
> The CMAKE_CXX_STANDARD must be set in every project, which exports an compiled target. For header-only libraries it does not work this way.

:+1:

> there is no documentation, what needs to be done to make the build work on platforms with older compilers

Please feel free to contribute the information to the ROS wiki wherever you see fit.

dirk-thomas:
I will go ahead and close this ticket.

CdeMills:
Excuse me, but CentOS only support C++-11.  What does mandate the change to C++-14 ?

cwecht:
Because C++14 is part of the target-plattform for melodic: http://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023

But currently, as far as I know, ros_comm does not use any C++14 feature, sou you should be able to build it with C++11 enabled.

CdeMills:
I found another option. There is a "Developper Toolset" which brings more recent versions of compiler. Full C++14 support is available in gcc-5.0; the active toolset 6 contains gcc-6.3. See https://www.softwarecollections.org/en/scls/rhscl/devtoolset-6/
on how to install and enable it.
Regards

PR #1630, opened 2019-02-23
/libros/node_handle: keep shared pointer to TopicManager and ServiceM…

### LINKED ISSUE #838: Exception boost::lock_error thrown from shutdown method
We see crashes like this as multi-threaded processes like `rqt_gui` shut down:

```
../../../../../ros-comm-c9e259436a855db5e5f50f3efe2cec60d1f5786e/clients/roscpp/src/libros/init.cpp in ros::shutdown at line 596

../../../../../ros-comm-c9e259436a855db5e5f50f3efe2cec60d1f5786e/clients/roscpp/src/libros/topic_manager.cpp in ros::TopicManager::shutdown at line 93

/usr/include/boost/thread/lock_types.hpp in unique_lock at line 124

/usr/include/boost/thread/lock_types.hpp in boost::unique_lock<boost::mutex>::lock at line 346

/usr/include/boost/thread/pthread/mutex.hpp in lock at line 116

/usr/include/boost/throw_exception.hpp in boost::throw_exception<boost::lock_error> at line 67

/usr/lib/x86_64-linux-gnu/libstdc++.so.6 in __cxa_throw ()

/usr/lib/x86_64-linux-gnu/libstdc++.so.6 in std::terminate() ()
```

This is the code:

```
void TopicManager::shutdown()
{
  boost::mutex::scoped_lock shutdown_lock(shutting_down_mutex_);
  if (shutting_down_)
  {
    return;
  }

  {
    boost::recursive_mutex::scoped_lock lock1(advertised_topics_mutex_);
    boost::mutex::scoped_lock lock2(subs_mutex_);
    shutting_down_ = true;
  }

  ...
```

It looks like the expected behaviour is to block until the mutex can be acquired, so this is maybe an issue with one thread tearing down the TopicManager and then other finding a mutex which is no longer valid? Maybe there's a better solution here than just wrapping a try block around the whole thing?

cwecht:
…anager to avoid boost::lock_error exceptions.

Fixes #838.
This solution does not change the object-layout of NodeHandle.

### DIFF
--- a/clients/roscpp/src/libros/node_handle.cpp
+++ b/clients/roscpp/src/libros/node_handle.cpp
@@ -56,8 +56,8 @@ int32_t g_nh_refcount = 0;
 bool g_node_started_by_nh = false;
 
 // postpone desctruction of the managers until every NodeHandle has been desctructed.
-TopicManagerPtr g_topic_manager_;
-ServiceManagerPtr g_service_manager_;
+TopicManagerPtr g_topic_manager_{nullptr};
+ServiceManagerPtr g_service_manager_{nullptr};
 
 class NodeHandleBackingCollection
 {
--- a/clients/roscpp/src/libros/node_handle.cpp
+++ b/clients/roscpp/src/libros/node_handle.cpp
@@ -55,6 +55,10 @@ boost::mutex g_nh_refcount_mutex;
 int32_t g_nh_refcount = 0;
 bool g_node_started_by_nh = false;
 
+// postpone desctruction of the managers until every NodeHandle has been desctructed.
+TopicManagerPtr g_topic_manager_;
+ServiceManagerPtr g_service_manager_;
+
 class NodeHandleBackingCollection
 {
 public:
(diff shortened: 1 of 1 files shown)


cwecht:
Well, this failed because of the flakyness of topic_tools throttle.test ...

clients/roscpp/src/libros/node_handle.cpp
```diff
@@ -55,6 +55,10 @@ boost::mutex g_nh_refcount_mutex;
 int32_t g_nh_refcount = 0;
 bool g_node_started_by_nh = false;
 
+// postpone desctruction of the managers until every NodeHandle has been desctructed.
+TopicManagerPtr g_topic_manager_;
```
scott-eddy: I'm guessing we want these explicitly initialized to NULL

scott-eddy:
This looks good to me

cwecht:
Alternatively, the shared_ptrs could be added NodeHandleBackingCollection. This would also avoid an API break. Since order of destruction of global object can be an issue, this might be a better solution.

cwecht:
The above mentioned alternative is astonishingly simple: https://github.com/cwecht/ros_comm/commit/39b9938dc28a4fddcea9ad3772a6d9d9ff08954f

This might actually be a better solution. The solution of this PR relies (probably?) on specific destruction order of globals/statics, which is unfortunate. The alternative version ensures, that the managers live at least as long as any NodeHandle-instance.

scott-eddy:
@cwecht can you close this in favor of #1656, so it doesn't take up maintainers bandwidth?

PR #1631, opened 2019-02-25
xmlrpcpp: fixed invalid zero index as suggested in #1547

cwecht:
This replaces #1547 .


dirk-thomas:
Thanks for the patch.

PR #1635, opened 2019-02-28
rostopic: repeatedly republish message from file

cwecht:
This replaces #1348

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -1945,6 +1945,13 @@ def stdin_publish(pub, msg_class, rate, once, filename, verbose):
     else:
         iterator = stdin_yaml_arg
 
+    exactly_one_message = False
+    for _ in iterator():
+        if exactly_one_message:
+            exactly_one_message = False
+            break
+        exactly_one_message = True
+
     r = rospy.Rate(rate) if rate is not None else None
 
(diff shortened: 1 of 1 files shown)


cwecht:
hm, looks like the build-infastructure had an hiccup ...

dirk-thomas:

tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1945,6 +1945,13 @@ def stdin_publish(pub, msg_class, rate, once, filename, verbose):
     else:
         iterator = stdin_yaml_arg
 
+    exactly_one_message = False
+    for _ in iterator():
+        if exactly_one_message:
+            exactly_one_message = False
+            break
+        exactly_one_message = True
```
dirk-thomas: Doesn't this alter the content returned by the iterator in the later loop - in the case of the iterator being `stdin_yaml_arg`?

cwecht: Yes, that's right. The fix makes not much sense for stdin stream anyway, so I limited it to actual files now.

cwecht:
flaky tests again...

cwecht:

dirk-thomas:

Ignoring the flaky test on CI...

pavloblindnology:
substitute_keywords argument isn't used in this scenario.

PR #1636, opened 2019-02-28
roscpp: added missin include path (for bazel workspaces)

### LINKED ISSUE #1548: unable to located ros/common.h with Bazel
some of the source files in this tree include ros/common.h whereas others include common.h. under certain circumstances (I am calling cmake within a bazel workspace, which means my working directory when I call cmake, as well as the directory where build results are written, are separate from the directory containing the source code), builds fail with `can't find file: ros/common.h`.

To fix this issue, I need to add `${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_INCLUDE_DESTINATION}` as an include dir (e.g. without the `ros` suffix)

I'd be happy to contribute a PR that does one of the following:
1. Standardizes on #include common.h
or 2. Adds both directories as include dirs so both #include ros/common.h and #include common.h will work

Let me know if this is a reasonable thing to do, or if I am missing something altogether. And thanks for maintaining this awesome software :)

cwecht:
This fixes #1548 .


dirk-thomas:
Thanks for creating a PR for the issue.

Ignoring the flaky test.

PR #1637, opened 2019-02-28
rosbag_storage: fixed dangeling if-else

cwecht:
This replaces #1306 (for melodic).


dirk-thomas:
Thanks for re-targeting the patch.

PR #1638, opened 2019-02-28
roscpp/service: use WallTime/WallDuration for waiting

### LINKED ISSUE #1492: service::waitForService stuck in ros::Duration::sleep when use_sim_time true without /clock publisher, because using ros::Duration::sleep instead of ros::WallDuration::sleep
The function service::waitForService in file libros/service.cpp 
uses ros::Duration::sleep instead of ros::WallDuration::sleep,
which leads to the side effect of beeing stuck in ros::Duration::sleep 
under the following conditions:
- use_sim_time is set as ros parameter 
- no simulated time publisher is active (no /clock publisher)
- the service provider is not yet available

This case is significantly critical, when:
- using a nodelet as /clock publisher
- starting nodelet manager and nodelet loader in 1 roslaunch
--> in 99/100 cases everything is fine
--> in  1/100 cases the nodelet loader gets stuck in service::waitForService -> ros::Duration::sleep

Although nodelet is used as an example, the main problem is the usage of ros::Duration::sleep at service client initialization instead of ros::WallDuration::sleep.

Quick Reproduction:
The problem can be quickly reproduced the following way by using the service "load_nodelet" of the nodelet manager as an example:
case A) everything is fine
- first: terminal 1: roscore
- then: terminal 2: rosparam set use_sim_time "true"
- then: terminal 3: rosrun nodelet nodelet manager __name:=nodelet_manager
- then: (wait 1s)
- then: terminal 4: rosrun nodelet nodelet load nodelet_tutorial_math/Plus nodelet_manager __name:=nodelet1 nodelet1/in:=foo _value:=1.1
- then: terminal 5: rosservice call /nodelet_manager/list
==> nodelet is loaded

case B) stuck, no nodelet loaded
- first: terminal 1: roscore
- then: terminal 2: rosparam set use_sim_time "true"
- then: terminal 3: rosrun nodelet nodelet load nodelet_tutorial_math/Plus nodelet_manager __name:=nodelet1 nodelet1/in:=foo _value:=1.1
- then: (wait 1s)
- then: terminal 4: rosrun nodelet nodelet manager __name:=nodelet_manager
- then: terminal 5: rosservice call /nodelet_manager/list
==> no nodelet loaded, empty list

(note: no /clock publisher is active, but ros::Duration::sleep is never executed in case A)

------
------

function
- bool service::waitForService(const std::string& service_name, ros::Duration timeout) 

in file 
- ros_comm/clients/roscpp/src/libros/service.cpp

uses
- ros::Duration::sleep

instead of
- ros::WallDuration::sleep

under the following conditions
- use_sim_time is set as ros parameter 
- no simulated time publisher is active (no /clock publisher)
- the service provider is not yet available

causing critical side effect:
- beeing stuck in service::waitForService -> ros::Duration::sleep

cwecht:
Fixes #1492

### DIFF
--- a/clients/roscpp/src/libros/service.cpp
+++ b/clients/roscpp/src/libros/service.cpp
@@ -88,7 +88,8 @@ bool service::waitForService(const std::string& service_name, ros::Duration time
 {
   std::string mapped_name = names::resolve(service_name);
 
-  Time start_time = Time::now();
+  const WallTime start_time = WallTime::now();
+  const WallDuration wall_timeout{timeout.toSec()};
 
   bool printed = false;
   bool result = false;
@@ -103,17 +104,17 @@ bool service::waitForService(const std::string& service_name, ros::Duration time
     {
       printed = true;
 
-      if (timeout >= Duration(0))
+      if (wall_timeout >= WallDuration(0))
       {
-        Time current_time = Time::now();
+        const WallTime current_time = WallTime::now();
 
-        if ((current_time - start_time) >= timeout)
+        if ((current_time - start_time) >= wall_timeout)
         {
           return false;
         }
       }
 
-      Duration(0.02).sleep();
+      WallDuration(0.02).sleep();
     }
   }
 


dirk-thomas:
Similar to #1543.

Thanks for the patch.

PR #1639, opened 2019-02-28
Fix wrong error handling in migration.

omangin:
self.(old|new)_types is supposed to be a dictionary, not a list which causes failures with cryptic messages

### DIFF
--- a/tools/rosbag/src/rosbag/migration.py
+++ b/tools/rosbag/src/rosbag/migration.py
@@ -281,7 +281,6 @@ class MessageUpdateRule(object):
             self.old_md5sum = self.old_class._md5sum
         except Exception as e:
             if not isinstance(e, self.EmptyType):
-                print("Warning: {}".format(e), file=sys.stderr)
                 traceback.print_exc(file=sys.stderr)
             self.old_types = {}
             self.old_class = None
@@ -294,7 +293,6 @@ class MessageUpdateRule(object):
             self.new_md5sum = self.new_class._md5sum
         except Exception as e:
             if not isinstance(e, self.EmptyType):
-                print("Warning: {}".format(e), file=sys.stderr)
                 traceback.print_exc(file=sys.stderr)
             self.new_types = {}
             self.new_class = None
--- a/tools/rosbag/src/rosbag/migration.py
+++ b/tools/rosbag/src/rosbag/migration.py
@@ -43,6 +43,7 @@ import itertools
 import os
 import string
 import sys
+import traceback
 
 import genmsg.msgs
 import genpy
@@ -280,7 +281,8 @@ class MessageUpdateRule(object):
             self.old_md5sum = self.old_class._md5sum
         except Exception as e:
             if not isinstance(e, self.EmptyType):
-                print("Warning: {}".format(e))
+                print("Warning: {}".format(e), file=sys.stderr)
+                traceback.print_exc(file=sys.stderr)
             self.old_types = {}
             self.old_class = None
             self.old_md5sum = ""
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please provide more context on the use case where you ran into the problem and steps to reproduce it.

omangin:
Hi @dirk-thomas. Unfortunately I do not know how this module works internally so I cannot provide more context on when this happens except that it happens sometime when `rosbag` attempts some migrations and it gets in the case where `self.old_type == ""`  or `self.new_type == ""` (the code fails to comment on when this occurs) or where any exception is raised by `genpy.dynamic.generate_dynamic(self.old_type, self.old_full_text)`.

You can easily see from the few changes in this PR why the current behavior is buggy:
1.  when the attribute mentioned above is an empty string—which seems to be intended to be a normal situation because the code explicitly handles it—it raises a generic `Exception` which it then catches and ignores,
2. hence, if another issue occurs—which is not expected—it is silently ignored,
3. the handling of any of these exceptions wrongly use a default values for `self.old/new_types` of `[]` while the attributes is expected to hold a dictionary (as returned by `genpy.dynamic.generate_dynamic`). This causes an issue later in the code when the attribute is indexed.

The fix addresses 1. and 2. by using a custom exception to treat differently the expected case and the unexpected exception (which it still ignores but not without a warning). It addresses 3. by using a default value of the expected type.

tools/rosbag/src/rosbag/migration.py
```diff
@@ -271,23 +274,26 @@ def __init__(self, migrator, location):
...
-                raise Exception
+                raise self.EmptyType
             self.old_types = genpy.dynamic.generate_dynamic(self.old_type, self.old_full_text)
             self.old_class = self.old_types[self.old_type]
             self.old_md5sum = self.old_class._md5sum
-        except:
-            self.old_types = []
+        except Exception as e:
+            if not isinstance(e, self.EmptyType):
+                print("Warning: {}".format(e))
```
dirk-thomas: Printing a full traceback might be more helpful (especially if it is unclear when this might happen). Also warnings / errors should go to `stderr`.

tools/rosbag/src/rosbag/migration.py
```diff
@@ -271,23 +274,26 @@ def __init__(self, migrator, location):
...
+                raise self.EmptyType
             self.new_types = genpy.dynamic.generate_dynamic(self.new_type, self.new_full_text)
             self.new_class = self.new_types[self.new_type]
             self.new_md5sum = self.new_class._md5sum
-        except:
-            self.new_types = []
+        except Exception as e:
+            if not isinstance(e, self.EmptyType):
+                print("Warning: {}".format(e))
+            self.old_types = {}
```
dirk-thomas: Is this supposed to set `self.new_types` instead?

dirk-thomas:
> it happens sometime when rosbag attempts some migrations 

Can you please share some reproducible steps where that is the case.

tools/rosbag/src/rosbag/migration.py
omangin: Sure for `stderr`.

Regarding the traceback, I can print it, yes. I was not sure because the previous behavior was to just silently ignore errors. I added the warning to be as close as possible to the previous behavior but I do not have example of when it happens. Maybe it would just be better to only catch the `EmptyType` exception instead.

tools/rosbag/src/rosbag/migration.py
omangin: Indeed, thanks.

omangin:
> Can you please share some reproducible steps where that is the case.

Unfortunately I cannot.

Regarding the print of the backtrace vs not catching the exception, let me know if you want the other solution implemented.

tools/rosbag/src/rosbag/migration.py
dirk-thomas: > Maybe it would just be better to only catch the `EmptyType` exception instead.

That would change / break behavior which would be problematic to be shipped into an existing ROS distro. I think a message with a traceback would be acceptable in terms of noise (with the advantage of actually knowing what failed).

dirk-thomas:
I removed the printing of the exception string since that is already part of the `traceback.print_exc()`: 644aa4e72f3fdfe0ac052daf7da0b0eaa072731c

dirk-thomas:
Thanks for the patch.

(Ignoring the flaky test on CI.)

PR #1641, opened 2019-03-01
rosbag modernization: replaced BOOST_FOREACH with range-based for-loo…

cwecht:
…ps, used algorithm where appropriated

### DIFF
--- a/tools/rosbag/src/player.cpp
+++ b/tools/rosbag/src/player.cpp
@@ -445,10 +445,12 @@ void Player::waitForSubscribers() const
     bool all_topics_subscribed = false;
     std::cout << "Waiting for subscribers." << std::endl;
     while (!all_topics_subscribed) {
-        all_topics_subscribed = std::all_of(std::begin(publishers_), std::end(publishers_),
-	                                    [](const PublisherMap::value_type& pub) {
-                                              return pub.second.getNumSubscribers() > 0;
-                                             });
+        all_topics_subscribed = std::all_of(
+                                        std::begin(publishers_),
+                                        std::end(publishers_),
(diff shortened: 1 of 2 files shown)


tools/rosbag/src/recorder.cpp
```diff
@@ -257,20 +254,16 @@ bool Recorder::shouldSubscribeToTopic(std::string const& topic, bool from_node)
...
-            boost::smatch what;
-            if (boost::regex_match(topic, what, e, boost::match_extra))
-                return true;
-        }
-    }
-    else {
-        foreach(string const& t, options_.topics)
-            if (t == topic)
-                return true;
+	return std::any_of(std::begin(options_.topics), std::end(options_.topics),
```
dirk-thomas: Same here.

tools/rosbag/src/player.cpp
```diff
@@ -448,10 +445,10 @@ void Player::waitForSubscribers() const
     bool all_topics_subscribed = false;
     std::cout << "Waiting for subscribers." << std::endl;
     while (!all_topics_subscribed) {
-        all_topics_subscribed = true;
-        foreach(const PublisherMap::value_type& pub, publishers_) {
-            all_topics_subscribed &= pub.second.getNumSubscribers() > 0;
-        }
+        all_topics_subscribed = std::all_of(std::begin(publishers_), std::end(publishers_),
```
dirk-thomas: Please wrap right after the first open parenthesis (`std::all_of(\n`) to avoid the aligned indendation of the following lines. All arguments can then be one level indented.

dirk-thomas:
Thanks for the patch.

PR #1642, opened 2019-03-01
test_rosbag modernization: replaced BOOST_FOREACH with range-based fo…

cwecht:
…r-loops


dirk-thomas:
Thanks for the patch.

Issue #1644, opened 2019-03-04
Rosbag duration can finish too early


eirik-strand:
I have not done a lot of tests with this issue, but I think it can be reproduced. I have a lot of topics, and will not record more than 1s of data. If it takes more than 1s to start the subscribers, everything just hangs. I guess that check_duration finishes before all subscribers got something on the topic. 

If you have a a long list of topics, doing:
`rosbag record --all --duration="0.1s" `
Will cause it to just hang and never finish
if not I guess you can try `0.01s `

cwecht:
I was able to reproduce this bug. The hangig is caused by a deadlock involving Recorders queue_mutex_ and TopicManagers advertised_topics_mutex_) caused by calling ros::shutdown in Recorder::checkDuration during the queue_mutex_ lock.

cwecht:
Ok, I was wrong about that, it is even worse: the deadlock involves TopicManagers::advertised_topics_mutex TopicManagers::subs_mutex_.

dirk-thomas:
Addressed by #1645.

PR #1645, opened 2019-03-05
roscpp/TopicManager: avoid deadlock

cwecht:
`TopicManager::subscribe` locks first `TopicManager::subs_mutex` and then `TopicManager::advertised_topics_mutex_`. `TopicManager::shutdown` locks them in reverse order. If the run concurrently, this can result in a deadlock, which is the cause of #1644 .

`TopicManager::shutdown` uses now [`boost::lock`](https://www.boost.org/doc/libs/1_69_0/doc/html/thread/synchronization.html#thread.synchronization.lock_functions.lock_multiple), which avoids deadlocks.

### DIFF
--- a/clients/roscpp/src/libros/topic_manager.cpp
+++ b/clients/roscpp/src/libros/topic_manager.cpp
@@ -97,9 +97,10 @@ void TopicManager::shutdown()
   }
 
   {
-    boost::recursive_mutex::scoped_lock lock1(advertised_topics_mutex_);
-    boost::mutex::scoped_lock lock2(subs_mutex_);
+    boost::lock(subs_mutex_, advertised_topics_mutex_);
     shutting_down_ = true;
+    subs_mutex_.unlock();
+    advertised_topics_mutex_.unlock();
   }
 
   // actually one should call poll_manager_->removePollThreadListener(), but the connection is not stored above


dirk-thomas:
> `TopicManager::subscribe` locks first `TopicManager::subs_mutex` and then `TopicManager::advertised_topics_mutex_`.

:+1: the second mutex is locked in the called `TopicManager::registerSubscriber` method.

Have you tried to just swap the order of the locks in `TopicManager::shutdown`?

```
// same order as in the subscribe / registerSubscriber methods
boost::mutex::scoped_lock lock1(subs_mutex_);
boost::recursive_mutex::scoped_lock lock2(advertised_topics_mutex_);
```

cwecht:
yes I did and it worked as well. I just thought, that this might be the better solution, because with boost::lock the locking-order does not matter. It will work even if anywhere else the locking order is different.

dirk-thomas:
Thanks for the patch. I was mostly wondering if there might be other cases where the order is incorrect.

dirk-thomas:

dirk-thomas:
Please use the phrase "fixes #1644" in the description in future tickets. That will make sure the referenced issue is closed automatically when the PR gets merged. Thanks.

PR #1646, opened 2019-03-06
Only specify the not-compensated missing dependency in roslaunch file check error

mbeutelspacher:
Since only missing dependencies which cannot be compensated by other package.xml lead to a failure only those missing dependencies should be specified in the error as this eases debugging. All the other missing dependencies which can be compensated are nonetheless in the log.

### DIFF
--- a/tools/roslaunch/src/roslaunch/rlutil.py
+++ b/tools/roslaunch/src/roslaunch/rlutil.py
@@ -220,7 +220,7 @@ def check_roslaunch(f, use_test_depends=False):
             miss_all = True
         if miss_all:
             roslaunch.core.printerrlog("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)))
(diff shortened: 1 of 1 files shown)


tools/roslaunch/src/roslaunch/rlutil.py
```diff
@@ -220,7 +220,7 @@ def check_roslaunch(f, use_test_depends=False):
             miss_all = True
         if miss_all:
             roslaunch.core.printerrlog("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)))
-            errors.append("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss)))
+            errors.append("Missing package dependencies: %s/package.xml: %s"%(pkg, ', '.join(miss_all)))
```
dirk-thomas, line 223: Since `miss_all` could be `True` (line 220) this line could result in a `TypeError: can only join an iterable`.

PR #1647, opened 2019-03-08
Add set_rate service

efernandez:
This adds a `set_rate` service to `rosbag play` and also captures the `+` and `-` keys to scale up or down the current rate by a factor of `2`.

### DIFF
--- a/tools/rosbag/CMakeLists.txt
+++ b/tools/rosbag/CMakeLists.txt
@@ -5,11 +5,15 @@ if(NOT WIN32)
   set_directory_properties(PROPERTIES COMPILE_OPTIONS "-Wall;-Wextra")
 endif()
 
-find_package(catkin REQUIRED COMPONENTS rosbag_storage rosconsole roscpp std_srvs topic_tools xmlrpcpp)
+find_package(catkin REQUIRED COMPONENTS message_generation rosbag_storage rosconsole roscpp std_srvs topic_tools xmlrpcpp)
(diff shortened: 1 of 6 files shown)


beetleskin:
Why would you want to scale up and down the playback speed on the fly? This looks like a very special use-case to me. Do you have in mind, that there is this `--rate-control-topic`?

efernandez:
@beetleskin The main use case of the `+` and `-` keys is for doing triage on long bag files. It makes it easier to start the bag, speed it up while looking for the issue or issues (so a single `-s` is not always very convenient), and slow it down when the issue happens, so it's easier to see what's going on.

Regarding `--rate-control-topic`, I guess it'll allow to implement something similar into a client node, but IMHO it's more convenient to control the rate directly on the terminal where `rosbag play` is running, if it's available to us. For any case, like running unattended on a server, `--rate-control-topic` is still useful.

The `set_rate` I provided here could be split on another commit and either removed and moved to another PR if that's the part you consider redundant with `--rate-control-topic`.

beetleskin:
Alright, thanks for the clarification.

However, if you want to have this feature stable, I suggest to add some unit and integration-tests.

dirk-thomas:
Since the patch breaks ABI I added the label "consider-future-release".

dirk-thomas:
This needs to be rebased to resolve conflicts.

dirk-thomas:
@efernandez Closing due to no further reply. Please feel free to open a new PR against `noetic-devel`.

PR #1648, opened 2019-03-11
fix topic message count for rosbag indexed v1.2.

kmiku7:
We have many indexed v1.2 bags. But command `rosbag info` can not display message count per topic in bags.
The current output like this:
> topics:      /velodyne_packets                    0 msg  @  10.0 Hz : av_msgs/VelodyneScan

The frequency is right, but the message count is always zero.


dirk-thomas:
Thanks for the patch.

PR #1649, opened 2019-03-11
rostest: use AnyMsg in publishtest

beetleskin:
Right now publishtest only works correctly when a publisher of the given topic exists: If no publisher exists, the retrieval of the topic class blocks until the test times out. This leads to aborted tests when one would expect a failed one with correct error message, stating that no message on the given topic was published. Using `rospy.AnyMsg` fixes that.

This also makes it impossible to test that a certain topic is not being published, when no publisher exists.

### DIFF
--- a/tools/rostest/nodes/publishtest
+++ b/tools/rostest/nodes/publishtest
@@ -72,10 +72,8 @@ class PublishChecker(object):
         self.topic_name = topic_name
         self.negative = negative
         self.deadline = rospy.Time.now() + rospy.Duration(timeout)
-        msg_class, _, _ = rostopic.get_topic_class(
-            rospy.resolve_name(topic_name), blocking=True)
         self.msg = None
-        self.sub = rospy.Subscriber(topic_name, msg_class, self._callback)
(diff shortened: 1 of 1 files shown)


beetleskin:
Uff .. the failing test is a hztest, without retry argument. This is unfortunate, this never works reliably on CI with shared resources. I could add a retry argument, but this change is not about fixing those tests for CI.

Other instances of this from the recent PRs:
* https://github.com/ros/ros_comm/pull/1647
* https://github.com/ros/ros_comm/pull/1606
* .. and then the Jenkins history is cut off (I guess, you get 404 for older Details/test results)

How can I just trigger the build again?

beetleskin:
> How can I just trigger the build again?

https://stackoverflow.com/a/28703197/1880234
.. but now we have this empty commit in there :/

dirk-thomas:
I am not sure why you closed the ticket. The empty commit is not a problem - at the end the PR will be squash merged anyway. Do you want to reopen this?

Nvm, just saw that you also deleted the fork...

beetleskin:
messed up my fork, there you go: https://github.com/ros/ros_comm/pull/1659

PR #1652, opened 2019-03-13
Pickleable rosbag exceptions (#1210 revisited).

### LINKED ISSUE #1209: ROSBagException should call the base constructor

cwecht:
This fixes  #1209 in a API preserving way. It replaces #1210. Also tests for the pickleability have been added.

### DIFF
--- a/test/test_rosbag/test/test_bag.py
+++ b/test/test_rosbag/test/test_bag.py
@@ -432,6 +432,20 @@ class TestRosbag(unittest.TestCase):
 
                 print(bag._OP_CODES.get(op, op))
 
+    # #1209
+    def test_rosbag_exceptions_are_pickleable(self):
+        #bag_exception = rosbag.ROSBagException("msg string")
+        def test(bag_exception):
+            import pickle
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thanks for the patch.

PR #1455, opened 2018-07-12
make roslaunch-check respect arg remappings with command line argument (:=)

k-okada:
the example
```
  <arg name="commandline_true_arg" default="true" />
  <group unless="$(arg commandline_true_arg)">
    <node pkg="doesnt_exist" type="nope" name="nope" />
  </group>
```
want to fail  `roslauch-check` script with
```
$ scripts/roslaunch-check resources/example.launch commandline_true_arg:=false
```

this PR added 
```
    context = { 'arg': load_mappings(sys.argv) }
```
to respect remappings with command line argument.

I'm not sure if we need to change 'subcontext', for now, 
```
    subcontext = {'arg': {}}

```
https://github.com/k-okada/ros_comm/blob/d206841691f45d3fd972149afdd9d5eb133e157e/tools/roslaunch/src/roslaunch/depends.py#L109-L111

For now, I could not find example which need to set 'load_remapping(sys.argv)' to 'subcontext'
```
$ cat hoge.launch 
<launch>
  <arg name="hoge" default="true" />
  <include file="./fuga.launch" >
    <arg name="hoge2" value="$(arg hoge)" />
  </include>
  <group if="$(arg hoge)">
    <node pkg="hoge" name="fuga" type="fuga" />
  </group>
</launch>
$ cat fuga.launch 
<launch>
  <arg name="hoge2" default="true" />
  <group if="$(arg hoge2)">
    <node pkg="hoge2" name="fuga2" type="fuga2" />
  </group>
</launch>
```

PR #1653, opened 2019-03-13
 #1455 reviseted: make roslaunch-check respect arg remappings with command line argument

cwecht:
Fix the changes of #1455 . It should not fail anymore.


dirk-thomas:
Thanks for providing the updated patch.

PR #1655, opened 2019-03-14
#1468 revisited: add option to hide summary from roslaunch output.

cwecht:
This replaces #1468 with the suggested changes.

### DIFF
--- a/tools/roslaunch/src/roslaunch/parent.py
+++ b/tools/roslaunch/src/roslaunch/parent.py
@@ -91,7 +91,7 @@ class ROSLaunchParent(object):
         @type  port: int
         @param verbose: (optional) print verbose output
         @type  verbose: boolean
-        @param show_summary: (optional) hether to show a summary or not
+        @param show_summary: (optional) whether to show a summary or not
         @type  show_summary: boolean
         @param force_screen: (optional) force output of all nodes to screen
         @type  force_screen: boolean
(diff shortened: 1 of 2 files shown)


tools/roslaunch/src/roslaunch/parent.py
```diff
@@ -73,7 +73,7 @@ class ROSLaunchParent(object):
     """
 
     def __init__(self, run_id, roslaunch_files, is_core=False, port=None, local_only=False, process_listeners=None,
-            verbose=False, force_screen=False, force_log=False, is_rostest=False, roslaunch_strs=None, num_workers=NUM_WORKERS, timeout=None, master_logger_level=False):
+            verbose=False, show_summary=True, force_screen=False, force_log=False, is_rostest=False, roslaunch_strs=None, num_workers=NUM_WORKERS, timeout=None, master_logger_level=False):
```
dirk-thomas: Please add the new argument at the end.

tools/roslaunch/src/roslaunch/parent.py
```diff
@@ -91,6 +91,8 @@ def __init__(self, run_id, roslaunch_files, is_core=False, port=None, local_only
         @type  port: int
         @param verbose: (optional) print verbose output
         @type  verbose: boolean
+        @param no_summary: (optional) hide summary output
+        @type  no_summary: boolean
```
dirk-thomas: The doc doesn't match the added argument.

dirk-thomas:
Thanks for the patch.

PR #1660, opened 2019-03-15
update approximate time filter to work with python2 and python3

rosslannen:


dirk-thomas:
Thanks for the improvement.

PR #1663, opened 2019-03-18
#1470 revisited: Use node namespace when looking up topic 

cwecht:
Replaces #1470. Adds the suggested changes.

### DIFF
--- a/tools/topic_tools/test/test_transform.py
+++ b/tools/topic_tools/test/test_transform.py
@@ -44,20 +44,15 @@ NAME = 'transform_sub'
 
 class TransformSub(unittest.TestCase):
 
-  def msg_cb(self, msg):
-    self.msg = msg
-
   def test_transform_sub(self):
     rospy.init_node(NAME)
     value = rospy.get_param("~value", 1.0)
 
-    self.msg = None
-    sub = rospy.Subscriber("input", Float32, self.msg_cb)
-
-    while not self.msg:
-      rospy.sleep(1.0)
-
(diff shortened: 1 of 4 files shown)


tools/topic_tools/test/test_transform.py
```diff
@@ -0,0 +1,63 @@
...
+
+  def test_transform_sub(self):
+    rospy.init_node(NAME)
+    value = rospy.get_param("~value", 1.0)
+
+    self.msg = None
+    sub = rospy.Subscriber("input", Float32, self.msg_cb)
+
+    while not self.msg:
+      rospy.sleep(1.0)
```
beetleskin: magic `1.0`  .. this might introduce random fails on any hardware, when it is busy.

I suggest to use [wait_for_message](https://docs.ros.org/melodic/api/rospy/html/rospy.client-module.html#wait_for_message) with a high timeout parameter, and let rostests's `time-limit` handle the case that the message does not arrive at all.

cwecht: `wait_for_message` makes the test actually much easier. the test is inherently a bit flaky. It seems, that sometimes rostopic does not advertise the topic early enough. Therefor the tests are configured to do three retries. The timeout in the tests should be no problem here.

cwecht:
This fixes #1427

dirk-thomas:
Thanks for the patch.

dirk-thomas:

dirk-thomas:
Thanks for the patch.

PR #1512, opened 2018-09-26
Add "param" substitution in tags

tahsinkose:
This PR implements the functionality proposed in #723. I only tested with `double`, `int` and `string` types. Actually, it converts any specified parameter of YAML into string to conform to the native behavior in `roslaunch`. It might require additional testing for different use cases.

PR #1665, opened 2019-03-19
roslaunch/test_xmlloader: conforming mock

cwecht:
The ROSLaunchConfigMock's params member is a dict now (not an array), to match ROSLaunchConfig.

This will help to make the test in #1512 because it requires params to be a dictionary.

Also the mocks name has been adapted to make clear, which class is mocked.

### DIFF
--- a/tools/roslaunch/test/unit/test_xmlloader.py
+++ b/tools/roslaunch/test/unit/test_xmlloader.py
@@ -45,8 +45,8 @@ def get_test_path():
 def get_example_path():
     return os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..', 'resources'))
 
-## Fake RosLaunch object
-class RosLaunchMock(object):
+## Fake ROSLaunchConfig object
+class ROSLaunchConfigMock(object):
     def __init__(self):
         self.nodes_core = []
         self.nodes = []
@@ -92,7 +92,7 @@ class TestXmlLoader(unittest.TestCase):
         
     def _load(self, test_file):
         loader = roslaunch.xmlloader.XmlLoader()
-        mock = RosLaunchMock()
+        mock = ROSLaunchConfigMock()
         self.assert_(os.path.exists(test_file), "cannot locate test file %s"%test_file)
         loader.load(test_file, mock)
         return mock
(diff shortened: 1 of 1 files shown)


tools/roslaunch/test/unit/test_xmlloader.py
```diff
@@ -192,8 +189,8 @@ def test_params(self):
         mock = self._load(os.path.join(self.xml_dir, 'test-params-valid.xml'))
 
         param_d = {}
-        for p in mock.params:
-            param_d[p.key] = p.value
+        for k, v in mock.params.iteritems():
```
dirk-thomas: This doesn't work with Python 3: use `.items()` instead.

Same below.

tools/roslaunch/test/unit/test_xmlloader.py
```diff
@@ -202,9 +199,9 @@ def test_params(self):
         self.assertEquals(2, param_d['/someinteger2'])
         self.assertAlmostEquals(3.14159, param_d['/somefloat1'], 2)
         self.assertAlmostEquals(5.0, param_d['/somefloat2'], 1)
-        self.assertEquals("a child namespace parameter 1", param_d['/wg/wgchildparam'], p.value)
-        self.assertEquals("a child namespace parameter 2", param_d['/wg2/wg2childparam1'], p.value)
-        self.assertEquals("a child namespace parameter 3", param_d['/wg2/wg2childparam2'], p.value)
+        self.assertEquals("a child namespace parameter 1", param_d['/wg/wgchildparam'])
+        self.assertEquals("a child namespace parameter 2", param_d['/wg2/wg2childparam1'])
+        self.assertEquals("a child namespace parameter 3", param_d['/wg2/wg2childparam2'])
```
dirk-thomas, line 204: Why is the third argument removed here?

cwecht: Because ``p`` does not exist anymore. Actually I don't really know, how this could have worked in the first place.

cwecht: Ok it worked, because loop-variables are not bound to the loops-scope in python ([see here](https://stackoverflow.com/questions/3611760/scoping-in-python-for-loops)). Anyway, I'd argue, that the asserts are sufficient the way they are.

dirk-thomas:
Please rebase the patch to resolve conflicts.

dirk-thomas:
@cwecht Closing due to no response. Please consider to open a new pull request targeting `noetic-devel`.

PR #1666, opened 2019-03-21
#961 revisited: Add default ROS_MASTER_URI

cwecht:
This is replaces #961. The suggested changes have been made:
  * added `ros::getDefaultMasterURI()` to roscpp
  * moved `DEFAULT_MASTER_URI` from rospy to rosgraph.rosenv, because it is needed there and import it from rospy is not possible, since this would introduce a circular dependency.

### DIFF
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -626,4 +626,9 @@ void shutdown()
   Time::shutdown();
 }
 
+const std::string& getDefaultMasterURI() {
+  static const std::string uri = "http://localhost:11311";
+  return uri;
+}
+
 }
--- a/tools/rosgraph/src/rosgraph/rosenv.py
+++ b/tools/rosgraph/src/rosgraph/rosenv.py
@@ -40,6 +40,9 @@ ROS_IPV6         ="ROS_IPV6"
 ROS_HOSTNAME     ="ROS_HOSTNAME"
 ROS_NAMESPACE    ="ROS_NAMESPACE"
 
+DEFAULT_MASTER_PORT=11311 #default port for master's to bind to
+DEFAULT_MASTER_URI = 'http://localhost:%s/'%DEFAULT_MASTER_PORT
+
 def get_master_uri(env=None, argv=None):
     """
     Get the :envvar:`ROS_MASTER_URI` setting from the command-line args or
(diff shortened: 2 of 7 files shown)


clients/roscpp/src/libros/init.cpp
```diff
@@ -612,4 +612,9 @@ void shutdown()
   Time::shutdown();
 }
 
+const std::string& getDefaultMasterURI() {
+  static const std::string uri = "http://localhost:11311";
```
mikepurvis, line 630: In the spirit of not spreading around the constant (as that was raised in the original review as a concern), what if we used `execute_process` from CMake to generate the implementation for this function based on a Python script, which could then import the `rosgraph.rosenv.DEFAULT_MASTER_URI` value?

cwecht: Although this would be a feasible way, I'd argue, that this would a require a solution, which is too complex for this particular case. There might be other constants, which are currently defined in roscpp and rospy (or anywhere else but in two different places). I think, it would be beneficial to find a solution for all these cases.

I'd suggest to merge this PR as it is for now and then think about a general solution for unifying the constants of C++ and Python components.

mikepurvis: I don't think it would really be that complicated, but I also don't feel strongly about it. :)

jspricke (approved):

tools/rosgraph/src/rosgraph/rosenv.py
```diff
@@ -69,7 +72,7 @@ def get_master_uri(env=None, argv=None):
                 if not val:
                     raise ValueError("__master remapping argument '%s' improperly specified"%arg)
                 return val
-        return env[ROS_MASTER_URI]
+        return env.get(ROS_MASTER_URI, DEFAULT_MASTER_URI)
     except KeyError as e:
```
dirk-thomas: I assume there is no case left which might raise a `KeyError`. If that is the case please remove the `try` / `except`.

dirk-thomas:
@cwecht Please comment on the PR after committing changes to the branch in the future. Otherwise nobody gets notified to revisit the PR.

dirk-thomas (approved):

dirk-thomas:
Thanks for the patch.

PR #1667, opened 2019-03-21
rosservice: use myargv()

cwecht:
There is no need to reinvent the wheel here.


dirk-thomas:
Thanks for the patch.

PR #1669, opened 2019-03-22
topic_tools/transfom: create publisher before subscriber

### LINKED ISSUE #1273: topic_tools transform publisher subscriber order
I believe that the script should call publisher before subscriber. I have has a situation in my launch file where the subscriber callback is call before the publisher is created and when the callback calls publish I get an error.

        self.pub = rospy.Publisher(args.output_topic, self.output_class, queue_size=1)
        self.sub = rospy.Subscriber(input_topic, input_class, self.callback)

cwecht:
because callback may use the publisher. This fixes #1273

### DIFF
--- a/tools/topic_tools/scripts/transform
+++ b/tools/topic_tools/scripts/transform
@@ -87,8 +87,8 @@ class TopicOp:
 
         self.output_class = roslib.message.get_message_class(args.output_type)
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for the patch!

PR #1674, opened 2019-03-25
#943 revisited: [rostopic] Unified list with subscribers and publishers

cwecht:
This replaces #943. The new verbose behavior supersedes the old now. Also a bit of cleanup has been done, and a bug in the number of pubs/subs calculation has been fixed.


sloretz:
Thank you for the PR!

PR #1680, opened 2019-04-01
roscpp/timer_manager: use steady time for timed waiting on condition …

### LINKED ISSUE #1558: roscpp spin() vs Timer
Hi,

I've been testing a lot moving the system clock backward and forward in time, while having a timer callback running in the node.

The code comes down to the following lines:
```
  timer_ = nh.createTimer(ros::Duration(1.0/50.0), &VescDriver::timerCallback, this);
  ros::Rate loop_rate(50);
  while (ros::ok())
    // ros::spin();  
  {
    loop_rate.sleep();
    ros::spinOnce();
  }
...
void VescDriver::timerCallback(const ros::TimerEvent& event)
{
  ROS_INFO("TIMER CALLBACK"); //TODO: REMOVE (TESTING TIMER RESET)
}
```

I noticed that if I have a `ros::spin()` enabled and move the time back using a command `sudo date +%T -s "10:23:00" ` the timerCallback stops firing.

If on the other hand I execute the busy waiting loop manually and call `ros::spinOnce()` after each pass of the loop the event is firing as expected when I move the time (both forward and backward).

Is this an expected behaviour? If yes I think it would be worth documenting this.

cwecht:
…variable

This fixes #1558

### DIFF
--- a/clients/roscpp/include/ros/timer_manager.h
+++ b/clients/roscpp/include/ros/timer_manager.h
@@ -27,6 +27,9 @@
 
 #ifndef ROSCPP_TIMER_MANAGER_H
 #define ROSCPP_TIMER_MANAGER_H
+#if !defined(__APPLE__) && !defined(WIN32)
+#define BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC
+#endif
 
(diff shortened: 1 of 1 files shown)


cwecht:

cwecht:
This is part of #1651.

PR #1681, opened 2019-04-01
roslaunch added --required option

### LINKED ISSUE #1678: [Feature request] --all-nodes-required
roslaunch can handle a node attribute required:

> required="true"(optional)
> 
> ROS 0.10: If node dies, kill entire roslaunch.

However in some situations all nodes a required. E.g. when running as a systemd service (systemd can then trigger a restart if the process died with an error code)
Or when running CI to check if your roslaunch works.

But enabling required=true for all nodes is impractical
- when launching locally, you might not want this behavior
- when using launch files provided by others.

Therefor I suggest an argument `--all-nodes-required`

cwecht:
Fixes #1678. If a node has `respawn` set, a ValueError-exception is thrown.


dirk-thomas:
Thanks for the patch.

PR #1682, opened 2019-04-01
/topic_tools/relay_field: added --tcpnodely

### LINKED ISSUE #1679: Topic Tools Relay: no option for tcp_nodelay
Hi,
please add an option to the relay script to activate tcp_nodelay, or activate it by default, since not having it can lead to very strange behavior of the system. 
The option would have to be passed when initiating the subscriber as parameter "txp_nodelay=true" in following line
https://github.com/ros/ros_comm/blob/af0fbaac0448a8ce215467cc343805b2e4bc8f8f/tools/topic_tools/scripts/relay_field#L93

cwecht:
Fixes #1679

### DIFF
--- a/tools/topic_tools/scripts/relay_field
+++ b/tools/topic_tools/scripts/relay_field
@@ -74,6 +74,10 @@ class RelayField(object):
             '--wait-for-start', action='store_true',
             help='Wait for input messages.'
             )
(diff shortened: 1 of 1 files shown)


cwecht:

dirk-thomas:
Thanks for the patch.

Issue #1671, opened 2019-03-22
rosgraph/test_names.py: test_load_mappings uses illegal remaps?


cwecht:
`test_names.py` contains this line:

    assert {'a': '1', 'b': '2', 'c': '3'} == load_mappings(['c:=3', 'c:=', ':=3', 'a:=1', 'b:=2'])

If I understand the wiki page about [Remapping Argument](http://wiki.ros.org/Remapping%20Arguments) correctly, these are illegal mappings.
These are neither remaps of names, because '1', '2' and '3' are no legal names as specified [here](http://wiki.ros.org/Names), nor they are parameter assignments, because they lack the leading underscore.

Currently this basically does not matter, but in order of my attempts to fix #1459 I'd like to create a function, which performs more extensive checks, to determine if a given string is a remap, than to check, if the string contains ':='. With this test present, this function can only check, if the substring in front of ':=' is a valid name; the later substring must basically be ignored.

cwecht:
`test_rospy_client.py` in `test_rospy` does something similar: it requires `rospy.myargv()` to strip out `-bar:=foo` even though it should strip out remaps only.

dirk-thomas:
What is the status of this ticket in relation to the merged PR #1683?

cwecht:
This is resolved by #1683.

PR #1683, opened 2019-04-01
added is_legal_remap() to rosgraph to make remap-detection more precise

### LINKED ISSUE #1459: Cannot send data containing `:=` through rostopic or rosservice
We cannot send `:=` data using the command line tools such as rosservice or rostopic.

```bash
$ rostopic pub /data std_msgs/String "data: ':='" 
> Usage: rostopic pub /topic type [args...]
> 
> rostopic: error: Please specify message values
```

When I escape the value, the string still contains the escaping backslash.

cwecht:
This fixes #1459 . Remaps are no longer only defined by the presence of *:='.

The criterion, a string need to satisfy to be considered as an remap, could be more precise (according to https://wiki.ros.org/Remapping%20Arguments), if #1671 gets fixed.

### DIFF
--- a/clients/rospy/src/rospy/client.py
+++ b/clients/rospy/src/rospy/client.py
@@ -82,7 +82,7 @@ def myargv(argv=None):
     """
     if argv is None:
         argv = sys.argv
-    return [a for a in argv if not rosgraph.names.REMAP in a]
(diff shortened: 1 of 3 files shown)


cwecht:

cwecht:

dirk-thomas:
Thanks for the patch.

PR #1685, opened 2019-04-03
[Kinetic] Adding 'template' to resolve compile error

kunaltyagi:
This syntax has already been used a few lines above in the same file. It prevents the compiler from assuming that `add` is being compared to `i` using the operator `<`.

Hoping to see this land soon :smile: (or am I the only one using Kinetic now :cold_sweat: )

Checked melodic-devel, and this bug [has been resolved there](https://github.com/ros/ros_comm/commit/65e4ce3de9747e359f38db65cf78763f75bcfbd1)

### DIFF
--- a/utilities/message_filters/include/message_filters/synchronizer.h
+++ b/utilities/message_filters/include/message_filters/synchronizer.h
@@ -355,7 +355,7 @@ private:
   template<int i>
   void cb(const typename mpl::at_c<Events, i>::type& evt)
   {
-    this->add<i>(evt);
+    this->template add<i>(evt);
   }
 
   uint32_t queue_size_;


dirk-thomas:
This is a backport of #1388.

> or am I the only one using Kinetic now

I assume you are either using Fedora or a custom g++ version on Ubuntu.

> Hoping to see this land soon

It will likely be a while since releasing a new version into Melodic is certainly higher on the priority list.

kunaltyagi:
> It will likely be a while since releasing a new version into Melodic is certainly higher on the priority list.

No problem.

You mentioned correctly that this is a backport of #1388. I accidentally linked my fork instead of ros repo for the commit (corrected now). Is there a separate process for backports or does this suffice?

And yes, I'm using an upgraded gcc on Ubuntu. I was just quipping because Kinetic is quite old now (and not receiving updates on many packages, rightfully so) and hopefully other people have planned the move to Melodic. It's hard to convey tone via text. 

Cheers :smiley:

dirk-thomas:
> Is there a separate process for backports or does this suffice?

Usually all backports are being merged in a single PR which enumerates all changes and mentions if / why the are (not) backported, e.g. #1477.

This is fine as is.

lucbettaieb:
@dirk-thomas Can this be merged please?  Message filters is broken on kinetic without this.

lucbettaieb:
@dirk-thomas @gerkey @mikepurvis ping ping <3

dirk-thomas:
Merging out-of-band since it is already part of a Melodic release and the original change won't be consider for backporting again.

PR #1687, opened 2019-04-03
fix windows build for rosbag_storage

kejxu:


dirk-thomas:
Thanks for the fix.

For the record: the regression was introduce in #1499.

PR #1688, opened 2019-04-03
Switch to yaml.safe_load(_all) to prevent YAMLLoadWarning

### LINKED ISSUE #1686: rosparam YAMLLoadWarning
Recently I've been seeing the following warning when I run `rosparam set` from commandline.
```
$ rosparam set /use_sim_time true
/opt/ros/melodic/lib/python2.7/dist-packages/rosparam/__init__.py:370:
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as
the default Loader is unsafe. Please read https://msg.pyyaml.org/load for
full details.
  set_param_raw(param, yaml.load(value), verbose=verbose)
```

As far as I could find this is related to the new release of pyyaml deprecating the use of `yaml.load(..)` as is described here: https://github.com/yaml/pyyaml/pull/257

Good to note is that I have PyYAML version 5.1 installed in my user directory through pip, this won't happen when using pyyaml from the Ubuntu repositories, since that's currently still version 3.12.

mbuijs:
Fixes #1686.

In https://github.com/yaml/pyyaml/issues/265 it's stated that `safe_load` always existed for this reason, I did a quick double check and it traces back at least to 2006, so this might even be safe to backport to older releases.

### DIFF
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -99,7 +99,7 @@ def convert_value(value, type_):
         raise ValueError("%s is not a '%s' type"%(value, type_))
     elif type_ == 'yaml':
         try:
-            return yaml.load(value)
+            return yaml.safe_load(value)
         except yaml.parser.ParserError as e:
             raise ValueError(e)
     else:
@@ -410,7 +410,7 @@ class Loader(object):
             if rosparam is None:
                 import rosparam
             try:
-                data = yaml.load(text)
+                data = yaml.safe_load(text)
                 # #3162: if there is no YAML, load() will return an
                 # empty string.  We want an empty dictionary instead
                 # for our representation of empty.
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -405,7 +405,7 @@ class Loader(object):
             # - lazy import: we have to import rosparam in oder to to configure the YAML constructors
             import rosparam
             try:
-                data = rosparam.yaml_load(text)
+                data = yaml.safe_load(text)
                 # #3162: if there is no YAML, load() will return an
                 # empty string.  We want an empty dictionary instead
                 # for our representation of empty.
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -47,9 +47,6 @@ from roslaunch.core import Param, RosbinExecutable, RLException, PHASE_SETUP
 
 from rosgraph.names import make_global_ns, ns_join, PRIV_NAME, load_mappings, is_legal_name, canonicalize_name
 
-#lazy-import global for yaml and rosparam
-rosparam = None
-
 class LoadException(RLException):
     """Error loading data as specified (e.g. cannot find included files, etc...)"""
     pass
@@ -406,11 +403,9 @@ class Loader(object):
                 text = subst_function(text)
             # parse YAML text
             # - lazy import: we have to import rosparam in oder to to configure the YAML constructors
-            global rosparam
-            if rosparam is None:
-                import rosparam
+            import rosparam
             try:
-                data = yaml.safe_load(text)
+                data = rosparam.yaml_load(text)
                 # #3162: if there is no YAML, load() will return an
                 # empty string.  We want an empty dictionary instead
                 # for our representation of empty.
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -47,6 +47,9 @@ from roslaunch.core import Param, RosbinExecutable, RLException, PHASE_SETUP
 
 from rosgraph.names import make_global_ns, ns_join, PRIV_NAME, load_mappings, is_legal_name, canonicalize_name
 
+#lazy-import global for yaml and rosparam
+rosparam = None
+
 class LoadException(RLException):
     """Error loading data as specified (e.g. cannot find included files, etc...)"""
     pass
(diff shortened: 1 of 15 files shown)


dirk-thomas:
Thanks for the patch. Can you maybe also update the other places in this repo where these functions are used?

mbuijs:
I can have a look. Would you prefer this also to be updated in tests?

dirk-thomas:
> Would you prefer this also to be updated in tests?

Yes, that would be good.

mbuijs:
So I had not realized about the following (from https://wiki.ros.org/rosparam):
> There are also special converters for angle radian/degree representations. Any Python-legal mathematical expression can be used with the radian value, with pi used to represent pi.

I'm doing some more changes to support this functionality while using the `SafeLoader` provided by pyyaml.

Furthermore I'll also try to see if there are other places where more features than the `SafeLoader` provides, so far I had just done a rapid search and replace...

mbuijs:
Additional question: which other tools aside of `rosparam` and `roslaunch` should have the ability to do the degrees and radians conversion functionality when evaluating YAML strings?

Edit: never mind, I see that also currently it is only supported in these 2, so I'll stick to that.

mbuijs:
So tests are passing now.
My experience with python is limited, I'm open to feedback, especially on how I applied the change in `rosparam` and `roslaunch`
1. attempting to not change the `SafeLoader` instance in PyYAML
2. adding the convencience functions `yaml_load` and `yaml_load_all` inside `rosparam`.

marting87:
Changes look to me, since I'm getting a lot of **YAMLLoadWarning** in my console I hope to see this merged.

tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -406,11 +403,9 @@ def load_rosparam(self, context, ros_config, cmd, param, file_, text, verbose=Tr
                 text = subst_function(text)
             # parse YAML text
             # - lazy import: we have to import rosparam in oder to to configure the YAML constructors
-            global rosparam
-            if rosparam is None:
-                import rosparam
+            import rosparam
```
dirk-thomas: Why has this import logic been changed? It looks unrelated to the patch.

tools/rosparam/src/rosparam/__init__.py
```diff
@@ -150,6 +151,12 @@ def print_params(params, ns):
     
 # yaml processing
 
+def yaml_load(str):
+    return yaml.load(str, Loader=ROSSafeLoader)
+
+def yaml_load_all(str):
+    return yaml.load_all(str, Loader=ROSSafeLoader)
```
dirk-thomas: The naming of both functions is misleading since it doesn't mention that it is actually using a safe loader.

tools/rosparam/src/rosparam/__init__.py
```diff
@@ -631,12 +638,12 @@ def yamlmain(argv=None):
...
+ROSSafeLoader.add_constructor(u'!radians', construct_angle_radians)
+ROSSafeLoader.add_constructor(u'!degrees', construct_angle_degrees)
 
 # allow both !degrees 180, !radians 2*pi
 pattern = re.compile(r'^deg\([^\)]*\)$')
-yaml.add_implicit_resolver(u'!degrees', pattern, first="deg(")
+ROSSafeLoader.add_implicit_resolver(u'!degrees', pattern, first="deg(")
 pattern = re.compile(r'^rad\([^\)]*\)$')
-yaml.add_implicit_resolver(u'!radians', pattern, first="rad(")
+ROSSafeLoader.add_implicit_resolver(u'!radians', pattern, first="rad(")
```
dirk-thomas: Before this change all the added constructors and implicit resolvers affected the global yaml module. If any other existing code relies on that behavior it would break with this change.

tools/roslaunch/src/roslaunch/loader.py
mbuijs: You're right, it's unrelated to the functional change in this patch.

The reason I touched it was that I was checking if/where `rosparam` was imported within `roslaunch` globally and finding the global `rosparam` variable confused me. Then I googled for this construction and came across https://stackoverflow.com/a/4177780/177643, where it's pointed out that python already caches imports, so this optimization would not really help performance.
Aside of that, it pollutes the interface of this python file, by exposing `rosparam` which might be `None` or the loaded `rosparam` module, depending on if this function has been called or not. Since this is the only use of rosparam within this module, I thought it would be a good idea to clean up.

It's fine for me to roll this back, you can decide.

tools/rosparam/src/rosparam/__init__.py
mbuijs: Good point, I'll adjust it to just modify the global `SafeLoader`. Then I can also remove the added yaml functions that I had added.

mbuijs:

tools/rosparam/src/rosparam/__init__.py
mbuijs: As mentioned below I will remove these functions and instead add the resolvers to the `SafeLoader` object in the `yaml` module.

mbuijs:
@dirk-thomas I processed your remarks and all checks have passed. Would you mind doing another check and merging if it's good?

tools/roslaunch/src/roslaunch/loader.py
dirk-thomas: Please remove the change from this PR since the title won't let you see that this is part of the patch. Please feel free tp create a separate PR with this change.

tools/rosparam/src/rosparam/__init__.py
```diff
@@ -149,7 +150,6 @@ def print_params(params, ns):
         print(params)
     
 # yaml processing
-
```
dirk-thomas: Please revert this unrelated white space change.

dirk-thomas:
Thank you for iterating on this.

PR #1692, opened 2019-04-11
add rostopic as run_depend of rostest

furushchev:
This is required to run `publishtest`.

https://github.com/ros/ros_comm/blob/melodic-devel/tools/rostest/nodes/publishtest#L63


dirk-thomas:
Please build and test a patch locally before creating a pull request. As the CI failure indicates this is not a viable dependency to add:

> Error: Circular dependency in subset of packages: rosbag, rosbag_storage, rostopic, topic_tools

PR #1698, opened 2019-04-15
rosgraph/network: use urlparse for parsing the port, whick makes ipv6 possible

### LINKED ISSUE #1694: parse_http_host_and_port function does not support IPv6 address
https://github.com/ros/ros_comm/blob/71872217c441280f189544aa0c2cf4909774f953/tools/rosgraph/src/rosgraph/network.py#L75

This function doesn't support IPv6 addresses of the form:
http://[<link address>%<network interface]:<port>

where link address is of the form: aaaa:bbbb:cccc:dddd:eeee:ffff:gggg:hhhh:iiii

cwecht:
This requires python 2.5.

This fixes #1694.

### DIFF
--- a/tools/rosgraph/src/rosgraph/network.py
+++ b/tools/rosgraph/src/rosgraph/network.py
@@ -83,11 +83,10 @@ def parse_http_host_and_port(url):
     :returns: hostname and port number in URL or 80 (default), ``(str, int)``
     :raises: :exc:`ValueError` If the url does not validate
     """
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thanks for the patch.

I removed the assert checking the version number since Python versions < 2.7 have been EOL for quite some time. Also the comment wasn't necessary anymore since the condition is not self explaining.

PR #1701, opened 2019-04-23
Fix error handling for Topic constructor

scooler:
There is no `s` variable in scope - and we clearly wanna display `reg_type`

To reproduce:
```
rospy.topics.Topic('/topic', rospy.AnyMsg, 'aa')
```
I get
```
raise TypeError("invalid reg_type: %s"%s)
NameError: global name 's' is not defined
```


dirk-thomas:
Thanks for the fix.

PR #1702, opened 2019-04-26
Random Access View for single topic

sujiwo:
This patch contains a new class (RandomAccessBag) for accessing a view in random fashion (not sequentially) by caching index entry after updates.


tools/rosbag_storage/include/rosbag/message_instance.h
```diff
@@ -93,7 +93,6 @@ class ROSBAG_STORAGE_DECL MessageInstance
     //! Size of serialized message
     uint32_t size() const;
 
-private:
```
dirk-thomas, line 96: I am a bit concerns of publically exposing these private members. Can you please elaborate why you made that change and what was missing from the public interface to not require this.

dirk-thomas:
@sujiwo Closing due to no response. Please consider to open a new pull request targeting `noetic-devel`.

PR #1703, opened 2019-04-26
Added possibility to pass rospy.Duration as timeout to wait_for_service and wait_for_message.

### LINKED ISSUE #1658: Confusing: C++ waitForService timeout is milisec, Python wait_for_service timeout is seconds
The function docs say it, but it still seems to me confusing that the C++ part takes in milisec and Python seconds.
C++ has another method that takes in `ros::Duration` as timeout. However, Python only accepts doubles. 

I understand that it's impossible to change either behavior of the C++ or the Python calls regarding doubles, but could the Python version try to parse the argument as `rospy.Duration`?

peci1:
Fixes https://github.com/ros/ros_comm/issues/1658.


dirk-thomas:
Thanks for the patch.

Issue #158, opened 2013-01-31
roscpp: roscpp topic unsubscribe callbacks not called if no messages published  (ros ticket #3403)


tfoote:
I've noticed that the unsubscribe callbacks in roscpp, set up with advertise options, don't get called if the publisher doesn't publish any messages before the disconnection. 

Specifically:
- Set up publisher with subscribe/unsubscribe callbacks
- Subscribe to topic, subscribe callback gets called
- Don't publish messages
- Unsubscribe to topic, unsubscribe callback doesn't get called

Attached is a modified version of talker.cpp to test this problem. To verify:
- Run modified talker.cpp
  - It won't publish
- Run "rostopic echo chatter", then ctrl-C a few times
- Run "rostopic info chatter" to make sure topic not subscribed to
- Unsubscribe callback never gets hit.

trac data:
- Owner: **straszheim**
- Reporter: **watts**
- Reported at: **Tue Mar 15 13:41:00 2011**
- URL: https://code.ros.org/trac/ros/ticket/3403

tfoote:
[watts] I should add that if messages actually publish from a publisher, then the unsubscribe callback does get hit.

tfoote:
[watts] I've noticed several problems with subscriber connection callbacks using the openni_launch Kinect driver. Some of the nodelets use lazy subscriptions for Kinect data. I have noticed a warning message that the topic "camera/depth/image_rect_raw" wasn't sync'd. It turns out that this topic was not being published, because the nodelet wasn't subscribed to "camera/depth/image_raw". (Nodelet is image_proc/src/nodelets/rectify.cpp). 

Patrick, I'm using the Electric openni_kinect stack from "default", version 445. 

Troy, can you investigate this problem and see if  we have any other problems with these connection callbacks?

tfoote:
[mihelich] I've noticed this issue with unsubscribe callbacks too. However, it's not true that they're never called; they _do_ get called exactly 60s after you subscribe.

In practice this shouldn't normally be an issue. In each nodelet we use connect callbacks to control whether it hooks into a data source (camera or ROS topic). As long as it's getting data, it publishes some processed output. So if all subscribers to the output unsubscribe, the next call to `publish()` triggers the disconnect callback, which unhooks from the data source. So as long as you're streaming data, disconnects will be timely.

[...]

PR #1704, opened 2019-04-26
roscpp/transport_tcp: enable poll event POLLRDHUP to detect dead

### LINKED ISSUE #1699: roscpp: getNumSubscribers() does not return 0 when a topic is unsubscribed
Tested in melodic.

It seems that the method getNumSubscribers() does not return 0 when a topic is unsubscribed **if a roscpp publisher is not publishing**.

**How to reproduce:** Have a simple publisher running that does not publish anything but only checks the number of subscribers
```cpp
{
  ros::init(argc, argv, "test_get_num_subscribers");

  ros::NodeHandle nh;
  ros::Publisher pub = nh.advertise<std_msgs::String>("test_topic", 1000);

  ros::Rate r(1);
  while(ros::ok())
  {
    if (pub.getNumSubscribers() > 0)
    {
      ROS_INFO("Someone is listening :)");
    } else {
      ROS_INFO("I am alone :(");
    }
    ros::spinOnce();
    r.sleep();
  }

  return 0;
}
```
and then subcribe to and unsubscribe from that topic, e.g, with

```sh
rostopic echo /test_topic && date +%s
^C
1555934665
```

**What should happen**:  The publisher node should recognize the new subscriber, print out `Someone is listening :)` as long as the subscriber is active, and more or less immediately realize that there are no subscribers any more and switch back to `I am alone :(`.

**What happens:** The publisher recognizes a new subscriber but does not notice that the subscriber unsubscribes shortly after:

```sh
...
[ INFO] [1555934661.098353493]: I am alone :(
[ INFO] [1555934662.098477395]: I am alone :(
[ INFO] [1555934663.098413898]: Someone is listening :)
[ INFO] [1555934664.098555165]: Someone is listening :)
[ INFO] [1555934665.098341645]: Someone is listening :)  (subscriber is shut off again) 
[ INFO] [1555934666.098435134]: Someone is listening :)
[ INFO] [1555934667.098370738]: Someone is listening :)
[ INFO] [1555934668.098414957]: Someone is listening :)
[ INFO] [1555934669.098372386]: Someone is listening :)
[ INFO] [1555934670.098341077]: Someone is listening :)
...
```

Only after approx. 120 seconds the publisher realizes that there is no subscriber any longer:

```sh
...
[ INFO] [1555934784.098598787]: Someone is listening :)
[ INFO] [1555934785.098545304]: Someone is listening :) (120 seconds later)
[ INFO] [1555934786.098460942]: Someone is listening :)
[ INFO] [1555934787.098624760]: I am alone :(
[ INFO] [1555934788.098622298]: I am alone :(
[ INFO] [1555934789.098441571]: I am alone :(
...
```

**Notes:**

- behavior is different when node actually publishes messages when someone listens: then `getNumSubscribers()` returns 0 immediately after subscribers are gone
- could not reproduce with rospy
- the 120 seconds is quite reproducible, so I guess there is a simple timeout set somewhere to update the subscription info

cwecht:
connections properly.

The poll event ``POLLRDHUP`` is needed to detect half-closed connections ([source](http://lkml.iu.edu/hypermail/linux/kernel/0603.0/2162.html)). This is needed to avoid the delay reported in #1699.

### DIFF
--- a/clients/roscpp/src/libros/transport/transport_tcp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_tcp.cpp
@@ -138,6 +138,10 @@ bool TransportTCP::initializeSocket()
   {
     ROS_DEBUG("Adding tcp socket [%d] to pollset", sock_);
     poll_set_->addSocket(sock_, boost::bind(&TransportTCP::socketUpdate, this, _1), shared_from_this());
+#if defined(POLLRDHUP) // POLLRDHUP is not part of POSIX!
+    // This is needed to detect dead connections. #1704
+    poll_set_->addEvents(sock_, POLLRDHUP);
+#endif
   }
 
   if (!(flags_ & SYNCHRONOUS))
@@ -696,6 +700,9 @@ void TransportTCP::socketUpdate(int events)
 
   if((events & POLLERR) ||
      (events & POLLHUP) ||
+#if defined(POLLRDHUP) // POLLRDHUP is not part of POSIX!
+     (events & POLLRDHUP) ||
+#endif
      (events & POLLNVAL))
   {
     uint32_t error = -1;


flixr:
That look great!

flixr (approved):
LGTM.

flixr:
I can confirm that this fixes #1699 on Linux

flixr:
This might also fix the long standing #158

flixr:
@dirk-thomas friendly ping.... could this please be merged?
I also reviewed and tested this and can confirm it fixes the problem (on Linux).

dirk-thomas:
Thanks for the patch.

protobits:
Hi, I'm on melodic on Ubuntu Bionic with ros-comm 1.14.3-0bionic.20191008.215948. I'm experiencing https://github.com/ros/ros_comm/issues/158 (disconnect callback is invoked after long time). I understand this PR was merged after the 1.14.3 release, correct? If so, any idea when the next release (containing this fix) is due?

Thanks

PR #1707, opened 2019-04-27
roscpp: TransportTCP: Allow socket() to return 0

g-arjones:
In situations that an application has stdin closed, socket() may (and will) return 0. This is totally fine.

### DIFF
--- a/clients/roscpp/src/libros/transport/transport_tcp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_tcp.cpp
@@ -389,7 +389,7 @@ bool TransportTCP::listen(int port, int backlog, const AcceptCallback& accept_cb
     sa_len_ = sizeof(sockaddr_in);
   }
 
-  if (sock_ <= 0)
+  if (sock_ == ROS_INVALID_SOCKET)
   {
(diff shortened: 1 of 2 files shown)


g-arjones:
I would really appreciate if this was also backported to Kinetic.... I will gladly open a PR for the corresponding branch if that's the case. Thank you.

cwecht:
Just to backup the argument up: ``socket()`` is [specified](http://man7.org/linux/man-pages/man2/socket.2.html#RETURN_VALUE) to return either a valid file descriptor (which is nonnegative) or -1 in the error case.

The same issue can be found in ``TransportUDP::createIncoming`` and should be fixed there as well.

g-arjones:
@dirk-thomas ?

dirk-thomas:
> The same issue can be found in TransportUDP::createIncoming and should be fixed there as well.

Please update the PR to do so.

g-arjones:
Done

ivan1993br:
I am facing the same problem, any updates?

ivan1993br:
@dirk-thomas

dirk-thomas:

dirk-thomas:
Thanks for the patch.

PR #1708, opened 2019-04-30
rosbag fix: keep latched topics latched

### LINKED ISSUE #1675: rosbag fix drops latching
Hi

This is related to some of the issues mentioned before, namely #679 and #1049. Fundamentally this is the same issue, and the fix has been implemented in the underlying rosbag API. However specifically the fix has not seem to be propagated to the `rosbag fix` command, which still drops latching for /tf_static. Here's a minimal example.

[test_2019-03-25-10-59-22.zip](https://github.com/ros/ros_comm/files/3004850/test_2019-03-25-10-59-22.zip)

then run 
```
rosbag fix test_2019-03-25-10-59-22.bag test_2019-03-25-10-59-22_fixed.bag placeholder.bmr
```

`placeholder.bmr` is an empty file just to make the tool happy.

The result should be that the `/tf_static` topic going from latching to non latching between the old and new bag.

A cursory look at the python script seems to indicate that the `fixbag2` function seems to still be using the old interface which is causing this issue.

cwecht:
This fixes #1675.

### DIFF
--- a/tools/rosbag/src/rosbag/migration.py
+++ b/tools/rosbag/src/rosbag/migration.py
@@ -118,6 +118,13 @@ def checkmessages(migrator, messages):
             
     return migrations
 
+def _migrate_connection_header(conn_header, new_msg_type):
+    conn_header['type'] = new_msg_type._type
+    conn_header['md5sum'] = new_msg_type._md5sum
+    conn_header['message_definition'] = new_msg_type._full_text
+
+    return conn_header
+
 ## Fix a bag so that it can be played in the current system
 #
 # @param migrator The message migrator to use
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch.

tools/rosbag/src/rosbag/migration.py
```diff
@@ -158,13 +166,14 @@ def fixbag2(migrator, inbag, outbag, force=False):
...
+        for topic, msg, t, conn_header in bag.read_messages(raw=True, return_connection_header=True):
             new_msg_type = migrator.find_target(msg[4])
             if new_msg_type != None:
                 mig_msg = migrator.migrate_raw(msg, (new_msg_type._type, None, new_msg_type._md5sum, None, new_msg_type))
-                rebag.write(topic, mig_msg, t, raw=True)
+                new_conn_header = _migrate_connection_header(conn_header, new_msg_type)
+                rebag.write(topic, mig_msg, t, connection_header=new_conn_header, raw=True)
             else:
-                rebag.write(topic, msg, t, raw=True)
+                rebag.write(topic, msg, t, connection_header=connection_header, raw=True)
```
dirk-thomas, line 176: `connection_header` doesn't seem to be defined here. See #1952.

PR #1709, opened 2019-04-30
XmlRpcValue added bool assignment operator

### LINKED ISSUE #1632: ros_comm/utilities/xmlrpcpp wrong representation/conversion of bool vectors with ros::param::set
Problem noticed with used ros version : kinetic
Version of the xmlrpcpp tested for problem: 1.12.12
Found on system: Ubuntu 16.04

While reading and writing a lot of parameters form and to the parameterserver I noticed a strange behavoir with arrays of boolean values from the cpp side. I'm querrying all the parameters via the overloaded getParam function with a XmlRpcValue. 

For the python version when I set a parameter  with:
`rospy.set_param("list_of_bools",[True,False,False,True])`
I will get the return from rosparam get  as the following:
`[true, false, false, true]`

and  get the output on the cpp side as the xml string as the following:
`<value><array><data><value><boolean>1</boolean></value><value><boolean>0</boolean></value><value><boolean>0</boolean></value><value><boolean>1</boolean></value></data></array></value>`

This is the output I would expect as these types are bools.

After trying to set the list the same way from the cpp side i got the following results:
`[1, 0, 0, 1]`
and for the cpp getParam:
`<value><array><data><value><i4>1</i4></value><value><i4>0</i4></value><value><i4>0</i4></value><value><i4>1</i4></value></data></array></value>`

so at first look it seemed as if the boolean values got converted to integers somewhere on the way to or from the parameter server. Since they normally should be bool as shown in the example above.

So I investigated it a bit further and also tried it out with some single boolean values. The result of this was that single bools don't seem to have that issue.

Then I drillt into the setParam for the vectors a bit deeper and I could reproduce the problem in a few lines when parsing them the same way as it would happen in ros::param::set. For reproducing the problem and narrowing it down to the xmlrpcpp class /part/package I did the following : 

for single bool values:

```
bool tmp = true;

XmlRpc::XmlRpcValue v(tmp);
std::string xml = v.toXml();

```
which returned the xml representation as expected.
And for the array of boolean values:

 ```
std::vector<bool> temp_bool;
 temp_bool.push_back(true);
 temp_bool.push_back(false);
 temp_bool.push_back(false);
 temp_bool.push_back(true);

      
 XmlRpc::XmlRpcValue xml_vec;
 xml_vec.setSize(temp_bool.size());

 // Copy the contents into the XmlRpcValue
 for(size_t i=0; i < temp_bool.size(); i++) 
{
           xml_vec[i] = temp_bool.at(i);
 }
 std::string xml =xml_vec.toXml();
```

The reproduction snipped does the same as if the bool vector array was processed by the following functions:

```
 void ros::param::set	(	const std::string & 	key,
const std::vector< bool > & 	vec 
)  //param.cpp 165
//->
void ros::param::setImpl | ( | const std::string & | key,
  |   | const std::vector< T > & | vec
  | )//param.cpp 130
```

I hope this was the right place to post it and that this problem can be addressed.

cwecht:
This fixes #1632. The missing  assignment operator caused booleans to be converted to ints.

### DIFF
--- a/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcValue.h
+++ b/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcValue.h
@@ -80,6 +80,7 @@ namespace XmlRpc {
 
     // Operators
     XmlRpcValue& operator=(XmlRpcValue const& rhs);
+    XmlRpcValue& operator=(bool const& rhs) { return operator=(XmlRpcValue(rhs)); }
     XmlRpcValue& operator=(int const& rhs) { return operator=(XmlRpcValue(rhs)); }
     XmlRpcValue& operator=(double const& rhs) { return operator=(XmlRpcValue(rhs)); }
     XmlRpcValue& operator=(const char* rhs) { return operator=(XmlRpcValue(std::string(rhs))); }


dirk-thomas:
Thanks for the patch.

Issue #1711, opened 2019-04-30
Performance enhancements for rosbag filter


mikepurvis:
Currently, `rosbag filter` deserializes every message in a bag, in a single threaded process, with a fresh `eval` each time. This is all enormously expensive and slow:

https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/tools/rosbag/src/rosbag/rosbag_main.py#L374-L383

Some possibilities for how to improve this situation:

- Lazy deserialization: Instead of deserializing every time, pass a proxy object which uses __getattr__ to deserialize on demand and pass through accesses. This would prevent unnecessary deserialization expense for the use-cases which only filter on topic name or timestamp.
- Wrap the `eval` in a lambda, so that it needs to be compiled just once rather than on each use: https://stackoverflow.com/a/12467755/109517
- Use multiprocessing to parallelize the eval invocations—would have to check how much of a gain there is to be had here, particularly with the added cost of managing the work queue to maintain sequence.

dirk-thomas:
Addressed by #1712.

PR #1712, opened 2019-04-30
Wrap the rosbag filter eval in a lambda

mikepurvis:
As proposed in #1711, and suggested here: https://stackoverflow.com/a/12467755/109517

In testing with some of our bags here, I got 978.9 kb/s and 942.9 kb/s before the change, and 1147.2 kb/s and 1126.6 kb/s after, or about 18% faster.


cwecht:

dirk-thomas:
Thanks for the patch.

PR #1720, opened 2019-05-23
Do not raise socket exception during shutdown

yli-cpr:
When using rospy with gevent, there is always socket exception raised during shutdown procedure, and it can flood logs. This simple change fixes the problem. Do this kind of exceptions have any value?

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -158,7 +158,8 @@ class TCPServer(object):
                 (e_errno, msg) = e.args
                 if e_errno == errno.EINTR: #interrupted system call
                     continue
-                raise
+                if not self.is_shutdown:
+                    raise
             if self.is_shutdown:
                 break
             try:


mikepurvis:
This looks reasonable to me.

yli-cpr:
Thanks @mikepurvis. The stack trace looks like this:
```Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
  File ".../rospy/impl/tcpros_base.py", line 154, in run
    (client_sock, client_addr) = self.server_sock.accept()
  File "/usr/lib/python2.7/dist-packages/gevent/_socket2.py", line 194, in accept
    self._wait(self._read_event)
  File "/usr/lib/python2.7/dist-packages/gevent/_socket2.py", line 179, in _wait
    self.hub.wait(watcher)
  File "/usr/lib/python2.7/dist-packages/gevent/hub.py", line 627, in wait
    result = waiter.get()
  File "/usr/lib/python2.7/dist-packages/gevent/hub.py", line 875, in get
    return self.hub.switch()
  File "/usr/lib/python2.7/dist-packages/gevent/hub.py", line 606, in switch
    return greenlet.switch(self)
error: [Errno 9] File descriptor was closed in another greenlet
```
PS: gevent closes all sockets during a graceful shutdown. That's why the ROS TCP server thread gets the error.

dirk-thomas:
Thanks for the patch.

PR #1722, opened 2019-06-07
Fix bare pointer in topic_tools::ShapeShifter

amiller27:
The copy constructor and copy assignment operator on the ShapeShifter class were wrong because the class was using a bare pointer for its internal buffer.  This was manifesting itself as double-free's for copy-assigned ShapeShifter objects.  This PR switches to using a std::vector, so the constructors are automatically handled correctly.

All the tests pass, and this has been verified to fix the issue running in a real node with the AddressSanitizer enabled.

### DIFF
--- a/tools/topic_tools/include/topic_tools/shape_shifter.h
+++ b/tools/topic_tools/include/topic_tools/shape_shifter.h
@@ -99,10 +99,7 @@ private:
   std::string md5, datatype, msg_def, latching;
   bool typed;
 
-  uint8_t *msgBuf;
-  uint32_t msgBufUsed;
-  uint32_t msgBufAlloc;
-  
+  std::vector<uint8_t> msgBuf;
 };
   
 }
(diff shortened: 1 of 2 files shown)


cwecht:
Although this is a nice improvement, this introduces an ABI-break. But the alternative would be to implement copy-constructors and such for ``ShapeShifter``, which would also break API, so this might be the best possible fix.

mikepurvis:
+1 to this change. ABI break is a bummer, but the most significant consumers of this class are `rosbag` and the tools in this same package, so it should be possible to bump it without affecting too many users, provided there's a `conflict` added to this package for the current rosbag, forcing them to upgrade together.

Other affected users I'm aware of include `rosserial_server`, `ros_type_introspection`, and `variant`.

dirk-thomas:
Thanks for the patch.

peci1:
@dirk-thomas @jacobperron I have locally a fix for the bare pointer handling for melodic which should retain ABI. I've just added proper copy- and move- constructors and assignment operators and now ShapeShifter behaves correctly. Is my reasoning correct that adding these 4 functions will not change ABI? Should I invest time into making a PR to melodic-devel, or is it already doomed with the EOL date closing in?

PR #1725, opened 2019-06-10
[Windows][melodic-devel] Use taskkill to terminate node process tree in LocalProcess.stop()

seanyen:
There are many [restrictions](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/signal?view=vs-2019) when using `signal` on Windows. For example, `SIGTERM` is not actually generated and `SIGINT` is not supported for Win32 applications. In order to terminate the child processes from roslaunch, I proposed to use [`taskkill`](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill) instead on Windows.


dirk-thomas:
Thanks for the patch.

seanyen:
fixes https://github.com/ms-iot/ROSOnWindows/issues/62

PR #1731, opened 2019-06-20
Add ability to increase and decrease the playback speed while rosbag is running

bobhenz-jabil:


dirk-thomas:
The current patch break API and doesn't contain any description...

bobhenz-jabil:
Sorry for the half-hearted PR. First of all, I am working in ROS Kinetic (on Ubuntu 16.04) and I have not been able to build ros_comm on my system.

To describe the changes: When playing back large bag-files I have often wished I could speed through the "boring" parts and then slow way down to watch key pieces of the file. I can do this manually by specifying a faster rate on the command line, noting where the interesting pieces show up, then stopping and restarting play back with a starting point and a slower rate specified on the command line but that is cumbersome. This change allows the user to press the "+" and "-" to speed up or slow down playback (respectively) _during_ playback.

bobhenz-jabil:
I do not understand your comment about the patch breaking the API. I did not intend to modify the existing user interface of rosbag. Can you please elaborate?

dirk-thomas:
> I do not understand your comment about the patch breaking the API.

Any C++ API not private is considered a public interface. Changing the signature in an incompatible way or even removing existing function is potentially breaking existing code and therefore not desired.

bobhenz-jabil:
Oh wow. OK I understand now. Unfortunately I don't think I can replicate the old interface while simplifying the internals in the way I needed to in order to add this feature.

Is there a branch out there that does _not_ have to be API compatible?

dirk-thomas:
> Is there a branch out there that does not have to be API compatible?

The to-be-created branch for Noetic can break ABI. But breaking API in such a fundamental way that downstream packages must be updated doesn't seem realistic. Instead you might want to try to realize the new feature while trying to keep existing API (as much as possible), e.g. by introducing new parallel API you need for the new feature.

dirk-thomas:
@bobhenz-jabil Closing due to no further reply. Please feel free to open a new PR against `noetic-devel`.

PR #1733, opened 2019-06-28
[roslaunch] Missing args

DLu:
If you don't pass the args down to the `get_ros_namespace` call, then arguments may not get used when using a non-command-line set of args.


dirk-thomas:
Thanks for the fix.

PR #1739, opened 2019-07-03
added archive desination for lzpy library

willdzeng:
In order to build static libraries by:
```
catkin_make -DBUILD_SHARED_LIBS=0
```
without this PR, it fails at:
```
CMake Error at /opt/ros/kinetic/share/catkin/cmake/custom_install.cmake:13 (_install):
  install TARGETS given no ARCHIVE DESTINATION for static library target
  "roslz4_py".
Call Stack (most recent call first):
  ros_comm/utilities/roslz4/CMakeLists.txt:59 (install)
```

Would be nice to have the archive destination in such that we can build static libraries.


willdzeng:
It's a very simple change on the CMake file, would be great if have someone can take a look and approve. Thanks.

dirk-thomas:
Thanks for the fix.

dirk-thomas:
I changes the target branch to `melodic-devel`. Please rebase your branch so that the patch applies cleanly.

willdzeng:
It seems like the changes are already in melodic-devel

PR #1740, opened 2019-07-04
roscpp/transport_udp: zero-initialize sockaddr_in object

### LINKED ISSUE #1736: sockaddr_in.in transport_udp.cpp is not initialized
Coverity report Using uninitialized value "sin". Field "sin.sin_zero" is uninitialized when calling "connect".
https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/clients/roscpp/src/libros/transport/transport_udp.cpp#L160-L208

I remember the uninitialized sin_zero might cause undefined behavior, not sure if it is still valid

cwecht:
Fixed #1736.


dirk-thomas:
Thanks for the patch.

PR #1741, opened 2019-07-04
rosbag/record: fix potential signed int overflow

### LINKED ISSUE #1738: CVE-2019-13445: potential integer overflow
Potentially overflowing expression "1048576 * S" with type "int" (32 bits, signed) is evaluated using 32-bit arithmetic, and then used in a context that expects an expression of type "uint64_t" (64 bits, unsigned).
https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/tools/rosbag/src/record.cpp#L129

To avoid overflow, cast either "1048576" or "S" to type "uint64_t".

cwecht:
Fixed #1738.


cwecht:

mikepurvis:
Looks like a safe change, thanks!

PR #1750, opened 2019-07-10
Check for fclose returns 0

### LINKED ISSUE #1743: Coverity - Calling "fclose" without checking return value
Event check_return: | Calling "fclose" without checking return value
https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/tools/rosbag_storage/src/chunked_file.cpp#L103

artivis:
Fix #1743


cwecht:

PR #1751, opened 2019-07-10
ServiceManager::unregisterService returns result of master::execute("unregisterService")

### LINKED ISSUE #1744: Coverity - Calling "execute" without checking return value
Event check_return: | Calling "execute" without checking return value
https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/clients/roscpp/src/libros/service_manager.cpp#L203

artivis:
`ServiceManager::unregisterService` returns result of `master::execute("unregisterService")`

fix #1744


dirk-thomas:
Thanks for the patch.

PR #1759, opened 2019-07-16
Added const indexer for xmlrpc

Vuwij:
- Missing the constant version for the [] operator


mikepurvis:
Great, thanks! I've had frustrations in the past with xmlrpcpp not being more const-friendly.

PR #1763, opened 2019-07-17
fixed potential buffer overflow

### LINKED ISSUE #1752: Potential buffer overflow
Copy the Internet host address of address->sin_addr to local variable with checking the buffer length, can cause potential buffer overflow if using a IP address with a long hostname that is obtained via a reverse DNS lookup

https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/clients/roscpp/src/libros/transport/transport_tcp.cpp#L287-L291

danielwangksu:
fix https://github.com/ros/ros_comm/issues/1752

change `strcpy` to `strncpy` and initialize `namebuf` to avoid non-NULL-terminated string

### DIFF
--- a/clients/roscpp/src/libros/transport/transport_tcp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_tcp.cpp
@@ -276,7 +276,7 @@ bool TransportTCP::connect(const std::string& host, int port)
 
     bool found = false;
     struct addrinfo* it = addr;
-    char namebuf[128] = {0};
+    char namebuf[128] = {};
     for (; it; it = it->ai_next)
     {
       if (!s_use_ipv6_ && it->ai_family == AF_INET)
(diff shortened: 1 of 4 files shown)


clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -276,7 +276,7 @@ bool TransportTCP::connect(const std::string& host, int port)
 
     bool found = false;
     struct addrinfo* it = addr;
-    char namebuf[128];
+    char namebuf[128] = {0};
```
cwecht: This is somewhat misleading. One might assume that `char numebuf [128] = {1}` would initialize all values to 1, but will only initialize the first value with 1 und every other with 0. `char numebuf [128] = {}` might be the better choice as it makes the [zero initialization](https://en.cppreference.com/w/cpp/language/zero_initialization) obvious.

danielwangksu: you are right. I will update it. Thanks

cwecht:

cwecht:
LGTM now!

mikepurvis:
Thanks for the contribution! This was already covered by the changes from #1771; closing now.

PR #1765, opened 2019-07-17
fixed check null return from malloc

### LINKED ISSUE #1753: unimplemented null return - malloc may return NULL
XXH_malloc, wrapper of malloc may return NULL,  state need to be checked
https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/utilities/roslz4/src/xxhash.c#L333-L334

danielwangksu:
fix https://github.com/ros/ros_comm/issues/1753


danielwangksu:
close: same as https://github.com/ros/ros_comm/pull/1778

PR #1770, opened 2019-07-19
Small fixes for roslaunch-check on Python 3.

mikepurvis:
Stuff that came up in the course of our migration work.

### DIFF
--- a/tools/roslaunch/scripts/roslaunch-check
+++ b/tools/roslaunch/scripts/roslaunch-check
@@ -113,7 +113,7 @@ if __name__ == '__main__':
 
     if error_msg:
         print("FAILURE:\n%s"%error_msg, file=sys.stderr)
-        with open(test_file, 'w') as f:
+        with open(test_file, 'wb') as f:
             message = "roslaunch check [%s] failed"%(roslaunch_path)
             f.write(rosunit.junitxml.test_failure_junit_xml(test_name, message, stdout=error_msg,
                     class_name="roslaunch.RoslaunchCheck", testcase_name="%s_%s" % (pkg, outname)))
@@ -121,6 +121,6 @@ if __name__ == '__main__':
         sys.exit(1)
     else:
         print("passed")
-        with open(test_file, 'w') as f:
+        with open(test_file, 'wb') as f:
             f.write(rosunit.junitxml.test_success_junit_xml(test_name,
                     class_name="roslaunch.RoslaunchCheck", testcase_name="%s_%s" % (pkg, outname)))


tools/roslaunch/scripts/roslaunch-check
```diff
@@ -113,14 +113,14 @@ if __name__ == '__main__':
 
     if error_msg:
         print("FAILURE:\n%s"%error_msg, file=sys.stderr)
-        with open(test_file, 'w') as f:
+        with open(test_file, 'wb') as f:
```
mikepurvis, line 116: I believe this is the intended use given that the `junitxml` functions explicitly return encoded strings/bytes, see:

https://github.com/ros/ros/blob/9f2051dd809352fa11bab4671f47d95129125a2e/tools/rosunit/src/rosunit/junitxml.py#L495

And the xml output elsewhere has tests which confirm that it returns bytes, eg: https://github.com/ros/ros/pull/212

mikepurvis:
Ugh, the new logic only works for Python 3's minidom, as the 2.7 version doesn't supply an implementation of `__contains__`:

```
  File "/tmp/ws/src/ros_comm/tools/roslaunch/src/roslaunch/depends.py", line 99, in _check_ifunless
    if 'if' in tag.attributes:
  File "/usr/lib/python2.7/xml/dom/minidom.py", line 522, in __getitem__
    return self._attrs[attname_or_tuple]
```

I don't see how to get out of making this a versioned check, possibly even as simple as `hasattr(dict, 'has_key')`. Will update.

tools/roslaunch/src/roslaunch/depends.py
```diff
@@ -84,23 +84,31 @@ def __repr__(self):
     def __str__(self):
         return "nodes: %s\nincludes: %s\npkgs: %s"%(str(self.nodes), str(self.includes), str(self.pkgs))
 
+def _tag_has_attribute(tag, attribute):
+    if hasattr(tag.attributes, '__contains__'):
+        # Python 3's minidom supports the in operator.
+        return attribute in tag.attributes
+    else:
+        # Python 2 uses a named method instead.
+        return tag.attributes.has_key(attribute)
```
mikepurvis: This could also be handled by patching a `__contains__` method onto the affected objects, but the wrapper function approach is simple, easy to understand, and still won't be too difficult to unwind in the unlikely scenario that Python 2 support is ever fully dropped from the ROS 1 codebase.

mikepurvis: Nevermind, this part of the change got covered in #1782.

dirk-thomas:
Thanks for the fix.

PR #1772, opened 2019-07-23
Fixing zip python3 error

kalyanvasudev:
zip returns an iterator in Python3 where as it returns list in Python2. 
Thus changing it to be compatible with both Python2 and Python3.

Useful resources - https://stackoverflow.com/questions/27431390/typeerror-zip-object-is-not-subscriptable

### DIFF
--- a/utilities/message_filters/src/message_filters/__init__.py
+++ b/utilities/message_filters/src/message_filters/__init__.py
@@ -279,7 +279,7 @@ class ApproximateTimeSynchronizer(TimeSynchronizer):
                 return
             topic_stamps = sorted(topic_stamps, key=lambda x: x[1])
             stamps.append(topic_stamps)
-        for vv in itertools.product(*[zip(*s)[0] for s in stamps]):
+        for vv in itertools.product(*[next(iter(zip(*s))) for s in stamps]):
             vv = list(vv)
             # insert the new message
             if my_queue_index is not None:


dirk-thomas:
I retargeted the `melodic-devel` branch. Please rebase your patch to cleanly apply.

Tobias-Fischer:
Could this please be merged? I am happy to open a new PR if I should base this on top of a particular branch.
@kalyanvasudev @dirk-thomas

mikepurvis:
@Tobias-Fischer Yes, that would be great. Please submit a patch for `melodic-devel`.

Tobias-Fischer:
@mikepurvis I just realised that it has already been fixed in https://github.com/ros/ros_comm/pull/1660

So I think this PR should be closed.

As a side note, however, the bug is still contained in the latest ROS packages on Ubuntu 18.04. How long does it take before the packages pick up those bugfixes?

mikepurvis:
The last batch of releases for ros_comm was in August 2018, see: https://github.com/ros-gbp/ros_comm-release

So yes, we're overdue for a release. As far as testing, Clearpath's internal devel fleet was switched to d487d5877049d61fe2ebcc6487e9a7a72d9f0346 about a month ago, and our systems includes both Python 2 and Python 3 usage. We'll have a major internal release in a few weeks after which we can update again.

So depending on @dirk-thomas's availability, I'd suggest a ros_comm release around the end of November. If others want to run ros_comm from source and vouch for the current state of the `kinetic-devel` branch, that's all the more helpful.

dirk-thomas:
Resolved by ##1660.

PR #1775, opened 2019-07-23
rosbag record: Fix recording timestamps

davrsky:
Problem:
When we are recording large volumes of data, we have it set to split to a new bag file every 30GB. The problem we see is that the recorded timestamps at the beginning of the second (and every subsequent bagfile) appear incorrect, there are no messages recorded for a few seconds and then a bunch of timestamps all bunched up.

Solution:
Use the timestamp from the message event, instead of ros::time::now -- this should help fix the problem of getting late timestamps in theory.

Longer explanation, of my understanding of everything:

ros subscribers:
In the background, ros creates a thread (or multiple threads?) that accept tcp connections, recieve messages, and store into an internal buffer / queue, one queue per topic
(length of queue set by the queue_size(number of messages) & buffer_size(number of bytes) options on ros subscriber constructor).

Then, the received messages are dispatched to subscriber callbacks. this either happens single threaded (if you call spin or spinOnce), or multithreaded (if you use asyncspinner).

rosbag record:
In rosbag record, it's using asyncspinner w/ 10 threads, and it subscribes to many topics, all to the same callback doQueue, which when called, checks the current time, and adds the message + timestamp onto a queue.

There's a separate thread in rosbag record that handles reading from that queue and writing the messages sequentially to disk.

There's a lock that synchronizes between the threads, so that no two threads are writing to the queue at the same time & corrupting it. However this means while writing messages out to disk, new messages can't be added on to the queue, so those threads will pause at the line of code that acquires the lock. Since this is AFTER capturing the timestamp, you'd think this would be ok...but the problem is we only have 10 threads handling callbacks, so if writing messages out to disk takes longer than the time for 10 messages to come in, the messages will be stuck on the internal ros queue BEFORE calling our callback, meaning the ros::Time::now call wont be called until later, getting a later timestamp. We don't actually care about the time the message was moved from the internal ros queue onto the rosbag record queue, what we care about is when the message was actually received.

### DIFF
--- a/tools/rosbag/src/record.cpp
+++ b/tools/rosbag/src/record.cpp
@@ -89,6 +89,7 @@ rosbag::RecorderOptions parseOptions(int argc, char** argv) {
 
     if (vm.count("help")) {
       std::cout << desc << std::endl;
+      std::cout << ">> Starsky modified rosbag <<" << std::endl;
       exit(0);
     }
 
--- a/tools/rosbag/src/recorder.cpp
+++ b/tools/rosbag/src/recorder.cpp
@@ -155,16 +155,16 @@ int Recorder::run() {
     last_buffer_warn_ = Time();
     queue_ = new std::queue<OutgoingMessage>;
 
+    if (!ros::Time::waitForValid(ros::WallDuration(2.0)))
+      ROS_WARN("/use_sim_time set to true and no clock published.  Still waiting for valid time...");
+
+    ros::Time::waitForValid();
+
     // Subscribe to each topic
     if (!options_.regex) {
     	foreach(string const& topic, options_.topics)
 			subscribe(topic);
     }
-
-    if (!ros::Time::waitForValid(ros::WallDuration(2.0)))
-      ROS_WARN("/use_sim_time set to true and no clock published.  Still waiting for valid time...");
-
-    ros::Time::waitForValid();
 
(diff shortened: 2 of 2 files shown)


cwecht:
The thread writing the rosbag does not hold the lock during writing. It aquires the lock, pops the last item in the queue, releases the lock and then writes. I'd argue, that the actual problem is a too small queue size or something similar. When the `doRecord` thread is busy closing the current file and opening the next one (I have no idea how long this might take), it can not pop items from the queue. As the queue size is limited by ``buffer_size``, increasing it might help.

This change might still be valid and reasonable though, but I think, this might have uncovered a deeper lying problem, so it would be nice, if you could further investigate this issue.

peci1:
This definitely needs to be an optional behavior.

1. Not all messages have headers
1. Some people are really interested in using the current time when writing the bag to e.g. examine network or processing latencies. Our processed and map-aligned pointclouds are coming up to 6 secs delayed, but still hold the timestamp of the original LaserScan that generated them. Using the time from the header would not lead to representing the system behavior as it was when the system was running.

dirk-thomas:
Please retarget the latest default branch (currently `melodic-devel`), remove debug output from the patch, as well as respond to the previous comments.

dirk-thomas:

PR #1777, opened 2019-07-26
Encrypted rosbag fixes for Python 3.

mikepurvis:
Two small unicode-related fixes for getting the key out of the header, which enables it to work from Python 3.


dirk-thomas:
Thanks for the fix.

PR #1779, opened 2019-08-01
add Timer::isValid() const

### LINKED ISSUE #1650: ros::Timer::isValid() is not a const operation
I believe `ros::Timer::isValid()` could be made `const`. It seems to only rely on `ros::Duration::isZero()` which is a `const` operation. Or am I missing something here?

Source: 
https://github.com/ros/ros_comm/blob/melodic-devel/clients/roscpp/include/ros/timer.h#L75 
https://github.com/ros/ros_comm/blob/melodic-devel/clients/roscpp/src/libros/timer.cpp#L52

artivis:
fix #1650

Signed-off-by: artivis <jeremie.deray@canonical.com>


cwecht:

dirk-thomas:
Thanks for the patch.

PR #1781, opened 2019-08-05
fix lower bound of signed int

dirk-thomas:
Follow up of ros/genpy#102.

Needs a new patch release of `genpy`.

### DIFF
--- a/test/test_rospy/test/unit/test_genmsg_py.py
+++ b/test/test_rospy/test/unit/test_genmsg_py.py
@@ -264,7 +264,7 @@ class TestGenmsgPy(unittest.TestCase):
         widths = [(8, Int8), (16, Int16), (32, Int32), (64, Int64)]
         for w, cls in widths:
             maxp = long(math.pow(2, w-1)) - 1
-            maxn = -long(math.pow(2, w-1)) + 1
+            maxn = -long(math.pow(2, w-1))
             self._test_ser_deser(cls(maxp), cls())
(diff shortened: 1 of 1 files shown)


dirk-thomas:

test/test_rospy/test/unit/test_genmsg_py.py
```diff
@@ -264,7 +264,7 @@ def test_std_msgs_SignedInt(self):
         widths = [(8, Int8), (16, Int16), (32, Int32), (64, Int64)]
         for w, cls in widths:
             maxp = long(math.pow(2, w-1)) - 1
-            maxn = -long(math.pow(2, w-1)) + 1
+            maxn = -long(math.pow(2, w-1))
```
sloretz, line 267: It looks like `long` doesn't exist in Python 3.

```
python3 -c 'long(0.5)'
  File "<string>", line 1, in <module>
NameError: name 'long' is not defined
```

dirk-thomas: That is why the type if already being defined in that case at the top of the file: https://github.com/ros/ros_comm/blob/6b1ee1a43b5b0a970d85612435e269818c89a152/test/test_rospy/test/unit/test_genmsg_py.py#L48-L51

sloretz: Ah got it, that's already on `melodic-devel`. This branch must have been created before that was added.

PR #1782, opened 2019-08-05
more Python 3 compatibility

dirk-thomas:
Following http://wiki.ros.org/python_2_and_3_compatible_code

### DIFF
--- a/tools/rosgraph/test/test_roslogging.py
+++ b/tools/rosgraph/test/test_roslogging.py
@@ -32,7 +32,10 @@
 
 import logging
 import os
-from StringIO import StringIO
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from io import StringIO
 import sys
 
 import re
--- a/tools/rosgraph/test/test_roslogging_user_logger.py
+++ b/tools/rosgraph/test/test_roslogging_user_logger.py
@@ -38,7 +38,10 @@ in rosgraph.roslogger.
 
 import logging
 import os
-from StringIO import StringIO
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from io import StringIO
 import sys
 
 from nose.tools import assert_regexp_matches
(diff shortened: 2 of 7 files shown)


tools/rosgraph/test/test_roslogging.py
```diff
@@ -32,7 +32,10 @@
 
 import logging
 import os
-from StringIO import StringIO
+try:
+    from cStringIO import StringIO
```
sloretz, line 36: It looks like the import `from io import StringIO` works in python 2.7. The API's for `StringIO` in both look the same, but I didn't try this locally.

* `io` [2.7 Documentation](https://docs.python.org/2.7/library/io.html#io.StringIO)
* `cStringIO` [2.7 documentation](https://docs.python.org/2/library/stringio.html#cStringIO.StringIO)

tools/rosgraph/test/test_roslogging_user_logger.py
```diff
@@ -38,7 +38,10 @@
 
 import logging
 import os
-from StringIO import StringIO
+try:
+    from cStringIO import StringIO
+except ImportError:
+    from io import StringIO
```
sloretz, line 44: Same comment about this import being in 2.7 and 3.x

tools/rosgraph/test/test_roslogging.py
sloretz: Nevermind, I failed to read this: http://wiki.ros.org/python_2_and_3_compatible_code#StringIO

tools/rosgraph/test/test_roslogging_user_logger.py
sloretz: nevermind, http://wiki.ros.org/python_2_and_3_compatible_code#StringIO

sloretz (approved):

tools/rosgraph/test/test_roslogging.py
dirk-thomas: Please see the comment on the ROS wiki page why this approach is used.

dirk-thomas:

PR #1783, opened 2019-08-06
more Python 3 compatibility

dirk-thomas:
Follow up of #1782.


test/test_rospy/test/unit/test_genmsg_py.py
```diff
@@ -95,8 +95,8 @@ def test_test_rospy_TestFixedArray(self):
         self.assertEquals([0., 0., 0.], m.f64_3)        
         self.assertEquals([0], m.i8_1)
         self.assertEquals([0, 0, 0], m.i8_3)        
-        self.assertEquals(chr(0), m.u8_1)
-        self.assertEquals(chr(0)*3, m.u8_3)        
+        self.assertEquals(chr(0).encode(), m.u8_1)
```
sloretz, line 98: It looks like this is a type change between python 2 and python 3. The new type `bytes` does make more sense to me for [`uint8`](https://github.com/ros/ros_comm/blob/8f22c20e418abe4abe23e789eef517a16a50604d/test/test_rospy/msg/TestFixedArray.msg#L7), just checking was that intentional?

```
sloretz@nxt ~> python -c 'print(type(chr(0).encode()))'
<type 'str'>
sloretz@nxt ~> python3 -c 'print(type(chr(0).encode()))'
<class 'bytes'>
```

tools/rosmsg/test/test_rosmsg_command_line.py
```diff
@@ -54,24 +54,24 @@ def test_cmd_help(self):
         
         for cmd in ['rosmsg', 'rossrv']:
             glob_cmd=[os.path.join(_SCRIPT_FOLDER, cmd)]
-            output = Popen(glob_cmd, stdout=PIPE, env=self.new_environ).communicate()[0]
+            output = Popen(glob_cmd, stdout=PIPE, env=self.new_environ).communicate()[0].decode()
```
sloretz, line 57: Just noting an alternative to decoding is to use "text mode" which makes `communicate()` return two strings. It can be done with the keyword arg `Popen(..., text=true)` in python 3.7+

sloretz (approved):

test/test_rospy/test/unit/test_genmsg_py.py
dirk-thomas: Yes, that is the type the Python code generator expands `uint8` arrays to. If possible I only want to modify tests to match the already existing in Python 3 (and not change that behavior).

tools/rosmsg/test/test_rosmsg_command_line.py
dirk-thomas: A Python 3.7 only feature couldn't land on the `melodic-devel` branch which targets Ubuntu Bionic.

sloretz: Roger. The [docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) say `text=True` is an alias for `universal_newlines=True`. `text` was added in 3.7, but it looks like `universal_newlines` has been around from the start.

PR #1787, opened 2019-08-08
fix Coverity forward null

### LINKED ISSUE #1746: Coverity - forward null
Event null_check: Comparing "f" to a null-like value implies that "f" might be null-like.
https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/tools/rosgraph/src/rosgraph/roslogging.py#L61-L62

Event property_access: | Accessing a property of null-like value "f".
https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/tools/rosgraph/src/rosgraph/roslogging.py#L76

There is a path for a null-valued variable to be dereferenced. But "f" might still be None

danielwangksu:
fix  #1746
Signed-off-by: Daniel Wang <daniel.wang@canonical.com>

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -66,7 +66,8 @@ class RospyLogger(logging.getLoggerClass()):
             filename = os.path.normcase(co.co_filename)
             if filename == file_name and f.f_lineno == lineno and co.co_name == func_name:
                 break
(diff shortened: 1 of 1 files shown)


cwecht:
This does not fix the issue with line 76 right?

dirk-thomas:
Thanks for the patch.

Issue #953, opened 2016-12-22
roslaunch_add_file_check doesn't support args


damonkohler:
We have some top-level launch files that expect args to be passed in on the command line. As far as I can tell, roslaunch_add_file_check only supports setting environment variables. So, the test fails when we have an undefined arg in a top-level launch file.

dirk-thomas:
Can you please clarify how it fails in your case. Maybe it is similar to other pending tickets.

damonkohler:
I'm not sure exactly what further clarification you're looking for. I've pasted a failing log snippet below though in case that helps.

```
cd /usr/local/google/home/damonkohler/catkin_ws/build_isolated/cartographer_ros && /usr/local/google/home/damonkohler/catkin_ws/build.../cartographer_ros/roslaunch-check_launch.xml'\ '/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch'\
FAILURE:
[/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch/offline_backpack_2d.launch]:
        [/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch/offline_backpack_2d.launch] requires the 'bag_filenames' arg to be set
[/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch/demo_backpack_3d.launch]:
        [/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch/demo_backpack_3d.launch] requires the 'bag_filename' arg to be set
[/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch/demo_backpack_2d.launch]:
        [/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch/demo_backpack_2d.launch] requires the 'bag_filename' arg to be set
[/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch/demo_pr2.launch]:
        [/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch/demo_pr2.launch] requires the 'bag_filename' arg to be set
[/usr/local/google/home/damonkohler/catkin_ws/src/cartographer_ros/cartographer_ros/launch/demo_taurob_tracker.launch]:
[...]

PR #1788, opened 2019-08-09
[roslaunch] Add ignore default args option to roslaunch-check

collinscrib:
Addresses https://github.com/ros/ros_comm/issues/953

Adds an option to roslaunch-check to ignore the error thrown when a `<arg>` is declared but doesn't have a default variable.

### Usage

From the roslaunch-check script: `rosrun roslaunch roslaunch-check [-i / --ignore-default-args] foo.launch`
From a CMakeLists.txt file: `roslaunch_add_file_check(launch/foo.launch IGNORE_DEFAULT_ARGS)`

### Example
The example given is as follows:
```
<launch>
  <arg name="foo" />    <!-- no `default="something" here! --> 
  <node name="bar" pkg="bar" type="bar">
    <param name="foo" value="$(arg foo)" />
  </node>
</launch>
```

which, when run with roslaunch-check, will output 
```
$ rosrun roslaunch roslaunch-check foo.launch
FAILURE:
[foo.launch]:
	[foo.launch] requires the 'foo' arg to be set
```

The change adds an option to ignore this error with `-i` or `--ignore-default-args` like so
```
$ rosrun roslaunch roslaunch-check --ignore-default-args foo.launch
passed
```


130s:
Just FYI for reviewers, we've been running tests using this PR (on ROS Kinetic, not Melodic) at Plus One Robotics and we'll keep doing so. We're also working on setting up any publicly visible test results (which I hope doesn't block the review).

cwecht:
The name of the new flag is a bit misleading, right? It does not ignore default arguments but ignores not set arguments.

collinscrib:
> The name of the new flag is a bit misleading

Maybe something like `--ignore-unset-args` is more appropriate?

cwecht:
> Maybe something like `--ignore-unset-args` is more appropriate?

Yes, this is better!

tools/roslaunch/test/unit/test_roslaunch_rlutil.py
```diff
@@ -89,3 +89,13 @@ def test_roslaunch_check_pass_all_args(self):
...
+        
+    def test_check_roslaunch(self):
+        from roslaunch.rlutil import check_roslaunch
+
+        test_ignore_unset_args_p = os.path.join(get_test_path(), 'test', 'xml', 'test-ignore-unset-args.launch')
+        try:
+            result = check_roslaunch(test_ignore_unset_args_p, ignore_unset_args=True)
+            self.assertEqual(result, None)
+        except roslaunch.RLException:
+            pass
```
130s: Don't we want the test to fail when an exception is caught here?

collinscrib: You're probably right. 

Not sure why I put that in there in the first place but it might've been to ignore specific read/write errors from check_roslaunch(). In any case, passing over any instance of RLException is probably a bad idea.

dirk-thomas:
Thanks for the patch.

PR #1789, opened 2019-08-09
[kinetic] Added rosbag::Bag::isOpen

### LINKED ISSUE #1448: rosbag_storage - Backport isOpen() to kinetic
@racko developed the convenient function for lunar and above, but this change was not backported to kinetic. Since I still have projects which require kinetic, I am stuck being unable to obtain the status of the bag file (since the bag file is private, and unable to be queried), and I have to track the bag file's status manually.

artivis:
Backport #1190 to kinetic-devel.

Fix #1448.

### DIFF
--- a/tools/rosbag_storage/include/rosbag/bag.h
+++ b/tools/rosbag_storage/include/rosbag/bag.h
@@ -176,6 +176,8 @@ public:
 
     void swap(Bag&);
 
+    bool isOpen() const;
+
 private:
     Bag(const Bag&);
     Bag& operator=(const Bag&);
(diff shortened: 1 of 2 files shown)


artivis:
Just found out that #1190 was marked as 'not backported' in https://github.com/ros/ros_comm/pull/1205. 
While it is a 'new feature' it is also useful and ridiculously small enough not to be considered a maintenance burden. Up to the maintainers.

dirk-thomas:
Accepting the backport since it has worked fine in Melodic for a long time and is API/ABI compatible.

PR #1791, opened 2019-08-13
Fix dead loop if accept connection error in XmlRpcServer

### LINKED ISSUE #1762: Dead loop if accept connection error in XmlRpcServer
If there is an error when accepting connections, `_accept_error` will be set to [true](https://github.com/ros/ros_comm/blob/melodic-devel/utilities/xmlrpcpp/src/XmlRpcServer.cpp#L196), and `XmlRpcServer` will be [add to XmlRpcDispatch](https://github.com/ros/ros_comm/blob/melodic-devel/utilities/xmlrpcpp/src/XmlRpcServer.cpp#L166). but when accept connection is OK  `_accept_error` is not set to true, and `XmlRpcServer` will still be add to `XmlRpcDispatch` in each work loop.

randoms:
Fix #1762

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcServer.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcServer.cpp
@@ -209,6 +209,7 @@ XmlRpcServer::acceptConnection()
   }
   else  // Notify the dispatcher to listen for input on this source when we are in work()
   {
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for the patch.

PR #1792, opened 2019-08-13
use condition attributes to specify Python 2 and 3 dependencies

dirk-thomas:
Similar to ros/catkin#1025.

### DIFF
--- a/test/test_rosbag/package.xml
+++ b/test/test_rosbag/package.xml
@@ -1,4 +1,8 @@
-<package format="2">
+<?xml version="1.0"?>
+<?xml-model
+  href="http://download.ros.org/schema/package_format3.xsd"
+  schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>test_rosbag</name>
   <version>1.14.3</version>
   <description>
@@ -7,10 +11,15 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>BSD</license>
 
-  <url>http://ros.org/wiki/rosbag</url>
+  <url type="website">http://wiki.ros.org/rosbag</url>
+  <url type="bugtracker">https://github.com/ros/ros_comm/issues</url>
+  <url type="repository">https://github.com/ros/ros_comm</url>
   <author>Tim Field</author>
   <author>Jeremy Leibs</author>
   <author>James Bowman</author>
+
+  <depend>rosbag</depend>
+  <depend>rostest</depend>
 
   <buildtool_depend version_gte="0.5.68">catkin</buildtool_depend>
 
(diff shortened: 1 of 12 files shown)


test/test_rosbag/package.xml
```diff
@@ -7,36 +11,39 @@
...
   <buildtool_depend version_gte="0.5.68">catkin</buildtool_depend>
 
   <build_depend>boost</build_depend>
   <build_depend>bzip2</build_depend>
   <build_depend>cpp_common</build_depend>
   <build_depend>message_generation</build_depend>
-  <build_depend>python-imaging</build_depend>
-  <build_depend>rosbag</build_depend>
+  <build_depend condition="$ROS_PYTHON_VERSION == 2">python-imaging</build_depend>
+  <build_depend condition="$ROS_PYTHON_VERSION == 3">python3-imaging</build_depend>
```
sloretz: I think the equivalent python3 key is [`python3-pil`](https://github.com/ros/rosdistro/blob/0ea30fed0bcb564b9e4d5b2d2efd8783e8e3afa9/rosdep/python.yaml#L5107). There's no `python3-imaging` key (nor is there a debian package with that name).

tools/rosbag/package.xml
```diff
@@ -9,38 +13,39 @@
...
   <buildtool_depend version_gte="0.5.78">catkin</buildtool_depend>
 
-  <build_depend>boost</build_depend>
   <build_depend>cpp_common</build_depend>
-  <build_depend>python-imaging</build_depend>
-  <build_depend>rosbag_storage</build_depend>
-  <build_depend>rosconsole</build_depend>
-  <build_depend>roscpp</build_depend>
+  <build_depend condition="$ROS_PYTHON_VERSION == 2">python-imaging</build_depend>
+  <build_depend condition="$ROS_PYTHON_VERSION == 3">python3-imaging</build_depend>
```
sloretz: `python3-pil` here too

tools/rosbag/package.xml
```diff
@@ -9,38 +13,39 @@
...
-  <run_depend>roscpp</run_depend>
-  <run_depend>roslib</run_depend>
-  <run_depend>rospy</run_depend>
-  <run_depend>std_srvs</run_depend>
-  <run_depend>topic_tools</run_depend>
-  <run_depend>xmlrpcpp</run_depend>
+  <exec_depend>genmsg</exec_depend>
+  <exec_depend>genpy</exec_depend>
+  <exec_depend condition="$ROS_PYTHON_VERSION == 2">python-crypto</exec_depend>
+  <exec_depend condition="$ROS_PYTHON_VERSION == 3">python3-crypto</exec_depend>
```
sloretz: There's no `python3-crypto` key, and the project is dead. [`python3-cryptography`](https://github.com/ros/rosdistro/blob/0ea30fed0bcb564b9e4d5b2d2efd8783e8e3afa9/rosdep/python.yaml#L4925-L4929) is a possible replacement #1599 https://github.com/ros/ros_comm/pull/1606

sloretz (changes requested):

test/test_rosbag/package.xml
dirk-thomas: Good catch: 85394ffe89f981469bbd0954d1483382aff20df8.

tools/rosbag/package.xml
dirk-thomas: Good catch: 85394ffe89f981469bbd0954d1483382aff20df8.

tools/rosbag/package.xml
dirk-thomas: Alternative PRs: #1609 or #1793. Currently I am leaning towards the former one due to the much simpler diff.

dirk-thomas: With #1609 merged I rebased this patch and updated the dependency to `python(3)-pycryptodome`.

sloretz (approved):

PR #1793, opened 2019-08-14
use python-cryptography instead of python-crypto

dirk-thomas:
Redo of #1606 since the fork has been delete and the patch needed some fixes.

The passing CI for the original PR which had bugs which would have failed at runtime shows that this code isn't covered by any unit tests. So this needs manual testing to ensure it doesn't break anything.

A few pending questions inline.

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -50,8 +50,11 @@ import threading
 import time
 import yaml
 
-from Crypto import Random
-from Crypto.Cipher import AES
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.ciphers import Cipher
+from cryptography.hazmat.primitives.ciphers.algorithms import AES
+from cryptography.hazmat.primitives.ciphers.modes import CBC
+
 import gnupg
 
 try:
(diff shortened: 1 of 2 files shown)


tools/rosbag/src/rosbag/bag.py
```diff
@@ -75,6 +78,9 @@
         'LZ4 compression will not be available.')
     found_lz4 = False
 
+AES_BLOCK_SIZE = 16
```
dirk-thomas, line 81: Why use this constant over `AES.block_size` (which is `128`)?

tools/rosbag/src/rosbag/bag.py
```diff
@@ -349,7 +360,7 @@ def get_info_str(self):
     def _build_symmetric_key(self):
         if not self._gpg_key_user:
             return
-        self._symmetric_key = Random.new().read(AES.block_size)
+        self._symmetric_key = os.urandom(AES.block_size)
```
dirk-thomas, line 363: Why `AES.block_size` here instead of `AES_BLOCK_SIZE`?

dirk-thomas:
This is currently blocking the Python 3 migration (see https://github.com/ros/ros_comm/pull/1792#discussion_r313638653).

dirk-thomas:
@mikepurvis Since you commented on the original ticket do you have a use case for this? If yes, could you manually test this change to ensure in doesn't break compatibility / behavior?

dirk-thomas:
As an alternative I am considering #1609 since it is much simpler and therefore has a lower chance of regressions.

dirk-thomas:
For now we went with `pycryptodome` in #1609. I will go ahead and close this.

PR #1794, opened 2019-08-15
fix line endings to be LF

dirk-thomas:
Change file with CRLF back to LF.

PR #1795, opened 2019-08-15
more Python 3 compatibility

dirk-thomas:
Follow up of #1785.

~~The first commit is in a separate PR #1794 and will be removed from this one as soon as it is merged. Instead of looking at the changes to `clients/rospy/src/rospy/core.py` (which are fixing the line endings in the file) it is best to look at the actual change in f1680fd7c3bb38df6b42cf6b81544ce7f8abeca2.~~ The other PR has been merged and this branch been rebased.

### DIFF
--- a/test/test_rosmaster/test/client_verification/test_slave_api.py
+++ b/test/test_rosmaster/test/client_verification/test_slave_api.py
@@ -124,7 +124,7 @@ class TestSlaveApi(unittest.TestCase):
                 time.sleep(0.1)
         if not self.node_api:
             self.fail("master did not return XML-RPC API for [%s, %s]"%(self.caller_id, self.test_node))
-        print "[%s] API  = %s"%(self.test_node, self.node_api)
+        print("[%s] API  = %s"%(self.test_node, self.node_api))
         self.assert_(self.node_api.startswith('http'))
         self.node = ServerProxy(self.node_api)
 
--- a/test/test_rosmaster/test/client_verification/test_slave_api.py
+++ b/test/test_rosmaster/test/client_verification/test_slave_api.py
@@ -173,7 +173,10 @@ class TestSlaveApi(unittest.TestCase):
         """
         validates a URI as being http(s)
         """
-        import urlparse
+        try:
+            import urllib.parse as urlparse
+        except ImportError:
+            import urlparse
         parsed = urlparse.urlparse(uri)
         self.assert_(parsed[0] in ['http', 'https'], 'protocol [%s] is [%s] invalid'%(parsed[0], uri))
         self.assert_(parsed[1], 'host missing [%s]'%uri)
@@ -323,7 +326,10 @@ class TestSlaveApi(unittest.TestCase):
         self.check_uri(uri)
 
         # check against env, canonicalize for comparison
-        import urlparse
+        try:
+            import urllib.parse as urlparse
+        except ImportError:
+            import urlparse
         master_env = rosgraph.get_master_uri()
         if not master_env.endswith('/'):
             master_env = master_env + '/'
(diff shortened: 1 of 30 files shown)


sloretz (dismissed):
f1680fd LGTM

test/test_rosmaster/test/client_verification/test_slave_api.py
```diff
@@ -124,7 +124,7 @@ def setUp(self):
                 time.sleep(0.1)
         if not self.node_api:
             self.fail("master did not return XML-RPC API for [%s, %s]"%(self.caller_id, self.test_node))
-        print "[%s] API  = %s"%(self.test_node, self.node_api)
+        print("[%s] API  = %s"%(self.test_node, self.node_api))
```
sloretz, line 127: Just noting usually `from __future__ import print_function` would be used, but it looks like there's no difference in output in this case.

tools/rostest/nodes/advertisetest
```diff
@@ -135,19 +133,19 @@ class AdvertiseTest(unittest.TestCase):
...
                         t_name_set.remove(t_name)
                         topic = self.topics[t_name]
-                        assert_true(t_elapsed < topic['timeout'],
-                                    'Topic [%s] is advertised before timeout [%s] secs' % (topic['name'], topic['timeout']))
-                        assert_false(topic['negative'],
-                                     'Topic [%s] is not advertised' % (topic['name']))
+                        assert t_elapsed < topic['timeout'], \
+                            'Topic [%s] is advertised before timeout [%s] secs' % (topic['name'], topic['timeout'])
+                        assert not topic['negative'], \
+                            'Topic [%s] is not advertised' % (topic['name'])
```
sloretz, line 139: If this is asserting `not topic['negative']`, would an assertion failure mean the topic was unexpectedly advertised?

tools/rostest/nodes/advertisetest
```diff
@@ -168,20 +166,20 @@ class AdvertiseTest(unittest.TestCase):
...
                         s_name_set.remove(s_name)
                         service = self.services[s_name]
-                        assert_true(t_elapsed < service['timeout'],
-                                    'Service [%s] is advertised before timeout [%s] secs' % (service['name'], service['timeout']))
-                        assert_false(service['negative'],
-                                     'Service [%s] is not advertised' % (service['name']))
+                        assert t_elapsed < service['timeout'], \
+                            'Service [%s] is advertised before timeout [%s] secs' % (service['name'], service['timeout'])
+                        assert not service['negative'], \
+                            'Service [%s] is not advertised' % (service['name'])
```
sloretz, line 172: Is the service not advertised, or unexpectedly advertised?

sloretz (approved):
LGTM, just a couple questions about assertion messages that were already there before this PR

test/test_rosmaster/test/client_verification/test_slave_api.py
dirk-thomas: As long as the code only prints a single string argument you don't need the future import.

When the code passes multiple string arguments you might want the future import in order to not change the behavior change what separator is used between the arguments.

When the code uses `file=sys.stderr` you certainly need the future import to make it work in Python 2.

See http://wiki.ros.org/python_2_and_3_compatible_code#Print_is_now_a_function which describes those cases.

tools/rostest/nodes/advertisetest
dirk-thomas: Correct.

tools/rostest/nodes/advertisetest
dirk-thomas: I think this check ensure that such a a service is not advertised. Similar to the topic test above.

sloretz: > I think this check ensure that such a a service is not advertised.

It looks like if the test fails the output will be `AssertionError: Service [foo] is not advertised`. If that's what it's checking I think it should say the opposite.

PR #1796, opened 2019-08-16
more Python 3 compatibility

dirk-thomas:
Follow up of #1795.

### DIFF
--- a/test/test_rospy/nodes/listener.py
+++ b/test/test_rospy/nodes/listener.py
@@ -55,7 +55,7 @@ def listener():
     # run simultaenously.
     rospy.init_node('listener', anonymous=True)
 
-    rospy.Subscriber("chatter", String, callback)
+    sub = rospy.Subscriber("chatter", String, callback)
 
     # spin() simply keeps python from exiting until this node is stopped
     rospy.spin()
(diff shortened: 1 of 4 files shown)


test/test_rospy/nodes/listener.py
```diff
@@ -55,7 +55,7 @@ def listener():
     # run simultaenously.
     rospy.init_node('listener', anonymous=True)
 
-    rospy.Subscriber("chatter", String, callback)
+    sub = rospy.Subscriber("chatter", String, callback)
```
sloretz, line 58: was the subscriber getting garbage collected before?

sloretz (approved):

test/test_rospy/nodes/listener.py
dirk-thomas: Yes, that is my assumption.

test/test_rospy/test/unit/test_rospy_msg.py
```diff
@@ -224,7 +225,6 @@ def validate_vals(vals, teststrs=teststrs):
         #deserialize multiple values with max_msgs
         max_msgs = 5
         b.truncate(0)
-        b.seek(0)
```
dirk-thomas, line 227: This one is not necessary since line 221 already ensures that the read position is zero.

sloretz (approved):

PR #1807, opened 2019-09-18
trim stdout lines to less than 80 characters

christian-rauch:
The stdout output from `rosbag play` (.e.g. ` [RUNNING]  Bag Time: 1555604159.419860   Duration: 2.043711 / 81.562797`) is typically longer than 80 characters since it adds unnecessary whitespace at the end. For standard sized terminals widths (80 characters), this leads to a new line for every printout and thus spams the terminal.

This PR simply removes the padded whitespace and should fix the issue (unless you have log files that are longer than 99999 seconds).

### DIFF
--- a/tools/rosbag/src/player.cpp
+++ b/tools/rosbag/src/player.cpp
@@ -40,6 +40,8 @@
   #include <sys/select.h>
 #endif
 
+#include <sys/ioctl.h>
+
 #include <boost/format.hpp>
 
 #include "rosgraph_msgs/Clock.h"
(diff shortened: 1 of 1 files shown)


tools/rosbag/src/player.cpp
```diff
@@ -385,16 +385,16 @@ void Player::printTime()
 
         if (paused_)
         {
-            printf("\r [PAUSED ]  Bag Time: %13.6f   Duration: %.6f / %.6f               \r", time_publisher_.getTime().toSec(), d.toSec(), bag_length_.toSec());
+            printf("\r [PAUSED ]  Bag Time: %13.6f   Duration: %.6f / %.6f", time_publisher_.getTime().toSec(), d.toSec(), bag_length_.toSec());
```
dirk-thomas: The diff removes the trailing `\r` which it shouldn't. Same below.

dirk-thomas (commented):
Some trailing spaces are necessary to ensure that a previous line which might have been longer is being cleared and doesn't leave the last character around. I would assume the current patch breaks that behavior.

tools/rosbag/src/player.cpp
christian-rauch: I removed the second carriage return (`\r`) in order for the active carriage to be placed after the text and not at the beginning.

christian-rauch:
> Some trailing spaces are necessary to ensure that a previous line which might have been longer is being cleared and doesn't leave the last character around. I would assume the current patch breaks that behavior.

Isn't this the case with the current state too? At the moment, a constant number of spaces are appended to the end of the text. If the previous text is longer than the new text + spaces, you will have the same issue.

I think the core problem is, that the amount of spaces added is fixed and arbitrary chosen. If you properly want to clear that line, you would need to determine the width of the terminal. But this might involve adding a dependency on `ncurses` or similar.

dirk-thomas:
> Isn't this the case with the current state too?

No, it is not because the fixed amount of trailing spaces is long enough to cover the variability of the non-empty message length.

> If you properly want to clear that line, you would need to determine the width of the terminal.

No, you don't need to know the terminal width. The only information you need is the previous message length and if the new message length is shorter then pad it with delta spaces.

christian-rauch:
I am not sure what you mean with
>long enough to cover the variability of the non-empty message length

since the previous message that was printed on the terminal can have an arbitrary length.
If you really just want to clear the previous message, you either (A) need to determine its length or (B) enforce a maximum length of messages, e.g. 80, and clear this amount of characters before writing again.

In any case, the current implementation writes more than 80 characters, which is the default terminal width on most systems. It would be useful to restrict the number of characters written per line to the terminal width.

What do you suggest to prevent writing behind the width of the terminal?

christian-rauch:
@dirk-thomas What do you have in mind to prevent writing more than 80 characters?

dirk-thomas:
> > long enough to cover the variability of the non-empty message length

> since the previous message that was printed on the terminal can have an arbitrary length.

The printed lines in question can't have an arbitrary length. There are no unbounded strings in the formatted string. Number are formatted with a specific number of decimals. So you kind of know the upper boundary.

> What do you suggest to prevent writing behind the width of the terminal?

It would be possible to reduce the line length, e.g. by abbreviating some parts (I guess the numbers shouldn't be shortened).

christian-rauch:
Well, I am now clearing the previous line explicitly by printing empty characters over the width of the terminal. I still see some glitches for the very first printed line, e.g. it says `[DELAYED]` instead of `[RUNNING]` but it does not flood the terminal anymore.

dirk-thomas:
As mentioned above I would rather not introduce logic which detects the width of the terminal. Simply because it is not strictly required.

The current patch is probably not working on e.g. Windows.

christian-rauch:
> As mentioned above I would rather not introduce logic which detects the width of the terminal. Simply because it is not strictly required.

I think we either need logic to determine the previously written line length and the newly to-be-written line length, or logic for determining the width of the terminal to prevent writing beyond the terminal.

I find the latter option less complex. The former option requires to write into buffers first and compare buffer sizes.

> The current patch is probably not working on e.g. Windows.

True. I did not have the Windows case in mind. There is probably a way to achieve the same on Windows. Would you accept the "terminal-width" solution if I provide the ifdef for Windows?

dirk-thomas:
> I find the latter option less complex. The former option requires to write into buffers first and compare buffer sizes.

I come to the opposite conclusion. I think remembering the previous line length is much easier to achieve and requires much less extra logic (which would need to be tested and verified in many different scenarios, Linux, Windows, interactive terminal, non-interactive, ...).

dirk-thomas:

PR #1809, opened 2019-09-19
add platform check for roslaunch --args code path

kejxu:
changes in this pull request:

* since `roslaunch --args` prints out commands that contain environment configuration, use `cmd /c` to execute the command in a separate cmd windows for isolated environment configuration
* use `set someenv=somevalue&& command` for Windows cmd

PR #1810, opened 2019-09-20
wrap env prefix with double quotes

kejxu:
adding double quotes will support environment variables that contain space; meantime, it does not affect environment variables that does not contain space.

test case (not ROS-specific):
this will fail, `value: command not found`
```bash
someenv=somevalue someenv2=some value python -c "import os;print os.environ['someenv'];print os.environ['someenv2']"
```

this would work:
```bash
someenv="somevalue" someenv2="some value" python -c "import os;print os.environ['someenv'];print os.environ['someenv2']"
```


dirk-thomas:
Thanks for the patch.

dirk-thomas:
@kejxu With #1809 merged this needs to be rebased to resolve the conflicts.

kejxu:
> @kejxu With #1809 merged this needs to be rebased to resolve the conflicts.

merged `master` into feature branch (instead of rebasing) for quick resolution =)

tools/roslaunch/src/roslaunch/node_args.py
```diff
@@ -198,9 +198,9 @@ def get_node_args(node_name, roslaunch_files):
         return ["set %s=%s&&"%(k, v) for k, v in env.items()] + args
     else:
         # sys.platform.startswith('linux')
-        # join environment vars are bash prefix args
-        return ["%s=%s"%(k, v) for k, v in env.items()] + args
-    
+        # join environment vars are bash prefix args, wrap with double quotes for variables that contains space
+        return [r'%s="%s"'%(k, v) for k, v in env.items()] + args
```
dirk-thomas: Why is the r-prefix used here?

kejxu: remove unnecessary r-prefix

dirk-thomas:
Thanks for the patch.

PR #1813, opened 2019-09-24
Wrap file names in quotes so they can be click selected in terminal

lucasw:
It's convenient to double click on the printed rosbag name to copy it and use as a parameter elsewhere, but the period will get included in the click (at least in Ubuntu 18.04 + gnome terminal).
Alternatively could eliminate the periods in all but one of the cases above.


dirk-thomas:
The extra space before the period is pretty awkward / non-standard. How about wrapping the path in single quotes instead?

lucasw:
Changed to quotes.

dirk-thomas:
Thanks for the patch.

PR #1816, opened 2019-09-26
[Windows][melodic-devel] workaround WSAPoll doesn't report failed connections.

### LINKED ISSUE #151: ros::master::check() hanging
**Describe the bug**
I tried to use ros::master::check() to see if roscore is ready. But it kept hanging in this function .

**To Reproduce**
I simply ran the program quoted in https://answers.ros.org/question/236678/check-master-ros-node-is-started-c/. If I didn't start roscore at the beginning and ran it, the program never output the "waiting..." line or anything. The program got hanging forever. If I started roscore before, it would output "master started!" as expected. 

**Environment information:**
Standard installation and catkin_make followed ROSonWindows page.

seanyen:
This fixes https://github.com/ms-iot/ROSOnWindows/issues/151

The original problem manifests as `ros::master::check()` never return back when the `master` is not running. It turns out that it is waiting for [`WSAPoll`](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsapoll) to be returned, which is at this below line:

https://github.com/ros/ros_comm/blob/d487d5877049d61fe2ebcc6487e9a7a72d9f0346/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp#L125

However, for a non-blocking socket which has a failed connection, `WSAPoll` exhibits a behavior which is not to signal this error for the caller, and with the infinite timeout, it just waits forever. There are many [discussions](https://www.bing.com/search?q=%22Windows+8+Bugs+309411+%E2%80%93+WSAPoll+does+not+report+failed+connections%22) about this behavior.

To workaround this issue, I am proposing for Windows platform adding a connection readiness pre-checks for writable sockets, which can be done by calling [`select`](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-select) and it will come back with a result immediately whether it is an `connection-error` socket or `writable-ready` socket. After we have the result, we can decide to error out or go ahead with the original `WSAPoll` to finish its job.

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -14,6 +14,7 @@
 static int poll( struct pollfd *pfd, int nfds, int timeout)
 {
   // workaround: "Windows 8 Bugs 309411 – WSAPoll does not report failed connections"
+  // https://curl.haxx.se/mail/lib-2012-10/0038.html
   // the following logic is to use select() to check all writable socket connnection status.
   // if all the sockets to be checked are not connected, it reports SOCKET_ERROR to
   // error out, instead of going to WSAPoll() which causes infinitely wait situation.
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -11,9 +11,56 @@
 
 #if defined(_WINDOWS)
 # include <winsock2.h>
-static inline int poll( struct pollfd *pfd, int nfds, int timeout)
+static int poll( struct pollfd *pfd, int nfds, int timeout)
 {
-  return WSAPoll(pfd, nfds, timeout);
+  // workaround: "Windows 8 Bugs 309411 – WSAPoll does not report failed connections"
+  // the following logic is to use select() to check all writable socket connnection status.
+  // if all the sockets to be checked are not connected, it reports SOCKET_ERROR to
+  // error out, instead of going to WSAPoll() which causes infinitely wait situation.
+  FD_SET writable;
+  FD_SET error;
+  FD_ZERO(&writable);
+  FD_ZERO(&error);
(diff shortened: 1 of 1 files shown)


seanyen:
@mikepurvis This is ready for review and merge. Thanks!

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
```diff
@@ -11,9 +11,57 @@
...
+    // error out if all sockets are failed to connect.
+    return SOCKET_ERROR;
+  }
+  else
+  {
+    return WSAPoll(pfd, nfds, timeout);
+  }
 }
 
 # define USE_FTIME
```
mikepurvis, line 67: I know it's not directly related here, but in the process of reviewing this, I noticed some if/endif mismatching going on. The `#endif` at the bottom of the file is labeled for `USE_FTIME` but I believe actually corresponds to `WINDOWS`.

@seanyen Do you mind looking into this and correcting the comments if applicable?

seanyen: By checking the [commit history](https://github.com/ros/ros_comm/commit/79bfbe8f05fcf73840c88b14b6f80ee48c5a78b9), it seems to me that `USE_FTIME` was first introduced because `Windows` doesn't have `gettimeofday` so it ended up with diverged implementations. However, two years ago, a [PR](https://github.com/ros/ros_comm/pull/1249) was merged and now non-Windows implementation is in favor of `ros::ros_steadytime`.

I would say perhaps we should see if there is any opportunity to converge the implementations now, since `ros::ros_steadytime` is a portable thing and theoretically it should behave the same across the platforms.

@mikepurvis, what do you think?

dirk-thomas: Since this is orthogonal please follow up on this separately form this PR.

dirk-thomas:
Thanks for the patch.

PR #1820, opened 2019-10-01
[windows][melodic] use c++11 std::snprintf

seanyen:
This pull request is in favor of `C++11` `std::snprintf` usage and to remove MSVC specific CRT functions [`_snprintf_s`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-s-snprintf-s-l-snwprintf-s-snwprintf-s-l?view=vs-2019).

This change is motivated by a name conflicts against `Boost`. Beginning from `Boost v1.69`, a new addition [`system_category_win32.hpp`](https://github.com/boostorg/system/blob/develop/include/boost/system/detail/system_category_win32.hpp) is added to `Boost.System`, and in this header, [it](https://github.com/boostorg/system/blob/3a4fff686e81320d346304b978f866c4136bd863/include/boost/system/detail/system_category_win32.hpp#L52) refers to `std::snprintf` in a code path. And because `snprintf` has been redefined as `_snprintf_s` in many places in `ros_comm`, it subsequently replaces `std::snprintf` with `std::_snprintf_s` and results in compiler errors.

### DIFF
--- a/clients/roscpp/src/libros/file_log.cpp
+++ b/clients/roscpp/src/libros/file_log.cpp
@@ -34,14 +34,6 @@
 
 #include <boost/filesystem.hpp>
 
-#ifdef _MSC_VER
-  #ifdef snprintf
-    #undef snprintf
-  #endif
-  #define snprintf _snprintf_s
-#endif
-
-
 namespace fs = boost::filesystem;
 
 namespace ros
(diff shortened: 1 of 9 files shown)


seanyen:
@mikepurvis This is ready for review too. The change basically is in favor of standard C++11 over MSVC specific APIs.

mikepurvis:
Great, thanks! I'm assuming that since ROS 1 on Windows is all officially still at an experimental stage, there's minimal commitment to supporting older MSVCs where this might not work?

PR #1821, opened 2019-10-01
[windows][melodic-devel] Add alternative lz4 library name

seanyen:
This pull request is to add another commonly seen `lz4` library name on Windows. This change is motivated by that `CMake` fails to find `lz4`, when building it against `lz4` built from its [visual studio solution file](https://github.com/lz4/lz4/tree/dev/visual/VS2017).

### DIFF
--- a/utilities/roslz4/CMakeLists.txt
+++ b/utilities/roslz4/CMakeLists.txt
@@ -13,7 +13,7 @@ if (NOT lz4_INCLUDE_DIRS)
   message(FATAL_ERROR "lz4 includes not found")
 endif()
 
-find_library(lz4_LIBRARIES NAMES lz4)
+find_library(lz4_LIBRARIES NAMES lz4 liblz4)
 if (NOT lz4_LIBRARIES)
   message(FATAL_ERROR "lz4 library not found")
 endif()


mikepurvis:
Is this typical? Normally the `lib` prefix is something that [CMake takes care of](https://cmake.org/cmake/help/latest/command/find_library.html):

> Each library name given to the `NAMES` option is first considered as a library file name and then considered with platform-specific prefixes (e.g. `lib`) and suffixes (e.g. `.so`). Therefore one may specify library file names such as `libfoo.a` directly. This can be used to locate static libraries on UNIX-like systems.

I'm not opposed to adding this in if it helps people building on Windows, but if nothing else, it would be good to understand why this isn't being handled elsewhere. And especially to confirm that this is _only_ an `lz4` thing, and not something that's going to end up being requested in a whole bunch of other places.

seanyen:
> I'm not opposed to adding this in if it helps people building on Windows, but if nothing else, it would be good to understand why this isn't being handled elsewhere. And especially to confirm that this is _only_ an `lz4` thing, and not something that's going to end up being requested in a whole bunch of other places.

At least per my testing, the `CMake` for Windows doesn't automatically search `NAMES` prefix. I think the situation is case by case, and it really depends on the naming rule of the upstream packages. For now, this naming divergence happens at `lz4`. 

JFYI, other projects also seemed to encounter the similar problem, and they ended up adding the additional commonly seen `liblz4` name. https://gitlab.kitware.com/vtk/vtk/issues/17314

mikepurvis:
Okay, this has my 👍 if it's something upstream CMake has had to do too.

seanyen:
@mikepurvis This one is ready for review and merge. Thanks!

PR #1823, opened 2019-10-03
Remove (potentially) recursive DEBUG statements

mjcarroll:
This is an alternative fix to #1241.   This also addresses ros/rosconsole#30

When using a log level of `debug`, every logging statement will have console output: 
```Warning: recursive print statement has occurred.  Throwing out recursive print.```

The logic introduced in #1241 uses `getCached` which will cause a second logging statement to occur recursively when log level is set to debug.

Edit: This has been updated to remove the offending DEBUG statements so that it will work in all cases.

### DIFF
--- a/clients/roscpp/src/libros/param.cpp
+++ b/clients/roscpp/src/libros/param.cpp
@@ -266,14 +266,11 @@ bool getImpl(const std::string& key, XmlRpc::XmlRpcValue& v, bool use_cache)
       {
         if (it->second.valid())
         {
-          ROS_DEBUG_NAMED("cached_parameters", "Using cached parameter value for key [%s]", mapped_key.c_str());
-
           v = it->second;
           return true;
         }
         else
         {
-          ROS_DEBUG_NAMED("cached_parameters", "Cached parameter is invalid for key [%s]", mapped_key.c_str());
           return false;
         }
       }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mjcarroll** [comment]
Alternatively: this is the offending log statement: https://github.com/ros/ros_comm/blob/d487d5877049d61fe2ebcc6487e9a7a72d9f0346/clients/roscpp/src/libros/param.cpp#L299

Is it likely that people are debugging the cache in normal operation?  Is this something that could be removed?

File: clients/roscpp/src/libros/param.cpp [hunk 1]
```diff
@@ -472,12 +462,12 @@ bool getCached(const std::string& key, XmlRpc::XmlRpcValue& v)
 	return getImpl(key, v, true);
 }
 
-template <class T> T xml_cast(XmlRpc::XmlRpcValue xml_value) 
+template <class T> T xml_cast(XmlRpc::XmlRpcValue xml_value)
```
dirk-thomas: Please revert this urelated whitespace change.

Same below.

mjcarroll: Whitespace changes removed.

flixr:
These debug messages can be useful sometimes, but while I personally prefer #1597 this is also acceptable to me.
The current state makes debugging impossible though and we really need to fix this one way or another!

dirk-thomas:

PR #1837, opened 2019-11-18
fix escape sequences in regular expressions

tgpfeiffer:
Using backslashes as escape sequences for regular expressions requires either escaping them from Python or using raw strings. (The current code causes warnings in recent Python versions.)


mikepurvis:
Thanks for this fix.

PR #1838, opened 2019-11-19
compatibility: RospyLogger findCaller arguments

Darnor:
In python 3.8 the findCaller() function will take an additional `stacklevel` argument (see: https://docs.python.org/3/library/logging.html#logging.Logger.findCaller). Since we are ignoring the stack_info argument already, we can just use *args and *kwargs to be compatible with python 3.7 and also python 3.8.

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -49,12 +49,12 @@ from rospkg.environment import ROS_LOG_DIR
 class LoggingException(Exception): pass
 
 class RospyLogger(logging.getLoggerClass()):
-    def findCaller(self, stack_info=False):
+    def findCaller(self, *args, **kwargs):
         """
         Find the stack frame of the caller so that we can note the source
         file name, line number, and function name with class name if possible.
         """
-        file_name, lineno, func_name = super(RospyLogger, self).findCaller()[:3]
+        file_name, lineno, func_name = super(RospyLogger, self).findCaller(*args, **kwargs)[:3]
         file_name = os.path.normcase(file_name)
 
         f = inspect.currentframe()


jwhendy:
Well, dangit. I wish I hadn't suspected that this was a stale python-3.7 lib dependency and rebuilt all of ROS from scratch today. This morning I found I can no longer `roslaunch` anything.

I'm on arch linux, which just updated to python-3.8.

```
$ roslaunch turtle_tf turtle_tf_demo.launch 
... logging to /home/jwhendy/.ros/log/051227fc-0c27-11ea-a276-0024d6f9ae16/roslaunch-voltaur-2726355.log
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.

  File "/opt/ros/melodic/lib/python3.8/site-packages/roslaunch/__init__.py", line 322, in main
    p.start()
  File "/opt/ros/melodic/lib/python3.8/site-packages/roslaunch/parent.py", line 273, in start
    self.logger.info("starting roslaunch parent run")
  File "/usr/lib/python3.8/logging/__init__.py", line 1434, in info
    self._log(INFO, msg, args, **kwargs)
  File "/usr/lib/python3.8/logging/__init__.py", line 1565, in _log
    fn, lno, func, sinfo = self.findCaller(stack_info, stacklevel)
TypeError: findCaller() takes from 1 to 2 positional arguments but 3 were given
Error in atexit._run_exitfuncs:
  File "/usr/lib/python3.8/logging/__init__.py", line 1565, in _log
    fn, lno, func, sinfo = self.findCaller(stack_info, stacklevel)
TypeError: findCaller() takes from 1 to 2 positional arguments but 3 were given
```

jwhendy:
If secondary input/replication helps, this patch was tested and fixed the python3.8 issue on arch linux. Thanks @Darnor !

mikepurvis:
Looks like a harmless fix, thanks for testing on Python 3.8!

PR #1842, opened 2019-11-28
[kinetic] Switch to yaml.safe_load

130s:
# Problem

File parsing with unsafe methods can be a threat.

`yaml.load` is recommended to be avoided in e.g. [security.openstack.org](https://security.openstack.org/guidelines/dg_avoid-dangerous-input-parsing-libraries.html).

# Potential solution

For the particular method in question (yaml.load), there's also a recommended method in the same link. So use that.

In fact in `melodic-devel` this is already done https://github.com/ros/ros_comm/pull/1688. This PR is essentially a cherry-pick of that.


dirk-thomas:
The original fix hasn't been released into Melodic yet (which will happen soon). After that all changes from the `melodic-devel` branch will be considered for backporting to `kinetic-devel`. Therefore closing this as a duplicate.

PR #1844, opened 2019-12-02
 #2 Buffer_overflow

sec-matsunaga:
Correct the indication by the static analysis tool(Klocwork).
The function 'strcpy' that does not check the buffer boundary is used.
Use the same function as the IPv6 method(inet_ntop).


sec-matsunaga:
I made a mistake and sent Pulliku to the head family. I'm sorry. Close

sec-matsunaga:
I made a mistake and sent pull request to the head family. I'm sorry. Close

PR #1850, opened 2019-12-13
Add option to repeat latched messages at the start of bag splits

tappan-at-git:
Split bagfiles lose all latched data. Many long-running ROS applications will include some kind of latched data that's only published once on launch (e.g. maps, static transforms), but prefer split bagfiles for ease of use, meaning they will have a difficult time actually using latched data without playing back the entire series of bagfiles.

This PR adds the --repeat-latched option to `rosbag record`. When this option is enabled, `rosbag::Recorder` remembers up to one message for each <topic, publisher> pair and prepends each latched message to each new bag file.

### DIFF
--- a/tools/rosbag/src/rosbag/rosbag_main.py
+++ b/tools/rosbag/src/rosbag/rosbag_main.py
@@ -368,7 +368,7 @@ The following variables are available:
                 else:
                     print('NO MATCH', verbose_pattern(topic, msg, t))          
 
-                total_bytes += len(serialized_bytes)
+                total_bytes += len(serialized_bytes) 
                 meter.step(total_bytes)
         else:
             for topic, raw_msg, t, conn_header in inbag.read_messages(raw=True, return_connection_header=True):
@@ -502,9 +502,9 @@ def check_cmd(argv):
         sys.exit(1)
 
     mm = MessageMigrator(args[1:] + append_rule, not options.noplugins)
-       
-    migrations = checkbag(mm, args[0])
 
+    migrations = checkbag(mm, args[0])
+       
     if len(migrations) == 0:
         print('Bag file does not need any migrations.')
         exit(0)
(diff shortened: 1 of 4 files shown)


mikepurvis:
This looks like a great addition; would you consider adding a simple test for the functionality, however? See several other bits of latch-related functional tests here:

https://github.com/ros/ros_comm/tree/melodic-devel/test/test_rosbag/test

dirk-thomas:
@tappan-at-git Friendly ping.

tools/rosbag/src/rosbag/rosbag_main.py
```diff
@@ -368,7 +370,7 @@ def eval_fn(topic, m, t):
                 else:
                     print('NO MATCH', verbose_pattern(topic, msg, t))          
 
-                total_bytes += len(serialized_bytes) 
+                total_bytes += len(serialized_bytes)
```
dirk-thomas: Please revert unrelated whitespace changes.

Same below.

dirk-thomas:
Since this patch changes ABI I will mark it to be considered for a future distro only.

tools/rosbag/src/rosbag/rosbag_main.py
tappan-at-git: I really thought I caught all of the EOL changes my editor snuck in. Will address along with the requested tests when I have the cycles.

oceanusxiv:
This resolves #706 if I recall correctly.

tools/rosbag/src/rosbag/rosbag_main.py
dirk-thomas: @tappan-at-git Please follow up on this soon if you want this patch to be considered for Noetic.

dirk-thomas: > Please follow up on this soon if you want this patch to be considered for Noetic.

@paulbovbel FYI

dirk-thomas: @tappan-at-git Friendly ping.

tappan-at-git:
Sorry about the force-push spam. Github was still showing one of the whitespace changes after I had pushed a fix. Unless Github is _seriously_ messing with me, all the unrelated whitespace changes should be squashed at this point.

I can't commit to adding full testing for this feature. Unless I'm missing something [rosbag's existing latched message test](https://github.com/ros/ros_comm/blob/noetic-devel/test/test_rosbag/test/latched_pub.py) doesn't seem to check if its latched messages were recorded correctly, and I don't see any tests for splits. I just don't have time to add those.

dirk-thomas:
Thank you for the patch.

It is unfortunate that it doesn't come with any kind of test coverage. The fact that some other parts of existing code aren't covered is not a reason to not implement tests for new features.

Since it is ABI breaking I will go ahead and merge it for the upcoming Noetic release anyway sice otherwise this feature couldn't land afterwards.

dirk-thomas:

PR #1854, opened 2019-12-23
Fix (#1516), prevent indefinitely trapping in unknown error state

akifh:
Even with https://github.com/ros/ros_comm/pull/1517 we still observe unknown network error time out, but this time, indefinitely (https://github.com/ros/ros_comm/issues/1516#issuecomment-565846185). Adding false return on `ValueError` seems to be the only fix for now.

### DIFF
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -365,6 +365,7 @@ def rosnode_ping(node_name, max_count=None, verbose=False):
                     return False
                 except ValueError:
                     print("unknown network error contacting node: %s"%(str(e)))
+                    return False
             if max_count and count >= max_count:
                 break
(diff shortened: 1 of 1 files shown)


akifh:
Shall we merge this PR? @mikepurvis @dirk-thomas

mikepurvis:
@dirk-thomas reviewed the previous patch that this builds on, so I'd prefer he look here as well.

akifh:
@mikepurvis, you're right. By the way, failed checks seem to be irrelevant.

mikepurvis:
The Noetic checks are expected to fail (for now) since none of Noetic has been bloomed other than `catkin` itself.

If the Melodic checks pass, it's fine.

flixr (approved):

dirk-thomas:
Thanks for the patch.

PR #1856, opened 2020-01-07
Read GPG passphrase from an environment variable

nnarain-cpr:
The `gnupg` library requires users to supply the passphrase for their key as a string into the decrypt function. There currently is no way to supply a passphrase through the `rosbag` command line tool, meaning only keys without passphrases can be used to decrypt encrypted bags.

This PR enables fetching the passphrase from the environment variable `ROSBAG_GPG_PASSPHRASE`.

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -194,7 +194,7 @@ class _ROSBagAesCbcEncryptor(_ROSBagEncryptor):
         # Encrypted symmetric key
         self._encrypted_symmetric_key = None
 
-    def initialize(self, bag, gpg_key_user):
+    def initialize(self, bag, gpg_key_user, passphrase=None):
         """
         Initialize encryptor by composing AES symmetric key.
         @param bag: bag to be encrypted/decrypted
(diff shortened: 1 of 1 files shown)


tools/rosbag/src/rosbag/bag.py
```diff
@@ -202,6 +204,7 @@ def initialize(self, bag, gpg_key_user):
         @raise ROSBagException: if GPG key user has already been set
         """
         if bag._mode != 'w':
+            self._gpg_passphrase = os.getenv('ROSBAG_GPG_PASSPHRASE', None)
```
mikepurvis: Can there be an option to pass it in here, for programmatic use? Eg,

```
def initialize(self, bag, gpg_key_user, passphrase=None):
    ...
    self._gpg_passphrase = passphrase or os.getenv('ROSBAG_GPG_PASSPHRASE', None)
```

nnarain-cpr: Updated

mikepurvis (changes requested):
Thanks for addressing this. Down the road, we need to figure out how to get the Python API to rosbag transparently using the gpg agent for the passphrase so that it works the same as how the C++ API does. But this is fine in the short term.

mikepurvis (approved):

mikepurvis:
@dirk-thomas Any concerns here? I doubt anyone else was using gpg keys with the passphrase set to `clearpath`, so I'd expect this is a safe change.

dirk-thomas:
Thanks for the patch.

Issue #1863, opened 2020-01-22
Forward declaration of struct in network.cpp on FreeBSD


dodsonmg:
`network.cpp` in `roscpp` includes `ifaddrs.h`.

On Linux, `ifaddrs.h` includes `sys/socket.h`, which defines `struct sockaddr`.

On FreeBSD, `ifaddrs.h` does *not* include `sys/socket.h`, though `sys/socket.h` still defines `struct sockaddr`.

When I try to build on FreeBSD, I get a series of `forward declaration of struct` errors when building `network.cpp` associated with pointers to `sockaddr` structures in the code block from lines 126 to 176.

The problem is fixed by including `sys/socket.h`, but presumably this isn't desirable, as the target OS for ROS is Ubuntu, and this include statement would be superfluous.

Ideally, I'd like to modify the include block of `network.cpp` as follows:

```
#ifdef HAVE_IFADDRS_H
    #include <ifaddrs.h>
    #include <sys/socket.h>  // added
#endif
```

How can I do this smarter, to support an acceptable PR?

mikepurvis:
The proposed modification looks fine to me. Just include a comment noting to future maintainers that it's there because FreeBSD needs it, so someone doesn't mistakenly "clean it up" down the road.

We have plenty of other similar hacks to ease building on non-Ubuntu systems, especially Windows and macOS.

dodsonmg:
Closing, as this is now tracked by PR #1864.

PR #1864, opened 2020-01-22
explicit include of socket.h to support FreeBSD

dodsonmg:
`network.cpp` in `roscpp` includes `ifaddrs.h`.

On Linux, `ifaddrs.h` includes `sys/socket.h`, which defines `struct sockaddr`.

On FreeBSD, `ifaddrs.h` does *not* include `sys/socket.h`, though `sys/socket.h` still defines `struct sockaddr`.

When I try to build on FreeBSD, I get a series of forward declaration of struct errors when building `network.cpp` associated with pointers to `sockaddr` structures in the code block from lines 126 to 176.

The problem is fixed by explicitly including `sys/socket.h`.

`transport_udp.cpp` and `transport_tcp.cpp` have similar issues.  Without explicitly including `sys/socket.h`, several variables and functions are undefined and result in errors during build (e.g., `AF_INIT`, `AF_INIT6`, `SOCK_DGRAM`).

See issue #1863

### DIFF
--- a/clients/roscpp/src/libros/transport/transport_tcp.cpp
+++ b/clients/roscpp/src/libros/transport/transport_tcp.cpp
@@ -42,6 +42,7 @@
 #include <boost/bind.hpp>
 #include <fcntl.h>
 #include <errno.h>
+#include <sys/socket.h>  // explicit include required for FreeBSD
 namespace ros
 {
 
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Thanks for the patch.

clients/roscpp/src/libros/transport/transport_tcp.cpp
```diff
@@ -42,6 +42,7 @@
 #include <boost/bind.hpp>
 #include <fcntl.h>
 #include <errno.h>
+#include <sys/socket.h>  // explicit include required for FreeBSD
```
seanyen, line 45: @dirk-thomas sorry for the late on this, but it seems to including `sys/socket.h` directly without any conditional guards would break the Windows build....

Issue #1866, opened 2020-01-23
Inability to find and link libgpgme when building rosbag_storage on FreeBSD


dodsonmg:
I am building `rosbag_storage` on FreeBSD.

I get the error: `ld: cannot find -lgpgme` though `libgpgme.so` is sitting happily in `/usr/local/lib`.

I've read that FreeBSD doesn't automatically include `/usr/local/lib` in the linker search path.

The relevant code is in `CMakeList.txt:28-48`

```
set(AES_ENCRYPT_SOURCE "")
set(AES_ENCRYPT_LIBRARIES "")
if(NOT WIN32)
  set(AES_ENCRYPT_SOURCE "src/aes_encryptor.cpp" "src/gpgme_utils.cpp")
  set(AES_ENCRYPT_LIBRARIES "crypto" "gpgme") <== set here
endif()

add_library(rosbag_storage
  src/bag.cpp
  src/bag_player.cpp
  src/buffer.cpp
  src/bz2_stream.cpp
  src/lz4_stream.cpp
  src/chunked_file.cpp
  src/message_instance.cpp
  src/query.cpp
  src/stream.cpp
  src/view.cpp
  src/uncompressed_stream.cpp
)
target_link_libraries(rosbag_storage ${catkin_LIBRARIES} ${Boost_LIBRARIES} \
${BZIP2_LIBRARIES} ${console_bridge_LIBRARIES} ${AES_ENCRYPT_LIBRARIES})  <== used here
```
Adding `link_directories(/usr/local/lib)` before setting `AES_ENCRYPT_LIBRARIES` solves the problem by (as I understand it) adding `/usr/local/lib` to the linker search path.  This is unnecessary for other platforms, so I can also wrap it in a conditional, checking for FreeBSD:

```
if(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
    link_directories(/usr/local/lib)
endif()
```

But this is still a bit hacky, and is adding possible dependencies that would need to be maintained.

I'm probably missing a simpler or more generic solution here.  I'd like to be able to submit a PR that makes it easy to build `rosbag_storage` (and ROS generally) on FreeBSD with the fewest possible changes to the existing instructions for installing from source (e.g., I'd like to avoid a long list of command line arguments being passed to cmake via catkin).

I've also asked about this at rosanswers.org and have been talking to Gijs van der Hoorn:
https://answers.ros.org/question/342313/failure-to-find-libgpgmeso-while-building-rosbag_storage/

[...]

PR #1867, opened 2020-01-23
rosbag_storage:  use find_library for abs path of crypto and gpgme libraries

dodsonmg:
On FreeBSD, `/usr/local/lib` is not in the default search path for the linker.

This results in the error `ld: cannot find -lgpgme` when building `rosbag_storage`, though `libgpgme.so` is present in `/usr/local/lib`.

The existing code sets `AES_ENCRYPT_LIBRARIES` to contain the library names (`crypto` and `gpgme`) and relies on the linker to find them.

This PR uses the `find_library` function to append `/usr/local/lib` to the default linker search path and set `AES_ENCRYPT_LIBRARIES` to the absolute paths of the `crypto` and `gpgme` libraries.

This is related to issue #1866.


dirk-thomas:
Thanks for the patch.

PR #1868, opened 2020-01-24
Do not display error message if poll yields EINTR

### LINKED ISSUE #63: " poll failed with error Interrupted system call" when debuging with C++
Hi all,

When I use the C++ debugger, roboware throws this error to me all the time. 
I am wondering anyone else having this issue.
According to this post, I am not alone. The difference is that I could debug in eclipse without any issues.
https://github.com/ros/ros_comm/issues/1370
gdb version: (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1

Thank you for your help.

### LINKED ISSUE #1370: gdb debugging generates error messages "[ERROR] [<TIME_STAMP>]: poll failed with error Interrupted system call"
summary:
--------
In the current version of ros_comm, when debugging with gdb using either single-steps or breakpoints, the error message "[ERROR] [<TIME_STAMP>]: poll failed with error Interrupted system call" is generated repeatedly (by the ros macro ROS_ERROR_STREAM).
The error messages were not generated in older ros_comm versions.
Only since end of february / beginning of march 2018 this problem seem to have occured.

severity of problem:
--------------------
A lot of error messages are generated during debugging.
Important error messages are therefore harder to see and may be overseen.

setup:
------
Ubuntu 16.04.4 LTS
ros-kinetic
ros-kinetic-roscpp_1.12.13-0xenial-20180222-205956-0800_amd64
gdb 7.11.1
gdbserver 7.11.1

details to reproduce problem:
-----------------------------
Debugging was performed with two simple publisher/subscriber nodes similar to talker/listener nodes of roscpp_tutorials.
Eclipse was used as main IDE.
See reference [1] how to setup eclipse project from catkin workspace.
(Note: problem can also be reproduced with command line gdb without eclipse).
The behaviour of error message generation dependends on different debugging setups.
The following list provides an overview of debugging setups (gdb with eclipse) and the occurance of error messages.

    Local C/C++ Application:
    - all-stop-mode
        -- error messages on single-step
    - non-stop-mode
        -- no error messages on single-step nor on breakpoints
        -- error messages, when thread #2 is suspended during poll and then resumed

    Remote C/C++ Application on localhost (with gdbserver):
    - non-stop-mode
        -- error messages on single-steps and on breakpoints

    Remote C/C++ Application on remote machine (with gdbserver):
    - non-stop-mode
        --- error messages on single-steps and on breakpoints
        

        
hints:
------
Building ros_comm package in debug mode, the source of the error messages was locatable.
See reference [2] how to build ros from source.
The library libroscpp.so generates the error messages.
The problem is located in the following files, lines and functions (as of writing 2018-04-20 16:00):

    - https://github.com/ros/ros_comm/blob/kinetic-devel/clients/roscpp/src/libros/io.cpp 
        -- line: 
            --- ~165
            --- 304 - 316
        -- function:
            --- pollfd_vector_ptr poll_sockets(int epfd, socket_pollfd *fds, nfds_t nfds, int timeout)
    - https://github.com/ros/ros_comm/blob/kinetic-devel/clients/roscpp/src/libros/poll_set.cpp
        -- line:
            --- ~184
        -- function:
            --- void PollSet::update(int poll_timeout)
            
In the simple publisher/subscriber test applications, the above mentioned functions were executed in thread number 2.
During debugging the main thread, in io.cpp, function poll_sockets, the system call epoll_wait is interrupted.

    - io.cpp line 306:    int fd_cnt = ::epoll_wait(epfd, ev, nfds, timeout);

A comment indicates, that an interrupted system call is not an error.

    - io.cpp line 310:    // EINTR means that we got interrupted by a signal, and is not an error

The gdb online documentation also states, that interrupted system calls are not errors, see reference [3].
The ROS_ERROR macro is NOT called in io.cpp.
But in poll_set.cpp, function PollSet::update (the caller of io.cpp, function poll_sockets), the return value of poll_sockets in case if interrupted system call is still handled as an error and the macro ROS_ERROR_STREAM is called.
This leads to many error messages with "poll failed with error Interrupted system call" during debugging.

The issue may be fixable by wrapping the system call epoll_wait into a while loop, see reference [4][5].

references:
-----------
references for setup:
    [1] http://wiki.ros.org/IDEs#Eclipse
    [2] http://wiki.ros.org/Installation/Source

reference about interrupted system calls can be found in gdbs online documentation:
    [3] https://sourceware.org/gdb/current/onlinedocs/gdb/Interrupted-System-Calls.html#Interrupted-System-Calls

similar issue of interrupted system calls and approach to solve:
    [4] https://stackoverflow.com/questions/6813629/how-do-i-use-gdb-for-multi-threaded-networking-program
    [5] https://stackoverflow.com/a/6815333

Arusekk:
Closes #1370
closes TonyRobotics/RoboWare#63
closes ericsantii/alexa-turtlesim-ros#2

### DIFF
--- a/clients/roscpp/src/libros/poll_set.cpp
+++ b/clients/roscpp/src/libros/poll_set.cpp
@@ -191,7 +191,10 @@ void PollSet::update(int poll_timeout)
   boost::shared_ptr<std::vector<socket_pollfd> > ofds = poll_sockets(epfd_, &ufds_.front(), ufds_.size(), poll_timeout);
   if (!ofds)
   {
-    ROS_ERROR_STREAM("poll failed with error " << last_socket_error_string());
+    if (errno != EINTR)
+    {
+      ROS_ERROR_STREAM("poll failed with error " << last_socket_error_string());
+    }
   }
   else
   {
--- a/clients/roscpp/src/libros/poll_set.cpp
+++ b/clients/roscpp/src/libros/poll_set.cpp
@@ -191,7 +191,7 @@ void PollSet::update(int poll_timeout)
   boost::shared_ptr<std::vector<socket_pollfd> > ofds = poll_sockets(epfd_, &ufds_.front(), ufds_.size(), poll_timeout);
   if (!ofds)
   {
-    if (errno != EINTR)
+    if (last_socket_error() != EINTR)
     {
       ROS_ERROR_STREAM("poll failed with error " << last_socket_error_string());
     }


clients/roscpp/src/libros/poll_set.cpp
```diff
@@ -191,7 +191,10 @@ void PollSet::update(int poll_timeout)
   boost::shared_ptr<std::vector<socket_pollfd> > ofds = poll_sockets(epfd_, &ufds_.front(), ufds_.size(), poll_timeout);
   if (!ofds)
   {
-    ROS_ERROR_STREAM("poll failed with error " << last_socket_error_string());
+    if (errno != EINTR)
```
dirk-thomas: Shouldn't this check `last_socket_error()`?

Arusekk:
This is possible, since I do not have any experience with the abstraction layer that is used here. I tested it, though, and works as expected (running an app using ros_comm, like rviz, hitting Ctrl-Z in terminal and then resuming with fg does not print any errors).

clients/roscpp/src/libros/poll_set.cpp
dirk-thomas: Please test the patch again with the above suggested change.

dirk-thomas:
Thanks for the patch.

PR #1870, opened 2020-01-25
Use setuptools instead of distutils

sloretz:
On Debian Buster and Ubuntu Focal [distutils has been split to a separate package](https://packages.debian.org/buster/python3-setuptools). With https://github.com/ros/catkin/pull/1048 catkin will prefer to use `setuptools` instead of `distutils`. This PR switches to `setuptools` to match catkin's preference. It uses conditional dependencies so it still works when `ROS_PYTHON_VERSION` is 2 to enable the target branch to be released to earlier ROS distros than Noetic.

I also removed shebangs from the setup.py since [`setup.py` isn't meant to be executed directly](http://docs.ros.org/melodic/api/catkin/html/user_guide/setup_dot_py.html)

### DIFF
--- a/clients/rospy/package.xml
+++ b/clients/rospy/package.xml
@@ -33,6 +33,8 @@
   <author>Ken Conley</author>
 
   <buildtool_depend version_gte="0.5.78">catkin</buildtool_depend>
+  <buildtool_depend condition="$ROS_PYTHON_VERSION == 2">python-setuptools</buildtool_depend>
+  <buildtool_depend condition="$ROS_PYTHON_VERSION == 3">python3-setuptools</buildtool_depend>
 
   <exec_depend>genpy</exec_depend>
   <exec_depend condition="$ROS_PYTHON_VERSION == 2">python-numpy</exec_depend>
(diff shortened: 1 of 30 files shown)


dirk-thomas:
Waiting for Noetic PR to be passing...

sloretz:

dirk-thomas:
Should this PR target `noetic-devel` instead? Assuming yes, I think the Python 2 dependency can be removed since Noetic only supports Python 3 and the Python 3 one shouldn't be necessary since catkin provides that dependency implicitly (https://github.com/ros/catkin/blob/19c653f118e49415c51326d9ca0de2ffec0cdda1/package.xml#L32).

sloretz:
> Should this PR target noetic-devel instead? Assuming yes, I think the Python 2 dependency can be removed since Noetic only supports Python 3 and the Python 3 one shouldn't be necessary since catkin provides that dependency implicitly

Retargeted at `noetic-devel`, rebased, and removed explicit dependency on setuptools in d9bc2fa1d031d3aff5fe23436f09a5e9b4c8ad43

dirk-thomas:
The failing tests are unfortunately know to be flaky.

PR #1871, opened 2020-01-26
[noetic] Restrict boost dependencies to components used

mikaelarguedas:
To be targeted at the noetic-devel branch

depends on https://github.com/ros/rosdistro/pull/23620 and https://github.com/ros/rosdistro/pull/23622

### DIFF
--- a/clients/roscpp/package.xml
+++ b/clients/roscpp/package.xml
@@ -24,6 +24,9 @@
   <buildtool_depend version_gte="0.5.78">catkin</buildtool_depend>
 
   <build_depend version_gte="0.3.17">cpp_common</build_depend>
+  <build_depend>libboost-chrono-dev</build_depend>
+  <build_depend>libboost-filesystem-dev</build_depend>
+  <build_depend>libboost-system-dev</build_depend>
   <build_depend>message_generation</build_depend>
   <build_depend>pkg-config</build_depend>
   <build_depend>rosconsole</build_depend>
@@ -37,6 +40,9 @@
 
   <run_depend version_gte="0.3.17">cpp_common</run_depend>
   <run_depend>message_runtime</run_depend>
+  <run_depend>libboost-chrono-dev</run_depend>
+  <run_depend>libboost-filesystem-dev</run_depend>
+  <run_depend>libboost-system-dev</run_depend>
   <run_depend>rosconsole</run_depend>
   <run_depend>roscpp_serialization</run_depend>
   <run_depend version_gte="0.3.17">roscpp_traits</run_depend>
--- a/tools/rosbag/package.xml
+++ b/tools/rosbag/package.xml
@@ -20,7 +20,11 @@
   <author>Jeremy Leibs</author>
   <author>James Bowman</author>
 
-  <depend>boost</depend>
+  <depend>libboost-date-time-dev</depend>
+  <depend>libboost-filesystem-dev</depend>
+  <depend>libboost-program-options-dev</depend>
+  <depend>libboost-regex-dev</depend>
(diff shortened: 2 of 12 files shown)


gavanderhoorn:
Should this target a (to be created) `noetic-devel`?

tools/rosbag/package.xml
```diff
@@ -20,7 +20,11 @@
   <author>Jeremy Leibs</author>
   <author>James Bowman</author>
 
-  <depend>boost</depend>
+  <depend>libboost-date-time-dev</depend>
+  <depend>libboost-filesystem-dev</depend>
+  <depend>libboost-program-options-dev</depend>
+  <depend>libboost-regex-dev</depend>
+  <depend>libboost-thread-dev</depend>
```
gavanderhoorn, line 27: Could we split this into `exec_depend` and the various `build*_depend`?

Did you not do this on purpose?

mikaelarguedas:
> Should this target a (to be created) noetic-devel?

Indeed, I put it in the PR description but I should have put it in the title for more visibility, editing now

tools/rosbag/package.xml
mikaelarguedas: Purposeful :grimacing: , currently I just replaced the keys without changing the scope of the tags.

I figured that when splitting the `depend`s tags, we should decide which to put in `build_depend` vs `exec_depend` vs `build_export_depend` which would require a closer audit of each package rather than just duplicating them all in build and exec depends.

It could be done in this PR though, it was just out of scope of my first pass

dirk-thomas:

dirk-thomas:

dirk-thomas:
@mikaelarguedas Please retarget the `noetic-devel` branch.

mikaelarguedas:
Done,

Note: CI will still install all boost packages because some PRs on underlying packages are pending https://github.com/ros/rospack/pull/115

clients/roscpp/package.xml
```diff
@@ -37,6 +40,9 @@
 
   <run_depend version_gte="0.3.17">cpp_common</run_depend>
   <run_depend>message_runtime</run_depend>
+  <run_depend>libboost-chrono-dev</run_depend>
+  <run_depend>libboost-filesystem-dev</run_depend>
+  <run_depend>libboost-system-dev</run_depend>
```
mikepurvis, line 45: Moving to format 2 would permit deduplicating these.

dirk-thomas:
> CI will still install all boost packages because some PRs on underlying packages are pending ros/rospack#115

With `rospack` recently released and rebuilt with that change I will trigger another round of CI for this:

dirk-thomas:
Thanks for the patch.

clients/roscpp/package.xml
mikaelarguedas: Agreed moving to format 2 for all packages would be a nice improvement.

Regarding deduplication: in the context of [Generating ‘dev’ and runtime artefacts from ROS packages](https://discourse.ros.org/t/generating-dev-and-runtime-artefacts-from-ros-packages/12448) we're aiming for not using the `<depend>` tag to be able to generate pure runtime packages without bringing in all the dev dependencies.

mikaelarguedas:
> With rospack recently released and rebuilt with that change I will trigger another round of CI for this:

Note: The resulting debs are still pulling all the boost packages because this repo depends on pluginlib and https://github.com/ros/pluginlib/pull/171 hasn't been merged and released.

ros-discourse:

https://discourse.ros.org/t/rfc-restricting-the-size-of-ros-docker-images/13252/1

PR #1873, opened 2020-01-30
Be able to have separate logger call rospy.loginfo and still get proper file and line number

lucasw:
I have a custom python logger that calls rospy.loginfo after doing custom operations with arguments, but anything that calls it sees the file and line number of the logger file rather than the original calling file and line number in the regular rosout.

I've modified a couple files here to make a parameter out of how far back to go with `inspect`, it works but it would be better not to modify ros_comm- inherit from it- or modify it in a better way?

This is a simple example of using it:

```
#!/usr/bin/env python
import rospy


class CustomLogger:
    def __init__(self):
        pass

    def debug(self, msg, special_kv={}):
        # do other stuff with special key values and msg, put them into a database, the cloud, etc.
        # ...
        rospy.logdebug(msg + " " + str(special_kv), num_frames_back=4)

    def info(self, msg, special_kv={}):
        rospy.loginfo(msg + " " + str(special_kv), num_frames_back=4)

    def warn(self, msg, special_kv={}):
        rospy.logwarn(msg + " " + str(special_kv), num_frames_back=4)

    def error(self, msg, special_kv={}):
        rospy.logerror(msg + " " + str(special_kv), num_frames_back=4)

    def fatal(self, msg, special_kv={}):
        rospy.logfatal(msg + " " + str(special_kv), num_frames_back=4)

if __name__ == '__main__':
    rospy.init_node('test_log')
    custom_logger = CustomLogger()

    for i in range(10):
        rospy.loginfo_once('test once ' + str(i))
        rospy.loginfo_throttle(1, 'test throttle ' + str(i))
        rospy.sleep(0.5)
        # custom_logger.info_once("custom message1", {"foo": "bar", "num": 2})

    rospy.logdebug('test0')
    custom_logger.debug("custom message0", {"foo": "bar", "num": 2})
    rospy.loginfo('test1')
    custom_logger.info("custom message1", {"foo": "bar", "num": 2})
    rospy.logwarn('test2')
    custom_logger.warn("custom message2", {"foo": "bar"})
    rospy.logwarn('test3')
    custom_logger.warn("custom message3", {"foo": "bar"})
    rospy.logfatal('test4')
    custom_logger.fatal("custom message4", {"foo": "bar"})
```

### DIFF
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -162,6 +162,11 @@ def _base_logger(msg, *args, **kwargs):
     level = kwargs.pop('logger_level', None)
     once = kwargs.pop('logger_once', False)
     throttle_identical = kwargs.pop('logger_throttle_identical', False)
+    num_frames_back = kwargs.get('num_frames_back', 2)
+    frame = inspect.currentframe()
+    for _ in range(num_frames_back):
+        if frame.f_back is not None:
+            frame = frame.f_back
 
     rospy_logger = logging.getLogger('rosout')
     if name:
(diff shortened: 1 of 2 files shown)


dirk-thomas:
@lucasw The patch is currently failing a unit test and needs to be rebased.

dirk-thomas:
@lucasw Closing due to no response. Please consider to open a new pull request targeting `noetic-devel`.

PR #1874, opened 2020-01-31
[rosbag] Catch exceptions by const ref.

mikepurvis:
GCC9 was throwing some `catch-value` warnings at me over this.


dirk-thomas:
Thanks for the patch.

PR #1875, opened 2020-02-04
install talker.py/listener.py under rospy/test_nodes

k-okada:
since launch files under https://github.com/ros/ros_comm/tree/1.14.3/tools/roslaunch/resources uses a lot of
```
    <node name="talker1" pkg="rospy" type="talker.py">
```
scripts, I think we should install talker.py/listener.py under rospy package directory


dirk-thomas:
I will go ahead and close this due to age and for targeting Melodic. I also don't think this is something which needs to change with Noetic likely being the last ROS 1 distro already being released.

PR #1876, opened 2020-02-04
[Windows][melodic-devel] Conditionally guard sys/socket.h for Windows.

seanyen:
A recent [merge](https://github.com/ros/ros_comm/pull/1864) results in build breaks on Windows side, and this pull request is to address this issue.


seanyen:
@dirk-thomas This is ready for review and merge. Thanks!

dirk-thomas:
Thanks for the quick fix.

PR #1877, opened 2020-02-04
test_rosbag: add target dependency to fix unit test failures in parallel builds

dirk-thomas:
Separated from https://github.com/ros/ros_comm/pull/1651#issuecomment-482148146.


dirk-thomas:

PR #1878, opened 2020-02-05
Drop custom implementation of boost::condition_variable to fix busy-wait spinning

meyerj:
An extension of https://github.com/ros/ros_comm/pull/1651, which addresses #1343 (_Regression in 1.12.13: rosout node spins rather than blocking and waiting, using 100% CPU_):

**Summary:**

The patch fixes an [ODR violation](https://en.cppreference.com/w/cpp/language/definition#One_Definition_Rule) causing undefined behavior, which - depending on the compiler flags for `roscpp` and compiler and linker internals - can make nodes either spin at 100% CPU or block indefinitely waiting on a `boost::condition_variable`. This happens if the symbol of the `boost::condition_variable` constructor is linked from a compilation unit where `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` has been defined, and symbols of member functions are linked from another compilation unit where `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` has not been defined, or vice-versa.

Actually this macro is meant as a build-time setting in Boost (in [boost/thread/detail/config.hpp](https://github.com/boostorg/thread/blob/32229388f51feba5af16eb5c99c746b459d14569/include/boost/thread/detail/config.hpp#L16)) and defining it in user packages is always dangerous. It is important to never expose `boost::condition_variable` instances  (or other Boost Thread types influenced by that flag?) in the public API, and to make sure that the constructor and member functions are only called internally by `roscpp` compilation units, where the `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` is defined. Strictly speaking, it is still an ODR violation because the macro effectively changes the implementation of the same symbols seen by different compilation units, and the patch only works because compilers tend to inline the relevant functions.

This problem will be only solved by Boost 1.67 or above, where `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` (renamed to `BOOST_THREAD_INTERNAL_CLOCK_IS_MONO`) became the default on platforms that support it (https://github.com/boostorg/thread/commit/1e84b978b2bb0aae830cc14533dea3b7ddda5cde).

**Extension to #1651:**

As requested in https://github.com/ros/ros_comm/pull/1651#issuecomment-582606542 I now completely removed the custom implementation of `boost::condition_variable` copied from Boost 1.61:

> > So instead of changing the namespace and selecting which implementation of `boost::condition_variable` to select depending on the system Boost version it would be straight-forward to drop the alternative implementation as a whole, as long as all target platforms have Boost 1.61 or above.
> 
> [REP 3](https://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023) specifies the targeted platforms and for Melodic that is:
> 
> * Ubuntu Artful (which is already EOL and can therefore be ignored),
> * Ubuntu Bionic (coming with Boost 1.65.1),
> * Debian Stretch (coming with Boost 1.62), and
> * Fedora 28 (which we don't build binaries for and can therefore also ignore).
> 
> Since that means all relevant platforms have Boost 1.61 or above we should do exactly that for the `melodic-devel` branch. Can you (or someone) please create a pull request with that proposed change?


meyerj:
ABI-compatibility report: [compat_report-fix-1343-melodic-devel-2.html](https://cdn.intermodalics.eu/compat_report-fix-1343-melodic-devel-2.html)

100% compatibility according to [abi-compliance-checker](https://lvc.github.io/abi-compliance-checker/) (on Ubuntu 18.04 amd64, with gcc 7.4.0). Only symbols exposed in public headers have been considered - as recommended on that page.

test/test_rosbag/bag_migration_tests/CMakeLists.txt
```diff
@@ -97,3 +97,7 @@ configure_file(test/random_play_sim.xml.in
   ${PROJECT_BINARY_DIR}/test/random_play_sim.xml)
 add_rostest(${PROJECT_BINARY_DIR}/test/random_play_sim.xml
   DEPENDENCIES run_tests_test_rosbag_rostest_test_random_record.xml)
+
+# Make sure that rostest random_record.xml is not executed twice, running in
+# parallel (https://github.com/ros/ros_comm/pull/1651#issuecomment-482148146):
+add_dependencies(run_tests_test_rosbag_rostest_test_random_play.xml run_tests_test_rosbag_rostest_test_random_play_sim.xml)
```
dirk-thomas: This change is already present on the target branch. Please remove it from this PR.

meyerj: Sorry, I did not notice that you force-pushed another version of my patch on branch `fix-1343-melodic-devel` (from #1651). I now cherry-picked the new commit on top of that and force-pushed it to here. I assume that is what you meant.

clients/roscpp/include/ros/rosout_appender.h
```diff
@@ -35,10 +35,11 @@
 #ifndef ROSCPP_ROSOUT_APPENDER_H
 #define ROSCPP_ROSOUT_APPENDER_H
 
+#include <ros/common.h>
 #include <ros/message_forward.h>
-#include "common.h"
```
dirk-thomas, line 39: Nitpick: I assume this change isn't necessary and can be removed?

clients/roscpp/include/ros/service_server_link.h
```diff
@@ -35,8 +35,9 @@
 #ifndef ROSCPP_SERVICE_SERVER_LINK_H
 #define ROSCPP_SERVICE_SERVER_LINK_H
 
-#include "ros/common.h"
+#include <ros/common.h>
```
dirk-thomas: Same here.

clients/roscpp/include/ros/timer_manager.h
```diff
@@ -578,14 +586,14 @@ void TimerManager<T, D, E>::threadFunc()
...
+        timers_cond_.wait_for(lock, boost::chrono::milliseconds(1));
       }
       else
       {
         // On system time we can simply sleep for the rest of the wait time, since anything else requiring processing will
         // signal the condition variable
-        int64_t remaining_time = std::max<int64_t>((sleep_end - current).toSec() * 1000.0f, 1);
-        timers_cond_.timed_wait(lock, boost::posix_time::milliseconds(remaining_time));
+        typename TimerManagerTraits<T>::time_point end_tp(boost::chrono::nanoseconds(sleep_end.toNSec()));
+        timers_cond_.wait_until(lock, end_tp);
```
dirk-thomas, line 596: What motivated the change to use `wait_until(end)` instead of `wait_for(duration)`? Without this change it seems `TimerManagerTraits` wouldn't be necessary either.

clients/roscpp/include/ros/rosout_appender.h
meyerj: I think it was a leftover from a previous version, where I modified `common.h.in` to  set the `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` flag there, which was a really bad idea because then the dangerous macro also affects user packages that include this header.

clients/roscpp/include/ros/timer_manager.h
meyerj: Yes, that should also work, indeed. I copied this section from [steady_timer.cpp](https://github.com/ros/ros_comm/pull/1878/files#diff-e304e755ded42c610ed4975cd99efd75L106-L109) and introduced the traits to avoid having to specialize the whole method.

Unfortunately I don't have time right now to test the proposed patch. It is important not to use `timed_wait()`, but `wait_for(duration)` as suggested.

dirk-thomas: I implemented that idea in 2a67cf6780c32fef3af96fac9da405f4e6eb0298.

flixr:
:+1: for removing the backported boost headers on melodic and up.

ahoarau:
@meyerj Still I have all my CPUs going crazy if I build with clang++-6 and c++14 enabled , on Ubuntu 18.04 with Boost 1.65.1.

meyerj:
> @meyerj Still I have all my CPUs going crazy if I build with clang++-6 and c++14 enabled , on Ubuntu 18.04 with Boost 1.65.1.

It is still possible that the symbols `boost::condition_variable()` and `boost::condition_variable::wait_for()` in `libroscpp.so`, which are compiled with `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` being defined, interfere with symbols of the same name generated in other compilation units where the macro has not been defined. So at the end the behavior depends on the linking order and compiler flags for `libroscpp.so` and other linked libraries (see description of this PR).

I have not tested this idea yet, but one option to get rid of the ODR violation - at least when building shared libraries - is to compile roscpp with hidden symbols ([\<LANG>_VISIBILITY_PRESET](https://cmake.org/cmake/help/v3.0/prop_tgt/LANG_VISIBILITY_PRESET.html#lang-visibility-preset) or [VISIBILITY_INLINES_HIDDEN](https://cmake.org/cmake/help/v3.0/prop_tgt/VISIBILITY_INLINES_HIDDEN.html#visibility-inlines-hidden) cmake properties). Then the Boost symbols and other internally used symbols will not be exposed in the dynamic symbol table and therefore cannot affect other libraries' behavior. This approach requires that all public symbols are exported explicitly using the `ROSCPP_DECL` macro, but because roscpp is ready for Windows where symbols are hidden by default, it could also work for Linux.

meyerj:
> To be clear, replacing `wait_until(sleep_end)` with `wait_for(sleep_end - current)` is not exactly the same.
> But I guess for how these Timers are supposed to be used it's ok.

I agree. If you check how [boost::condition_variable::wait_for()](https://github.com/boostorg/thread/blob/boost-1.65.0/include/boost/thread/pthread/condition_variable.hpp#L342) is implemented for the case `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` is defined, it basically calls `steady_clock::now()` again and then [forwards to wait_until()](https://github.com/boostorg/thread/blob/boost-1.65.0/include/boost/thread/pthread/condition_variable.hpp#L348) with the given duration added to the steady clock stamp. For [wait_until()](https://github.com/boostorg/thread/blob/boost-1.65.0/include/boost/thread/pthread/condition_variable.hpp#L354) and an argument already given in steady-clock time (for the `SteadyTime` instantiation) it directly calls the pthread implementation without an extra call to `now()`. For `WallTime` the two alternatives are mostly equivalent, with a slight advantage for [boost::condition_variable::wait_for()](https://github.com/boostorg/thread/blob/boost-1.65.0/include/boost/thread/pthread/condition_variable.hpp#L342) because it does not require [an extra call to `Clock::now()`](https://github.com/boostorg/thread/blob/boost-1.65.0/include/boost/thread/pthread/condition_variable.hpp#L335).

meyerj:
> I have not tested this idea yet, but one option to get rid of the ODR violation - at least when building shared libraries - is to compile roscpp with hidden symbols ([\<LANG>_VISIBILITY_PRESET](https://cmake.org/cmake/help/v3.0/prop_tgt/LANG_VISIBILITY_PRESET.html#lang-visibility-preset) or [VISIBILITY_INLINES_HIDDEN](https://cmake.org/cmake/help/v3.0/prop_tgt/VISIBILITY_INLINES_HIDDEN.html#visibility-inlines-hidden) cmake properties).

abi-compliance-checker is not happy with a `libroscpp.so` binary compiled with one of these options enabled. Many symbols are missing compared to the current version: [compat_report-fix-1343-melodic-devel-2-2a67cf67-vibility-preset-hidden.html](https://cdn.intermodalics.eu/compat_report-fix-1343-melodic-devel-2-2a67cf67-vibility-preset-hidden.html)

All those symbols are defined in a roscpp header or implicitly by the compiler, so normally derived packages should not depend on them and compile their own versions instead. This check has been done by comparing debug versions of the `libroscpp.so` binaries. In release mode the compiler would probably have inlined some of them and they would also not appear in the symbol table either.

The good thing is that also all `boost::condition_variable`-related symbols disappear from the dynamic symbol table and hence there is no risk that the versions affected by `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` are linked elsewhere.

dirk-thomas:
@ahoarau Can you please double check this patch again? Several others have reported in the past that it fixes the problem for them. If it doesn't for you can please share the complete and exact steps you did to try it.

dirk-thomas:
I have convinced myself that the current patch fixes the problem for Melodic (and does the right thing by getting rid of the copied Boost files).

* Using the following Dockerfile I confirmed the CPU usage is high in vanilla Kinetic:

<details>
<summary>Dockerfile - vanilla Kinetic:</summary>
FROM ros:kinetic-ros-base-xenial

RUN mkdir -p /catkin_ws/src && \
    cd /catkin_ws/src && \
    git clone https://github.com/ros/ros_comm.git -b kinetic-devel && \
    git clone https://github.com/ros/roscpp_core.git -b kinetic-devel

RUN cd /catkin_ws && \
    /ros_entrypoint.sh catkin_make_isolated

RUN echo '#!/bin/bash\n\
set -e\n\
source "/catkin_ws/devel_isolated/setup.bash"\n\
exec "$@"\n' > /devel_entrypoint.sh && chmod a+x /devel_entrypoint.sh

ENTRYPOINT ["/devel_entrypoint.sh"]
CMD ["roscore"]
</details>

* Using the following Dockerfile I confirmed the CPU usage is high in vanilla Melodic:

<details>
<summary>Dockerfile - vanilla Melodic:</summary>
FROM ros:melodic-ros-base-bionic

RUN mkdir -p /catkin_ws/src && \
    cd /catkin_ws/src && \
    git clone https://github.com/ros/ros_comm.git -b melodic-devel && \
    git clone https://github.com/ros/roscpp_core.git -b kinetic-devel

RUN cd /catkin_ws && \
    /ros_entrypoint.sh catkin_make_isolated

RUN echo '#!/bin/bash\n\
set -e\n\
source "/catkin_ws/devel_isolated/setup.bash"\n\
exec "$@"\n' > /devel_entrypoint.sh && chmod a+x /devel_entrypoint.sh

ENTRYPOINT ["/devel_entrypoint.sh"]
CMD ["roscore"]
</details>

* Using the following Dockerfile I confirmed the CPU usage is **not** high anymore using this patch in Melodic:

<details>
<summary>Dockerfile - patched Melodic:</summary>
FROM ros:melodic-ros-base-bionic

RUN mkdir -p /catkin_ws/src && \
    cd /catkin_ws/src && \
    git clone https://github.com/meyerj/ros_comm.git -b fix-1343-melodic-devel-2 && \
    git clone https://github.com/ros/roscpp_core.git -b kinetic-devel

RUN cd /catkin_ws && \
    /ros_entrypoint.sh catkin_make_isolated

RUN echo '#!/bin/bash\n\
set -e\n\
source "/catkin_ws/devel_isolated/setup.bash"\n\
exec "$@"\n' > /devel_entrypoint.sh && chmod a+x /devel_entrypoint.sh

ENTRYPOINT ["/devel_entrypoint.sh"]
CMD ["roscore"]
</summary>

I will go ahead and merge this patch by the end of the day (if there are no objections). Thank you to everyone who has contributed to this - especially @meyerj for creating and iterating on the patch.

matthew-reynolds:
>Several others have reported in the past that it fixes the problem for them

I was one of those people, but I only compiled using GCC, so I just re-ran your two Melodic Dockerfiles (Didn't bother w/ Kinetic) using `clang-6.0`/`clang++-6.0`.

I got the same behaviour with `clang` as you did with `gcc`:
 - Without the patch, very high CPU usage.
 - With the patch, near 0% CPU usage.

The Dockerfile is using the Ubuntu Bionic defaults: `clang-6.0`, `boost 1.65.1`. Perhaps if @ahoarau is still seeing poor behaviour there's something else at play than just a different compiler.

dirk-thomas:
Moving forward and merging this. Please feel free to continue commenting on the closed ticket.

dirk-thomas:
Special thanks to @meyerj for working on multiple iterations of this important fix as well as everyone else who helped with this.

Issue #1830, opened 2019-10-15
[py3] test_rostest test node run with wrong python version


sloretz:
## Problem description

It looks like `rostest` tests in this repo are using the wrong python version when `ROS_PYTHON_VERSION` is `3`. To reproduce, edit [`test_rostest` `just_advertise`](https://github.com/ros/ros_comm/blob/melodic-devel/test/test_rostest/test/just_advertise) to check the python version.

```diff
sloretz@nxt ~/w/n/s/ros_comm> git diff
diff --git a/test/test_rostest/test/just_advertise b/test/test_rostest/test/just_advertise
index 48507030f..6bda8abe5 100755
--- a/test/test_rostest/test/just_advertise
+++ b/test/test_rostest/test/just_advertise
@@ -44,6 +44,11 @@ import roslib.message
 
 
 if __name__ == '__main__':
+    import os
+    import sys
+    sys.stderr.write(repr(sys.version) + " " + repr(os.environ['ROS_PYTHON_VERSION']) + "\n")
+    assert sys.version[0] == os.environ['ROS_PYTHON_VERSION']
+
     rospy.init_node('just_advertise')
     msg_name = rospy.get_param('~msg_name')
     msg_class = roslib.message.get_message_class(msg_name)
```

In a `noetic` Python 3 workspace (I'm using Buster, but I'd expect the same issue on Bionic), run the tests using Python 3.

```
$ echo $ROS_PYTHON_VERSION
3
$ python3 src/catkin/bin/catkin_make_isolated --pkg test_rostest --cmake-args -DCATKIN_ENABLE_TESTING=1 --make-args run_tests
```

The test node is using python 2

```
'2.7.16 (default, Apr  6 2019, 01:42:57) \n[GCC 8.3.0]' '3'
Traceback (most recent call last):
  File "/home/sloretz/ws/noetic/src/ros_comm/test/test_rostest/test/just_advertise", line 50, in <module>
    assert sys.version[0] == os.environ['ROS_PYTHON_VERSION']
```

<details><summary>Full test output</summary>

```
==> Processing catkin package: 'test_rostest'
==> Building with env: '/home/sloretz/ws/noetic/devel_isolated/test_rosservice/env.sh'
Makefile exists, skipping explicit cmake invocation...
==> make cmake_check_build_system in '/home/sloretz/ws/noetic/build_isolated/test_rostest'
[...]

PR #1879, opened 2020-02-07
[noetic] wrap rostest call to add `python` pointing to sys.executable in PATH

sloretz:
Another possible solution to ros/ros_comm#1830. This adds a temporary directory with a symlink to python to the path variable.

This uses [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) which is only available in python 3.2 and above, so it must not be merged onto the `melodic-devel` branch. With this PR (and #1870) I don't see any `ros_comm` tests using Python 2, though many tests raise [`ModuleNotFoundError`](https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError) when trying to import other packages.

### DIFF
--- a/tools/rostest/scripts/rostest
+++ b/tools/rostest/scripts/rostest
@@ -53,9 +53,8 @@ class PythonPathOverride:
 
 
     def __enter__(self):
-
         self.dir = tempfile.TemporaryDirectory(prefix="rostest_bin_hook")
-        os.symlink(os.path.abspath(sys.executable), os.path.join(self.dir.name, "python"))
+        os.symlink(sys.executable, os.path.join(self.dir.name, "python"))
         os.environ['PATH'] = self.dir.name + ":" + os.environ['PATH']
 
     def __exit__(self, t, v, tb):
--- a/tools/rostest/scripts/rostest
+++ b/tools/rostest/scripts/rostest
@@ -33,4 +33,30 @@
 
 from rostest import rostestmain
 
-rostestmain()
+import os
+import sys
+import tempfile
+
+
+class PythonPathOverride:
+    """Replace python so test scripts with unmodified shebangs use correct python version."""
+
+    def __enter__(self):
+        # skip this hack on windows
+        if 'nt' == os.name:
+            return
+
+        self.dir = tempfile.TemporaryDirectory(prefix="rostest_bin_hook")
+        os.symlink(os.path.abspath(sys.executable), os.path.join(self.dir.name, "python"))
+        os.environ['PATH'] = self.dir.name + ":" + os.environ['PATH']
+
+    def __exit__(self, t, v, tb):
+        # skip this hack on windows
+        if 'nt' == os.name:
+            return
+
+        self.dir.cleanup()
+
+
+with PythonPathOverride():
+    rostestmain()
--- a/tools/rostest/scripts/rostest
+++ b/tools/rostest/scripts/rostest
@@ -39,14 +39,14 @@ import tempfile
 
 
 class PythonPathOverride:
-    """Replace python so test scripts with unmodified shebangs use correct python version."""
+    """Replace python so test scripts with unmodified shebangs use correct Python version."""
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> though many tests raise ModuleNotFoundError when trying to import other packages.

Can you please reference some of the failing tests here to be able to determine what might cause this.

tools/rostest/scripts/rostest
```diff
@@ -33,4 +33,30 @@
...
+class PythonPathOverride:
+    """Replace python so test scripts with unmodified shebangs use correct python version."""
+
+    def __enter__(self):
+        # skip this hack on windows
+        if 'nt' == os.name:
+            return
+
+        self.dir = tempfile.TemporaryDirectory(prefix="rostest_bin_hook")
+        os.symlink(os.path.abspath(sys.executable), os.path.join(self.dir.name, "python"))
```
dirk-thomas: Nitpick: [sys.executable](https://docs.python.org/3/library/sys.html#sys.executable) is already an absolute path.

sloretz:
> Can you please reference some of the failing tests here to be able to determine what might cause this.

```
ERROR: test_validate_master_launch (unit.test_roslaunch_launch.TestRoslaunchLaunch)
  File "/home/sloretz/ws/noetic/src/ros_comm/tools/roslaunch/test/unit/test_roslaunch_launch.py", line 47, in test_validate_master_launch
    import roslaunch.launch
  File "/home/sloretz/ws/noetic/src/ros_comm/tools/roslaunch/src/roslaunch/__init__.py", line 62, in <module>
    from .scriptapi import ROSLaunch
  File "/home/sloretz/ws/noetic/src/ros_comm/tools/roslaunch/src/roslaunch/scriptapi.py", line 42, in <module>
    import roslaunch.parent
  File "/home/sloretz/ws/noetic/src/ros_comm/tools/roslaunch/src/roslaunch/parent.py", line 54, in <module>
    import roslaunch.server
  File "/home/sloretz/ws/noetic/src/ros_comm/tools/roslaunch/src/roslaunch/server.py", line 79, in <module>
    from rosgraph_msgs.msg import Log
ModuleNotFoundError: No module named 'rosgraph_msgs'
```

dirk-thomas:
That test (`roslaunch/test/unit/test_roslaunch_launch.py`) isn't run by `rostest` but `nose`, right? So this change wouldn't have any effect.

tools/rostest/scripts/rostest
sloretz: resolved in 6e57fa0

sloretz:
> That test (roslaunch/test/unit/test_roslaunch_launch.py) isn't run by rostest but nose, right? So this change wouldn't have any effect.

Yeah it seems unrelated to this PR. `rosgraph_msgs` doesn't appear to be part of the underlay when building with `catkin_make_isolated`. No clue why yet, but I'll move that to a separate issue.

dirk-thomas:
> `rosgraph_msgs` doesn't appear to be part of the underlay when building with `catkin_make_isolated`.

I am surprised since it is a declared dependency: https://github.com/ros/ros_comm/blob/aedcece97cdf9c1de627e3610cd2ffd241c29531/test/test_rosbag/package.xml#L35

dirk-thomas (approved):
lgtm

sloretz:

edit: test failures, rerunning to check for flakiness, these are the old tests

* http://build.ros.org/job/Mpr__ros_comm__ubuntu_bionic_amd64/284/
* http://build.ros.org/job/Mpr_ds__ros_comm__debian_stretch_amd64/275/
* http://build.ros.org/job/Npr__ros_comm__ubuntu_focal_amd64/117/
* http://build.ros.org/job/Npr_db__ros_comm__debian_buster_amd64/116/

sloretz:

dirk-thomas (approved):
Just waiting for the noetic-devel branch to be created.

Issue #1855, opened 2019-12-27
Network latency affects performance of rospy logger function


furushchev:
Hi,

I tested the performance of ros on network with huge latency (up to 2000msec for round trip) and found that network latency affects execution time of logging functions on rospy.
(For example, In network 2000msec latency, `rospy.loginfo` takes more than 2000msec)
I remapped `/rosout` to other topic name, which is not subscribed by any node, but still the same issue happens.
I can guess `rospy.loginfo` invokes any blocking communication with rosmaster internally but I still could not track it.
On the other hand, in `roscpp` logging functions return immediately even in network with latency.

Does anyone know something about this issue?
Thank you very much in advance!

paulbovbel:
I'll bet I know what this is - from https://github.com/ros/ros_comm/issues/1575, rospy logging will currently hit rosmaster for the `/rosout_disable_topics_generation` on every logging invocation. This can be quite.. unpleasant.

According to https://github.com/ros/ros_comm/pull/1597#pullrequestreview-198270096, it would not be acceptable to only check this parameter at init time.

One way to resolve this would be to pull in the rospy cached param API PR from https://github.com/ros/ros_comm/pull/1515, and then query the param in cached mode as in here https://github.com/locusrobotics/ros_comm/pull/10/commits/5b39fbfd0bbecfbf4a44ba6d2ec00fa0af85a7da.

furushchev:
@paulbovbel Thank you for very helpful comment! I'll try the patch and see it solves the issue. Anyway, all the best for the new year! 👍

furushchev:
@paulbovbel Hi, I tested on the environment with huge latency (where it takes 2 seconds for round trip).
As far as I see the code, there are two calling rosmaster in one logging invocation:
1. `rospy.get_param("/rosout_disable_topics_generation")`
2. `get_topic_manager().get_topics()`

The hypotheses are:
As for the call 1, I expect that we can save calling every time by applying #1515.
As for the call 2, after applying the patch AND setting param `/rosout_disable_topics_generation` to `true`, the call is skipped.

And I confirmed it works as expected.

With application of #1515, I changed as follows:

```diff
diff --git a/clients/rospy/src/rospy/impl/rosout.py b/clients/rospy/src/rospy/impl/rosout.py
index a1ce4d84c..ba7d135dd 100644
--- a/clients/rospy/src/rospy/impl/rosout.py
+++ b/clients/rospy/src/rospy/impl/rosout.py
[...]

PR #1881, opened 2020-02-10
Use cached parameter for rosout_disable_topics_generation

paulbovbel:
Resolve issue noted in https://github.com/ros/ros_comm/issues/1855

Currently, every rospy.log* event triggers a param fetch from master.


dirk-thomas:
Thanks for the patch.

PR #1885, opened 2020-02-16
Allow empty machine arg in node tag.

### LINKED ISSUE #274: Allow empty machine attribute
Allowing empty machine attribute of node tag would help to make a launch-files with optional machine argument.

An example:

```
<arg name="machine" default=""/>
<node name="some_name" pkg="some_pkg" type="some_type" machine="$(arg machine)"/>
```

peci1:
Fixes #274. Tested locally on my machine and it seems to work. Are there any integration tests that could really check the working of the machine tags?


dirk-thomas:
Thanks for the patch.

PR #1894, opened 2020-02-21
fix dictionary changed size during iteration

dirk-thomas:
Search for `changed size during iteration` in http://build.ros.org/view/Ndev/job/Ndev__ros_comm__ubuntu_focal_amd64.


dirk-thomas:
While the builds are still not green the build output doesn't contain the error message anymore.

ros-discourse:

https://discourse.ros.org/t/new-packages-for-melodic-2020-03-22/13329/3

PR #1899, opened 2020-02-25
make statistics.test less flaky

dirk-thomas:


dirk-thomas:

Issue #1880, opened 2020-02-09
Roscpp won't compile on Arch Linux because it expects Boost 1.61 or above and fails to recgonize my Boost 1.72.0 version.


djohn49:
I am currently trying to build Ros on Arch Linux through yay/aur as documented [here](wiki.ros.org/melodic/Installation/ArchLinux). When yay gets to roscpp, however, the build fails with the following error (happens about a half hour after starting the command to install the full desktop package set):
```
-- Found Boost: /usr/lib64/cmake/Boost-1.72.0/BoostConfig.cmake (found version "1.72.0") found components: chrono filesystem system 
CMake Error at CMakeLists.txt:30 (message):
  roscpp requires Boost 1.61 or above.


-- Configuring incomplete, errors occurred!
See also "/home/david/.cache/yay/ros-melodic-roscpp-git/src/build/CMakeFiles/CMakeOutput.log".
See also "/home/david/.cache/yay/ros-melodic-roscpp-git/src/build/CMakeFiles/CMakeError.log".
==> ERROR: A failure occurred in build().
    Aborting...
error: target not found: ros-melodic-roslint
error: target not found: ros-melodic-urdf-parser-plugin
error: target not found: ignition-cmake
error: target not found: ignition-tools
```
It looks like whatever version check is being used can't work correctly with Boost 1.72.0 for some reason.

dirk-thomas:
You might want to print the Boost version as reported by CMake by adding the following line above the line resulting in the fatal error:

```message("Boost_VERSION ${Boost_VERSION}")```

**Comment by john01dav** [comment]
Isn't that what's already shown in the first line of my log? It says that the version is 1.72.0.

**Comment by dirk-thomas** [comment]
I will close the ticket for now. Please feel free to comment on the closed ticket with the requested information and if necessary the ticket can be reopened.

**Comment by tspicer01** [comment]
@dirk-thomas   We just upgrade boost (1.71) and hit the same issue.
It looks like there is a new policy on how the boost version is defined
https://cmake.org/cmake/help/v3.15/policy/CMP0093.html?highlight=cmp0093

I've changed my local CMakeList.txt to version check on Boost_VERSION_MACRO instead of Boost_VERSION and that is consistent with the old behaviour.

```
Boost Version found 1.71.0
Boost Version Macro found 107100
```

PR #1902, opened 2020-02-27
CMP0093 in findBoost changes how Boost_VERSION is reported

tspicer01:
This should fix the Boost version checks for the newer version of Boost.

Newer version (1.71+) report Boost_VERSION in x.y.z format, the older policy reported the Boost_VERSION_MACRO.  Set the check to always use the Boost_VERSION_MACRO

See https://cmake.org/cmake/help/v3.15/policy/CMP0093.html?highlight=cmp0093

This should fix issue https://github.com/ros/ros_comm/issues/1880

Tested using Boost version 1.71 and 1.67 on Ubuntu focal

### DIFF
--- a/clients/roscpp/CMakeLists.txt
+++ b/clients/roscpp/CMakeLists.txt
@@ -26,10 +26,10 @@ find_package(Boost REQUIRED COMPONENTS chrono filesystem system)
 
 # Make sure we use CLOCK_MONOTONIC for the condition variable wait_for if not Apple.
 if(NOT APPLE AND NOT WIN32)
-  if(Boost_VERSION LESS 106100)
+  if(Boost_VERSION_MACRO LESS 106100)
     message(FATAL_ERROR "${PROJECT_NAME} requires Boost 1.61 or above.")
   endif()
-  if(Boost_VERSION LESS 106700)
+  if(Boost_VERSION_MACRO LESS 106700)
     # CLOCK_MONOTONIC became the default in Boost 1.67:
     # https://github.com/boostorg/thread/commit/1e84b978b2bb0aae830cc14533dea3b7ddda5cde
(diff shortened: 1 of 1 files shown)


mikepurvis:
This would also be fixable by using VERSION_LESS for the comparison, right? Wondering if that way might be more compatible.

----

Oh shoot no, that won't work either because you still need to compare against the other style of version number on the right hand side. Ugh, it looks like this came in as a consequence of CMP0093 in CMake 3.15:

https://cmake.org/cmake/help/v3.15/policy/CMP0093.html

If `noetic-devel` is supposed to support Debian Buster and Ubuntu Focal (per https://github.com/ros-infrastructure/rep/pull/202), then we're dealing with CMake 3.13.4 and 3.16.3. TBH I think the easiest fix for this might be to just disable CMP0093.

----

A little bit of fun further discovery— CMP0093 only applies to the FindBoost shipped by CMake, which on Focal is in `/usr/share/cmake-3.16/Modules/FindBoost.cmake`. However, in Boost 1.71, Boost itself also ships its _own_ module, in `/usr/lib/x86_64-linux-gnu/cmake/Boost-1.71.0/BoostConfig.cmake`.

The native Boost version is first in the search path, so setting CMP0093 in either direction does nothing— it doesn't exist in CMake 3.13, and by 3.16, Boost's own module aliases the place where it would have had an effect.

My suggestion would be to do something like this:

```
# Compatibility with Boost < 3.15.
if (NOT Boost_VERSION_STRING)
  set(Boost_VERSION_STRING "${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}")
endif()
```

And then do comparisons against `Boost_VERSION_STRING` using `VERSION_LESS` and friends.

dirk-thomas:
Please see #1903 for an alternative approach since the conditional logic isn't necessary anymore.

dirk-thomas:
Closing in favor of #1903.

PR #1903, opened 2020-02-28
remove Boost version check since Noetic only targets platforms with 1.67+

dirk-thomas:
Alternative to #1902.


mikepurvis:
Ah, well, that works too. :)

dirk-thomas:

sloretz (approved):
built fine in a focal docker container created today

dirk-thomas:

PR #1914, opened 2020-03-19
roslaunch-check: Search dir recursively

mortenfyhn:
Hi! What do yall think about something like this?

This will simply check all .launch files in the directory you pass, including
subdirectories. I recently set up roslaunch-check on a CI server, and
missed having a feature like this.

However, the directory you pass must still be a ROS package
or a subdirectory of a ROS package. If you have, say, a dir
containing many packages, and you want to check every launch
file in every package in that dir, then you'll need to invoke
roslaunch-check once for each package.

### DIFF
--- a/tools/roslaunch/scripts/roslaunch-check
+++ b/tools/roslaunch/scripts/roslaunch-check
@@ -53,6 +53,7 @@ def check_roslaunch_file(roslaunch_file, use_test_depends=False, ignore_unset_ar
 def check_roslaunch_dir(roslaunch_dir, use_test_depends=False, ignore_unset_args=False):
     roslaunch_files = []
     for dirpath, dirnames, filenames in os.walk(roslaunch_dir):
+        dirnames[:] = [d for d in dirnames if not d.startswith(".")]
         for filename in fnmatch.filter(filenames, "*.launch"):
             roslaunch_files.append(os.path.join(dirpath, filename))
     error_msgs = []
--- a/tools/roslaunch/scripts/roslaunch-check
+++ b/tools/roslaunch/scripts/roslaunch-check
@@ -33,6 +33,7 @@
 
 from __future__ import print_function
 
+import fnmatch
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


Rayman:
~~This already works like this? Try to give a directory as argument:~~
```
rosrun roslaunch roslaunch-check launch/
checking *.launch in directory launch/
...
```
Edit: You are right, it's normally does not recurse into subdirectories

mortenfyhn:
Either I'm very confused, or something caused it to not work that way on my machine. Hang on, I'll double check.

mortenfyhn:
(I'm running Melodic.)

It seems that if I have a package that for instance looks like this
```
my_package
├── CMakeLists.txt
├── launch
│   ├── foo.launch
│   ├── bar.launch
│   └── baz
│       └── foobar.launch
└── package.xml
```
then
```
rosrun roslaunch roslaunch-check my_package/launch
```
will check `foo.launch` and `bar.launch`, but not `foobar.launch`, presumably because it's in a subdirectory.

dirk-thomas:
Please target the default branch (which is `noetic-devel`).

mortenfyhn:
Done!

mortenfyhn:
I'm also very interested in what you think about a further change to allow passing a non-ROS-package dir (to check all launch files in an entire workspace, for instance).

tools/roslaunch/scripts/roslaunch-check
```diff
@@ -50,12 +51,13 @@ def check_roslaunch_file(roslaunch_file, use_test_depends=False, ignore_unset_ar
         return "[%s]:\n\t%s"%(roslaunch_file,error_msg)
 
 def check_roslaunch_dir(roslaunch_dir, use_test_depends=False, ignore_unset_args=False):
+    roslaunch_files = []
+    for root, dirnames, filenames in os.walk(roslaunch_dir):
```
dirk-thomas: This loop should probably ignore subdirectories starting with a `.` in order to e.g. not recursively search in `.git` / `.svn` directories (which in the latter case would contain the same launch files again).

mortenfyhn: You're right.

dirk-thomas:

dirk-thomas:
Thanks for the patch.

dirk-thomas:
Thanks for the patch.

mortenfyhn:
Thanks for merging!

PR #1918, opened 2020-03-20
rospy: fix tcp reconnect case

dandedrick:
If we hit one of the cases where we get an exception during the
connection sequence and a reconnection is desired then we need to close
the socket and set it to None. This will allow the connect loop in
robust_connect_subscriber() to continue for another iteration. Without
these changes the connect loop would exit because conn.socket was not
None and then enter the receive loop without successfully having
negotiated a connection.

This can cause a deadlock between a publisher and subscriber if the
connection was established but the ros header was not written out. This
will leave the publisher trying to receive the ros header from the
subscriber and the subscriber waiting to receive a message from the
publisher. When this case is hit, for whatever reason, it's best to just
close the current connection and establish a new connection.

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -586,6 +586,14 @@ class TCPROSTransport(Transport):
                 #
                 # no reconnection as error is not 1.-4.
                 self.close()
+            else:
+                try:
+                    self.socket.shutdown(socket.SHUT_RDWR)
+                except:
+                    pass
+                finally:
+                    self.socket.close()
+                self.socket = None
             raise TransportInitError(str(e)) #re-raise i/o error
                 
     def _validate_header(self, header):


dirk-thomas:
Please provide a reproducible example as well as information what specific exception this patch is handling.

dandedrick:
Sorry it's been over a month since I was reproducing this and I'm not completely sure exactly which exception this was as there was several issues around this that I was hitting. I'm going to see if I can dig up the specific exception that I was hitting in this case. I'll update when I've gotten back to where I was with more detailed explanation. The reproducible scenario I think was somewhat timing based though so I don't think it would happen every time. 

The fundamental problem is that the socket in not None but the header exchange had not occurred when the exception occurred. It would not hit either the if or elif case in the exception handler and then it would exit with the socket not None and done is False and go into the recieve_loop on the subscriber side. The publisher would still be in _tcp_server_callback waiting for the ros header forever, which was never sent due to the exception and the fact that it left the connection open waiting to receive a message despite having never exchanged headers.

Basically if we have failed to fully complete the connect function we should never try continue on and accept messages. Instead we should either attempt to reestablish the connection or give up. The if and elif cases handle when we want to give up but the non-existent else that I added handles the case where we actually want to re establish the connection.

dirk-thomas:
> The reproducible scenario I think was somewhat timing based though so I don't think it would happen every time.

Any reproducible steps - even if they need to be run N times to make it fail - would be helpful.https://guides.github.com/features/mastering-markdown/

dandedrick:
The exception that I'm seeing when I can reproduce this is
```
[rospy.internal][WARNING] 2020-04-23 10:43:00,369: Unknown error initiating TCP/IP socket to USLEXM02287574:43345 (http://USLEXM02287574:44553/): Traceback (most recent call last):
  File ".../ros_comm/clients/rospy/src/rospy/impl/tcpros_base.py", line 555, in connect
    self.socket.connect((dest_addr, dest_port))
  File "/usr/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
timeout: timed out
```

I reproduce this with a simple publisher:
```
#!/usr/bin/env python
import rospy
import std_msgs

rospy.init_node("Foo", anonymous=True)
pub = rospy.Publisher("Foo", std_msgs.msg.String, queue_size=1)

rate = rospy.Rate(1)
    pub.publish(std_msgs.msg.String("foo"))
    rate.sleep()
```

I start this publisher and then determine what tcp port it's listening on. I usually do this with strace but I'll provide the more elegant solution.  Use `rostopic info /Foo` to get the http URI and use it in `xmlrpc $FOO_URI requestTopic s/rostopic_123 s/Foo "array/(array/(s/TCPROS))" `  where $FOO_URI is the http URI. This should return the tcp port number to use.

Now run `nc $TCP_HOSTNAME $TCP_PORT` where $TCP_HOSTNAME and $TCP_PORT correspond the tcp port number found in the last step. This netcat command will just hold the Publisher in it's accept loop and prevent it from accepting new connections. This simulates something like a process being put to sleep mid connection attempt. One example might be someone doing control-z mid rostopic call or attaching gdb to a process and freezing it for some period of time.

The next step is to spawn a bunch of rostopic echo processes this just simulates a number of connections coming in while the publisher is trapped in the accept loop. I do the following `for i in {0..32}; do rostopic echo -n 1 /Foo & done`. You can adjust the number of processes you start but I've found its much easier to hit when you start this many. I think that's just because it increases the probability of hitting some timing correctly.

After you have started all of these processes you need to wait about 1 minute (I think this allows the processes to hit their connection timeout at least once) and then kill the netcat process you started earlier (control-d) should stop it. When you hit this issue you will find that some of the processes do finish correctly but one will hit the above mentioned exception and jam the accept loop for the publisher. This prevents any of the other subscribers from being able to fully establish their connection.

Some additional information. In this state I've connected gdb to both the publisher and the subscriber that was stuck in the receive loop and used py-bt to backtrace the accept thread.

The publisher is stuck here:
```
Traceback (most recent call first):
  File ".../rosgraph/network.py", line 362, in read_ros_handshake_header
    d = sock.recv(buff_size)
  File ".../ros_comm/clients/rospy/src/rospy/impl/tcpros_base.py", line 321, in _tcp_server_callback
    header = read_ros_handshake_header(sock, StringIO(), buff_size)
  (frame information optimized out)
  File "/usr/lib/python2.7/threading.py", line 754, in run
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
```

and the subscriber is stuck here:
```
Traceback (most recent call first):
  File ".../ros_comm/clients/rospy/src/rospy/impl/tcpros_base.py", line 103, in recv_buff
    d = sock.recv(buff_size)
  File ".../ros_comm/clients/rospy/src/rospy/impl/tcpros_base.py", line 732, in receive_once
    self.stat_bytes += recv_buff(sock, b, p.buff_size)
  File ".../ros_comm/clients/rospy/src/rospy/impl/tcpros_base.py", line 792, in receive_loop
    msgs = self.receive_once()
  (frame information optimized out)
  File "/usr/lib/python2.7/threading.py", line 754, in run
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
```

Using gdb and lsof I can confirm that both processes are waiting on the same socket demonstrating that the publisher is waiting for the ros header while the subscriber has skipped over that step and is now waiting for a topic message.

dirk-thomas:
Blindly (`else`) allowing a reconnect to happen sounds potentially undesired to me. With all the special case consideration already happening in the lines above the patch I think it would be good to know which exact case this patch is trying to handle: what is the exception type, what is the `errno`, etc. to then narrowly allow a retry for this specific case.

dirk-thomas:

PR #1920, opened 2020-03-27
Fix brief description comments after members

elektrokokke:
Current description comments after members start with '//!', resulting in broken documentation showing the descriptions on the wrong members. Change that to '//!<' according to doxygen documentation to have descriptions correctly associated with the respective members.


dirk-thomas:
Thanks for the fix.

PR #1922, opened 2020-04-01
[roslaunch] remove pycrypto import (not used)

jspricke:
pycrypto used to be installed through rosbag. This was dropped in #1609
making roslaunch on remote hosts fail.


dirk-thomas:
Thanks for the patch.

PR #1928, opened 2020-04-12
Do not count free FDs in userland.

### LINKED ISSUE #1927: XmlRpcServer::countFreeFDs() takes too long inside docker container
Hi there,

We just upgrade to ROS melodic from kinetic recently, and found that node connections take too long to establish, sometimes even over 2 seconds.
Our application is very time sensitive, and this affects a lot.

After some debug, I found that the issue is caused by `countFreeFDs()` function in xmlrpcpp.

It's comments say:
```
int XmlRpcServer::countFreeFDs() {
  // NOTE(austin): this function is not free, but in a few small tests it only
  // takes about 1.2mS when querying 50k file descriptors.
```

But inside docker container, file descriptors number usually is 1048576:
```
$ python -c "import resource; print(resource.getrlimit(resource.RLIMIT_NOFILE))"
(1048576, 1048576)
```
that takes a huge time to query.

docker version we are using is
```
$ docker --version
Docker version 19.03.5, build 633a0ea838
```

Any idea how to fix this? Thanks

fujitatomoya:
fix https://github.com/ros/ros_comm/issues/1927, just get rid of counting free file descriptors.

Signed-off-by: Tomoya.Fujita <tomoya.fujita825@gmail.com>

### DIFF
--- a/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcServer.h
+++ b/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcServer.h
@@ -96,9 +96,6 @@ namespace XmlRpc {
     //! Create a new connection object for processing requests from a specific client.
     virtual XmlRpcServerConnection* createConnection(int socket);
 
-    //! Count number of free file descriptors
-    int countFreeFDs();
-
     // Whether the introspection API is supported by this server
     bool _introspectionEnabled;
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Just removing these checks (which obviously have been introduced for a reason) doesn't seem to make sense. Since you also opened #1929 I will go ahead and close this as a duplicate.

PR #1929, opened 2020-04-12
Check if enough FDs are free, instead counting the total free FDs.

### LINKED ISSUE #1927: XmlRpcServer::countFreeFDs() takes too long inside docker container

fujitatomoya:
fix https://github.com/ros/ros_comm/issues/1927,

requirement seems that there is enough free file descriptors or not, that is it.
so no need to count the number of  free file descriptors.

Signed-off-by: Tomoya.Fujita <tomoya.fujita825@gmail.com>

### DIFF
--- a/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcServer.h
+++ b/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcServer.h
@@ -96,8 +96,8 @@ namespace XmlRpc {
     //! Create a new connection object for processing requests from a specific client.
     virtual XmlRpcServerConnection* createConnection(int socket);
 
-    //! Count number of free file descriptors
-    int countFreeFDs();
+    //! Check if enough number of free file descriptors
+    bool enoughFreeFDs();
 
     // Whether the introspection API is supported by this server
     bool _introspectionEnabled;
--- a/utilities/xmlrpcpp/src/XmlRpcServer.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcServer.cpp
@@ -197,7 +197,7 @@ XmlRpcServer::acceptConnection()
     _accept_retry_time_sec = _disp.getTime() + ACCEPT_RETRY_INTERVAL_SEC;
     return 0; // Stop monitoring this FD
   }
-  else if( countFreeFDs() < FREE_FD_BUFFER )
+  else if ( !enoughFreeFDs() )
   {
     XmlRpcSocket::close(s);
     XmlRpcUtil::error("XmlRpcServer::acceptConnection: Rejecting client, not enough free file descriptors");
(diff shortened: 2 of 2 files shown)


flixr:
You are still counting the number of free FDs...
I don't see how this could help at all... or what am I missing?

fujitatomoya:
@flixr 

> You are still counting the number of free FDs...

actually it is, once it hits the enough number, it returns. I guess that this could be faster in most cases against checking all.

flixr:
True, my oversight... and I guess since it polls fds with increasing number it should first go through all used fds, then unused ones (at which point it exits early enough free ones are found).

wenbin1989:
thanks @fujitatomoya 
it works fine, the connections init latency reduces a lot.
but I still have concerns about the `std::vector<struct pollfd> pollfds` vector.
it resizes to `max_files` at https://github.com/ros/ros_comm/blob/87b2baf36ca4d4183b2f982c5fb2d707b146c352/utilities/xmlrpcpp/src/XmlRpcServer.cpp#L45
which is `1048576` inside docker container. it takes extra 8MB memory, I think we don't need allocate the total FDs there as well?

fujitatomoya:
@wenbin1989 

thanks for checking, appreciate it.

> std::vector<struct pollfd> pollfds

nice catch, to be honest, i did miss it. thanks!

but i do not have any idea to reduce it. since we cannot know how many/what events in poll() returns, i guess that we have to keep the maximum.

harlowja:
Possible to allocate and check `pollfds` in chunks of smaller sizes?

fujitatomoya:
> Possible to allocate and check pollfds in chunks of smaller sizes?

I don't think so,

we do not know which file descriptor returns with POLLNVAL, actually that is exactly what we are trying to get here. so i believe that we need to scan with max index for pollfds, then once hits the requirement, return right away.

fujitatomoya:
@dirk-thomas 

do we have a go on this to merge?

dirk-thomas:
Thanks for the patch.

wenbin1989:
thanks! @dirk-thomas can we update the package version?

dirk-thomas:
> can we update the package version?

Since patch releases of the `ros_comm` are fairly expensive in terms of resources to rebuild almost full ROS distros as well as bandwidth for every user downloading new `.deb`s they happen whenever a significant enough amount of changes have accumulated. So it will likely be some time (I would guess a couple of weeks) until a new patch release will be available in Noetic and Melodic.

PR #1917, opened 2020-03-20
changes between 1.15.4 (Noetic) and 1.14.4 for backporting into Melodic

dirk-thomas:
The following list of changes has been integrated into ros_comm 1.15.4 (Noetic) since the last Melodic release (1.14.4).

**Backported:** (these changes are part of this PR)

* update test to pass with old and new yaml (#1893)

  * Improve testing

* fix flakyness of transform test (#1890)

  * Improve testing

* fix dictionary changed size during iteration (#1894)

  * Improve testing

* fix flakyness of test_topic_statistics (#1896)

  * Improve testing

* increase time limit of advertisetest/publishtest.test to reduce flakyness (#1897)

  * Improve testing

* make statistics.test less flaky (#1899)

  * Improve testing

* Fix "Unable to report rosout: __exit__" (#1915)

  * Bug fix

* Fix mac rospy polling issue (#1907)

  * Bug fix

* Fix Python 3 mac uncaught exception (#1908)

  * Bug fix

**Not backported:**

* wrap rostest call to add `python` pointing to sys.executable in PATH (#1879)

  * Noetic / Python 3 specific change

* use setuptools instead of distutils #1870

  * Python 3 related, requiring changes in other packages

* fix missing boost dependencies (#1895)

  * Not need / undesired in Melodic

* Add timestamp formatting for rosconsole (#1892)

  * New feature,

* export missing Boost dependency (#1898)

  * Not need / undesired in Melodic

* remove Boost version check since Noetic only targets platforms with 1.67+ (#1903)

  * Not applicable to Melodic

* Restrict boost dependencies to components used (#1871)

  * Not need / undesired in Melodic

PR #1931, opened 2020-04-14
Added const indexer for xmlrpc (#1759)

ebimor:
Cherry-pick of the commit #1759  into the kinetic-devel branch.


dirk-thomas:
I have added the "consider-backport" label to the original pull request. Whenever a new Kinetic release is done all patches from Melodic will be considered and decided which ones are being backported. See e.g. #1917.

With the amount of changes it is impracticable to backport individual changes on their own since e.g. the order or changes might make a big difference. Therefore I am going to close this individual backport.

Issue #111, opened 2013-01-09
rospy: rospy.Publisher.unregister does not work  (ros ticket #3900)


tfoote:
Testing on latest fuerte. When Publisher.unregister() is called, the following error starts popping up periodically (from another thread):

{{{
[WARN] [WallTime: 1332168660.580924] Could not process inbound connection: ...
}}}

This makes it difficult to change the publisher of the same topic...

trac data:
- Owner: **kwc**
- Reporter: **rdiankov**
- Reported at: **Mon Mar 19 07:54:54 2012**
- URL: https://code.ros.org/trac/ros/ticket/3900

tfoote:
[kwc] in r16546 I change the logger level to an internal logger that should be less noisy.  The actual error message appears to be edited in your ticket, but my guess is that this is usually caused by a subscriber continuing to attempt to connect to a publisher (with stale registration information).

in r16547 I changed a potential stale reconnection logic that could cause the above.

tfoote:
[rdiankov] i'm not sure if it is normal to have that message, but even after putting in changes r16547, the message still pops up.

What i'm doing is unregistering the goal/cancel publishers of an actionlib ActionClient.

tfoote:
[kwc] per comment, you need r16546

tfoote:
[rdiankov] right, that only changes whether the messages shows on the screen or not. from your response, i thought these stale connections shouldn't be happening at all, but if they are a normal part of the operation then i have no complaints.

tfoote:
[kwc] From the perspective of the publisher, the message is an expected race condition, which is why the patch mostly addresses visibility.  The other patch addresses one possible way in which the message could be raised, but there are others as well

If it happens periodically, then it means something, then something is happening on the subscriber side.  There's not enough information in this ticket to debug that side of the issue, so I am marking 'fixed'.  If you wish to open another ticket with something that can reproduce this, happy to look into it more.

tfoote:
[rdiankov] i'm attaching example scripts that reproduces the stale connections.  To run it do:

{{{
roscore
python fibonacci_server.py
python testclient2.py
}}}

The stale connection junk should be printed on the client side:

{{{
[...]

PR #1933, opened 2020-04-15
Avoid warning messages when a publisher unregisters

lucasw:
This doesn't fix the underlying problem of unregister from #111, just hides it.

I assume that unregistering a publisher works sufficiently well that it is safe to do (and re-registering and unregistering any amount of times), that the only downside to unregistering beyond the warning message is the lingering memory and processing overhead that leads to the warning message- clearly state is not returned to the same as where the publisher never existed in the first place.

Maybe some additional logic could be introduced that could distinguish between a 'real' TransportInitError (caused by network issues?) and the unregister case?  The `if not rospy.core.is_shutdown_requested():` line  (https://github.com/ros/ros_comm/blob/melodic-devel/clients/rospy/src/rospy/impl/tcpros_base.py#L342-L347) is already handling node shutdown differently, maybe something else could happen if unregistration can be detected.


dirk-thomas:

dirk-thomas:
> Maybe some additional logic could be introduced that could distinguish between a 'real' TransportInitError (caused by network issues?) and the unregister case?

Without that separation I don't think it is a good idea to demote the message from `warning` to `debug`.

dirk-thomas:
@lucasw This seems to need a rebase / resolve conflicts to be mergable and not show unrelated commits.

dirk-thomas:
I will go ahead and close this for now due to no further responses. Please feel free to comment on the closed ticket once the feedback has been addressed and the ticket can be reopened.

Issue #1939, opened 2020-04-27
protect drop_signal_ with drop_mutex_


fujitatomoya:
### Problem Observed

- stack trace

```
gsignal
_gnu_cxx::_verbose_terminate_handler()
std::rethrow_exception(std::__exception_ptr::exception_ptr)
std::terminate()
__cxa_throw
void boost::throw_exception<boost::bad_weak_ptr>(boost::bad_weak_ptr const&)
ros::TransportSubscriberLink::onConnectionDropped(boost::shared_ptr<ros::Connection> const&)
boost::signals2::detail::signal_impl<void (boost::shared_ptr<ros::Connection> const&, ros::Connection::DropReason), boost::signals2::optional_last_value<void>, int, std::less<int>, boost::function<void (boost::shared_ptr<ros::Connection> const&, ros::Connection::DropReason)>, boost::function<void (boost::signals2::connection const&, boost::shared_ptr<ros::Connection> const&, ros::Connection::DropReason)>, boost::signals2::mutex>::operator()(boost::shared_ptr<ros::Connection> const&, ros::Connection::DropReason)
ros::Connection::drop(ros::Connection::DropReason)
ros::TransportUDSStream::close()
ros::TransportUDSStream::socketUpdate(int)
ros::PollSet::update(int)
ros::PollManager::threadFunc()
```

### Code Analysis

https://github.com/ros/ros_comm/blob/d82f3f09f2b9b55cbbee68ba25f36174ec2c7002/clients/roscpp/src/libros/transport_subscriber_link.cpp#L122-L135

there is a possibility that TransportSubscriberLink could be deleted already via Publication::dropAllConnections.

https://github.com/ros/ros_comm/blob/d82f3f09f2b9b55cbbee68ba25f36174ec2c7002/clients/roscpp/src/libros/publication.cpp#L306-L323

### Suggested Fix

- Protect drop_signal_ with drop_mutex_, so if TransportLink is already deleted, drop_signal_ is also disconnected.

Barry-Xu-2018:
@fujitatomoya 

I submitted a commit to fix this issue. #1940

fujitatomoya:
@dirk-thomas @Barry-Xu-2018 

closing this, since PR https://github.com/ros/ros_comm/pull/1940 is merged.

PR #1940, opened 2020-04-27
Fix bug that connection drop signal related funtion throw a bad_weak …

Barry-Xu-2018:
fix exception which is mentioned in #1939

TransportSubscriberLink::onConnectionDropped throws a bad_weak exception
while call shared_from_this() if TransportSubscriberLink already destroyed

### DIFF
--- a/clients/roscpp/src/libros/connection.cpp
+++ b/clients/roscpp/src/libros/connection.cpp
@@ -93,9 +93,7 @@ boost::signals2::connection Connection::addDropListener(const DropFunc& slot)
 void Connection::removeDropListener(const boost::signals2::connection& c)
 {
   boost::recursive_mutex::scoped_lock lock(drop_mutex_);
-  if (c.connected()) {
-    c.disconnect();
-  }
+  c.disconnect();
 }
 
 void Connection::onReadable(const TransportPtr& transport)
--- a/clients/roscpp/src/libros/connection.cpp
+++ b/clients/roscpp/src/libros/connection.cpp
@@ -93,7 +93,9 @@ boost::signals2::connection Connection::addDropListener(const DropFunc& slot)
 void Connection::removeDropListener(const boost::signals2::connection& c)
 {
   boost::recursive_mutex::scoped_lock lock(drop_mutex_);
-  c.disconnect();
+  if (c.connected()) {
+    c.disconnect();
+  }
 }
 
 void Connection::onReadable(const TransportPtr& transport)
@@ -337,7 +339,10 @@ void Connection::drop(DropReason reason)
 
   if (did_drop)
   {
-    drop_signal_(shared_from_this(), reason);
+    {
+      boost::recursive_mutex::scoped_lock lock(drop_mutex_);
+      drop_signal_(shared_from_this(), reason);
+    }
     transport_->close();
   }
 }
--- a/clients/roscpp/src/libros/service_server_link.cpp
+++ b/clients/roscpp/src/libros/service_server_link.cpp
@@ -340,14 +340,14 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
 
   bool immediate = false;
   {
-    boost::mutex::scoped_lock lock(call_queue_mutex_);
-
     if (connection_->isDropped())
     {
       ROSCPP_LOG_DEBUG("ServiceServerLink::call called on dropped connection for service [%s]", service_name_.c_str());
       info->call_finished_ = true;
       return false;
     }
+
+    boost::mutex::scoped_lock lock(call_queue_mutex_);
 
     if (call_queue_.empty() && header_written_ && header_read_)
     {
--- a/clients/roscpp/src/libros/transport_subscriber_link.cpp
+++ b/clients/roscpp/src/libros/transport_subscriber_link.cpp
@@ -52,6 +52,7 @@ TransportSubscriberLink::TransportSubscriberLink()
 TransportSubscriberLink::~TransportSubscriberLink()
 {
   drop();
+  connection_->removeDropListener(dropped_conn_);
 }
 
 bool TransportSubscriberLink::initialize(const ConnectionPtr& connection)
--- a/clients/roscpp/include/ros/transport_publisher_link.h
+++ b/clients/roscpp/include/ros/transport_publisher_link.h
@@ -32,6 +32,8 @@
 #include "publisher_link.h"
 #include "connection.h"
 
+#include <boost/signals2/connection.hpp>
+
 namespace ros
 {
 class Header;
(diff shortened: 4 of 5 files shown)


fujitatomoya:
@Barry-Xu-2018 

could you check the test failure?

http://build.ros.org/job/Npr__ros_comm__ubuntu_focal_amd64/226/testReport/junit/rostest.runner/RosTest/testservice_call_zombie/
http://build.ros.org/job/Npr_db__ros_comm__debian_buster_amd64/226/testReport/junit/rostest.runner/RosTest/testservice_call_zombie/

Barry-Xu-2018:
> @Barry-Xu-2018
> 
> could you check the test failure?
> 
> http://build.ros.org/job/Npr__ros_comm__ubuntu_focal_amd64/226/testReport/junit/rostest.runner/RosTest/testservice_call_zombie/
> http://build.ros.org/job/Npr_db__ros_comm__debian_buster_amd64/226/testReport/junit/rostest.runner/RosTest/testservice_call_zombie/

I am checking this failure.

fujitatomoya:
@ros-pull-request-builder

Barry-Xu-2018:
FYI  
In my local environment, I cannot reproduce failure case `RosTest.testpubsub_n_fast_udp`.

fujitatomoya (approved):
LGTM, but somebody else needs to check.

fujitatomoya:
@mjcarroll 

would you take a look at this?

fujitatomoya:
@mjcarroll 

friendly reminder.

fujitatomoya:
we've confirmed 3 days our production test pass without error.

mjcarroll:

clients/roscpp/src/libros/connection.cpp
```diff
@@ -337,7 +339,10 @@ void Connection::drop(DropReason reason)
 
   if (did_drop)
   {
-    drop_signal_(shared_from_this(), reason);
```
dirk-thomas, line 340: Should this line better be moved up into the `if (!dropped_)` block to avoid the mutex being locked twice?

[I would assume there was a reason why the call was originally not within the locked mutex scope and am a bit concerned this could lead to a regression.]

clients/roscpp/src/libros/connection.cpp
```diff
@@ -93,7 +93,9 @@ boost::signals2::connection Connection::addDropListener(const DropFunc& slot)
 void Connection::removeDropListener(const boost::signals2::connection& c)
 {
   boost::recursive_mutex::scoped_lock lock(drop_mutex_);
-  c.disconnect();
+  if (c.connected()) {
```
dirk-thomas: This check doesn't seem to be necessary? The `disconnect` method already checks for this.

clients/roscpp/src/libros/transport_subscriber_link.cpp
```diff
@@ -52,6 +52,7 @@ TransportSubscriberLink::TransportSubscriberLink()
 TransportSubscriberLink::~TransportSubscriberLink()
 {
   drop();
+  connection_->removeDropListener(dropped_conn_);
```
dirk-thomas, line 55: Isn't the `drop()` method already handling this? https://github.com/ros/ros_comm/blob/35c0a105c6d7ed6601f54487462b08743e84b8c2/clients/roscpp/src/libros/transport_subscriber_link.cpp#L237-L244

clients/roscpp/src/libros/service_server_link.cpp
```diff
@@ -340,15 +340,15 @@ bool ServiceServerLink::call(const SerializedMessage& req, SerializedMessage& re
...
-    boost::mutex::scoped_lock lock(call_queue_mutex_);
-
     if (connection_->isDropped())
     {
       ROSCPP_LOG_DEBUG("ServiceServerLink::call called on dropped connection for service [%s]", service_name_.c_str());
       info->call_finished_ = true;
       return false;
     }
 
+    boost::mutex::scoped_lock lock(call_queue_mutex_);
```
dirk-thomas, line 350: How is this change related to the title of the PR?

clients/roscpp/src/libros/connection.cpp
Barry-Xu-2018: Yes. I will update codes.

clients/roscpp/src/libros/connection.cpp
Barry-Xu-2018: > Should this line better be moved up into the if (!dropped_) block to avoid the mutex being locked twice?

If there is no PR https://github.com/ros/ros_comm/pull/1950, could use only one lock and move the drop_signal_ and close() up.
So it seems that it needs lock twice here, the modification will be as follows,

```
void Connection::drop(DropReason reason)

  if (did_drop)
  {
    transport_->close();
    {
      boost::recursive_mutex::scoped_lock lock(drop_mutex_);
      drop_signal_(shared_from_this(), reason);
    }
  }

```

moving the transport->close() before the first lock might be work, but I am not sure if there is a restriction.

Barry-Xu-2018: > [I would assume there was a reason why the call was originally not within the locked mutex scope and am a bit concerned this could lead to a regression.]

Think so too. Not very clear about the original thought. So keep the original logic here but to add lock if necessary.

clients/roscpp/src/libros/service_server_link.cpp
Barry-Xu-2018: > How is this change related to the title of the PR?

After using lock before "drop_signal_(shared_from_this(), reason);", there will be a deadlock in two threads

```
thread1:
    ros::ServiceServerLink::call
        boost::mutex::scoped_lock lock(call_queue_mutex_);            # lock A

        ros::Connection::isDropped

            boost::recursive_mutex::scoped_lock lock(drop_mutex_);    # lock B
       
thread2:

    ros::Connection::drop
        boost::recursive_mutex::scoped_lock lock(drop_mutex_);        # lock B
       
        ros::ServiceServerLink::onConnectionDropped
            boost::mutex::scoped_lock lock(call_queue_mutex_);        # lock A
```

clients/roscpp/src/libros/transport_subscriber_link.cpp
Barry-Xu-2018: TransportSubscriberLink::~TransportSubscriberLink() -> drop() -> to call connection_->drop(Connection::Destructing);  
This path will not call "connection_->removeDropListener"

Barry-Xu-2018:
After checking failure of  `Npr_db__ros_comm__debian_buster_amd64` (build 253),  I find below 2 failure cases are the same as http://build.ros.org/job/Npr_db__ros_comm__debian_buster_amd64/251/. So the cause of failure should not be related to this PR.
```
rostest.runner.RosTest.teston_shutdown
__main__.TestOnShutdown.test_notify
```

dirk-thomas:

dirk-thomas:
> So the cause of failure should not be related to this PR.

Those tests are unfortunately flaky. I retriggered the PR builds and both passed.

dirk-thomas:
Thanks for the patch.

fujitatomoya:
@dirk-thomas 

appreciate for the quick support!

PR #1943, opened 2020-04-29
Fix throttle publish first message

efernandez:
The `throttle` node from `topic_tools` fails to publish the first message because it doesn't wait for the subscribers.

Technically, it doesn't have to wait for the subscribers, but it should at least wait some time for them. I've added a `1.0s` wait time, which seems reasonable to me, but I'm happy to make that a param if that makes more sense. I'm open to discuss that part of this change.

The use case is illustrated in a new `rostest`, that basically tries to only publish the first message, by using a very low frequency with `throttle messages`.

### DIFF
--- a/tools/topic_tools/src/throttle.cpp
+++ b/tools/topic_tools/src/throttle.cpp
@@ -32,11 +32,15 @@
 // message stream density, etc., rather than just being greedy and stuffing
 // the output as fast as it can. 
 
+#include <algorithm>
 #include <cstdio>
 #include <cstdlib>
 #include <deque>
+#include <ostream>
+#include <set>
 #include "topic_tools/shape_shifter.h"
 #include "topic_tools/parse.h"
+#include <XmlRpcValue.h>
 
 using std::string;
 using std::vector;
@@ -60,6 +64,9 @@ bool g_lazy;
 bool g_force_latch = false;
 bool g_force_latch_value = true;
 ros::TransportHints g_th;
+ros::WallDuration g_wait_for_subscribers_timeout{ 1.0 };
+ros::WallDuration g_wait_for_subscribers_connect_time{ 0.0 };
+vector<string> g_wait_for_subscribers;
 
 class Sent
 {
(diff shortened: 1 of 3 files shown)


efernandez:
I've just sent another commit that improve the wait for subscribers logic. The previous approach was only able to wait for a single subscriber, that could be anyone. The new one allows to wait for specific subscribers provided in a ROS param list: https://github.com/ros/ros_comm/pull/1943/commits/ac841a3e2024a558b385b3c5d5fb104b4745c8cb

tools/topic_tools/src/throttle.cpp
```diff
@@ -58,6 +62,9 @@ ros::Time g_last_time;
 bool g_use_wallclock;
 bool g_lazy;
 ros::TransportHints g_th;
+ros::WallDuration g_wait_for_subscribers_timeout{ 1.0 };
+ros::WallDuration g_wait_for_subscribers_connect_time{ 1.0 };
```
dirk-thomas: Neither of these two default values seems to replicate the previous behavior. In favor of not changing existing behavior I would argue the defaults should be zero.

tools/topic_tools/src/throttle.cpp
```diff
@@ -58,6 +62,9 @@ ros::Time g_last_time;
...
+vector<string> g_wait_for_subscribers = { "*" };
```
dirk-thomas: Why not use an empty list as the default?

tools/topic_tools/src/throttle.cpp
efernandez: I believe I only have to set `g_wait_for_subscribers_connect_time` to `0.0`.

The default value for `g_wait_for_subscribers_timeout` won't matter if I default `g_wait_for_subscribers` to empty.

tools/topic_tools/src/throttle.cpp
efernandez: I was actually changing the previous behaviour under the assumption the new behaviour was better. That was the reason.

I've changed it. Actually all the defaults, so now the behaviour of this change should match the previous one.

dirk-thomas:

dirk-thomas:
@efernandez Please check the failing test which seems to be related to the patch.

efernandez:
Rebased and fixed conflicts

dirk-thomas:
> Please check the failing test which seems to be related to the patch.

This comment is still pending.

dirk-thomas:
Closing for now due to no further replies. Please feel free to comment on the closed ticket once the feedback has been addressed and the ticket can be reopened.

efernandez:
@dirk-thomas Sorry for not addressing the broken test sooner. I just updated the branch and openned https://github.com/ros/ros_comm/pull/2088

PR #1944, opened 2020-04-29
Add latch param to throttle

efernandez:
This allows to force the advertised throttled publisher to be latched or not, instead of using the same **latching** property value as the input topic.

It defaults to the previous behaviour, i.e. it uses the same **latching** property value as the input topic.

:point_right: Disable white space changes to get a better diff: https://github.com/ros/ros_comm/pull/1944/files?diff=unified&w=1

### DIFF
--- a/tools/topic_tools/src/throttle.cpp
+++ b/tools/topic_tools/src/throttle.cpp
@@ -57,6 +57,8 @@ bool g_use_messages;
 ros::Time g_last_time;
 bool g_use_wallclock;
 bool g_lazy;
+bool g_force_latch = false;
+bool g_force_latch_value = true;
 ros::TransportHints g_th;
 
 class Sent
@@ -87,6 +89,21 @@ void conn_cb(const ros::SingleSubscriberPublisher&)
   }
 }
 
+bool is_latching(const boost::shared_ptr<const ros::M_string>& connection_header)
+{
+  if (connection_header)
+  {
+    ros::M_string::const_iterator it = connection_header->find("latching");
+    if ((it != connection_header->end()) && (it->second == "1"))
+    {
+      ROS_DEBUG("input topic is latched; latching output topic to match");
+      return true;
+    }
+  }
+
+  return false;
+}
+
 void in_cb(const ros::MessageEvent<ShapeShifter>& msg_event)
 {
   boost::shared_ptr<ShapeShifter const> const &msg = msg_event.getConstMessage();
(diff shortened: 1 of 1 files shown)


tools/topic_tools/src/throttle.cpp
```diff
@@ -57,6 +57,7 @@ bool g_use_messages;
 ros::Time g_last_time;
 bool g_use_wallclock;
 bool g_lazy;
+int g_latch = -1;
```
dirk-thomas: Please use a boolean parameter for this and rename it to `force_latch`.

efernandez: I actually need to distinguish between three values:
1. Make the publisher `latching` property the same as the subscriber
2. Make the publisher latched, regardless of the subscriber `latching` property
3. Make the publisher non-latched, regardless of the subscriber `latching` property

So I can't use a `bool force_latch` only. If you prefer, I can have two variables:
1. `bool force_latch;` : Whether to force the user given `latch` value, or use the subscriber one
2. `bool latch` : The desired `latching` value when forcing it

This means I'd have to read two ROS params, which is fine. They'd default to `false`, to keep the current behaviour by default.

Regarding the name, I was trying to follow the `g_` pattern used for the other global variables, but I'm happy to get rid of the global variables and create a class. But I'd do that in another PR. Thoughts? :smiley:

dirk-thomas: > So I can't use a `bool force_latch` only.

Why not? If the parameter is not set the behavior stays as it was before - mimic the property of the subscriber.

If the parameter is set it overrides the property with the provided boolean.

dirk-thomas:
@efernandez Friendly ping.

tools/topic_tools/src/throttle.cpp
efernandez: You can do that for the ROS parameter:
1. If the ROS parameters isn't set: previous behaviour, i.e. output topic latching == input topic latching
2. If the ROS parameter is set to `false`, output topic latching == false
3. If the ROS parameter is set to `true`, output topic latching == true

So we can use a single `force_latch` ROS parameter.

That's ok, but, as I said, I need a copule of `bool` variables to handle those three states.

I've done that. I think we weren't talking about the same thing.

If tested things out for the 6 possible combinations:

| Force Latch? | Input Topic Latch? | Output Topic Latch? |
|--------------|--------------------|---------------------|
| True         | True               | True                |
| True         | False              | True                |
| False        | True               | False               |
| False        | False              | False               |
| Not set      | True               | True                |
| Not set      | False              | False               |

I used:
1. `rostopic pub -r 1 /test std_msgs/Float32 1` for a non-latched input topic
2. `rostopic pub -l -r 1 /test std_msgs/Float32 1` for a non-latched input topic

And:
1. `rosrun topic_tools throttle messages /test 0.1 _force_latch:=true` to force the output topic to be latched
2. `rosrun topic_tools throttle messages /test 0.1 _force_latch:=false` to force the output topic to be non-latched
3. `rosrun topic_tools throttle messages /test 0.1` to set the output topic latching the same as the input topic one (previous bahaviour)

efernandez:
@dirk-thomas Sorry for the delay. I just addressed all your comments here and in #1943

dirk-thomas:
Thanks for the enhancement.

Issue #1351, opened 2018-03-16
rospy can only support one subscriber type per topic


tfoote:
If I create two subscribers for a single topic with different types. The 2nd callback gets the type of the first registered callback. 

This is a refiling of: https://github.com/ros/geometry2/issues/288

Reproduction script
```
#!/usr/bin/env python
import rospy
from tf2_msgs.msg import TFMessage
import tf2_ros

def callback(data):
    rospy.loginfo(rospy.get_caller_id() + "I heard something...AnyThing?")
    print("Type of data is", type(data))


def typed_callback(data):
    who = data._connection_header.get('callerid', "default_authority")
    print("Type of data is", type(data))
    for transform in data.transforms:
        print("Iterating data transforms", transform)


def topic_subscriber():
    sub2 = rospy.Subscriber("/tf", TFMessage, typed_callback)
    sub1 = rospy.Subscriber("/tf", rospy.AnyMsg, callback)


if __name__ == '__main__':
    rospy.init_node('test_callback_types', anonymous=True)
    topic_subscriber()
    rospy.spin()
```

To see the errors run the above with this data source

```
$ rostopic pub /tf tf2_msgs/TFMessage "transforms:
- header:
    seq: 0
    stamp:
      secs: 0
      nsecs: 0
    frame_id: ''
  child_frame_id: ''
  transform:
    translation:
      x: 0.0
      y: 0.0
      z: 0.0
    rotation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 0.0"  -r 10
```

Depending on the order of creating the Subscribers you will get both callback in either `AnyMsg` or `TFMessage` regardless of what the prototype you subscribed with.

```
[INFO] [1521221823.348291]: /test_callback_types_7127_1521221822176I heard something...AnyThing?
('Type of data is', <class 'tf2_msgs.msg._TFMessage.TFMessage'>)
('Type of data is', <class 'tf2_msgs.msg._TFMessage.TFMessage'>)
('Iterating data transforms', header: 
  seq: 0
  stamp: 
    secs: 0
    nsecs:         0
  frame_id: ''
child_frame_id: ''
transform: 
  translation: 
    x: 0.0
    y: 0.0
    z: 0.0
  rotation: 
    x: 0.0
    y: 0.0
    z: 0.0
    w: 0.0)
[INFO] [1521221823.448275]: /test_callback_types_7127_1521221822176I heard something...AnyThing?
[...]

PR #1947, opened 2020-05-07
Warn if rospy clients subscribe more than once to a topic.

MichaelGrupp:
See #1351 - as long as this is not supported there should at least be a warning IMO,
because the following behavior is not obvious at all.

Is it fine to use `getLogger('rosout')` in this context? (worked fine in my tests and showed the warning visible for the user).

### DIFF
--- a/clients/rospy/src/rospy/topics.py
+++ b/clients/rospy/src/rospy/topics.py
@@ -97,6 +97,7 @@ from rospy.impl.tcpros import get_tcpros_handler, DEFAULT_BUFF_SIZE
 from rospy.impl.tcpros_pubsub import QueuedConnection
 
 _logger = logging.getLogger('rospy.topics')
+_rosout_logger = logging.getLogger('rosout')
 
 # wrap genpy implementation and map it to rospy namespace
 import genpy
@@ -1266,7 +1267,11 @@ class _TopicManager(object):
         else:
             raise TypeError("invalid reg_type: %s"%reg_type)
         with self.lock:
-            impl = rmap.get(resolved_name, None)            
+            impl = rmap.get(resolved_name, None)
+            if impl and reg_type == Registration.SUB:
+                _rosout_logger.warning("more than one subscription to %s - "
+                    "subscribing twice to the same topic in one process is not "
+                    "supported properly, see: https://github.com/ros/ros_comm/issues/1351" % resolved_name)
             if not impl:
                 impl = impl_class(resolved_name, data_class)
                 self._add(impl, rmap, reg_type)


clients/rospy/src/rospy/topics.py
```diff
@@ -97,6 +97,7 @@ def isstring(s):
 from rospy.impl.tcpros_pubsub import QueuedConnection
 
 _logger = logging.getLogger('rospy.topics')
+_rosout_logger = logging.getLogger('rosout')
```
fujitatomoya, line 100: I do not understand why it needs to change logger name. I believe that it should stick with "rospy.topics".

fujitatomoya (changes requested):

clients/rospy/src/rospy/topics.py
MichaelGrupp: It's not changing it, but referencing the `rosout` logger instance as an additional one.

The `rospy.topics` logger seems to be not logging visibly by default for normal users, so I added the rosout logger to make sure this warning ends up clearly visible for the user. It works, but as I said I don't know about the desired logger hierarchy in this package.

clients/rospy/src/rospy/topics.py
```diff
@@ -1266,7 +1267,11 @@ def acquire_impl(self, reg_type, resolved_name, data_class):
         else:
             raise TypeError("invalid reg_type: %s"%reg_type)
         with self.lock:
-            impl = rmap.get(resolved_name, None)            
+            impl = rmap.get(resolved_name, None)
+            if impl and reg_type == Registration.SUB:
+                _rosout_logger.warning("more than one subscription to %s - "
+                    "subscribing twice to the same topic in one process is not "
```
MichaelGrupp, line 1273: TODO: only if requested types are different (see #1351), e.g. AnyMsg and LaserScan

clients/rospy/src/rospy/topics.py
fujitatomoya: IMO, since this message related to rospy.topics (Not rosout), i think it is better to be logger("rospy.topics").  if rospy.topics warning does not work, maybe we would make issue against it?

MichaelGrupp: Hmm... on second thought, I think this should actually be an exception rather than a warning. Because the program won't behave as expected by the user, so it won't be very useful to just warn.

>  if rospy.topics warning does not work, maybe we would make issue against it?

It works, but it seems to be unrelated to the standard rosout logger. This is fine if the `rospy.topics` logger is just intended for internal logging and not for end users.

dirk-thomas:

Issue #1949, opened 2020-05-11
assertion fail happens via boost::recursive_mutex when Connection drops.


fujitatomoya:
## Core Stack

```
gsignal+0x38/0x78 (libc-2.23.so +0x31528)
uselocale+0x2f4/0x68 (libc-2.23.so +0x2ac04)
__assert_fail+0x4c/0x4c (libc-2.23.so +0x2acac)
_ZN5boost15recursive_mutex4lockEv+0x78/0xa0
_ZN5boost11unique_lockINS_15recursive_mutexEE4lockEv+0x13c/0x230
_ZN3ros10Connection4dropENS0_10DropReasonE+0x84/0x2b4 (libroscpp.so +0xa1554)
```

### Crash Site
https://github.com/ros/ros_comm/blob/c7f7598e8a49b1b4eb6f3e42f698147e151a811d/clients/roscpp/src/libros/connection.cpp#L330

scoped_lock is using typedef boost::unique_lock, assertion with unique_lock is either no mutex or already owns the mutex in the same thread. in this case, expecting the former, which means Connection object owns this drop_mutex_ is gone at this time.

### Expected Scenario

1. Publisher::Impl::unadvertise (thread-A) Publication::dropAllConnections() here it will call ConnectionManager::onConnectionDropped to push the connection in disconnected list. during deleting TransportSubscriberLink list, it gets context switch. (TransportSubscriberLink also possess the shared pointer to connection.)
1. PollManager::threadFunc (thread-B) ConnectionManager::removeDroppedConnections() disconnected list is swapped, so deleting the shared pointers which belonged to ConnectionManager. (at this point, connection objects are still alive, because TransportSubscriberLink still has reference.)
1. PollManager::threadFunc (thread-B) Connection::drop(), at the entrance it gets context switched.
1. thread-A resumes, delete Connection::drop completely. this means reference to connection from TransportSubscriberLink is now gone. Connection object gone.
1. thread-B resumes, accessing Connection object members which are already freed. boom!!!!

Barry-Xu-2018:
Use the following sequences could cause an issue that use a destroyed connection object.
 
```
# thread A
 
Publisher::Impl::unadvertise
    TopicManager::instance()->unadvertise
        Publication::drop
            Publication::dropAllConnections
                TransportSubscriberLink::drop
                    Connection::drop
                        drop_signal_
                            ConnectionManager::onConnectionDropped           # step1: push conn into dropped_connections_
 
[...]

PR #1950, opened 2020-05-11
Fix a bug that using a destroyed connection object.

Barry-Xu-2018:
there is a case that accessing a connection object which was destroyed by
ConnectionManager and TransportSubscriberLink/TransportPublicationLink
while getting a close event in pollset to call Connection::drop.

Signed-off-by: Barry Xu <Barry.Xu@sony.com>

### DIFF
--- a/clients/roscpp/src/libros/connection.cpp
+++ b/clients/roscpp/src/libros/connection.cpp
@@ -337,11 +337,11 @@ void Connection::drop(DropReason reason)
 
   if (did_drop)
   {
+    transport_->close();
     {
       boost::recursive_mutex::scoped_lock lock(drop_mutex_);
       drop_signal_(shared_from_this(), reason);
     }
-    transport_->close();
   }
 }
 


Barry-Xu-2018:
This fixing is related to #1949.

fujitatomoya (approved):
LGTM, but requesting others to review.

fujitatomoya:
@dirk-thomas @mjcarroll

would you take a look at this?

mjcarroll (approved):
Looks reasonable to me. Thanks for the contribution.

dirk-thomas:
This change seems to be closely related to #1940. It doesn't seem that *both* patches should be applied. Please clarify.

dirk-thomas:
With #1940 merged this needs to be rebased. Please still clarify why this is (still) necessary.

Barry-Xu-2018:
@dirk-thomas 

> With #1940 merged this needs to be rebased. Please still clarify why this is (still) necessary.

Rebase action is done. 

#1940 is to fix that calling drop_signal_() might access a TransportSubscriberLink/TransportPublicationLink which is already destroyed.  
#1950 is to fix that calling Connection::drop() might access a Connection which is already destroyed

dirk-thomas:
Thanks for the patch as well as iterating on it.

venkisagunner93:
Just dropping by. I saw that the backporting is complete. If possible can someone tell me when the package will be available as a debian? I searched http://packages.ros.org/ros/ubuntu/pool/main/r/ros-melodic-ros-comm/ and I still see 1.14.5 not 1.14.6. Just wondering about the process and timeline as to when a pull request is made, how long until the package gets debianized? Trying to be understand the release cycles. @dirk-thomas

dirk-thomas:
Once a new version is tagged in the upstream repo it is released using `bloom`. The buildfarm will create new Debian packages shortly after that. Once all Debian packages for a ROS distro have been rebuild they are available in the `testing` repo (which you can use if you want to get new versions earlier). And every few weeks a sync is being made to the `main` repo most users use. That sync is announced on Discourse, e.g. https://discourse.ros.org/t/new-packages-for-melodic-2020-06-12/14640, which is also the annoucement mentioning the new version you are looking for.

venkisagunner93:
@dirk-thomas Thank you. Appreciate it !

PR #1952, opened 2020-05-14
Fix bag migration failures caused by typo in connection_header assignment

TSchildhauer:
Bag migration failed with following error:
NameError: global name 'connection_header' is not defined
This change fixes that by using local var conn_header


dirk-thomas:
Thanks for the fix. The regression was introduced in https://github.com/ros/ros_comm/pull/1708#discussion_r425315598.

PR #1953, opened 2020-05-15
Fix bag migration failures caused by typo in connection_header assignment

TSchildhauer:
Bag migration failed with follwoing errors:
NameError: global name 'connection_header' is not defined
This change fixes that by using local name conn_header

... now also for melodic


dirk-thomas:
Duplicate of #1952. The commit will be cherry-picked from the `noetic-devel` branch to the `melodic-devel` branch together with other tickets. That will happen in a single PR doing so for all the commit *in order* they have been merged to avoid problems with patches which build on top of each other.

PR #1954, opened 2020-05-15
check if async socket connect is success or failure before read and write in TransportT

dirk-thomas:
Reopening #1202 targeting Noetic.


gavanderhoorn:
Thanks for this @dirk-thomas.

I only noticed you'd already done this as my rebase conflicted with just about everything in this PR.

:+1:

PR #1961, opened 2020-05-22
[noetic] Fixing Windows build break.

seanyen:
Fixed Windows build break after https://github.com/ros/ros_comm/pull/1954.

```
error C2664: 'int getsockopt(SOCKET,int,int,char *,int *)': cannot convert argument 4 from 'int *' to 'char *'
```


seanyen:
@dirk-thomas This is ready for review and merge. Thanks.

dirk-thomas:
Thanks for the fix.

PR #1969, opened 2020-06-11
preformance: improve performance of writes

tsijs:
When writing a lot of messages sequentially in a bag, there are a lot of seemingly unnecessary seeks.
By creating a member state last_action with a corresponding enum, writing messages sequentially get a performance boost up to 80%.

Also ran a linter which explains all the whitespace changes.

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -68,17 +68,14 @@ import genpy.message
 
 import roslib.names  # still needed for roslib.names.canonicalize_name()
 import rospy
-
 try:
     import roslz4
-
     found_lz4 = True
 except ImportError:
     rospy.logwarn(
         'Failed to load Python extension for LZ4 support. '
         'LZ4 compression will not be available.')
     found_lz4 = False
-
 
 class RosBagAction(enum.Enum):
     """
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for creating this patch. Please provide reproducible instructions how you measured the performance before and after the change.

> Also ran a linter which explains all the whitespace changes.

Please revert those changes. It simply isn't feasible to have the style improved on one branch (`noetic-devel`) but not on other which patches need to be backported to.

tsijs:
> Thanks for creating this patch. Please provide reproducible instructions how you measured the performance before and after the change

I ran it in line_profiler outside of the ros environment for an example dataset that mattered to me that writes ~500000 messages to a bag without closing it. 

I added two files with a comparison on line-level:
[after.txt](https://github.com/ros/ros_comm/files/4769096/after.txt)
[before.txt](https://github.com/ros/ros_comm/files/4769097/before.txt)

As can be seen the seek line (R:606) is spend in 75% of the time in the before case, or 402202676.0us. The after case this is 561997.0us and 0.4% of the time.

> Please revert those changes. It simply isn't feasible to have the style improved on one branch (noetic-devel) but not on other which patches need to be backported to.

I will revert the style changes tonight.

dirk-thomas:
The results look very good. I was also asking for:

> Please provide reproducible instructions how you measured the performance before and after the change.

tsijs:
>Please provide reproducible instructions how you measured the performance before and after the  change.

I don't fully understand what you mean with this? You want an example script that can be run in line_profiler?

I guess something like https://github.com/ros/ros_comm/blob/noetic-devel/test/test_rosbag/bag_migration_tests/scripts/generate_data_4.py with a for loop around e.g. R66 would work. As I mentioned before, for profiling I used line_profiler. so then running this on both branches would do the trick:
```from test_rosbag.msg import *  
import rosbag

def generate_data():
    bag = rosbag.Bag("test/converged_gen4.bag", "w")
    m = Converged([1.2, 3.4, 5.6, 7.8], [SimpleMigrated(11), SimpleMigrated(22), SimpleMigrated(33), SimpleMigrated(44)])
    for i in range(100000):
        bag.write("converged", m, genpy.Time())
        bag.close()

if __name__ == '__main__':
    generate_data()
```

 with kernprof -l ./example.py would do the trick. 

btw: I also removed the linter modifications.

dirk-thomas:
With the above snippet I just get a: `ValueError: I/O operation on closed bag`. I assume the `bag.close()` should happen outside of the loop?

With that change and measuring the time the loop takes I can only see a small improvement: from 2.5s to 2.2s (~15% faster). Because of the above error I assume you didn't actually run the posted snippet. Can you please do so and share your performance results with that snippet - before and after your patch.

> I also removed the linter modifications.

Nitpick: there are still a couple left. But it was clean enough to get the essence of the patch.

I wonder why the enum is necessary. A simple boolean fields seems sufficient, e.g.:

<details>

<summary>Minimal diff:</summary>

```
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -454,6 +454,7 @@ class Bag(object):
         self._reader          = None
 
         self._file_header_pos = None
+        self._file_pos_is_end = False
         self._index_data_pos  = 0       # (1.2+)
 
         self._clear_index()
@@ -575,6 +576,8 @@ class Bag(object):
         if not self._file:
             raise ValueError('I/O operation on closed bag')
 
+        self._file_pos_is_end = False
+
         if self._chunk_open:
             self._stop_writing_chunk()
 
@@ -603,7 +606,9 @@ class Bag(object):
             t = genpy.Time.from_sec(time.time())
 
         # Seek to end (in case previous operation was a read)
-        self._file.seek(0, os.SEEK_END)
+        if not self._file_pos_is_end:
+            self._file.seek(0, os.SEEK_END)
+            self._file_pos_is_end = True
 
         # Open a chunk, if needed
         if not self._chunk_open:
@@ -708,6 +713,7 @@ class Bag(object):
         Reindexes the bag file.  Yields position of each chunk for progress.
         """
         self._clear_index()
+        self._file_pos_is_end = False
         return self._reader.reindex()
 
     def close(self):
```

</details>

That being said I am worried that this approach doesn't necessarily ensure that the necessary state information (`self._last_action`) is always up-to-date. E.g. there are several other cases where different `_BagReader` seek within the file without updating that state.

dirk-thomas:
@tsijs Are you planning on iterating on this patch?

dirk-thomas:
Closing due to no further replies. Please feel free to comment on the closed ticket and it can be reopened.

PR #1970, opened 2020-06-12
Update __init__.py

gachiemchiep:
Hello guys. 
I think this PR https://github.com/ros/ros_comm/pull/1660 should be applied into kinetic-devel branch too.


dirk-thomas:
Kinetic doesn't officially support Python 3 so there is no need to backport the referenced change.

There are far to many changes to correctly support Python 3 (which is the case for Noetic) to backport them to a distribution released 4 years ago.

PR #1978, opened 2020-06-24
Added const versions of XmlRpcValue converting operators.

peci1:
Followup of https://github.com/ros/ros_comm/pull/1759.


peci1:
It'd also be possible to return copies of the data instead of references when called on the const object. That might make some downstream code a little nicer, but const references seem to me like more versatile (or performance-friendly if ever anybody is interested in the performance of this library).

dirk-thomas:
Thanks for the patch.

peci1:
Thanks. That was lightning fast :) Is this eligible for backporting?

dirk-thomas:
> Is this eligible for backporting?

Sounds reasonable to me. I assumed it is API and ABI compatible so no problem to backport it before the next Melodic patch release.

PR #1982, opened 2020-06-26
Add test_depend on 'rostopic' in 'rostest' package

### LINKED ISSUE #1981: Missing dependency on 'rostopic' in 'rostest' package
`rostest` should exec_depend on `rostopic` because it is used in `publishtest` tool

https://github.com/ros/ros_comm/blob/535999d87238e22346b12fc1c6d7d28dc2a22ae8/tools/rostest/nodes/publishtest#L61

mateus-amarante:
Fix #1981

### DIFF
--- a/tools/rostest/nodes/publishtest
+++ b/tools/rostest/nodes/publishtest
@@ -58,7 +58,6 @@ import time
 import unittest
 
 import rospy
-import rostopic
 
 
 PKG = 'rostest'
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please try to build and verify a proposed patch before creating a pull request. The proposed change fails to even build because it introduces a circular dependencies between these packages.

mateus-amarante:
My bad. I underestimated the impact of such a simple change. I'm sorry for the precipitation.

I think the cyclic dependency problem is not so straightforward to solve. I think it is better to close this PR and maybe reopen it later on (I will test it locally before opening the PR next time).

PR #1983, opened 2020-06-26
Changed is_async_connected to use epoll when available

astere-cpr:
Scalability improvement to remove the 1024 socket limit imposed by select() on Linux

### DIFF
--- a/clients/roscpp/src/libros/io.cpp
+++ b/clients/roscpp/src/libros/io.cpp
@@ -380,6 +380,31 @@ pollfd_vector_ptr poll_sockets(int epfd, socket_pollfd *fds, nfds_t nfds, int ti
  * @return int : 1 on connected, 0 on connecting, -1 on failure.
  */
 int is_async_connected(socket_fd_t &socket, int &err) {
+#ifdef HAVE_EPOLL
+    int nfds = 1;
+    int epfd = create_socket_watcher();
+    add_socket_to_watcher(epfd, socket);
+    set_events_on_socket(epfd, socket, EPOLLIN | EPOLLOUT);
+    struct epoll_event ev[nfds];
+    int fd_cnt = ::epoll_wait(epfd, ev, nfds, 0);
+    close_socket_watcher(epfd);
+    if (fd_cnt < 0) {
+      if (errno != EINTR) {
+        ROS_ERROR("Error in epoll_wait! %s", strerror(errno));
+        err = errno;
+        return -1;
+      }
+    } else if (fd_cnt == 0) {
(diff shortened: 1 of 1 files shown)


clients/roscpp/src/libros/io.cpp
```diff
@@ -380,6 +380,26 @@ pollfd_vector_ptr poll_sockets(int epfd, socket_pollfd *fds, nfds_t nfds, int ti
...
+    if (fd_cnt < 0) {
+      if (errno != EINTR) {
+        ROS_ERROR("Error in epoll_wait! %s", strerror(errno));
+        err = errno;
+        return -1;
+      }
+    } else if (fd_cnt == 0) {
+      return 0;
+    } else if (ev[0].events & EPOLLERR || ev[0].events & EPOLLHUP) {
+      return -1;
```
dirk-thomas, line 405: Shouldn't `err` be set to something here?

clients/roscpp/src/libros/io.cpp
```diff
@@ -380,6 +380,26 @@ pollfd_vector_ptr poll_sockets(int epfd, socket_pollfd *fds, nfds_t nfds, int ti
...
+    int fd_cnt = ::epoll_wait(epfd, ev, nfds, 0);
+    close_socket_watcher(epfd);
+    if (fd_cnt < 0) {
+      if (errno != EINTR) {
+        ROS_ERROR("Error in epoll_wait! %s", strerror(errno));
+        err = errno;
+        return -1;
+      }
+    } else if (fd_cnt == 0) {
+      return 0;
```
dirk-thomas, line 399: The existing code sets `err` to `0` on success in this case. Should this branch do the same?

dirk-thomas (commented):
Beside the inline comments can you please comment on the amount of testing you have done with this patch.

astere-cpr:
> Beside the inline comments can you please comment on the amount of testing you have done with this patch.

The patch has been in continuous use since I submitted the PR on our large fleet testing cluster which runs 24/7. The busiest ROS node on that cluster has around 4K open sockets at any given time.

clients/roscpp/src/libros/io.cpp
astere-cpr: Done

clients/roscpp/src/libros/io.cpp
astere-cpr: I changed it to assign the error on the socket being polled, if any

dirk-thomas:
Thanks for the patch and the quick update.

PR #1986, opened 2020-06-29
rosnode_ping fix for issue #1967

### LINKED ISSUE #1967: [bug] python rosnode_ping failure after node restart
Problem:

bofetadas:
Fixes #1967 
rosnode cleanup now also deletes unavailable nodes from `_caller_apis` cache. This avoids buggy behaviour when nodes exit the ros network due to wifi issues and then come back later.
Programmers also have to possibility to skip that cache when calling `rosnode_ping` or `rosnode_ping_all`.

### DIFF
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -431,7 +431,7 @@ def cleanup_master_blacklist(master, blacklist):
                 service_api = master.lookupService(s)
                 master_n = rosgraph.Master(n)
                 master_n.unregisterService(s, service_api)
-	_caller_apis.pop(n, None)
+        _caller_apis.pop(n, None)
 
 def cleanup_master_whitelist(master, whitelist):
     """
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Please split these changes into two PRs: one for the new `skip_cache` argument and one for the fix to the `_caller_apis` (please include reproducible steps foo the latter).

dirk-thomas:
@mabaue Friendly ping.

bofetadas:
@dirk-thomas 
Sorry for the delay. I was busy and on vacation the last days. Will take care of it this week!

bofetadas:
@dirk-thomas Done. Two seperate PRs have been made

dirk-thomas:
> Two seperate PRs have been made

@mabaue Can you please reference them here and close this PR if it is obsolete.

bofetadas:
This PR was seperated into the following two PRs:
https://github.com/ros/ros_comm/pull/2009
https://github.com/ros/ros_comm/pull/2010

PR #1992, opened 2020-07-08
Add which node has been registered with the same name

cjds:
Was looking through logs of several nodes that had been launched at the same time and it was getting hard to tell which node was being duplicated. This would allow you do to see the name of the node that caused the issue

### DIFF
--- a/tools/rosmaster/src/rosmaster/registrations.py
+++ b/tools/rosmaster/src/rosmaster/registrations.py
@@ -464,7 +464,7 @@ class RegistrationManager(object):
             else:
                 bumped_api = node_ref.api
                 self.thread_pool.queue_task(bumped_api, shutdown_node_task,
-                                            (bumped_api, caller_id, "new node registered with same name"))
+                                            (bumped_api, caller_id, "new node registered with same name {}".format(node_ref.id)))
 
         node_ref = NodeRef(caller_id, caller_api)
         self.nodes[caller_id] = node_ref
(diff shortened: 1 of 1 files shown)


tools/rosmaster/src/rosmaster/registrations.py
```diff
@@ -464,7 +464,7 @@ def _register_node_api(self, caller_id, caller_api):
             else:
                 bumped_api = node_ref.api
                 self.thread_pool.queue_task(bumped_api, shutdown_node_task,
-                                            (bumped_api, caller_id, "new node registered with same name"))
+                                            (bumped_api, caller_id, "new node registered with same name {}".format(node_ref.id)))
```
dirk-thomas: Instead of modifying the `reason` arg how about changing `shutdown()` call in the `shutdown_node_task()` function instead? In that function the `caller_id` could be prepended to the provided reason which makes sure that the node name is shown independent how the shutdown was triggered.

cjds: Yeah I think that would solve the same problem in a better way. I'll make that change

dirk-thomas: @cjds Friendly ping.

cjds: Sorry I left this by the wayside. Made the change to `shutdown_node_task`. It now looks like

```
shutdown request: [/teleop_twist_keyboard] Reason: new node registered with same name
```

dirk-thomas:
Thanks for the improvement.

PR #1993, opened 2020-07-09
Add "param" substitution in tags

tahsinkose:
Noetic-version of https://github.com/ros/ros_comm/pull/1512.

### DIFF
--- a/clients/roscpp/package.xml
+++ b/clients/roscpp/package.xml
@@ -1,6 +1,6 @@
 <package>
   <name>roscpp</name>
-  <version>1.15.8</version>
+  <version>1.15.9</version>
   <description>
     roscpp is a C++ implementation of ROS. It provides
     a <a href="http://wiki.ros.org/Client%20Libraries">client
--- a/clients/rospy/package.xml
+++ b/clients/rospy/package.xml
@@ -4,7 +4,7 @@
   schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>rospy</name>
-  <version>1.15.8</version>
+  <version>1.15.9</version>
   <description>
     rospy is a pure Python client library for ROS. The rospy client
     API enables Python programmers to quickly interface with ROS <a
--- a/ros_comm/package.xml
+++ b/ros_comm/package.xml
@@ -1,6 +1,6 @@
 <package>
   <name>ros_comm</name>
-  <version>1.15.8</version>
+  <version>1.15.9</version>
   <description>
     ROS communications-related packages, including core client libraries (roscpp, rospy) and graph introspection tools (rostopic, rosnode, rosservice, rosparam).
   </description>
--- a/test/test_rosbag/package.xml
+++ b/test/test_rosbag/package.xml
@@ -4,7 +4,7 @@
   schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>test_rosbag</name>
-  <version>1.15.8</version>
+  <version>1.15.9</version>
   <description>
     A package containing the unit tests for rosbag.
   </description>
--- a/test/test_rosbag_storage/package.xml
+++ b/test/test_rosbag_storage/package.xml
@@ -1,6 +1,6 @@
 <package>
   <name>test_rosbag_storage</name>
-  <version>1.15.8</version>
+  <version>1.15.9</version>
   <description>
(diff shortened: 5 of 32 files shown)


tahsinkose:
@dirk-thomas could you have a look in here?

tahsinkose:
Let's merge this @dirk-thomas :slightly_smiling_face:

dirk-thomas:
The current patch still contains unrelated white space changes (as mentioned before https://github.com/ros/ros_comm/pull/1512#issuecomment-437538621). Please revert them to keep the diff minimal.

tahsinkose:
> The current patch still contains unrelated white space changes (as mentioned before [#1512 (comment)](https://github.com/ros/ros_comm/pull/1512#issuecomment-437538621)). Please revert them to keep the diff minimal.

Addressed.

tahsinkose:
I noticed that `node` tag still lacks this feature. Please wait for it too @dirk-thomas. I'll hopefully implement it today  :slightly_smiling_face:

tahsinkose:
Same functionality is implemented for `node`, `include` and `remap` tags and the tests are added. PTAL @dirk-thomas and let's have this finally merged :crossed_fingers:

tahsinkose:
Ready @dirk-thomas.

tools/roslaunch/src/roslaunch/xmlloader.py
```diff
@@ -186,6 +187,9 @@ def resolve_args(self, args, context):
         if args and '$' in args:
             # Populate resolve_dict with name of the current file being processed.
             context.resolve_dict['filename'] = context.filename
+            if 'param' not in context.resolve_dict:
+                context.resolve_dict['param'] = {}
+            context.resolve_dict['param'] = context.params
```
dirk-thomas: This line seems to overwrite the value of the conditional assignment from the line above?

tools/roslaunch/src/roslaunch/xmlloader.py
```diff
@@ -223,6 +227,27 @@ def _check_attrs(self, tag, context, ros_config, attrs):
...
+        params_restore_point = copy.deepcopy(context.params)
+        # Use context to emulate get_param functionality in substitution_args
+        if ros_config.params:
+            if type(ros_config.params) is list:
+                for p in ros_config.params:
+                    context.add_param(p)
+
+        retval = fn(context)
+
+        context.params = copy.deepcopy(params_restore_point)
```
dirk-thomas: Why does this need to b a `deepcopy` instead of a plain assignment?

tools/roslaunch/src/roslaunch/xmlloader.py
dirk-thomas: @tahsinkose Friendly ping.

tools/roslaunch/src/roslaunch/xmlloader.py
dirk-thomas: @tahsinkose Friendly ping.

tools/roslaunch/src/roslaunch/xmlloader.py
tahsinkose: Yes it does. I truely don't remember why this was the case. But after removal tests didn't fail :crossed_fingers:

tools/roslaunch/src/roslaunch/xmlloader.py
tahsinkose: You are right. One `deepcopy` should suffice :+1:

tahsinkose:
Addressed the review and also added invalidity test for missing param substitutions.

@dirk-thomas ping :smile:

peci1:
How does this implementation handle circular references?

```
<param name="a" value="$(arg b)" />
<arg name="b" value="$(param a)" />
```

So far, the parsing process (if I got it correctly) was to first resolve all args and includes, then all params, and then start launching nodes. This definitely had to change in this PR. What's the new process?

tahsinkose:
It doesn't handle circular references. In the snippet you provided, it would error out with something as "no arg b specified". Yes args and includes first, then follows params. 

EDIT: I don't think it should support any circular references at all. With this PR `param`s are made usable in exactly the same way of `arg`s inside launch files. That's it.

peci1:
So is the evaluation done in the order of appearance?

tahsinkose:
> So is the evaluation done in the order of appearance?

Yes

Issue #1496, opened 2018-08-27
help maintaining ros_comm packages


dirk-thomas:
If you would like to help maintaining the packages in this repository (or any other ROS repository) please consider to contribute to any of the following efforts:

* Go through open issues:
  * Check if it is still valid
  * If it is a bug try to reproduce the reported problems
  * Consider creating a pull request to fix the bug / implement the feature request (always [referencing](https://help.github.com/articles/closing-issues-using-keywords/) the original issue)
  * If you don't have the permission (yet) to set labels or close obsolete tickets please just comment with the suggestion and someone will follow up on it.

* Go through open pull requests:
  * Check if it is still valid
    * All pull requests must target the latest development branch which is commonly also the default branch (currently `melodic-devel`), the only exception is if a patch is only applicable to older branches and not necessary on the default branch
  * If the patch already has feedback which hasn't been addressed by the author consider creating a new PR addressing the pending feedback to make the patch mergable (always referencing the original ticket)
  * Review the patch which includes:
    * Check the diff for problems (unnecessary / unrelated changes, incorrect / incomplete logic, API / ABI breakage, code style matching surrounding code)
    * Reproduce the problem without the patch
    * Apply the patch
    * Reproduce that the problem has been addressed and try to make sure that it doesn't introduce any regressions
    * Ensure that it passes the tests

* Once in a while a new patch release is being made, always into the latest ROS distro first
  * After that has been synced and little soak time has passed these changes are being consider to be backported to the previous ROS distro, see #1477 for an example, usually backports are limited to bug fixes, enhancement are only considered when they are "easy" to review / argue that they don't introduce regressions and make life of either users or maintainers in older ROS distro more convenient
  * After e.g. a backport from Melodic to Lunar and a following patch release for Lunar the next step can be considered: backporting from Lunar to Kinetic (see e.g. #1490) and so forth

Please feel free to ask questions on this ticket.

dirk-thomas:
With maintenance passed to @jacobperron, @mjcarroll, and @sloretz I will go ahead and close this ticket.

PR #2003, opened 2020-07-21
XmlRpcValue::_doubleFormat should be used during write.

### LINKED ISSUE #1958: XmlRpcValue ignores the value of `_doubleFormat`
https://github.com/ros/ros_comm/search?q=_doubleformat&unscoped_q=_doubleformat
https://github.com/ros/ros_comm/search?q=getdoubleformat&unscoped_q=getdoubleformat

This value has no use at all.

    XmlRpc::XmlRpcValue::setDoubleFormat("%.2f");
    std::stringstream ss;
    ss << XmlRpc::XmlRpcValue(2.0);
    EXPECT_EQ("2.00", ss.str());

This fails. The output of `ss.str()` is "2".

fujitatomoya:
fix https://github.com/ros/ros_comm/issues/1958

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcValue.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcValue.cpp
@@ -615,16 +615,19 @@ namespace XmlRpc {
           int required_size;
           char buf[128]; // Should be long enough
           required_size = std::snprintf(buf, sizeof(buf)-1,
-                                        getDoubleFormat().c_str(), _value.asDouble);
-          if (required_size < (int) sizeof(buf)) {
+                            getDoubleFormat().c_str(), _value.asDouble);
+          if (required_size >= 0 && required_size < (int) sizeof(buf)) {
             buf[sizeof(buf)-1] = 0;
             os << buf;
-          } else {
+          } else if (required_size >= (int) sizeof(buf)) {
+            int ret;
             char required_buf[required_size+1];
-            std::snprintf(required_buf, required_size,
-                          getDoubleFormat().c_str(), _value.asDouble);
-            required_buf[required_size] = 0;
-            os << required_buf;
+            ret = std::snprintf(required_buf, required_size,
+                    getDoubleFormat().c_str(), _value.asDouble);
+            if (ret == required_size) {
+              required_buf[required_size] = 0;
+              os << required_buf;
+            }
           }
           break;
         }
@@ -632,11 +635,14 @@ namespace XmlRpc {
       case TypeDateTime:
         {
           struct tm* t = _value.asTime;
+          int ret;
           char buf[20];
-          std::snprintf(buf, sizeof(buf)-1, "%4d%02d%02dT%02d:%02d:%02d", 
-            t->tm_year,t->tm_mon,t->tm_mday,t->tm_hour,t->tm_min,t->tm_sec);
-          buf[sizeof(buf)-1] = 0;
-          os << buf;
+          ret = std::snprintf(buf, sizeof(buf)-1, "%4d%02d%02dT%02d:%02d:%02d",
+                  t->tm_year,t->tm_mon,t->tm_mday,t->tm_hour,t->tm_min,t->tm_sec);
+          if (ret > 0) {
+            buf[sizeof(buf)-1] = 0;
+            os << buf;
+          }
           break;
         }
       case TypeBase64:
--- a/utilities/xmlrpcpp/src/XmlRpcValue.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcValue.cpp
@@ -612,10 +612,20 @@ namespace XmlRpc {
       case TypeInt:      os << _value.asInt; break;
       case TypeDouble:
         {
-          char buf[100]; // Should be long enough
-          std::snprintf(buf, sizeof(buf)-1, getDoubleFormat().c_str(), _value.asDouble);
-          buf[sizeof(buf)-1] = 0;
-          os << buf;
+          int required_size;
+          char buf[128]; // Should be long enough
+          required_size = std::snprintf(buf, sizeof(buf)-1,
+                                        getDoubleFormat().c_str(), _value.asDouble);
+          if (required_size < (int) sizeof(buf)) {
+            buf[sizeof(buf)-1] = 0;
+            os << buf;
+          } else {
+            char required_buf[required_size+1];
+            std::snprintf(required_buf, required_size,
+                          getDoubleFormat().c_str(), _value.asDouble);
+            required_buf[required_size] = 0;
+            os << required_buf;
+          }
           break;
         }
       case TypeString:   os << *_value.asString; break;
--- a/utilities/xmlrpcpp/src/XmlRpcValue.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcValue.cpp
@@ -610,7 +610,14 @@ namespace XmlRpc {
       default:           break;
       case TypeBoolean:  os << _value.asBool; break;
       case TypeInt:      os << _value.asInt; break;
-      case TypeDouble:   os << _value.asDouble; break;
+      case TypeDouble:
+        {
+          char buf[100]; // Should be long enough
+          std::snprintf(buf, sizeof(buf)-1, getDoubleFormat().c_str(), _value.asDouble);
+          buf[sizeof(buf)-1] = 0;
+          os << buf;
+          break;
+        }
       case TypeString:   os << *_value.asString; break;
       case TypeDateTime:
         {
--- a/utilities/xmlrpcpp/src/XmlRpcValue.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcValue.cpp
@@ -1,4 +1,5 @@
 
+#include "ros/ros.h"
 #include "xmlrpcpp/XmlRpcValue.h"
 #include "xmlrpcpp/XmlRpcException.h"
 #include "xmlrpcpp/XmlRpcUtil.h"
@@ -14,6 +15,7 @@
 #endif
 
 #include <sstream>
+#include <mutex>
 
 namespace XmlRpc {
 
@@ -612,10 +614,19 @@ namespace XmlRpc {
       case TypeInt:      os << _value.asInt; break;
       case TypeDouble:
         {
-          int required_size;
+          static std::once_flag once;
+          int ret, required_size;
           char buf[128]; // Should be long enough
-          required_size = std::snprintf(buf, sizeof(buf)-1,
-                            getDoubleFormat().c_str(), _value.asDouble);
+          ret = std::snprintf(buf, sizeof(buf)-1,
+                  getDoubleFormat().c_str(), _value.asDouble);
+          if (ret < 0) {
+            std::call_once(once,
+              [](){ROS_WARN("Failed to format with %s, using default %%.16g format",
+                getDoubleFormat().c_str());});
+            required_size = std::snprintf(buf, sizeof(buf)-1, "%.16g", _value.asDouble);
+          } else {
+            required_size = ret;
+          }
           if (required_size >= 0 && required_size < (int) sizeof(buf)) {
             buf[sizeof(buf)-1] = 0;
             os << buf;
@@ -628,6 +639,8 @@ namespace XmlRpc {
               required_buf[required_size] = 0;
               os << required_buf;
             }
+          } else {
+            ROS_ERROR("Unexpected error to format %s", getDoubleFormat().c_str());
           }
           break;
         }
@@ -642,6 +655,8 @@ namespace XmlRpc {
           if (ret > 0) {
             buf[sizeof(buf)-1] = 0;
             os << buf;
+          } else {
+            ROS_ERROR("Unexpected error to format TypeDateTime");
           }
           break;
         }
--- a/utilities/xmlrpcpp/src/XmlRpcValue.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcValue.cpp
@@ -1,5 +1,4 @@
 
-#include "ros/ros.h"
 #include "xmlrpcpp/XmlRpcValue.h"
 #include "xmlrpcpp/XmlRpcException.h"
 #include "xmlrpcpp/XmlRpcUtil.h"
@@ -621,7 +620,7 @@ namespace XmlRpc {
                   getDoubleFormat().c_str(), _value.asDouble);
           if (ret < 0) {
             std::call_once(once,
-              [](){ROS_WARN("Failed to format with %s, using default %%.16g format",
+              [](){XmlRpcUtil::error("Failed to format with %s, using default %%.16g format",
                 getDoubleFormat().c_str());});
             required_size = std::snprintf(buf, sizeof(buf)-1, "%.16g", _value.asDouble);
           } else {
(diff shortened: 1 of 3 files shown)


utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,7 +610,14 @@ namespace XmlRpc {
       default:           break;
       case TypeBoolean:  os << _value.asBool; break;
       case TypeInt:      os << _value.asInt; break;
-      case TypeDouble:   os << _value.asDouble; break;
+      case TypeDouble:
+        {
+          char buf[100]; // Should be long enough
```
dirk-thomas: > Should be long enough

The integer part of a double can be much larger - at least 309 digits. And there can be a `-` sign in front. The problem is that the format string can extend the size with decimal digits, e.g. `%.999f`.

So I am not sure that `100` is a reasonable upper limit. Not sure if we should just use a higher boundary or if we should do this dynamically:
* determine the necessary buffer size,
* allocate that buffer dynamically,
* printf the double value,
* stream it into the out stream
* deallocate the temporary buffer.

peci1: I suggest an opportunistic approach. The [return value of snprintf](http://www.cplusplus.com/reference/cstdio/snprintf/#return) tells the required size of the buffer. So it should be possible to have the 100-byte (or whatever) buffer statically initialized, and if the formatted string fits inside, go on with this static buffer. If the buffer is too small, then dynamic allocation should create the buffer of the size snprintf tells.

This would result in a small slowdown for long strings, but would avoid dynamic allocation for more common cases.

peci1:
This PR should contain a test that would verify the behavior.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
dirk-thomas: That sounds like the best of both :+1:

fujitatomoya:
@dirk-thomas @peci1 

could you take a look at it when you got time?

i confirmed `catkin_make run_tests_xmlrpcpp_gtest_TestValues XmlRpc.testDouble` got green light.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,7 +610,24 @@ namespace XmlRpc {
       default:           break;
       case TypeBoolean:  os << _value.asBool; break;
       case TypeInt:      os << _value.asInt; break;
-      case TypeDouble:   os << _value.asDouble; break;
+      case TypeDouble:
+        {
+          int required_size;
+          char buf[128]; // Should be long enough
+          required_size = std::snprintf(buf, sizeof(buf)-1,
```
dirk-thomas: The return value should be checked for a negative value indicating an error.

Same below.

dirk-thomas (commented):
Beside the inline comment it seems that some unit tests are failing now in the PR builds.

fujitatomoya:
> Beside the inline comment it seems that some unit tests are failing now in the PR builds.

I will check on those,

peci1:
As the double format is static, it is shared by all code. After your test finishes, you should set doubleformat back to the default `"%.16g"`. This should get the other tests green.

fujitatomoya:
@peci1 

nice catch!!! thanks for the information, that helps :smile: (actually i did not have time to look into...)

fujitatomoya:
local test shows green light,

```
# catkin_make run_tests_xmlrpcpp_gtest_TestValues
Base path: /root/catkin_ws
Source space: /root/catkin_ws/src
Build space: /root/catkin_ws/build
Devel space: /root/catkin_ws/devel
Install space: /root/catkin_ws/install
####
#### Running command: "make cmake_check_build_system" in "/root/catkin_ws/build"
####
####
#### Running command: "make run_tests_xmlrpcpp_gtest_TestValues -j24 -l24" in "/root/catkin_ws/build"
####
[ 50%] Built target gtest
[100%] Built target xmlrpcpp
[100%] Built target TestValues
-- run_tests.py: execute commands
  /root/catkin_ws/devel/lib/xmlrpcpp/TestValues --gtest_output=xml:/root/catkin_ws/build/test_results/xmlrpcpp/gtest-TestValues.xml
[==========] Running 12 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 11 tests from XmlRpc
[ RUN      ] XmlRpc.Bool
[       OK ] XmlRpc.Bool (0 ms)
[ RUN      ] XmlRpc.testBoolean
[       OK ] XmlRpc.testBoolean (0 ms)
[ RUN      ] XmlRpc.testInt
[       OK ] XmlRpc.testInt (0 ms)
[ RUN      ] XmlRpc.testDouble
[       OK ] XmlRpc.testDouble (0 ms)
[ RUN      ] XmlRpc.testString
[       OK ] XmlRpc.testString (0 ms)
[ RUN      ] XmlRpc.testDateTime
[       OK ] XmlRpc.testDateTime (0 ms)
[ RUN      ] XmlRpc.testArray
[       OK ] XmlRpc.testArray (1 ms)
[ RUN      ] XmlRpc.testStruct
[       OK ] XmlRpc.testStruct (5 ms)
[ RUN      ] XmlRpc.base64
[       OK ] XmlRpc.base64 (0 ms)
[ RUN      ] XmlRpc.int_errors
[       OK ] XmlRpc.int_errors (0 ms)
[ RUN      ] XmlRpc.array_errors
[       OK ] XmlRpc.array_errors (0 ms)
[----------] 11 tests from XmlRpc (6 ms total)

[----------] 1 test from XmpRpc
[ RUN      ] XmpRpc.errors
[       OK ] XmpRpc.errors (0 ms)
[----------] 1 test from XmpRpc (0 ms total)

[----------] Global test environment tear-down
[==========] 12 tests from 2 test suites ran. (6 ms total)
[  PASSED  ] 12 tests.
-- run_tests.py: verify result "/root/catkin_ws/build/test_results/xmlrpcpp/gtest-TestValues.xml"
[100%] Built target run_tests_xmlrpcpp_gtest_TestValues
```

fujitatomoya:
@dirk-thomas 

thanks for checking on this. we got all green, we have a go?

and would you want me to work on backport for melodic?
I am willing to do that, just let me know which version you want this to be put back.

dirk-thomas:
> and would you want me to work on backport for melodic?

Not need to. Any backport will happen in a combined PR as described in #1496.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
dirk-thomas: The logic still doesn't check / handle the case when `required_size` is negative. Please cover that case as - also in a unit test.

fujitatomoya: okay! when it returns negative, we would put the error message in the stream instead? would you share your thought?

dirk-thomas: Assuming the function will only fail when the format string is invalid I would suggest to print a warning message (maybe only once?) and put the double into `os` without custom formatting.

fujitatomoya: that makes sense to me, fixed.

fujitatomoya:
@dirk-thomas 

could you take a look when you got time?

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,16 +611,52 @@ namespace XmlRpc {
       default:           break;
       case TypeBoolean:  os << _value.asBool; break;
       case TypeInt:      os << _value.asInt; break;
-      case TypeDouble:   os << _value.asDouble; break;
+      case TypeDouble:
+        {
+          static std::once_flag once;
+          int ret, required_size;
```
dirk-thomas: A single `int` variable (`required_size`) should be sufficient. It can also be declared at the same time when being defined.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,16 +611,52 @@ namespace XmlRpc {
...
+                  getDoubleFormat().c_str(), _value.asDouble);
+          if (ret < 0) {
+            std::call_once(once,
+              [](){XmlRpcUtil::error("Failed to format with %s, using default %%.16g format",
+                getDoubleFormat().c_str());});
+            required_size = std::snprintf(buf, sizeof(buf)-1, "%.16g", _value.asDouble);
+          } else {
+            required_size = ret;
+          }
+          if (required_size >= 0 && required_size < (int) sizeof(buf)) {
```
dirk-thomas: Please use C++ casts instead of C casts.

Same below.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,16 +611,52 @@ namespace XmlRpc {
...
+            buf[sizeof(buf)-1] = 0;
+            os << buf;
+          } else if (required_size >= (int) sizeof(buf)) {
+            int ret;
```
dirk-thomas: No need to (re-)declare the variable.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,16 +611,52 @@ namespace XmlRpc {
...
+            char required_buf[required_size+1];
+            ret = std::snprintf(required_buf, required_size,
+                    getDoubleFormat().c_str(), _value.asDouble);
+            if (ret == required_size) {
+              required_buf[required_size] = 0;
+              os << required_buf;
+            }
+          } else {
+            XmlRpcUtil::error("Unexpected error to format %s", getDoubleFormat().c_str());
```
dirk-thomas: This message would point to the user provided format string - even when the above logic has switched to use the default format string.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,16 +611,52 @@ namespace XmlRpc {
...
-          buf[sizeof(buf)-1] = 0;
-          os << buf;
+          ret = std::snprintf(buf, sizeof(buf)-1, "%4d%02d%02dT%02d:%02d:%02d",
+                  t->tm_year,t->tm_mon,t->tm_mday,t->tm_hour,t->tm_min,t->tm_sec);
+          if (ret > 0) {
+            buf[sizeof(buf)-1] = 0;
+            os << buf;
+          } else {
+            XmlRpcUtil::error("Unexpected error to format TypeDateTime");
+          }
```
dirk-thomas: Why should the date time logic we changed at all? Can you describe a case where the `snprintf()` would fail? If not, please leave this part of the code unchanged.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
fujitatomoya: correct, fixed.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
fujitatomoya: fixed

utilities/xmlrpcpp/src/XmlRpcValue.cpp
fujitatomoya: fixed.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
fujitatomoya: fixed.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,13 +611,45 @@ namespace XmlRpc {
...
+          static std::once_flag once;
+          bool use_default = false;
+          char buf[128]; // Should be long enough
+          int required_size = std::snprintf(buf, sizeof(buf)-1,
+                                getDoubleFormat().c_str(), _value.asDouble);
+          if (required_size < 0) {
+            use_default = true;
+            std::call_once(once,
+              [](){XmlRpcUtil::error("Failed to format with %s, using default %%.16g format",
+                getDoubleFormat().c_str());});
```
fujitatomoya: this leads not to detect second or after error format in the same process space. this is unlikely to happen so catching 1st error format would be okay. what would you think?

fujitatomoya:
@dirk-thomas 

could you check when you got time? thanks!

fujitatomoya:
@dirk-thomas 

ping friendly,

fujitatomoya:
@dirk-thomas 

i think this one was missed under your radar, ping friendly 😄

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,13 +611,45 @@ namespace XmlRpc {
...
+            required_size = std::snprintf(buf, sizeof(buf)-1, "%.16g", _value.asDouble);
```
dirk-thomas: If the custom format string wasn't valid why not just do what was done before: `os << _value.asDouble`?

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,13 +611,45 @@ namespace XmlRpc {
...
+          }
+          break;
+        }
       case TypeString:   os << *_value.asString; break;
       case TypeDateTime:
         {
           struct tm* t = _value.asTime;
           char buf[20];
-          std::snprintf(buf, sizeof(buf)-1, "%4d%02d%02dT%02d:%02d:%02d", 
+          std::snprintf(buf, sizeof(buf)-1, "%4d%02d%02dT%02d:%02d:%02d",
```
dirk-thomas: Please avoid unrelated white space changes.

utilities/xmlrpcpp/src/XmlRpcValue.cpp
fujitatomoya: that makes sense to me!

utilities/xmlrpcpp/src/XmlRpcValue.cpp
```diff
@@ -610,7 +611,28 @@ namespace XmlRpc {
...
+        {
+          static std::once_flag once;
+          char buf[128]; // Should be long enough
+          int required_size = std::snprintf(buf, sizeof(buf)-1,
+                                getDoubleFormat().c_str(), _value.asDouble);
+          if (required_size < 0) {
+            std::call_once(once,
+              [](){XmlRpcUtil::error("Failed to format with %s", getDoubleFormat().c_str());});
+            os << _value.asDouble;
+          } else if (required_size >= 0 && required_size < static_cast<int>(sizeof(buf))) {
```
dirk-thomas: The `required_size >= 0` part of the condition is redundant since the the previous `if` evaluated to `false` to enter this `else` block.

fujitatomoya: correct, thanks.

fujitatomoya:
@dirk-thomas 

I've addressed your comments, could you take a look?

dirk-thomas:
Thanks for the patch.

PR #2009, opened 2020-07-26
Add skip_cache parameter to rosnode_ping()

### LINKED ISSUE #1967: [bug] python rosnode_ping failure after node restart

bofetadas:
Fixes #1967
Programmers now have the possibility to skip looking into the node cache when calling rosnode_ping or rosnode_ping_all.

### DIFF
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -294,7 +294,7 @@ def rosnode_listnodes(namespace=None, list_uri=False, list_all=False):
     """
     print(_sub_rosnode_listnodes(namespace=namespace, list_uri=list_uri, list_all=list_all))
     
-def rosnode_ping(node_name, max_count=None, verbose=False):
+def rosnode_ping(node_name, max_count=None, verbose=False, skip_cache=False):
     """
     Test connectivity to node by calling its XMLRPC API
     @param node_name: name of node to ping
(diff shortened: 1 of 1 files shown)


tools/rosnode/src/rosnode/__init__.py
```diff
@@ -303,12 +303,14 @@ def rosnode_ping(node_name, max_count=None, verbose=False):
...
     @type  verbose: bool
+	@param skip_cache: flag to skip cached data and force to lookup from master
+	@type  skip_cache: bool
     @return: True if node pinged
     @rtype: bool
     @raise ROSNodeIOException: if unable to communicate with master
     """
     master = rosgraph.Master(ID)
-    node_api = get_api_uri(master,node_name)
+    node_api = get_api_uri(master, node_name, skip_cache)
```
dirk-thomas: Please pass keyword arguments with a keyword.

tools/rosnode/src/rosnode/__init__.py
```diff
@@ -303,12 +303,14 @@ def rosnode_ping(node_name, max_count=None, verbose=False):
     @type  max_count: int
     @param verbose: print ping information to screen
     @type  verbose: bool
+	@param skip_cache: flag to skip cached data and force to lookup from master
+	@type  skip_cache: bool
```
dirk-thomas: NItpick: indendentation needs to match (and use spaces).

dirk-thomas:
@mabaue Thanks for the patch and the updates. Please comment on the PR in the future since there is notification for added commits so it is unlikely anyone will revisit the ticket.

bofetadas:
Okay will do so!

Issue #1975, opened 2020-06-20
New Kinetic Release


DLu:
There are a handful of backports on the kinetic branch [dating back to early 2019](https://github.com/ros/ros_comm/commit/4c4ecec73e647512fd086ab880f93b38b71330c2) that have not been released for kinetic. 

It would be useful to cut a new release for these features.

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/preparing-for-kinetic-sync-2020-06-25/15074/2

DLu:
Paging @dirk-thomas

ros-discourse:
This issue has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/preparing-for-kinetic-sync-2020-07-05/15250/2

dirk-thomas:
I understand the desire to get a new Kinetic release of `ros_comm` out but in the light of #1496 and me being the only person at Open Robotics doing ROS 1 maintenance work on ROS core packages I unfortunately can't give an estimate when I will be able to catch up with Kinetic backports and a release. The little amount of time I can use towards maintenance is usually used by work for newer distros. I know this is not a satisfying answer but I just wanted to describe how it is and give a realistic expectation.

DLu:
@dirk-thomas I think we both understand the burdens of maintaining core ROS software. However, I'm not asking for additional backports, but just release of the repo as-is. I do not have the permissions to make the release, otherwise I would do it myself. I believe not doing the release would be a disservice to past-Dirk who [already](https://github.com/ros/ros_comm/pull/1685) [worked](https://github.com/ros/ros_comm/pull/1789) [hard](https://github.com/ros/ros_comm/pull/1730) [reviewing and](https://github.com/ros/ros_comm/pull/1576) [merging previous PRs](https://github.com/ros/ros_comm/pull/1578)

DLu:
It appears as though I *have* been given permissions to make the release. Any objection to me doing that next week?

dirk-thomas:
> It appears as though I have been given permissions to make the release. Any objection to me doing that next week?

I don't know if anything has changed regarding permissions recently. Please do no make a release yourself without the maintainer of the repository. I will do a release when I find the time and will try to get to it within the next weeks.

DLu:
![pretty please?](https://media.giphy.com/media/jLKvwWHIXUS4/giphy.gif)

[...]

PR #1932, opened 2020-04-15
Use an internal implementation of boost::condition_variable with monotonic clock

meyerj:
_Follow-up on and partially revert #1878_:

Unfortunately @ahoarau was right in https://github.com/ros/ros_comm/pull/1878#issuecomment-582898222: The patch did not solve the issue with busy-wait spinning or non-functional timers completely, because the definition of `boost::condition_variable()` provided by different compilation units might still have been compiled with non-matching settings of `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC`. So even with the patch the eventually conflicting definitions of the symbol is an ODR violation and can cause undefined behavior, depending on the compiler internals and flags.

For example, `libtf2_ros.so` from latest `ros-melodic-tf2-ros` provides a weak symbol `boost::condition_variable()` which was compiled without the macro `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC`. Therefore, as soon as `libtf2_ros.so` is linked into the process all condition variables are constructed with the standard system clock as their internal clock. The wait calls within roscpp, e.g. for timers, which expect that condition variables are configured with CLOCK_MONOTONIC internally, become no-ops then.

While the problem has finally been solved in Boost 1.67 and the remaining checks for older Boost versions have been completely removed in https://github.com/ros/ros_comm/pull/1903 for noetic, I propose to partially revert #1878  for melodic and older, and to re-add a custom stripped down version of `boost::condition_variable` as `ros::internal::condition_variable_monotonic`. This is more or less what I proposed in my earlier pull request https://github.com/ros/ros_comm/pull/1651, but rebased on the latest `melodic-devel` branch and simplified. The copy uses a different namespace and does not respect `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC`. It uses the monotonic clock unconditionally. This custom implementation is only selected for platforms that support it and if `BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC` or `BOOST_THREAD_INTERNAL_CLOCK_IS_MONO` are not already defined by Boost itself (and therefore would be effective for all compilation units). Otherwise `ros::internal::condition_variable_monotonic` is an alias for `boost::condition_variable`:

```cpp
#if !defined(BOOST_THREAD_PLATFORM_PTHREAD) || \
    defined(BOOST_THREAD_HAS_CONDATTR_SET_CLOCK_MONOTONIC) || \
[...]

PR #2011, opened 2020-07-31
Use an internal implementation of boost::condition_variable with monotonic clock [kinetic-devel]

meyerj:
Backports the following two pull requests to `kinetic-devel`:
- Drop custom implementation of boost::condition_variable to fix busy-wait spinning (#1878)
- Use an internal implementation of boost::condition_variable with monotonic clock (#1932)
  - ... which partially reverted #1878.

Basically both together replace one custom implementation of `boost::condition_variable` by another, with the goal to support waiting on it with a monotonic clock source and without ODR violations. Both replaced earlier pull requests #1557 (kinetic) and #1651 (melodic). The original issue that they hopefully fixed now is #1343.

A small step towards _New Kinetic Release_ (https://github.com/ros/ros_comm/issues/1975)...


dirk-thomas:
Thanks for the patch!

PR #2016, opened 2020-08-07
[melodic] support Boost 1.66

at-wat:
interruption_checker::check() had been available only between Boost 1.65.0 and 1.65.1.
It has been renamed to unlock_if_locked() since 1.66.0.

ref: https://github.com/seqsense/aports-ros-experimental/pull/281 (build for Alpine Linux failed)


dirk-thomas:
Thanks for the patch.

@meyerj FYI.

PR #2017, opened 2020-08-10
Use heapq.merge instead of custom merge sort code

chaoflow:
Since Python 3.5 a library function exists that does what we need. I'm therefore proposing to replace custom merge sort code with calls to heapq.merge.

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -1333,7 +1333,7 @@ class Bag(object):
         """
         Yield index entries on the given connections in the given time range.
         """
-        for entry, _ in _mergesort(self._get_indexes(connections), key=lambda entry: entry.time):
+        for entry in heapq.merge(*self._get_indexes(connections), key=lambda x: x.time.to_nsec()):
             if start_time and entry.time < start_time:
                 continue
             if end_time and entry.time > end_time:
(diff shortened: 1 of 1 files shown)


chaoflow:
Is this good for melodic-devel? Shall I rebase to noetic-devel (and open a separate PR)?

chaoflow:
This needs at least python 3.5 and therefore cannot be used on melodic-devel.

The two failing checks, I don't understand.

dirk-thomas:
Please provide some context in the description of the PR why you are proposing this change.

chaoflow:
@dirk-thomas Just did.

dirk-thomas:
I couldn't find any documentation if `heapq.merge` is stable. Do you have any information on that?

Also what motivates the change? Is `heapq.merge` in any way better (e.g. faster) than the custom function?

chaoflow:
@dirk-thomas The heapq.merge implementation since python 3.5 https://github.com/python/cpython/blob/3.5/Lib/heapq.py is very similar to the the one in bag.py, with slight optimizations.

My motivation for the change is to replace custom code with equal or better code in Python's standard library.

In its current implementation it is stable. But given that it is not explicitly called *merge sort*, we should confirm with Python whether its stability can be taken granted also in future versions. I'm happy to do that if the change in general is desirable.

chaoflow:
It has been stable since it was added in 2.6 due to explicit code, not as a side effect. I think we can take it for granted that it will stay stable.

dirk-thomas:
Sounds good to me. Thanks for the patch.

PR #2022, opened 2020-08-12
Ignore underscores when parsing literal numeric values

jspricke:
Python 3, due to PEP515, allows numerical values to be (visually)
grouped via underscores. However, here, we want to consider literals
containing an underscore as a string.


jspricke:
Similar to https://github.com/ros/xacro/pull/247

dirk-thomas:
Thanks for the patch.

PR #2023, opened 2020-08-16
[roscpp] Update boost::placeholders usage for boost 1.73 (and later)

seanyen:
Since `boost 1.73`, when including `boost\bind.hpp`, the following message will be generated:

```
The practice of declaring the Bind placeholders (_1, _2, ...) 
in the global namespace is deprecated. Please use
<boost/bind/bind.hpp> + using namespace boost::placeholders,
or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.
```

And with the `boost 1.73`, some `cpp` files started to show the compilation errors,, such as:
```
C:\workspace\rcpp_ws\src\ros_comm\clients\roscpp\src\libros\connection_manager.cpp(69): error C2065: '_1': undeclared identifier
```

This pull request is to use the fully qualified name first.

Replacing `boost/bind.hpp` with `boost/bind/bind.hpp` directly will cause downstream projects build breaks before the downstream projects comply to the same migration.

### DIFF
--- a/clients/roscpp/include/ros/node_handle.h
+++ b/clients/roscpp/include/ros/node_handle.h
@@ -47,7 +47,7 @@
 #include "ros/init.h"
 #include "common.h"
 
-#include <boost/bind.hpp>
+#include <boost/bind/bind.hpp>
 
 #include <xmlrpcpp/XmlRpcValue.h>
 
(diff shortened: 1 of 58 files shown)


seanyen:
This is ready for review.

dirk-thomas:
Thanks for the patch.

PR #2035, opened 2020-09-02
Change error message

avwhite:
As far as I can tell, not being able to ping itself does not *cause* the issue (at least if ping refers to using the `ping` program to check if a host can be reached).

Not sure if this wording is the optimal.

The old wording caused me to spend quite a bit of time being confused about why `ping` was a dependency of ROS (launch) when the official ROS docker containers do not come with `ping` installed.


dirk-thomas:

dirk-thomas:
Thanks for the patch.

PR #2036, opened 2020-09-02
rostest: add service and filter testing

aremazeilles:
This PR extends `rostest` tools for checking if:

* a service server provides appropriate reply for a given request
* a filter-like node generates and publishes appropriate message after having received a given message.

The usage is aligned with the other `rostest` tools:
For service test:

```xml
<test test-name="test_service" pkg="node_test" type="test_service" >
        <rosparam>
          calls:
            - name: /trigger_spec
              input: None
              output: {'success': True, 'message': 'well done!'}
        </rosparam>
</test>
```

For filter test:

```xml
<test test-name="test_service" pkg="node_test" type="test_service" >
    <rosparam>
      calls:
        - name: /add_two_ints
          input: {'a': 0, 'b': 5.0}
          output: {'sum': 5}
    </rosparam>      
 </test>
```

In both cases, Input and output are provided as a _python dictionnary_, and it is then converted to the appropriate message format using [rospy_message_converter](https://github.com/uos/rospy_message_converter).

More description is available at the repository [ros_node_test](https://github.com/tecnalia-advancedmanufacturing-robotics/ros_node_test) that I created before considering proposing that PR to the upstream respository.

PR #2039, opened 2020-09-08
rostest: add service and filter testing (noetic)

aremazeilles:
Following the discussion in this [previous request](https://github.com/ros/ros_comm/pull/2036), This new pull request targets the latest branch `noetic-devel`.

This PR extends `rostest` tools for checking if:

* a service server provides appropriate reply for a given request
* a filter-like node generates and publishes appropriate message after having received a given message.

The usage is aligned with the other `rostest` tools:
For service test:

```xml
<test test-name="test_service" pkg="node_test" type="test_service" >
        <rosparam>
          calls:
            - name: /trigger_spec
              input: None
              output: {'success': True, 'message': 'well done!'}
        </rosparam>
</test>
```

For filter test:

```xml
<test test-name="test_service" pkg="node_test" type="test_service" >
    <rosparam>
      calls:
        - name: /add_two_ints
          input: {'a': 0, 'b': 5.0}
          output: {'sum': 5}
    </rosparam>      
 </test>
```

In both cases, Input and output are provided as a _python dictionnary_, and it is then converted to the appropriate message format using [rospy_message_converter](https://github.com/uos/rospy_message_converter).

More description is available at the repository [ros_node_test](https://github.com/tecnalia-advancedmanufacturing-robotics/ros_node_test) that I created before considering proposing that PR to the upstream respository.

### DIFF
--- a/tools/rostest/package.xml
+++ b/tools/rostest/package.xml
@@ -21,7 +21,7 @@
   <run_depend>rosmaster</run_depend>
   <run_depend>rospy</run_depend>
   <run_depend>rosunit</run_depend>
-
+  <run_depend>rospy_message_converter</run_depend>
   <export>
     <rosdoc config="rosdoc.yaml"/>
     <architecture_independent/>
--- a/tools/rostest/CMakeLists.txt
+++ b/tools/rostest/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(rostest)
 
-find_package(catkin COMPONENTS rosunit)
+find_package(catkin COMPONENTS rosunit rospy_message_converter)
 find_package(Boost COMPONENTS thread)
 
 include_directories(include ${Boost_INCLUDE_DIRS})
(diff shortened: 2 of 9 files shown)


dirk-thomas:
@aremazeilles The new test is failing.

aremazeilles:
@dirk-thomas , thanks for the info. I fixed the errors

tools/rostest/package.xml
```diff
@@ -21,7 +21,7 @@
   <run_depend>rosmaster</run_depend>
   <run_depend>rospy</run_depend>
   <run_depend>rosunit</run_depend>
-
+  <run_depend>rospy_message_converter</run_depend>
```
dirk-thomas, line 24: I am not sure this new external dependency in an already release ROS distro is desired in a package that low in the stack (being part of the ROS core).

I don't know how easy it would be to remove the need for this dependency or if it would be easier to release these new features as a standalone package?

dirk-thomas: @aremazeilles Any thought on this?

aremazeilles: Sorry for the delay in answering. Removing the dependency would require copying significant code from  `rospy_message_converter` into that package, which I am not sure it is the best strategy.
So maybe the second option you propose, making a standalone package, might be a better solution. What do you think?

dirk-thomas: > making a standalone package, might be a better solution. What do you think?

Yes, I would recommend that.

sloretz:
Thank you for the PR!

PR #2044, opened 2020-09-12
Fix log*_throttle with sim time

salihmarangoz:
When log*_throttle is used with bags or a simulation, logging stops when bag/simulation restarts. Solved this problem with resetting logging_time_table when ros time moved backward.

Testing code:
```py
import rospy

rospy.init_node('test_node')

    rospy.loginfo_throttle(100.0, "Throttled message!")
```

### DIFF
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -223,8 +223,9 @@ class LoggingThrottle(object):
               (now - last_logging_time) > rospy.Duration(period)):
             self.last_logging_time_table[caller_id] = now
             return True
-        elif (last_logging_time > now):
+        elif last_logging_time > now:
             self.last_logging_time_table = {}
+            self.last_logging_time_table[caller_id] = now
             return True
         return False
 
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -223,6 +223,9 @@ class LoggingThrottle(object):
               (now - last_logging_time) > rospy.Duration(period)):
             self.last_logging_time_table[caller_id] = now
             return True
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/core.py
```diff
@@ -223,6 +223,9 @@ def __call__(self, caller_id, period):
               (now - last_logging_time) > rospy.Duration(period)):
             self.last_logging_time_table[caller_id] = now
             return True
+        elif (last_logging_time > now):
```
dirk-thomas: Nitpick: please remove the parenthesis.

clients/rospy/src/rospy/core.py
```diff
@@ -223,6 +223,9 @@ def __call__(self, caller_id, period):
...
+            self.last_logging_time_table = {}
```
dirk-thomas, line 227: After resetting the dictionary shouldn't the `caller_id` still be added: `self.last_logging_time_table[caller_id] = now`?

salihmarangoz: Yes, you are right. I fixed now it should prevent losing a log message when the time goes back

clients/rospy/src/rospy/core.py
salihmarangoz: Done, thanks

salihmarangoz:
It should be ok now

clients/rospy/src/rospy/core.py
dirk-thomas: I don't think it would have lost a message but always also printed the next message. Anyway its fixes. Thanks.

dirk-thomas:
Thanks for the patch.

PR #2050, opened 2020-09-22
fix error handling with Python 3

seanyen:
This is motivated by the issue found here: https://github.com/ms-iot/ROSOnWindows/issues/275

The exception object in Python3 will unpack more than 2 outputs in Python3 and this pull request is to capture and discard the rest to avoid the error: `too many values to unpack`.

NOTE: This is a Python3 idiom to use wildcard for unpack, so this is not compatible with Melodic or older distributions.

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -157,7 +157,7 @@ class TCPServer(object):
             except ConnectionAbortedError:
                 continue
             except IOError as e:
-                (e_errno, msg, *discard) = e.args
+                (e_errno, msg, *_) = e.args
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -157,7 +157,7 @@ def run(self):
             except ConnectionAbortedError:
                 continue
             except IOError as e:
-                (e_errno, msg) = e.args
+                (e_errno, msg, *discard) = e.args
```
dirk-thomas: Please use `_` instead of `discard`.

seanyen: You mean using `*_` right? I renamed it in the new commit.

PR #2051, opened 2020-09-22
[Windows] Single quote string fix-up

seanyen:
This is motivated by the fact that many of launch files originally written for Linux are using single quotes to enclose the file path. For example, in the [XACRO tutorial](http://wiki.ros.org/action/fullsearch/urdf/Tutorials/Using%20Xacro%20to%20Clean%20Up%20a%20URDF%20File?action=fullsearch&context=180&value=linkto%3A%22urdf%2FTutorials%2FUsing+Xacro+to+Clean+Up+a+URDF+File%22), it is using this format:
image

Unfortunately, the single quotes on Windows has different meaning and it will be treated as part of the argument and passed to the program to run. It causes problems since the program doesn't expect the extra characters.

One way is to go edit the original launch files to replace the single quotes with `&quot;`, so it will be like:

```xml
<param name="robot_description"
   command="xacro --inorder &quot;$(find pr2_description)/robots/pr2.urdf.xacro&quot;" />
```

However it requires editing on all the files and ideally those changes need to be upstream'ed.

So this pull request is to propose the method to fix up the cases on-the-fly in the roslaunch internally, which helps the downstream project and we don't have to touch the downstream code extensively.

And please consider to back-port to `melodic-devel` if it makes sense.

### DIFF
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -501,6 +501,17 @@ class Loader(object):
                     # special handle the use of Python scripts in Windows environment:
                     # 1. search for a wrapper executable (of the same name) under the same directory with stat.S_IXUSR flag
                     # 2. if no wrapper is present, prepend command with 'python' executable
+                    def split_with_single_quote_enclosed_fixup(command):
+                        tokens = shlex.split(command, posix=False)  # use non-posix method on Windows
+                        if not ("'" in command):
+                            return tokens
+                        new_command = []
+                        for token in tokens:
+                            if token.startswith("'") and token.endswith("'"):
+                                new_command.append(token[1:-1])
+                            else:
+                                new_command.append(token)
+                        return new_command
 
                     cl = shlex.split(command, posix=False)  # use non-posix method on Windows
                     if os.path.isabs(cl[0]):
@@ -528,6 +539,7 @@ class Loader(object):
                                             executable_command = ' '.join([sys.executable, f])
                             if executable_command:
                                 command = command.replace(cl[0], executable_command, 1)
+                    command = split_with_single_quote_enclosed_fixup(command)
                 p = subprocess.Popen(command, stdout=subprocess.PIPE)
                 c_value = p.communicate()[0]
                 if not isinstance(c_value, str):
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -497,32 +497,28 @@ class Loader(object):
                 if os.name != 'nt':
                     command = shlex.split(command)
                 else:
+                    orig_command = command
+                    command = shlex.split(command, posix=False)  # use non-posix method on Windows
+
+                    # On Linux, single quotes are commonly used to enclose a path to escape spaces.
+                    # However, on Windows, the single quotes are treated as part of the arguments.
+                    # Special handling is required to remove the extra single quotes.
+                    if "'" in orig_command:
+                        command = [token[1:-1] if token.startswith("'") and token.endswith("'") else token for token in command]
+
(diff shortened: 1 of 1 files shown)


tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -528,6 +539,7 @@ def param_value(self, verbose, name, ptype, value, textfile, binfile, command):
                                             executable_command = ' '.join([sys.executable, f])
                             if executable_command:
                                 command = command.replace(cl[0], executable_command, 1)
+                    command = split_with_single_quote_enclosed_fixup(command)
```
dirk-thomas: Why should this only be applied after the long `if` block? Couldn't the same also apply to the first item of `command` and therefore affect the `if` block?

tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -501,6 +501,17 @@ def param_value(self, verbose, name, ptype, value, textfile, binfile, command):
                     # special handle the use of Python scripts in Windows environment:
                     # 1. search for a wrapper executable (of the same name) under the same directory with stat.S_IXUSR flag
                     # 2. if no wrapper is present, prepend command with 'python' executable
+                    def split_with_single_quote_enclosed_fixup(command):
+                        tokens = shlex.split(command, posix=False)  # use non-posix method on Windows
+                        if not ("'" in command):
```
dirk-thomas: Nitpick: this should be written as `if "'" not in command:`.

tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -501,6 +501,17 @@ def param_value(self, verbose, name, ptype, value, textfile, binfile, command):
                     # special handle the use of Python scripts in Windows environment:
                     # 1. search for a wrapper executable (of the same name) under the same directory with stat.S_IXUSR flag
                     # 2. if no wrapper is present, prepend command with 'python' executable
+                    def split_with_single_quote_enclosed_fixup(command):
```
dirk-thomas: Since this is a single use local function I would rather keep the code inline. Also in the current patch it seems to imply that the previously existing comment above now relates to the function but that isn't the case. Therefore please move the logic above that comment.

tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -501,6 +501,17 @@ def param_value(self, verbose, name, ptype, value, textfile, binfile, command):
...
+                        tokens = shlex.split(command, posix=False)  # use non-posix method on Windows
+                        if not ("'" in command):
+                            return tokens
+                        new_command = []
+                        for token in tokens:
+                            if token.startswith("'") and token.endswith("'"):
+                                new_command.append(token[1:-1])
+                            else:
+                                new_command.append(token)
```
dirk-thomas: This loop could be expressed with a list comprehension which would make it more readable.

dirk-thomas:
@seanyen Friendly ping.

seanyen:
@dirk-thomas Thanks for the ping. I reworked my change and hope now it is more readable.

tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -497,21 +497,28 @@ def param_value(self, verbose, name, ptype, value, textfile, binfile, command):
                 if os.name != 'nt':
                     command = shlex.split(command)
                 else:
+                    orig_command = command
+                    command = shlex.split(command, posix=False)  # use non-posix method on Windows
```
dirk-thomas: Please keep the variable name as `cl` as it was before the patch to avoid unnecessary renaming.

seanyen: I pushed another change to drop more renaming.

dirk-thomas:

Issue #2053, opened 2020-09-23
paramtest fails for parameters with valid None or empty string values


lewkoo:
**What did you try?**
Tried to test a parameter with an empty value using `paramtest` node. 

**What did you observe?**
Test timed out

**What did you expect?**
Test succeeded

**All said?**

When using paramtest node to test existence / correctness of a value of a parameter in a parameter server, the test will not finish for parameters which have a valid value of None or are just empty. 

In `paramtest` the condition is specified as `while not param_obtained` but it will also evaluate to True if param_obtained is an empty string and sometimes that is a correct parameter value. 

Solution: Replace `while not param_obtained` with `while param_obtained is None`.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #2054, opened 2020-09-23
Fix paramtest for empty values of a parameter

lewkoo:


lewkoo:
Relates to issue #2053

tools/rostest/nodes/paramtest
```diff
@@ -90,7 +90,7 @@ class ParamTest(unittest.TestCase):
 
         wallclock_timeout_t = time.time() + wait_time
         param_obtained = None
-        while not param_obtained and time.time() < wallclock_timeout_t:
+        while param_obtained in None and time.time() < wallclock_timeout_t:
```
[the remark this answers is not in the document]
dirk-thomas: Please use `is None` here. Currently the test is failing because `None` is not iterable. Please make sure to run the code with your changes locally to confirm it works as desired.

dirk-thomas: Please use `is None` here. As the failing tests show the current patch doesn't work because `None` isn't iterable. Please also make sure to run the modified code locally and confirm that it behaves as desired before creating a pull request.

lewkoo: I am sorry, my bad. I have amended my original commit with your suggestions after running the tests locally.

dirk-thomas:
Thanks for the patch.

PR #2055, opened 2020-09-23
Add `filter` and `reduce` among the commands allowed in `$(eval)`

peci1:
`map` is already there and it seems to me that is makes sense to also add `filter` and `reduce`. They can be used with lambdas (verified with `map`).


dirk-thomas:
Please target the current default branch `noetic-devel`.

peci1:
You're right. See #2070 instead.

PR #2038, opened 2020-09-04
Gracefully stop recording upon SIGTERM and SIGINT

dabonnie:
Resolves #2006

* Add handling for SIGTERM in recorder
* Preserve handling for SIGINT in recorder

PR #2058, opened 2020-09-24
[Kinetic] Backport #2038

### LINKED ISSUE #2006: rosbag leaves an ".active" file dangling when receiving SIGTERM
Currently, rosbag leaves a dangling ".active" file when stopped with SIGTERM.

1. Can we update `rosbag` so that the behavior of SIGTERM and SIGINT is the same?
2. Would it be acceptable to backport this behavior to Kinetic, and Melodic?

dabonnie:
Backport of #2038 , resolves #2006 

* Gracefully stop recording upon SIGTERM and SIGINT 
* Add rostopic test dependency


dabonnie:
@dirk-thomas please review

dirk-thomas:
Changes are backported together before a new patch release and not on an individual base since with the number of changes and their order that becomes unmanagable. See #1496 which describes the process. In this case the change first needs to land in a Noetic release, then be backported to Melodic and from there to Kinetic.

Issue #1884, opened 2020-02-16
roslaunch: scope of machine tag with default=true not respected


peci1:
From roslaunch/XML/machine docs:

> default="true|false|never" (optional)

> Sets this machine as the default to assign nodes to. The default setting only applies to nodes defined later in the same scope. NOTE: if there are no default machines, the local machine is used. You can prevent a machine from being chosen by setting default="never", in which case the machine can only be explicitly assigned. 

I've a pair of launch files:

a.launch
```
<launch>
  <node name="a" pkg="rostopic" type="rostopic" args="pub -r 1 /a std_msgs/Int32 'data: 1'" />
  <include file="b.launch" />
  <node name="c" pkg="rostopic" type="rostopic" args="pub -r 1 /a std_msgs/Int32 'data: 1'" />
</launch>
```

b.launch
```
<launch>
  <machine name="b" address="robot" user="robot" password="robot" env-loader="/home/robot/env.sh" default="true" />
  <node name="b" pkg="rostopic" type="rostopic" args="pub -r 1 /b std_msgs/Int32 'data: 1'" machine="b" />
  <node name="b2" pkg="rostopic" type="rostopic" args="pub -r 1 /b std_msgs/Int32 'data: 1'" />
</launch>
```

With this setup, when launching a.launch, nodes `b`, `b2` and `c` are launched on the remote machine.

So it seems that the "scope" the docs are talking about are the whole "include-resolved" launch file, which is highly undesirable and unintuitive. Even enclosing the nodes in b.launch doesn't limit the scope.

What I'd expect would be nodes `b` and `b2` launching on remote machine, while `a` and `c` being launched locally.

And I'm not alone with this expectation: https://answers.ros.org/question/206699/what-is-scope-in-roslaunch-default-machine-not-changing/ .

Why I'd like to do this is to be able to specify an arg in a launch file that would allow me to select to launch the node either without any machine tag, or, if the arg is nonempty, with a machine tag belonging to the selected machine. But with https://github.com/ros/ros_comm/issues/274 still unresolved, I cannot pass an empty machine attribute to node, so the only way to allow configurable remote/local launching is by typing the whole node code twice. That's not really nice...

git-afsantos:
I second this. A minimal example showing this behaviour could be:

```xml
<launch>
  <group>
    <machine name="willow" address="blah.willowgarage.com" default="true" />
  </group>

[...]

PR #1937, opened 2020-04-21
roslaunch: Allow passing _TIMEOUT_SIGINT and _TIMEOUT_SIGTERM as parameters

peci1:
Continuation of #1502 and #1670.

roslaunch tests pass for me locally, including the manual test that verifies that this behavior works even for remote machines (`test/manual-test-remote-timeouts.sh 5 10 localhost ~/.bashrc  $USER`).

PR #2059, opened 2020-09-24
Support "scope" for machine tag default semantics

luisrayas3:
Proposed fix for https://github.com/ros/ros_comm/issues/1884

### DIFF
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -155,7 +155,8 @@ class LoaderContext(object):
     """
     
     def __init__(self, ns, filename, parent=None, params=None, env_args=None, \
-                     resolve_dict=None, include_resolve_dict=None, arg_names=None):
+                     resolve_dict=None, include_resolve_dict=None, arg_names=None, \
+                     default_machine=None):
         """
         @param ns: namespace
         @type  ns: str
@@ -169,6 +170,9 @@ class LoaderContext(object):
         @type include_resolve_dict: dict
         @param arg_names: name of args that have been declared in this context
         @type  arg_names: [str]
+        @param default_machine: name of the machine to use as default only
+        within this context.
+        @type  default_machine: str
         """
         self.parent = parent
         self.ns = make_global_ns(ns or '/')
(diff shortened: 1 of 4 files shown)


luisrayas3:
Open to alternatives to `"scope"` for the value that enables this behavior. `"true_while_in_scope"` seemed a bit verbose

peci1:
I tested this PR and it works as advertised. Thanks!

I also added another value for the `default` argument - `parent`, which means that the machine becomes default in the parent context... that is useful if you want to define the machines in a separate launch file which is then included at relevant places in other launch files. It is just a few changed lines compared to this PR. Would you be interested in getting this feature here? On one hand, I consider this a useful feature, on the other hand, it might become a bit unintuitive when included files can change the default machine...

peci1 (approved):

peci1:
For those who do not want to wait until this PR is merged, I created a package that hacks roslaunch and adds this functionality: https://github.com/ctu-vras/scoped_roslaunch .

ameysutavani:
Hello all,
Touch basing on this PR. Is is possible to get it merged?

DanyResasco:
hi, I need this feature too. is it possible to get merged? @peci1 I'm tried to use your patch could you please tell me how to use it? Maybe  I installed it wrong.

130s:
I asked whether there's a room for merging a fix to the targeted issue https://github.com/ros/ros_comm/issues/1884#issuecomment-1078795224. That said, the way this PR suggests as of now seems to be adding a new capability ('scope'), which I'm not sure if we can still add to already-released core library. Just my feeling.

peci1:
If the wiki docs have been correct for the past few years, then this PR is just fixing a bug. It adds the "new" functionality in order to keep code using the wrong (old, wiki-nonconformant) implementation working.

@130s Do you have a reference to a decision where OR decided not adding any new functionality into released core ROS1 packages? In earlier versions, it was quite common that released distros gained new functionality - if its implementation did not break ABI and API. I know everything is ROS 2 now, but Noetic still has quite some lifespan ahead and it would be sad to know it will not gain even non-intrusive features. Also, there is recent new functionality that has been accepted to roslaunch: https://github.com/ros/ros_comm/pull/1937 (both into Melodic and Noetic). So it seems that adding new features should still be possible.

DanyResasco:
Hi, 
any update on this feature? The machine tag is not working as supposed to.

sloretz:
Thank you for the PR!

PR #2063, opened 2020-09-25
Use --substitute_keywords when repeatedly republishing single message from file

### LINKED ISSUE #2062: rostopic: --substitute_keywords isn't used when repeatedly republishing single message from file
#1635 introduced feature to repeatedly republish single message from file.
However, `--substitute_keywords` argument isn't used in this case,
which makes it impossible to simulate constant sensor readings.
Is there any point behind this?

pavloblindnology:
Fixes #2062 for melodic-devel


dirk-thomas:
Please target the default branch (`noetic-devel`) with this PR. After it has been merged it will be considered for backporting before the next Melodic patch release.

dirk-thomas:
@pavloblindnology Friendly ping.

sloretz:
Thank you for the PR!

PR #2065, opened 2020-09-30
Trap for overly large input to XmlRPCPP

SidFaber:
Oversize input to XmlRPCPP could cause problems with int <-> size_t conversions.

 - Recognize when incoming or outgoing data is too large, generate an error and discard the data when practical.
 - Use the safe strtol() rather than atoi() to decode an incoming content-length header and generate an error if the length is invalid or too large.
 - Prevent attempts to parse overly large XML input.
 - Add tests where they can reasonably be inserted into existing test routines.

Although this fix could be cleaner the update is written to make the update ABI compatible.

This fix addresses CVE-2020-16124 / Integer overflow in ros_comm.

Signed-off-by: Sid Faber <sid.faber@canonical.com>

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -458,7 +458,7 @@ XmlRpcClient::readHeader()
   return true;    // Continue monitoring this source
 }
 
-
+    
 bool
 XmlRpcClient::readResponse()
 {
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -458,6 +458,7 @@ XmlRpcClient::readHeader()
   return true;    // Continue monitoring this source
 }
 
+
 bool
 XmlRpcClient::readResponse()
 {
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -312,6 +312,13 @@ XmlRpcClient::generateRequest(const char* methodName, XmlRpcValue const& params)
                   header.length(), body.length());
 
   _request = header + body;
+  // Limit the size of the request to avoid integer overruns
+  if (_request.length() > size_t(__INT_MAX__)) {
+    XmlRpcUtil::error("XmlRpcClient::generateRequest: request length (%u) exceeds maximum allowed size (%u).",
+                      _request.length(), __INT_MAX__);
+    _request.clear();
+    return false;
+  }
   return true;
 }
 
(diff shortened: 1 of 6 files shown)


utilities/xmlrpcpp/test/TestValues.cpp
```diff
@@ -27,8 +27,9 @@
 #include <stdlib.h>
 #include <string>
 
-#include "xmlrpcpp/XmlRpcValue.h"
```
dirk-thomas, line 30: Nitpick: please avoid reordering existing code to keep the diff minimal.

utilities/xmlrpcpp/src/XmlRpcClient.cpp
```diff
@@ -448,7 +458,6 @@ XmlRpcClient::readHeader()
   return true;    // Continue monitoring this source
 }
 
-
```
dirk-thomas: Nitpick: please avoid unrelated changed to keep the diff minimal.

dirk-thomas (commented):
Thanks for the patch.

Atm CI is failing for two tests though.

SidFaber:
> Atm CI is failing for two tests though.

@dirk-thomas, thanks for the review. The overrun tests I added are causing memory allocation failures--they work on my local machine but I assume the CI tests don't expect to be handling 2Gb input values. Any suggestions on how best to handle, is it advisable to remove the overrun tests?

dirk-thomas:
> Any suggestions on how best to handle, is it advisable to remove the overrun tests?

Can the code be written in a way that it fails gracefully if that much memory isn't available?

dirk-thomas:
> Any suggestions on how best to handle, is it advisable to remove the overrun tests?

Can the code be written in a way that it fails gracefully if that much memory isn't available? Maybe marking the test as skipped.

utilities/xmlrpcpp/test/TestValues.cpp
SidFaber: Addressed with 6f8500dcb139f13c50060bc6e44a4f748799a5a1

utilities/xmlrpcpp/src/XmlRpcClient.cpp
SidFaber: Addressed with 4afa09b44e7419ee63cce7e171cdd0d411b149af

dirk-thomas:
Debian Buster only has Googletest 1.8.1 which doesn't have `GTEST_SKIP`.

The `readResponse_oversize` test still fails in CI.

SidFaber:
I removed GTEST_SKIP, printed an error message but otherwise let the test succeed. Also I removed one test that was causing problems. The test_client routines for readResponse still run when objects are finalized; since the test object could be in an unpredictable state I couldn't find a good way to ignore the results.

dirk-thomas:
> I removed GTEST_SKIP, printed an error message but otherwise let the test succeed.

That sounds like a good fall backback behavior. Can you make this conditional based on if `GTEST_SKIP` is available? If it is use it to give better feedback, otherwise fall back to the error message / succeed.

SidFaber:
@dirk-thomas, I added the conditional statement in one of the tests but had to rip another one out--the Ubuntu tests started running long and got killed causing the whole module to fail. I think this is as good as I can get it unless you have any ideas on how to safely & consistently write tests for huge input strings.

dirk-thomas:
Sounds good to me. Thanks for the patch and for iterating on it.

PR #2066, opened 2020-10-01
fix misspell.

fujitatomoya:
Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>


fujitatomoya:
@dirk-thomas 

only fix for spelling, could you take a look?

dirk-thomas:
Thanks for the patch.

PR #2070, opened 2020-10-05
Add `filter` and `reduce` among the commands allowed in `$(eval)`

peci1:
#2055 retargeted to noetic-devel.

`map` is already there and it seems to me that is makes sense to also add `filter` and `reduce`. They can be used with lambdas (verified with `map`).


dirk-thomas:
@peci1 Please check the test failures for the PR.

peci1:
Holding off before #2071 gets resolved and decided.

The failures were because of the fact that `reduce` is no longer a builtin in Python 3. But it can be imported from `functools`.

sloretz:
Thank you for the PR!

PR #2068, opened 2020-10-02
cached parameter should be unsubscribed

fujitatomoya:
address https://github.com/ros/ros_comm/issues/2067

PR #2073, opened 2020-10-06
[Melodic Backport] cached parameter should be unsubscribed (#2068)

fujitatomoya:
melodic backport for https://github.com/ros/ros_comm/pull/2068


dirk-thomas:
Changes are backported together before a new patch release and not on an individual base since with the number of changes and their order that becomes unmanagable. See #1496 which describes the process.

PR #2076, opened 2020-10-14
[Melodic] Update maintainers (#2075)

sloretz:
Cherry-pick backport of #2075

Previous: @dirk-thomas
New: @jacobperron, @mjcarroll, @sloretz

Signed-off-by: Shane Loretz<sloretz@openrobotics.org>
Signed-off-by: Shane Loretz <sloretz@osrfoundation.org>

### DIFF
--- a/clients/roscpp/package.xml
+++ b/clients/roscpp/package.xml
@@ -12,7 +12,9 @@
     roscpp is the most widely used ROS client library and is designed to
     be the high-performance library for ROS.
   </description>
-  <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
+  <maintainer email="jacob@openrobotics.org">Jacob Perron</maintainer>
+  <maintainer email="michael@openrobotics.org">Michael Carroll</maintainer>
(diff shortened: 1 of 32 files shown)


dirk-thomas:
I strongly suggest *not* to backport individual commits out of order but follow the process described in #1496.

sloretz:
> I strongly suggest not to backport individual commits out of order but follow the process described in #1496.

As in backports to older distros should be done periodically in big batches like #1477? Is this what that process looks like?

1. #2075 should be included in the next Noetic release
2. After some soak time it should be included in a batch of backports from Noetic -> Melodic
3. The backports should be released to Melodic
4. After some soak time the backports should be included in a batch from Melodic -> Kinetic
5. The backports should be releasd to Kinetic

Does the process have a gap between merging a PR and releasing it, or merging backports and releasing them?

dirk-thomas:
> Is this what that process looks like?

Correct.

> Does the process have a gap between merging a PR and releasing it, or merging backports and releasing them?

That is up to you and the confidence you have that it doesn't break anything. A longer soak time allows you to catch regressions in e.g. Noetic before even backporting and releasing a change into Melodic.

PR #2086, opened 2020-10-30
rosbag: start player in paused state

fevb:
- Initializes the rosbag player correctly in paused mode in case the `options.start_paused` parameter is set to true.

Otherwise one could run into timing issues because the `paused_` member variable was only set at a later stage when publishing messages from the bag:

https://github.com/ros/ros_comm/blob/090daa56f57d18c39946f107bcee5e3efefb2cac/tools/rosbag/src/player.cpp#L229

For instance: if one would set `options.start_paused = true` and immediately call the  `~/pause_playback` service, the service call could return an error stating that the bag was already being replayed.

### DIFF
--- a/tools/rosbag/src/player.cpp
+++ b/tools/rosbag/src/player.cpp
@@ -109,7 +109,7 @@ void PlayerOptions::check() {
 
 Player::Player(PlayerOptions const& options) :
     options_(options),
-    paused_(false),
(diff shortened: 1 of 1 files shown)


jacobperron (approved):
I wasn't able to reproduce the race described, but the change looks okay to me. Thanks!

PR #2089, opened 2020-11-11
Sort connection indices

durko:
This PR makes the python rosbag reader sort connection indices by time, like its C++ rosbag::View counterpart does. Without this messages from a single topic are not necessarily yielded in order of their timestamp, which is guaranteed by the C++ implementation.

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -2682,7 +2682,12 @@ class _BagReader200(_BagReader):
             self.bag._curr_chunk_info = chunk_info
             for i in range(len(chunk_info.connection_counts)):
                 connection_id, index = self.read_connection_index_record()
-                self.bag._connection_indexes[connection_id].extend(index)
+                self.bag._connection_indexes[connection_id].append(index)
+
+        self.bag._connection_indexes = {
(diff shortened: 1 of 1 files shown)


pjreed:
I just reviewed and tested this, and it looks good to me.  I think this is actually a really good fix; for a long time I've just been manually reordering bag files after recording them so that `rosbag play` would play messages in the right order.

The only concern I would have is that *technically* this breaks backwards-compatibility in terms of rosbag's functionality.  I highly doubt that anybody was intentionally depending on rosbag playing messages in the wrong order, but it's possible that, for example, somebody could have a suite of unit tests that use rosbag to play a bag file and expect the messages to be in a certain order, and this change could cause that order to change.

Personally I think the benefit of having them play in the right order outweighs the risk, especially since it now matches C++ rosbag's behavior, but it's something to keep in mind.

chaoflow:
@pjreed Thank you very much for the review! I missed the notification back then.

Note: we're using this code in production https://github.com/ternaris/marv-robotics/commit/3084f07ae79398a74e77c116aad8f510b571d592

jacobperron (approved):
The patch looks reasonable to me. AFAIK, there aren't any guarantees about the returned order of read messages, so I think any potential breaks for users depending on this is acceptable.

@sloretz @mjcarroll Thoughts?

sloretz (approved):
Thank you for the PR!

PR #2096, opened 2020-11-25
Replace message assertion with logging in order to have release modes…

### LINKED ISSUE #2095: ROS_ASSERT_MSG check on md5_sum of publish is ignored in release modes.
Not the same issue in https://github.com/ros/ros_comm/issues/1972, but the root cause is the same. CMake release modes skip assertion checks by implicitly specifying `-NDEBUG` flag, hence the check in https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/include/ros/publisher.h#L81 gets skipped. This allows the successful compilation of the following:

```
...
ros::Publisher chatter_pub = n.advertise<std_msgs::String>("chatter", 1000);
std_msgs::String msg;
msg.data = "hello world ";
std::vector<std_msgs::String> msgs{msg};
chatter_pub.publish(msgs);
...
```

In my case, the subscriber receives an empty `std_msgs/String`. 

The rationale in http://wiki.ros.org/CppStyleGuide is given as: 
_"It is typical to develop software with assertion-checking enabled, in order to catch violations. When nearing software completion and when assertions are found to always be true in the face of **extensive testing**, you build with a flag that removes assertions from compilation, so they take up no space or time."_

It is reasonable as all pub/sub relationships should be unit tested. However such testing is very non-trivial and tedious in some cases. Therefore I think that this particular check should not be skipped in release modes, which would facilitate the detection of such type erasure failures on published messages.

tahsinkose:
… to fail in compilation when msg type mismatches occur.

Fixes https://github.com/ros/ros_comm/issues/2095.

### DIFF
--- a/clients/roscpp/include/ros/publisher.h
+++ b/clients/roscpp/include/ros/publisher.h
@@ -77,11 +77,14 @@ namespace ros
           ROS_ASSERT_MSG(false, "Call to publish() on an invalid Publisher (topic [%s])", impl_->topic_.c_str());
           return;
         }
-
-      ROS_ASSERT_MSG(impl_->md5sum_ == "*" || std::string(mt::md5sum<M>(*message)) == "*" || impl_->md5sum_ == mt::md5sum<M>(*message),
-                     "Trying to publish message of type [%s/%s] on a publisher with type [%s/%s]",
-                     mt::datatype<M>(*message), mt::md5sum<M>(*message),
-                     impl_->datatype_.c_str(), impl_->md5sum_.c_str());
+        if (impl_->md5sum_ == "*" ||
+            std::string(mt::md5sum<M>(*message)) == "*" ||
+            impl_->md5sum_ == mt::md5sum<M>(*message)) {
+          ROS_DEBUG_ONCE("Trying to publish message of type [%s/%s] on a "
+                    "publisher with type [%s/%s]",
+                    mt::datatype<M>(*message), mt::md5sum<M>(*message),
+                    impl_->datatype_.c_str(), impl_->md5sum_.c_str());
+        }
 
       SerializedMessage m;
       m.type_info = &typeid(M);
@@ -110,11 +113,14 @@ namespace ros
           ROS_ASSERT_MSG(false, "Call to publish() on an invalid Publisher (topic [%s])", impl_->topic_.c_str());
           return;
         }
-
-      ROS_ASSERT_MSG(impl_->md5sum_ == "*" || std::string(mt::md5sum<M>(message)) == "*" || impl_->md5sum_ == mt::md5sum<M>(message),
-                     "Trying to publish message of type [%s/%s] on a publisher with type [%s/%s]",
-                     mt::datatype<M>(message), mt::md5sum<M>(message),
-                     impl_->datatype_.c_str(), impl_->md5sum_.c_str());
+        if (impl_->md5sum_ == "*" ||
+            std::string(mt::md5sum<M>(message)) == "*" ||
+            impl_->md5sum_ == mt::md5sum<M>(message)) {
+          ROS_DEBUG_ONCE("Trying to publish message of type [%s/%s] on a "
+                    "publisher with type [%s/%s]",
+                    mt::datatype<M>(message), mt::md5sum<M>(message),
+                    impl_->datatype_.c_str(), impl_->md5sum_.c_str());
+        }
 
       SerializedMessage m;
       publish(boost::bind(serializeMessage<M>, boost::ref(message)), m);


clients/roscpp/include/ros/publisher.h
```diff
@@ -77,11 +77,10 @@ namespace ros
...
         }
-
-      ROS_ASSERT_MSG(impl_->md5sum_ == "*" || std::string(mt::md5sum<M>(*message)) == "*" || impl_->md5sum_ == mt::md5sum<M>(*message),
-                     "Trying to publish message of type [%s/%s] on a publisher with type [%s/%s]",
-                     mt::datatype<M>(*message), mt::md5sum<M>(*message),
-                     impl_->datatype_.c_str(), impl_->md5sum_.c_str());
+      ROS_DEBUG("Trying to publish message of type [%s/%s] on a "
+                "publisher with type [%s/%s]",
+                mt::datatype<M>(*message), mt::md5sum<M>(*message),
+                impl_->datatype_.c_str(), impl_->md5sum_.c_str());
```
fujitatomoya: How about adding if statement to check message type, then call `ROS_FATAL` and `ROS_BREAK`? i think this is fatal information since it does not publish the correct message as user application expects.

clients/roscpp/include/ros/publisher.h
```diff
@@ -110,11 +109,10 @@ namespace ros
           ROS_ASSERT_MSG(false, "Call to publish() on an invalid Publisher (topic [%s])", impl_->topic_.c_str());
           return;
         }
-
-      ROS_ASSERT_MSG(impl_->md5sum_ == "*" || std::string(mt::md5sum<M>(message)) == "*" || impl_->md5sum_ == mt::md5sum<M>(message),
-                     "Trying to publish message of type [%s/%s] on a publisher with type [%s/%s]",
-                     mt::datatype<M>(message), mt::md5sum<M>(message),
-                     impl_->datatype_.c_str(), impl_->md5sum_.c_str());
+      ROS_DEBUG("Trying to publish message of type [%s/%s] on a "
```
fujitatomoya: same here.

clients/roscpp/include/ros/publisher.h
tahsinkose: With this, it won't even compile in case of such a type mismatch. `ROS_DEBUG` is just a macro that I'm sure of not being opt-out by the compiler in release mode :slightly_smiling_face: Any similar line could be replaced with it. The important bits are ` mt::datatype<M>(*message)` and `mt::md5sum<M>(*message)`, which fail when type mismatch occurs.

I mean, replacements with `ROS_FATAL` or `ROS_BREAK` won't compile either. Ultimately, any mismatches in the published messages will be captured in compile-time.

fujitatomoya: > ROS_DEBUG is just a macro that I'm sure of not being opt-out by the compiler in release mode

i do not think so, it actually depends on build environment.

https://github.com/ros/rosconsole/blob/854b3a1ec426e173383d2cf1d241c5626ec33715/include/ros/console.h#L314-L321

in default, it does not opt-out ROS_DEBUG nor ROS_FATAL. (So that we can see the ROS_INFO with release mode)

`ROS_BREAK` is the same with `ROS_ASSERT_MSG`, which i think it would be nice to abort when it is built with debug mode.

>  The important bits are mt::datatype<M>(*message) and mt::md5sum<M>(*message), which fail when type mismatch occurs.

that is exactly why i was thinking that using if statement to check `impl_->md5sum_ == "*" || std::string(mt::md5sum<M>(*message)) == "*" || impl_->md5sum_ == mt::md5sum<M>(*message)` then print message. and with debug built it goes abort.

tahsinkose: So, do you suggest guarding `ROS_DEBUG/ROS_BREAK` statement with an if-check, which would be never opt-out?

fujitatomoya: Yes. `ROS_BREAK` will be opt-out with release build as i mentioned previous comment. but as far as I see, this seems to be a fatal problem for user application because message cannot be delivered to the other side. of course this is just my idea, so I'd like to hear from others.

tahsinkose: If we guard with if-check, such delivery is irrelevant in here as it will never compile in release mode, too. I'm trying to make sure that the mismatches in delivery are noticed in compile-time.

fujitatomoya: okay, i see what you mean. type mismatch will be checked during compile time, then just prints the debug information if user configures loglevel into debug. that makes sense. thanks for iterating 👍

tahsinkose: Added. PTAL @fujitatomoya

tahsinkose:
@sloretz You seem to be the latest maintainer. Could you please have a look?

jacobperron (approved):
The change LGTM. Cheers!

PR #2101, opened 2020-12-06
Mitigate io reentrance error on roslaunch test

at-wat:
Python's IO stack is not reentrant.
Writing to a file in signal handling is unsafe and raises RuntimeError on Python 3 when flush() is called during flush().
Call flush() only on SIGTERM to reduce a chance of reentrance.
See https://bugs.python.org/issue24283

### DIFF
--- a/tools/roslaunch/test/signal_logger.py
+++ b/tools/roslaunch/test/signal_logger.py
@@ -12,9 +12,14 @@ if __name__ == '__main__':
 
     def handler(signum, _):
         log_stream.write("%i %s\n" % (signum, str(time.time())))
-        log_stream.flush()
 
         if signum == signal.SIGTERM:
+            # note: Python's IO stack is not reentrant.
+            #       Writing to a file in signal handling is unsafe and raises
+            #       RuntimeError on Python 3 when flush() is called during flush().
+            #       Call flush() only on SIGTERM to reduce a chance of reentrance.
+            #       See https://bugs.python.org/issue24283
+            log_stream.flush()
             log_stream.close()
 
 


tools/roslaunch/test/signal_logger.py
```diff
@@ -12,9 +12,14 @@
 
     def handler(signum, _):
         log_stream.write("%i %s\n" % (signum, str(time.time())))
-        log_stream.flush()
```
fujitatomoya, line 15: I think that this flush is intended to flush the previous write to the disk.

tools/roslaunch/test/signal_logger.py
```diff
@@ -12,9 +12,14 @@
...
 
         if signum == signal.SIGTERM:
+            # note: Python's IO stack is not reentrant.
+            #       Writing to a file in signal handling is unsafe and raises
+            #       RuntimeError on Python 3 when flush() is called during flush().
+            #       Call flush() only on SIGTERM to reduce a chance of reentrance.
+            #       See https://bugs.python.org/issue24283
+            log_stream.flush()
```
fujitatomoya, line 22: I do not think this is needed since `close` does flush before closing.

fujitatomoya (commented):
I think that to address the problem in [IO reentrant], it should not handle IO operation in signal handler. instead, we would want to create `EventMonitor` to see the signal event and notified, then do IO operation in that class.

btw, did you actually meet the problem during test?

at-wat:
> I think that to address the problem in [IO reentrant], it should not handle IO operation in signal handler. instead, we would want to create EventMonitor to see the signal event and notified, then do IO operation in that class.
>
> btw, did you actually meet the problem during test?

Yes, I'm managing ROS buildfarm for Alpine Linux and this error is reproduced almost 100% when building on AMD EPYC (32 vCPU) environment. (I haven't seen the error on Intel CPU with completely same build script. I don't know why.)
This is a workaround to make it pass for now and completely removing IO operation from signal handler is better.

sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #2102, opened 2020-12-09
Fix /use_sim_time parameter typo in roswtf error

nick-paul:
`roswtf` reports that `/use_simtime` is not set though it is actually checking for the parameter `/use_sim_time`


fujitatomoya (approved):

jacobperron (approved):
Thanks for the fix!

PR #2103, opened 2020-12-09
rosmsg: clean up test dependencies

kyrofa:
All dependencies must be declared in the package.xml, and all these dependencies must be in the ROS distro (see https://github.com/ros/rosdistro/issues/27359). rosmsg doesn't meet either of these requirements today, by using but not depending on both rostest and std_srvs, and by using test_rosmaster, which isn't in the ROS distro at all.

Update rosmsg to properly depend upon what it requires, and stop using test_rosmaster completely. In place of test_rosmaster, use diagnostic_msgs, which satisfies the same requirements of the existing tests by having both messages and services as well as standalone and recursive messages.


jacobperron (approved):
Seems reasonable to me, but I'd like to check-in with @sloretz and @mjcarroll for any reservations on introducing a test depend on diagnostic_msgs.

PR #2107, opened 2020-12-16
fix getNumPublishers() to only count fully connected

c-andy-martin:
There is code which uses getNumPublishers() being positive as a
guarantee that a call to the remote publisher's publish() is
guaranteed to send a message to the subscriber. However, because
connections which have not yet received their header are counted,
this is not true. One such user of getNumPublishers() is actionlib.
It relies on getNumPublishers() only returning postive when a publish
on the result topic would succeed.

Fix this by only counting connections with received headers. If we
have received the header, we know the remote publisher is tracking a
connection back to us and has accepted the connection and that a
call to publish() on its end will send the message.

Note that this issue is only relevant for the TCP transport, as the
UDP transport and intraprocess link have already setup their header
by the time they are added to the publication links. Since all
publisher link types update the header, including the caller ID,
counting those with caller IDs set to non-empty is sufficient to
solve the issue.

### DIFF
--- a/clients/roscpp/src/libros/subscription.cpp
+++ b/clients/roscpp/src/libros/subscription.cpp
@@ -141,8 +141,19 @@ void Subscription::getInfo(XmlRpc::XmlRpcValue& info)
 
 uint32_t Subscription::getNumPublishers()
 {
-	boost::mutex::scoped_lock lock(publisher_links_mutex_);
-	return (uint32_t)publisher_links_.size();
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
this makes sense to me. only happens with TCPROS but there is a possibility that header is not registered via `Connection::onHeaderLengthRead`. keeping the behavior consistent for `Subscription::getNumPublishers` makes sense.

jacobperron (approved):
LGTM, thanks for the fix!

PR #2108, opened 2020-12-18
Add missing library depends

roehling:
The rosbag encrypt tool uses boost_program_options, but does not explicitly link against it. This is masked by the considerable overlinking that happens upstream in other ROS packages (in this case, rospack "helpfully" provides boost_program_options to all its dependencies). Nevertheless, it's a good idea to make such dependencies explicit.


jacobperron (approved):
LGTM, thanks!

PR #2113, opened 2021-01-02
Added warning to sleep() when ROS Time Clock is not set.

valschmidt:
Added warning to sleep() when ROS Time clock is not set. Previously sleep would block indefinitely without any error or warning, which is confusing. 

If this is acceptable I'll submit a similar PR for roscpp.


sloretz:
Thank you for the PR!

Unfortunately I don't think we should merge this one. [ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160), and I'm not sure if this change will introduce log spam or not.

PR #2114, opened 2021-01-05
Fix Building Encrypted Bags on Mac

smnogar:
Swaps out crypto library name with openssl. I didn't test to see if this change would also work on ubuntu, but it might.

### DIFF
--- a/tools/rosbag_storage/CMakeLists.txt
+++ b/tools/rosbag_storage/CMakeLists.txt
@@ -22,21 +22,29 @@ catkin_package(
 # Support large bags (>2GB) on 32-bit systems
 add_definitions(-D_FILE_OFFSET_BITS=64)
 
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
would you mind sharing the actual problem that you have w/o this fix?

smnogar:
Ah, yes. I suppose I should provide more context. rosbag_storage currently fails to build by default on macOS because crypto is included in openssl when obtained via brew, so when looking for the `crypto` library cmake would fail. Previously I would just patch out encrypted bags altogether, but this change makes it so rosbag_storage and the encryption functionality correctly builds.

sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #2116, opened 2021-01-08
Unroll parameters when using type="yaml" with the param tag.

Hugal31:
Make the `<param>` tag unrolls the parameters when using `type="yaml"`, just like the `<rosparam>` tag does.

Closes ##2115

### DIFF
--- a/tools/roslaunch/src/roslaunch/xmlloader.py
+++ b/tools/roslaunch/src/roslaunch/xmlloader.py
@@ -274,9 +274,7 @@ class XmlLoader(loader.Loader):
                 p = Param(name, value)
                 context.add_param(p)
             else:
-                p = Param(ns_join(context.ns, name), value)
-                ros_config.add_param(Param(ns_join(context.ns, name), value), filename=context.filename, verbose=verbose)
-            return p
+                self.add_param(ros_config, ns_join(context.ns, name), value, verbose=verbose)
 
         except KeyError as e:
             raise XmlParseException(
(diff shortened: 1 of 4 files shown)


tools/roslaunch/src/roslaunch/xmlloader.py
```diff
@@ -274,9 +274,7 @@ def _param_tag(self, tag, context, ros_config, force_local=False, verbose=True):
                 p = Param(name, value)
                 context.add_param(p)
             else:
-                p = Param(ns_join(context.ns, name), value)
-                ros_config.add_param(Param(ns_join(context.ns, name), value), filename=context.filename, verbose=verbose)
-            return p
```
Hugal31, line 279: I remove this return in `XmlLoader._param_tag`, I am almost sure it is not used anywhere. Tell me if I am wrong.

sloretz: I'm not familiar with this code. What is this case doing? Is removing it necessary for this PR, or is it an unrelated change?

Hugal31: This function is XmlLoader._param_tag. This function is called "dynamically" (i.e. you will see no invocation of this function written as-it), along some other functions as _rosparam_tag, which doesn't return anything. This, and the fact that the tests are still working, make me think removing this return has no impact.

In this case, this function used to return the parsed parameter. However, this function can now parse multiple parameters (by unrolling them). The return don't make sense anymore.

mlanighan:
Is there anyway to help move this PR forward? This would be a nice feature addition.

sloretz:
Thank you for the PR!

Issue #1980, opened 2020-06-26
[BUG] (Risk of a) deadlock in ros::Timer impl?


CodeFinder2:
Hi all,

I've already described my issue here in detail: https://answers.ros.org/question/355644/possible-risk-of-a-deadlock-in-rostimer-impl/

But since I didn't get any reply, I would like to post it here as well since it may be a bug.

Please refer to my linked post above on ROS Answers for the details.
TL;DR: I am using async. spinners with multiple threads and a ROS timer. All my callbacks have to lock a specific mutex first before doing anything in the callbacks. But what happened was that my `timerCallback()` was invoked and the above mentioned mutex was already locked by another thread - perfectly fine so far. That other thread was calling `timer.stop()` and since my `timerCallback()` did not return (and cannot because it is still passively waiting for the mutex to be released), `timer.stop()` waited infinitely -> deadlock.

Im my opinion, this should not happen. I mean, when I issue a `timer.stop()` but the `timerCallback()` was already invoked slightly before, it should be allowed to continue without any locking. `timer.stop()`, once called, should prevent to invoke any *new* callbacks but shouldn't care about a callback / event that has been triggered already.

Please take a look and let me know what do you think!

This was happening on Ubuntu 20.04 LTS with ROS Noetic. Unfortunately, because this was a race condition happening only very rarely, I may not simply be able to reproduce it.

Thanks for taking a look!

CodeFinder2:
Any ideas on this? Facing this issue again in a different situation...really annoying.

Any help is (still) highly appreciated...

iwanders:
@CodeFinder2, problem is that the timer cannot be stopped when a callback associated to that timer is being executed. Combined with the external lock this can currently cause a deadlock.

We encountered this problem in a production system. I've reproduced it with the attached minimal-non-working example [here](https://gist.github.com/iwanders/ede48fb649fd47f9b1f9a52c527b463c).

I also have a proposed fix / mitigation for this which I'll be filing shortly.

CodeFinder2:
Thanks for your reply and for providing a fix! I also came up with a temporary solution (by creating a new thread in the `timerCallback()` to let it return immediately) but doesn't seem to be a good way to go ... 🙈

jacobperron:
Resolved in #2121

jovere:
@CodeFinder2 I realize this is quite old, but could you explain how your workaround was implemented?

Thanks,
Jeremy

CodeFinder2:
[...]

PR #2121, opened 2021-01-25
Fix for deadlock issue 1980

iwanders:
We encountered the issue sketched in #1980 in a production system last week.

To understand the issue at hand I've made a minimal-non-working example [here](https://gist.github.com/iwanders/ede48fb649fd47f9b1f9a52c527b463c). Looking at the source code it seems that the callback queue is actually at fault, not the timer code. The `removeByID` deadlocks when a callback by that id is currently being executed.

The first commit introduces a unit test that demonstrates this deadlock in a unit test.

The second commit is the proposed fix for this problem. We try to obtain the lock, if we fail to obtain the lock we defer removal until the next callback queue cycle using the already existing `marked_for_removal` boolean.

FYI @guillaumeautran , @mikepurvis , @jasonimercer

### DIFF
--- a/clients/roscpp/src/libros/callback_queue.cpp
+++ b/clients/roscpp/src/libros/callback_queue.cpp
@@ -166,19 +166,36 @@ void CallbackQueue::removeByID(uint64_t removal_id)
     }
 
     {
-      boost::unique_lock<boost::shared_mutex> rw_lock(id_info->calling_rw_mutex);
-      boost::mutex::scoped_lock lock(mutex_);
-      D_CallbackInfo::iterator it = callbacks_.begin();
-      for (; it != callbacks_.end();)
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
makes sense to me. I also confirmed that this PR fixes the deadlock problem introduced [here](https://gist.github.com/iwanders/ede48fb649fd47f9b1f9a52c527b463c).

iwanders:
@fujitatomoya , is there anything else I can do to help get this issue merged and resolved?

fujitatomoya:
@jacobperron @mjcarroll @sloretz friendly ping.

jacobperron (approved):
The fix looks good to me, thanks!

Also, thank you for the reproduction and unit test :bow:

Issue #2123, opened 2021-01-30
Function recv_buff(sock, b, buff_size) when the remote end is closed, resulting in abnormal reconnection.


jinmenglei:
When I use rosbridge, after I send a message, I close the client connected through rosbridge, which will result in `d = sock.recv(buff_size)`  After receiving an empty string.

```python
 def recv(self, buffersize, flags=None): # real signature unknown; restored from __doc__
        """
        recv(buffersize[, flags]) -> data
        
        Receive up to buffersize bytes from the socket.  For the optional flags
        argument, see the Unix manual.  When no data is available, block until
        at least one byte is available or until the remote end is closed.  When
        the remote end is closed and all data is read, return the empty string.
        """
        pass
```

An exception is thrown, and then the socket is reconnected. Finally, the waiting time exceeds 32S. It is difficult to connect, and the port will be wasted when reconnecting. I refer to the C + + code and find that the processing logic is different.

```cpp
int32_t TransportTCP::read(uint8_t* buffer, uint32_t size)
{
  {
    boost::recursive_mutex::scoped_lock lock(close_mutex_);

    if (closed_)
    {
      ROSCPP_LOG_DEBUG("Tried to read on a closed socket [%d]", sock_);
      return -1;
    }
  }

  ROS_ASSERT(size > 0);

  // never read more than INT_MAX since this is the maximum we can report back with the current return type
  uint32_t read_size = std::min(size, static_cast<uint32_t>(INT_MAX));
  int num_bytes = ::recv(sock_, reinterpret_cast<char*>(buffer), read_size, 0);
  if (num_bytes < 0)
  {
	if ( !last_socket_error_is_would_block() ) // !WSAWOULDBLOCK / !EAGAIN && !EWOULDBLOCK
    {
      ROSCPP_LOG_DEBUG("recv() on socket [%d] failed with error [%s]", sock_, last_socket_error_string());
      close();
    }
    else
    {
      num_bytes = 0;
    }
  }
  else if (num_bytes == 0)
  {
    ROSCPP_LOG_DEBUG("Socket [%d] received 0/%u bytes, closing", sock_, size);
    close();
    return -1;
  }

  return num_bytes;
}
```

 C + + receives zero and closes the connection directly instead of reconnecting. Can we consider closing it, like this:

```python
# add break here
                    if str(e) == "unable to receive data from sender, check sender's logs for details":
                        break
```

```python
    def receive_loop(self, msgs_callback):
        """
        Receive messages until shutdown
[...]

PR #2124, opened 2021-02-01
fix bug #2123:Do not raise exception if socket is busy in TCPROSTransport

jinmenglei:
It's the same as roscpp socket.recv Exception handling process.

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -801,6 +801,11 @@ class TCPROSTransport(Transport):
                     else:
                         self._reconnect()
 
+                except TransportTerminated as e:
+                    logdebug("[%s] failed to receive incoming message : %s" % (self.name, str(e)))
+                    rospydebug("[%s] failed to receive incoming message: %s" % (self.name, traceback.format_exc()))
+                    break
+
                 except TransportException as e:
                     # set socket to None so we reconnect
                     try:
@@ -814,14 +819,12 @@ class TCPROSTransport(Transport):
                     except:
                         pass
                     self.socket = None
-                    if str(e) == "unable to receive data from sender, check sender's logs for details":
-                        break
 
                 except DeserializationError as e:
                     #TODO: how should we handle reconnect in this case?
                     
-                    logerr("[%s] error deserializing incoming request: %s"%self.name, str(e))
-                    rospyerr("[%s] error deserializing incoming request: %s"%self.name, traceback.format_exc())
+                    logerr("[%s] error deserializing incoming request: %s" % (self.name, str(e)))
+                    rospyerr("[%s] error deserializing incoming request: %s" % (self.name, traceback.format_exc()))
                 except:
                     # in many cases this will be a normal hangup, but log internally
                     try:
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -62,6 +62,8 @@ from rospy.msg import deserialize_messages, serialize_message
 from rospy.service import ServiceException
 
 from rospy.impl.transport import Transport, BIDIRECTIONAL
+# add #2123
+from errno import EAGAIN, EWOULDBLOCK
 
 logger = logging.getLogger('rospy.tcpros')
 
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -100,12 +102,19 @@ def recv_buff(sock, b, buff_size):
...
+    # add #2123
+    try:
+        d = sock.recv(buff_size)
+        if d:
+            b.write(d)
+            return len(d)
+        else: #bomb out
+            raise TransportTerminated("unable to receive data from sender, check sender's logs for details")
+    except socket.error as ex:
+        if ex.args[0] in (EAGAIN, EWOULDBLOCK):
```
fujitatomoya: why it compares error message against error numbers? it is supposed to be `ex.errno`?

clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -100,12 +102,19 @@ def recv_buff(sock, b, buff_size):
...
+            raise TransportTerminated("unable to receive data from sender, check sender's logs for details")
```
fujitatomoya, line 114: if either EAGAIN or EWOULDBLOCK, it would want to retry. but raising exception with this particular message, it will break out the while loop in `receive_loop`.

clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -100,12 +102,19 @@ def recv_buff(sock, b, buff_size):
     @return: number of bytes read
     @rtype: int
     """
-    d = sock.recv(buff_size)
-    if d:
-        b.write(d)
-        return len(d)
-    else: #bomb out
-        raise TransportTerminated("unable to receive data from sender, check sender's logs for details")
+    # add #2123
```
fujitatomoya: i think we can delete this comment.

clients/rospy/src/rospy/impl/tcpros_base.py
```diff
@@ -806,6 +815,9 @@ def receive_loop(self, msgs_callback):
                     except:
                         pass
                     self.socket = None
+                    # add #2123
```
fujitatomoya: same here.

fujitatomoya (commented):
IMO, I was thinking that where we would want to change is `receive_loop` to catch exception `TransportTerminated` before `TransportException`.

clients/rospy/src/rospy/impl/tcpros_base.py
jinmenglei: i will delete it

clients/rospy/src/rospy/impl/tcpros_base.py
jinmenglei: This place is reference C++
// never read more than INT_MAX since this is the maximum we can report back with the current return type
  uint32_t read_size = std::min(size, static_cast<uint32_t>(INT_MAX));
  int num_bytes = ::recv(sock_, reinterpret_cast<char*>(buffer), read_size, 0);
  if (num_bytes < 0)
  {
	if ( !last_socket_error_is_would_block() ) // !WSAWOULDBLOCK / !EAGAIN && !EWOULDBLOCK
    {
      ROSCPP_LOG_DEBUG("recv() on socket [%d] failed with error [%s]", sock_, last_socket_error_string());
      close();
    }
    else
    {
      num_bytes = 0;
    }
  }
  else if (num_bytes == 0)
  {
    ROSCPP_LOG_DEBUG("Socket [%d] received 0/%u bytes, closing", sock_, size);
    close();
    return -1;
  }

clients/rospy/src/rospy/impl/tcpros_base.py
jinmenglei: I fixed this bug and had to deal with the logic

clients/rospy/src/rospy/impl/tcpros_base.py
jinmenglei: i delete this comment

clients/rospy/src/rospy/impl/tcpros_base.py
jinmenglei: i change “ex.args[0]” to "ex.errno"

jinmenglei:
@fujitatomoya I didn't find a good way to deal with `transportterminated` before `TransportException`.Do you have a better solution?

fujitatomoya:
how about the following?

```
diff --git a/clients/rospy/src/rospy/impl/tcpros_base.py b/clients/rospy/src/rospy/impl/tcpros_base.py
index 57f340f49..b4bfd402b 100644
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -800,6 +800,11 @@ class TCPROSTransport(Transport):
                     else:
                         self._reconnect()

+                except TransportTerminated as e:
+                    logerr("[%s] failed to receive incoming message : %s"%self.name, str(e))
+                    rospyerr("[%s] failed to receive incoming message: %s"%self.name, traceback.format_exc())
+                    break
+
                 except TransportException as e:
                     # set socket to None so we reconnect
                     try:
```

jinmenglei:
@fujitatomoya
I updated the code. Is this branch no longer maintained？
image

fujitatomoya (approved):

jinmenglei:
Is this PR to be closed because #2131 it has been merged?
@jacobperron

fujitatomoya:
http://wiki.ros.org/Distributions, kinetic is already E.O.L. I guess backport is not doable. so maybe just close this.

jinmenglei:
> http://wiki.ros.org/Distributions, kinetic is already E.O.L. I guess backport is not doable. so maybe just close this.

ok

jacobperron:
I will look into backporting #2131 to Melodic, following the next Noetic release.

Issue #2118, opened 2021-01-14
XMLRPC HTTP/1.1 causes performance degradation in rosmaster between Kinetic and Melodic


emersonknapp:
Related to https://github.com/ros/ros_comm/pull/371
Introduced in https://github.com/ros/ros_comm/pull/1287

## Environment

* ROS Distro: Melodic
* Platform: Ubuntu 18.04 Bionic
* Linux Kernel Version: 4.15
  * special note - this doesn't occur on Linux kernel 5.4. Ubuntu 18.04.5 ships by default with 4.15, so that should be an important consideration for ROS Melodic - a Focal-based distro (e.g. Noetic) won't have to worry about this, I am not able to reproduce it on Bionic+Kernel5.4

## Description

I am able to reliably cause the `rosmaster` to stop responding to service calls with `unable to contact master` - as raised by https://github.com/ros/ros_comm/blob/noetic-devel/clients/rospy/src/rospy/impl/tcpros_service.py#L467 (triggered specifically by the call to `master.lookupService`)

The reproduction workflow involves starting a single `rospy.Service` in one node (serving `std_srvs/SetBool`), and 11 `rospy.ServiceProxy` instances each in separate nodes. Each of these clients calls the service at 200Hz. After about 30 seconds, the `ServiceException: unable to contact master` starts to occur. The master does not crash, but is unreachable for several seconds. If the stress is stopped without stopping the master, then the situation is reproducible, suggesting there is no lasting damage done to the master process - just a temporary hang of some kind.

## Repro Instructions

I am running in a container, the image was build using the following Dockerfile

```
from osrf/ros:melodic-desktop

run apt-get update
run apt-get install -y python3-pip
run pip3 install -U pip setuptools
run pip3 install -U colcon-common-extensions
```

The test application sources are [testpkg.tar.gz](https://github.com/ros/ros_comm/files/5812028/testpkg.tar.gz)

I run the following workflow

```
$ docker build . -t melodic-desktop-dev
$ mkdir src/
# move testpkg into src/
$ docker run -it -v $(pwd):/ros_ws -w /ros_ws melodic-desktop-dev

# in the container now
$ source /opt/ros/melodic/setup.bash
$ colcon build

## open up a separate shell into the container via docker exec
$ source install/setup.bash
$ roscore

## open up a third shell into the container via docker exec
$ source install/setup.bash
$ roslaunch testpkg testapp.launch --screen --wait
```

[...]

PR #1287, opened 2017-12-31
Use HTTP/1.1 in XMLRPC Server

mgrrx:
This PR activates HTTP/1.1 in the XMLRPC Server and fixes the CTRL-C issues that were reported in https://github.com/ros/ros_comm/pull/371

Setting daemon_threads to True allows the program to terminate.

PR #2132, opened 2021-02-14
Fix HTTP for kernel < 4.16

jikawa-az:
Fixes bug #2118

Linux kernels 4.15 and older encounter performance issues with HTTP/1.1
I have tested this with a customer's 4.14 kernel workflow and it addressed their performance issues.
Will need to be backported to Melodic.

Signed-off-by: Jesse Ikawa <jikawa@amazon.com>

### DIFF
--- a/tools/rosgraph/src/rosgraph/xmlrpc.py
+++ b/tools/rosgraph/src/rosgraph/xmlrpc.py
@@ -79,25 +79,23 @@ def isstring(s):
 
 
 def _support_http_1_1():
-    """Linux kernels 4.15 and older encounter performance issues with HTTP/1.1
-    This function returns true if the Linux kernel version is greater than the
-    input or if the platform is not Linux.
     """
-    # HTTP/1.1 performance issue only detected on Linux
+    Determine whether HTTP 1.1 should be enabled for XMLRPC communications.
+
+    This will be true on non-Linux systems, and on Linux kernels at least as
+    new as 4.16. Linux kernels 4.15 and older cause significance performance
+    degradation in the roscore when using HTTP 1.1
+    """
     if platform.system() != 'Linux':
         return True
-
+    minimum_supported_major, minimum_supported_minor = (4, 16)
     release = platform.release().split('.')
-
-    major = 4
-    minor = 15
-
     platform_major = int(release[0])
     platform_minor = int(release[1])
-
-    if platform_major < major:
+    if platform_major < minimum_supported_major:
         return False
-    if platform_major == major and platform_minor <= minor:
+    if (platform_major == minimum_supported_major and
+        platform_minor < minimum_supported_minor):
         return False
     return True
 
--- a/tools/rosgraph/src/rosgraph/xmlrpc.py
+++ b/tools/rosgraph/src/rosgraph/xmlrpc.py
@@ -78,34 +78,35 @@ def isstring(s):
         return isinstance(s, str)
 
 
-def _check_linux_kernel(major, minor):
-    """Small helper version to check if the current Linux kernel is older than
-    the input version and return true. If not Linux return false.
+def _support_http_1_1(major, minor):
+    """Linux kernels 4.15 and older encounter performance issues with HTTP/1.1
+    This function returns true if the Linux kernel version is greater than the
+    input or if the platform is not Linux.
     """
+    # HTTP/1.1 performance issue only detected on Linux
     if platform.system() != 'Linux':
-        return False
+        return True
 
     release = platform.release().split('.')
 
     platform_major = int(release[0])
     platform_minor = int(release[1])
 
-    if platform_major < major:
+    if platform_major > major:
         return True
-    elif platform_major > major:
+    elif platform_major < major:
         return False
 
-    if platform_minor < minor:
+    if platform_minor > minor:
         return True
-    elif platform_minor > minor:
+    elif platform_minor < minor:
         return False
 
     return False
 
 
 class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
-    # Linux kernels 4.15 and older encounter performance issues with HTTP/1.1
-    if not _check_linux_kernel(4,15):
+    if _support_http_1_1(4,15):
         protocol_version = 'HTTP/1.1'
 
     def log_message(self, format, *args):
@@ -118,7 +119,7 @@ class ThreadingXMLRPCServer(socketserver.ThreadingMixIn, SimpleXMLRPCServer):
     requests via threading. Also makes logging toggleable.
     """
 
-    if not _check_linux_kernel(4,15):
+    if _support_http_1_1(4,15):
         daemon_threads = True
 
     def __init__(self, addr, log_requests=1):
--- a/tools/rosgraph/src/rosgraph/xmlrpc.py
+++ b/tools/rosgraph/src/rosgraph/xmlrpc.py
@@ -44,6 +44,7 @@ The common entry point for most libraries is the L{XmlRpcNode} class.
 
 import errno
 import logging
+import platform
 import select
 import socket
 
@@ -78,12 +79,9 @@ def isstring(s):
 
 
 def check_linux_kernel(major, minor):
-    """Small helper version to check that the current platform kernel is a
-    newer or the same as the input
+    """Small helper version to check that the current Linux kernel is a
+    newer or the same as the input.
     """
-    if platform.system()!='Linux':
-        return True
-
     release = platform.release().split('.')
 
     platform_major = int(release[0])
@@ -104,7 +102,7 @@ def check_linux_kernel(major, minor):
 
 class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
     # Linux kernels 4.15 and older encounter performance issues with HTTP/1.1
-    if check_linux_kernel(4,16):
+    if platform.system() == 'Linux' and check_linux_kernel(4,16):
         protocol_version = 'HTTP/1.1'
 
     def log_message(self, format, *args):
@@ -117,7 +115,7 @@ class ThreadingXMLRPCServer(socketserver.ThreadingMixIn, SimpleXMLRPCServer):
     requests via threading. Also makes logging toggleable.
     """
 
-    if check_linux_kernel(4,16):
+    if platform.system() == 'Linux' and check_linux_kernel(4,16):
         daemon_threads = True
 
     def __init__(self, addr, log_requests=1):
--- a/tools/rosgraph/src/rosgraph/xmlrpc.py
+++ b/tools/rosgraph/src/rosgraph/xmlrpc.py
@@ -95,21 +95,15 @@ def _support_http_1_1():
     platform_major = int(release[0])
     platform_minor = int(release[1])
 
-    if platform_major > major:
-        return True
-    elif platform_major < major:
+    if platform_major < major:
         return False
-
-    if platform_minor > minor:
-        return True
-    elif platform_minor < minor:
+    if platform_major == major and platform_minor <= minor:
         return False
-
-    return False
+    return True
 
 
 class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
-    if _support_http_1_1:
+    if _support_http_1_1():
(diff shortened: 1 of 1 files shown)


fujitatomoya:
@ros-pull-request-builder please run the test again

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +76,32 @@ def isstring(s):
...
+    if platform_minor > minor:
+        return True
+    elif platform_minor < minor:
+        return False
+
+    return True
+
+
+class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
+    if check_kernel(4,16):
```
fujitatomoya: I think that this problem only happens on linux platform. I am not windows user but this will break the windows? I was thinking that it would be nice to specify the linux platform on this fix.

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +76,32 @@ def isstring(s):
     except NameError:
         return isinstance(s, str)
 
-class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
 
-    protocol_version = 'HTTP/1.1'
+def check_kernel(major, minor):
+    """Small helper version to check that the current platform kernel is a
```
fujitatomoya: probably it would be nice to keep some description from https://github.com/ros/ros_comm/issues/2118#issue-785600749 for maintenance?

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +76,32 @@ def isstring(s):
...
+    newer or the same as the input
+    """
+    release = platform.release().split('.')
```
fujitatomoya: on windows, it will return `2008ServerR2` against `platform.release()`.

jikawa-az:
Thanks for the review! I will update the PR.

tools/rosgraph/src/rosgraph/xmlrpc.py
jikawa-az: checks specifically for linux now!

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +76,36 @@ def isstring(s):
...
         return isinstance(s, str)
 
-class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
 
-    protocol_version = 'HTTP/1.1'
+def check_linux_kernel(major, minor):
+    """Small helper version to check that the current platform kernel is a
+    newer or the same as the input
+    """
+    if platform.system()!='Linux':
```
fujitatomoya: nitpick

```suggestion
    if platform.system() != 'Linux':
```

fujitatomoya (approved):
one nitpick, lgtm

tools/rosgraph/src/rosgraph/xmlrpc.py
jikawa-az: I moved the check outside the function since it didn't seem right to return true if not Linux.
Also I forgot to import platform so hopefully the tests pass this time.

jikawa-az:
Looks like the checks passed, if it looks good can we merge? Also whats the process for backport, should I just open another PR for melodic-devel?

fujitatomoya:
> if it looks good can we merge?

i am good to go with this, but i am not maintainer... @sloretz @jacobperron friendly ping.

> Also whats the process for backport, should I just open another PR for melodic-devel?

i think backport will be taken care of CI.

jikawa-az:
Awesome, thanks for all the help!!

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +77,33 @@ def isstring(s):
...
+        return True
+    elif platform_minor < minor:
+        return False
+
+    return True
+
+
+class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
+    # Linux kernels 4.15 and older encounter performance issues with HTTP/1.1
+    if platform.system() == 'Linux' and check_linux_kernel(4,16):
```
emersonknapp: tiny nitpick - i think you could have the `platform.system() == 'Linux'` inside the `check_linux_kernel` function, it seems appropriate to have that utility `return False` on non-Linux systems

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +77,33 @@ def isstring(s):
     except NameError:
         return isinstance(s, str)
 
-class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
 
-    protocol_version = 'HTTP/1.1'
+def check_linux_kernel(major, minor):
```
emersonknapp: This could be communicated as purely internal `_check_linux_kernel`

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +77,36 @@ def isstring(s):
     except NameError:
         return isinstance(s, str)
 
-class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
 
-    protocol_version = 'HTTP/1.1'
+def _check_linux_kernel(major, minor):
+    """Small helper version to check if the current Linux kernel is older than
```
emersonknapp: This function's purpose / return value is not clear from the function name.. maybe `_linux_kernel_at_most` instead? A `:returns:` notice in the docstring would also help clarify.

You could also make the fn just be called `support_http_1_1` and then heavily comment your logic inside that function.

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +77,37 @@ def isstring(s):
...
+    if platform_minor > minor:
+        return True
+    elif platform_minor < minor:
+        return False
+
+    return False
+
+
+class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
+    if _support_http_1_1(4,15):
```
emersonknapp: these arguments don't make sense anymore given the new function name

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -106,7 +109,7 @@ def _support_http_1_1(major, minor):
 
 
 class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
-    if _support_http_1_1(4,15):
+    if _support_http_1_1:
```
emersonknapp: this is a function - don't you need to call it?

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +77,40 @@ def isstring(s):
...
+
+    release = platform.release().split('.')
+
+    major = 4
+    minor = 15
+
+    platform_major = int(release[0])
+    platform_minor = int(release[1])
+
+    if platform_major > major:
```
emersonknapp: Are we going for ">4.15" as the target of this logic? Just thinking out loud, but I had a mildly difficult time getting that from the blocks below - the following might be easier to read

```
if platform_major < major:
    return False
if platform_major == major and platform_minor <= minor:
    return False
return True

```

jinmenglei:
@jikawa-az 
if _support_http_1_1():
like this？

tools/rosgraph/src/rosgraph/xmlrpc.py
jikawa-az: arg, dumb error

jikawa-az:
Hi @sloretz @jacobperron, when you get a chance may we please have a review? I would like to have this fix merged to address a customer's performance issues.
Thank you!

tools/rosgraph/src/rosgraph/xmlrpc.py
```diff
@@ -76,9 +77,32 @@ def isstring(s):
...
 
-class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
 
-    protocol_version = 'HTTP/1.1'
+def _support_http_1_1():
+    """
+    Determine whether HTTP 1.1 should be enabled for XMLRPC communications.
+
+    This will be true on non-Linux systems, and on Linux kernels at least as
+    new as 4.16. Linux kernels 4.15 and older cause significance performance
```
emersonknapp:
```suggestion
    new as 4.16. Linux kernels 4.15 and older cause significant performance
```

emersonknapp (approved):

jikawa-az:
Friendly ping: @sloretz @jacobperron

jacobperron (approved):
Apologies for the delay. The change LGTM.

schadocalex:
Update after @Crcodlus comment
https://github.com/ros/ros_comm/issues/2118#issuecomment-845700027 
> the memory usage of rosmaster unfortunately increases continuously and the memory is not going to be released anymore.

I have the exact same problem. Any use of the rosmaster API, even a simple getPid, increases the memory without any release.

After days of investigation, I found out it was caused by ros_comm itself and not something in my project. I did a git-bisect and "8b4089917fef19ab9fd0e00f11ae06235ed7380b is the first bad commit". I work at a company with robots, and all robots updated in >= 1.15.10 with a linux kernel < 4.16 have the bug. In our project it increases the memory by 3GB/hour and it freezes sometimes the communication between nodes.

I will have to force 1.15.9 or update linux kernel on all affected robots, I think that is a bit critical!

emersonknapp:
That's not good! 
It's very strange - have you tried building from source, do you know if it's the HTTP protocol version, or perhaps the `daemon_threads` option, that is introducing the memory issue? Or only the combination of both together? 

This change to HTTP1.1 that went in from Kinetic to Lunar made it so that our service calls would reliably fail - is this application old enough that you happened to be running it on Kinetic? If so, did you experience any performance issues when you upgraded to Lunar/Melodic/Noetic?

This issue is especially hard to test, because even containerizing doesn't help since Docker shares the host kernel.

schadocalex:
Yes I was building from sources to test the 1.15.9 version and then to do the git-bisect.

I've just tested, and it is the `daemon_threads` option. When I force it to True, the memory issue disappears. We may force it all the time, I'm not really sure why it was conditioned by the http 1.1 support.

I can't tell you about any performance issue, I didn't notice anything in the past (we were on Kinetic then Melodic and now Neotic) since this memory issue (that also triggers communication timeouts between some nodes).

schadocalex:
Even if the code has changed, I think it is related to that: https://salsa.debian.org/debian/python-prometheus-client/commit/5aa256d8aab3b81604b855dc03f260342fc391fb

emersonknapp:
>  it is the `daemon_threads` option

That's good to know! I see no reason why `daemon_threads` should depend on the HTTP version. The only reason the changes were grouped together was to "switch off" the change in https://github.com/ros/ros_comm/pull/1287/files, which we observed introduced the problem. Perhaps you could open a followup PR making `daemon_threads` non-conditional?

schadocalex:
Sure: https://github.com/ros/ros_comm/pull/2165

peci1:
Please, see https://github.com/ros/ros_comm/issues/2182 - this fix has unwanted side-effects on 4.x systems: it does not allow setting parameters larger than 32 kB from roscpp.

peci1:
@jikawa-az would you be able to write down a way to tell whether the performance hit you observed is resolved or not? there are several other issues around the change this PR made and it would be good to know that solutions to those issues do not break the improvement your PR brought

peci1:
@jikawa-az or is the test app from #2118 still a good test for this?

peci1:
Okay, I confirm that the test app from #2118 still fails on a Jetson with kernel 4.9 if `_support_http_1_1()` is forced to return `True`. When it return `False` (as is done in this PR), the test app can run for a long time without failure. However, there's still issue #2182 caused by this PR.

emersonknapp:
Thanks for confirming that the test app from #2118 still fails. That's the only repro we currently have for the issue

PR #2134, opened 2021-02-16
Add a rosbag tool 'transform' to split, merge and transform bag files

Earthwings:
Post-processing bag files with `rosbag filter` or the rosbag Python API is very flexible. But when bag files grow large (dozens or hundreds of GB), the processing time becomes significant. This pull request adds a new rosbag tool `transform`, which is written in C++ and performs considerably faster. It provides common filter operations and additional transformations not possible with `rosbag filter` alone (namely bag file merging and topic renaming).

### Examples
```
  # remove the /clock and the /diagnostics topics
  rosbag transform -i input.bag --exclude /clock /diagnostics -o output.bag

  # rename topic from /foo to /bar
  rosbag transform -i input.bag -o output.bag -r /foo:=/bar

  # extract 5.5 seconds from input.bag 10 seconds after the start
  rosbag transform -i input.bag -s 10 -u 5.5 -o output.bag

  # merge everything from input1.bag and input2.bag
  rosbag transform -i input1.bag -i input2.bag -o merge.bag

  # split the /camera/video topic between bag files
  rosbag transform -i input.bag \
                   -o split1.bag --include /camera/video \
                   -o split2.bag --exclude /camera/video
```

### Main commit

> Add a rosbag tool 'transform' to split, merge and transform bag files
> 
> rosbag transform is a bit similar to rosbag filter with two main
> differences:
> * transform can work on multiple input and also output files, and this
>   way it can merge bag files and split bag files in a single invocation
> * transform is much faster since
>   * it does not have to process the whole input files always (for time
>     slice filters)
>   * it is written in C++
>   For example, a 4.5 GB bag file that merely gets copied takes 216s
>   with 'rosbag filter' on my machine and 122s with 'rosbag transform',
>   each with cold caches (echo 1 > /proc/sys/vm/drop_caches). For warm
>   caches, the times are 95s vs. 31s. For time slice extraction, the
>   difference is even larger.
> 
> Possible transformations to apply to input and output bag files are
> topic filtering (via inclusion and exclusion lists), topic renaming,
> time slice extraction (via start and duration) and compression.
> 
> Signed-off-by: Dennis Nienhüser <dennis.nienhueser@de.bosch.com>

### DIFF
--- a/tools/rosbag/src/rosbag/rosbag_main.py
+++ b/tools/rosbag/src/rosbag/rosbag_main.py
@@ -33,6 +33,7 @@
 from __future__ import print_function
 
 import optparse
+import argparse
 import os
 import shutil
 import signal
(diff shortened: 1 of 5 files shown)


tools/rosbag/src/rosbag/rosbag_main.py
```diff
@@ -33,6 +33,7 @@
 from __future__ import print_function
 
 import optparse
+import argparse
```
Earthwings, line 36: optparse has been deprecated in Python 3.2 in favour of argparse. I still tried using optparse in the first place for code consistency here, but optparse lacks nargs='+' support. To the user the difference shouldn't be visible.

Earthwings:
Thoughts, anyone?

saumilsdk:
Nice feature.

valschmidt:
+1 for this feature!.

sloretz:
Thank you for the PR!

Issue #2126, opened 2021-02-05
Duplicate messages when reindexing a rosbag with Python API


Ecophagy:
# Description
After using `reindex()` and then calling `read_messages()`, the resulting messages will be duplicated (so for 100 messages, the list will contain messages 1-100 followed again by 1-100). I had this happen specifically with a version 200 bag, but at a glance it looks like it will happen with all bag versions.

This is because as part of `reindex()`, `_clear_index()` is called. This sets `_connection_indexes_read = False`. The indexing then refills the `_connection_indexes` list but crucially leaves `_connection_indexes_read` as `False`. So when `read_messages` is called, and `_get_indexes()` [is called](https://github.com/ros/ros_comm/blob/f4ab23fe6d9e64fcc2d08cbd88ba368f5188b131/tools/rosbag/src/rosbag/bag.py#L1392) the key code is: (line 1396)
```python
if not self._connection_indexes_read:
    self._reader._read_connection_index_records()
```
`_connection_indexes_read` being False causes `_reader._read_connection_index_records()` to run despite the connection index having been populated in the reindexing. The loop-and-extend nature of the `_reader._read_connection_index_records()` means it just duplicates all the indexes (and therefore messages).

# Code to reproduce
```python
import rosbag

 in_bag = rosbag.Bag(path_bag_file, 'a', allow_unindexed=True)
 for offset in in_bag.reindex():
        pass
messages = list(in_bag.read_messages(topics=None)) 
```` 
You will see the messages contain duplicates. 

# Proposed Fix
Adding `_connection_indexes_read = True` somewhere in the `reindex()` process. Or if this is intended behaviour, then documenting the workaround below would be useful because I spent 3 days trying to track this down!

# Workaround
 You can avoid this problem by closing then reopening the bag after reindexing it. The process of opening a bag rebuilds the connection index and sets `_connection_indexes_read = True`.
```python
import rosbag

in_bag = rosbag.Bag(path_bag_file, 'a', allow_unindexed=True)
for offset in in_bag.reindex():
       pass
in_bag.close()
in_bag = rosbag.Bag(path_bag_file, 'r', allow_unindexed=False
messages = list(in_bag.read_messages(topics=None)) 
````

jinmenglei:
After my previous PR is completed, I will submit a new one to fix this problem

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
[...]

PR #2135, opened 2021-02-18
fix bug #2126:Duplicate messages when reindexing a rosbag with Python API

jinmenglei:
fix bug #2126 

when `_reindex_read_chunk()` is called, `_connection_indexes` is assigned, but `_connection_indexes_read`   is not set to True.

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -2302,7 +2302,7 @@ class _BagReader102_Indexed(_BagReader102_Unindexed):
                 _skip_record(f)
 
             offset = f.tell()
-            
+
         self.bag._connection_indexes_read = True
 
     def start_reading(self):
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
it seems to be right to me, but what about `_BagReader102_Unindexed` and `_BagReader102_Indexed` looks like they have the same problem? I think that this can only fixes `_BagReader200`.

jinmenglei:
> it seems to be right to me, but what about `_BagReader102_Unindexed` and `_BagReader102_Indexed` looks like they have the same problem? I think that this can only fixes `_BagReader200`.
Thanks for the reminder. I'll submit it again

jinmenglei:
@ros-pull-request-builder please run the test again

jinmenglei:
@fujitatomoya @jacobperron friendly ping.

sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #2143, opened 2021-03-08
fix issue 2141: raw_input does not exist in python 3

### LINKED ISSUE #2141: raw_input does not exist in python3
[rosbag_main.py ](https://github.com/ros/ros_comm/blob/b6c23e1282815823c8e07b47384391cf7cf0eab4/tools/rosbag/src/rosbag/rosbag_main.py#L554) and [makerule.py](https://github.com/ros/ros_comm/blob/b6c23e1282815823c8e07b47384391cf7cf0eab4/tools/rosbag/scripts/makerule.py#L136) still use `raw_input()`, which was renamed to `input()` in python 3. Both do not work on noetic with python 3.

bst:
This fixes issue #2141 by using `input()` instead of `raw_input()` and simulating python 3's `input() for python 2.

### DIFF
--- a/tools/rosbag/scripts/makerule.py
+++ b/tools/rosbag/scripts/makerule.py
@@ -42,6 +42,12 @@ import rosbag.migration
 import genpy.message
 import genpy.dynamic
 
+# raw_input was renamed to input in Python 3.x
+try:
+    input = raw_input  # Python 2.x: Simulate 3.x behavior.
+except NameError:
+    pass  # Python 3.x: There is no raw_input, use input directly.
+
 def print_trans(old, new, indent):
     from_txt = '%s [%s]' % (old._type, old._md5sum)
     if new is not None:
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
lgtm

tools/rosbag/scripts/makerule.py
```diff
@@ -42,6 +42,12 @@
 import genpy.message
 import genpy.dynamic
 
+# raw_input was renamed to input in Python 3.x
+try:
+    input = raw_input  # Python 2.x: Simulate 3.x behavior.
+except NameError:
+    pass  # Python 3.x: There is no raw_input, use input directly.
```
jacobperron: Do we need to make this backwards compatible?

As I understand, since ROS Noetic we're only expected to support Python 3 (see [REP 3](https://www.ros.org/reps/rep-0003.html#noetic-ninjemys-may-2020-may-2025).

cc/ @sloretz

sloretz: Correct, on the noetic branch there's no need for the backwards compatibility. Some people use Python 3 with older ROS distros (like arch melodic users), so mabye @bst added it in hopes for a backport to melodic?

fujitatomoya: Ah, i wasn't aware of that. thanks for pointing that out 👍

bst: Good point, I was actually not aware that we here do not need to support Python 2 anymore.

I have no preference in any direction. If you prefer I can also remove the workaround for supporting Python 2.

jacobperron: I would remove the workaround here, and if there is desire to backport to Melodic, we can consider the workaround in a backport PR.

fujitatomoya: agree. we would not want to keep something will never be used.

bst: Makes sense to me. I just removed the python2 compatibility workaround I had added before.

jacobperron (approved):
Thanks for the fix!

PR #2144, opened 2021-03-17
Fix for #152 add a an optional timeout on service calls

alainsanguinetti:
A timeout global is defined for XmlRpc connections.
Overrides are created to allow users to specify a timeout for specific calls.
In particular for ros::master::check a new version is created to allow users to check if the master is up without forever blocking their app.
This fix preserves current functionality but allows user to switch on a global timeout by setting it in the XmlRpcClient class.

### DIFF
--- a/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
+++ b/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
@@ -50,13 +50,28 @@ namespace XmlRpc {
     //!  @param method The name of the remote procedure to execute
     //!  @param params An array of the arguments for the method
     //!  @param result The result value to be returned to the client
-    //!  @return true if the request was sent and a result received 
+    //!  @return true if the request was sent and a result received
     //!   (although the result might be a fault).
     //!
     //! Currently this is a synchronous (blocking) implementation (execute
     //! does not return until it receives a response or an error). Use isFault()
     //! to determine whether the result is a fault response.
+    //!
+    //! This will call the method using the value of executeTimeout as the timeout parameter value
     bool execute(const char* method, XmlRpcValue const& params, XmlRpcValue& result);
+
+    //! Execute the named procedure on the remote server.
+    //!  @param method The name of the remote procedure to execute
+    //!  @param params An array of the arguments for the method
+    //!  @param result The result value to be returned to the client
+    //!  @param timeout The timeout that the poll implementation will use
+    //!  @return true if the request was sent and a result received
+    //!   (although the result might be a fault).
+    //!
+    //! Currently this is a synchronous (blocking) implementation (execute
+    //! does not return until it receives a response or an error, or the socket connection timed out).
+    //! Use isFault() to determine whether the result is a fault response.
+    bool execute(const char* method, XmlRpcValue const& params, XmlRpcValue& result, const double timeout);
 
     bool executeNonBlock(const char* method, XmlRpcValue const& params);
     bool executeCheckDone(XmlRpcValue& result);
@@ -70,7 +85,7 @@ namespace XmlRpc {
     virtual void close();
 
     //! Handle server responses. Called by the event dispatcher during execute.
-    //!  @param eventType The type of event that occurred. 
+    //!  @param eventType The type of event that occurred.
     //!  @see XmlRpcDispatch::EventType
     virtual unsigned handleEvent(unsigned eventType);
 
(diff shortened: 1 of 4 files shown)


utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
```diff
@@ -130,6 +145,14 @@ namespace XmlRpc {
     // Event dispatcher
     XmlRpcDispatch _disp;
 
+  protected:
+    //! general timeout passed to the work method of the dispatcher, to allow avoiding calls blocking forever
+    static double executeTimeout; // s, -1 means infinite wait
```
fujitatomoya, line 150: why this is declared as `protected`, not `private`?

fujitatomoya (commented):
this makes sense to me, poll(2) system call does not wait forever for the events via XmlRpcClient request. but this seems to be unlikely to merge into mainline because of breaking API/ABI, if i am not mistaken.

utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
alainsanguinetti: I aligned with the rest of the code, I didn't see any reasons to make it private instead of protected.
Maybe the code needs some fixing, since most members are public.

alainsanguinetti:
> this makes sense to me, poll(2) system call does not wait forever for the events via XmlRpcClient request. but this seems to be unlikely to merge into mainline because of breaking API/ABI, if i am not mistaken.

I think it only breaks ABI. But I don't know if ABI is a big concern in this library since it doesn't look like it was designed with this concern in mind, for example no use of d-pointer.

fujitatomoya:
right, only ABI breaks. to support the product, ABI compatibility is really important. i am not sure if the mainline takes this fix with breaking it. i would not do that though. but thanks for bringing this up, at least fix makes sense to me.

alainsanguinetti:
@jacobperron how does the review and integration process goes?

Hugal31:
Tell me if I am wrong, but I don't see any ABI-breaking changes in this PR.

This PR only adds functions, methods and some static members, and do not changes any existing function, nor adds, removes or reorders class members or virtual methods.

fujitatomoya:
> This PR only adds functions, methods and some static members

thanks for pointing that out, i think that is correct. this does not break ABI.

utilities/xmlrpcpp/src/XmlRpcClient.cpp
```diff
@@ -303,13 +314,13 @@ XmlRpcClient::generateRequest(const char* methodName, XmlRpcValue const& params)
       body += params.toXml();
       body += PARAM_ETAG;
     }
-      
+
```
jacobperron, line 317: Please avoid unrelated whitespace changes. This helps keep the diff small and focused.

jacobperron (commented):
Could you provide a reproduction of the original issue that this is supposed to fix? Ideally, as a test case.

alainsanguinetti:
> Could you provide a reproduction of the original issue that this is supposed to fix? Ideally, as a test case.

Hello @jacobperron the case is the following : client and master are on two different machines, if suddenly the wifi of the master disconnects, the service calls of the client can hang because execute call uses -1 as the timeout so it will not return until the socket sends an event and it will only happen after a quite long duration.
The proposition of this PR is to actually use the timeout feature intended with a global timeout value that the user can set.
Is it possible to add it as test case somewhere ? It can be reproduced with two docker instances I think

utilities/xmlrpcpp/src/XmlRpcClient.cpp
alainsanguinetti: You're right, I'll be more careful next time. I can edit the commit also if you want

jacobperron: Updating this PR to remove the unrelated whitespace changes would be much appreciated :slightly_smiling_face:

jacobperron:
I was hoping that we could write a unit test involving a single host. I think it might be overkill to introduce docker as a test dependency. I'll at least try to reproduce the setup locally with docker and report back here.

amilcarlucas:
ping. How is this progressing?

5730289021-NN:
This problem persisted since 2013!

It was first posted on #152 then the fix was continued on #738, #1165, #1271, #1282. Unfortunately, it was never got merged.
I'd like to lend a hand if needed. 

If not now, then this may never be patched forever. 

@jacobperron @fujitatomoya @sloretz @mjcarroll Please reconsider checking this PR.

sloretz:
Thank you for the PR!

alainsanguinetti:
@sloretz this is literally a bug fix ........

PR #2145, opened 2021-03-21
github action for noetic

lucasw:
Maybe there are other steps to do here to match more of what is happening in jenkins?  I can view raw console output in jenkins (e.g. https://build.ros.org/job/Npr__ros_comm__ubuntu_focal_amd64/455/console)  but I'm not sure where I can view the ros_comm jenkins script/s that made that happen- it's in another repo (buildfarm?)?

If there are other core/base ros packages that use github actions I can look at those and following yml naming schemes or copy in boilerplate.

Run tests, but don't run linting because no packages here use roslint and that results in an error (but some extra bash scripting could handle that), but leave it here in case one does in the future.

I notice the action is running in my fork, but not here in the PR, maybe that's a repo setting (or a consequence of the repo type or the `ros` github organization type- the action minutes would get billed?) or I need to add something to the action.  Even if the action doesn't run here it would still be useful to run in a fork prior to making a PR here, since the jenkins builds don't and shouldn't run in forks.

### DIFF
--- a/.github/workflows/ubuntu_20_04.yml
+++ b/.github/workflows/ubuntu_20_04.yml
@@ -2,6 +2,7 @@ name: Ubuntu 20.04 Noetic ros_comm
 
 on:
   push:
+  pull_request:
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/ubuntu_20_04.yml
+++ b/.github/workflows/ubuntu_20_04.yml
@@ -1,0 +1,79 @@
+name: Ubuntu 20.04 Noetic ros_comm
+
+on:
+  push:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    env:
+      ROS_CI_DESKTOP: "`lsb_release -cs`"  # e.g. [trusty|xenial|...]
+      ROS_DISTRO: noetic
+    steps:
+      - name: ros_comm
+        uses: actions/checkout@v2
+        with:
+          path: catkin_ws/src/ros_comm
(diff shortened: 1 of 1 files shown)


jacobperron:
To clarify, is the reason for adding this so that you can get CI in forks? My concern with this is the added maintenance burden for troubleshooting potential issues specific to the GitHub CI versus only Jenkins.  I think it's also easy for the different CI scripts to go out of sync.

lucasw:
I'd like to use it in forks, and it is easier to see the results in actions so it would be nice here too.  The maintenance shouldn't be too bad, an average github PR author/maintainer is going to be more familiar with actions than the buildfarm.  Long term maybe everything noetic gets switched over to actions, then no burden on the buildfarm?

jacobperron:
I'm still hesitant that the action will break at some point while the build farm jobs succeed, and then the action stays broken barring someones time to fix it (I've seen this happen in other repos that have both GitHub action CI and ROS build farm Pr jobs). We should get feedback from the other maintainers @sloretz @mjcarroll 

I doubt the Jenkins-based buildfarm is going away anytime soon as there's a lot going on there and I think migrating to a different system (like GitHub actions) would be an incredible undertaking. FWIW, here is a GitHub action for ROS CI that might be of interest: https://github.com/ros-tooling/action-ros-ci. I'm not sure that it mimics exactly what's happening in the build farm Pr jobs, but if we want to add a GitHub action for CI it might be better to leverage the existing action.

sloretz:
Thank you for the PR!

PR #2160, opened 2021-06-18
fix error in base64EncodedSize()

mjs973:
I have been debugging an issue where no messages are delivered over UDP Transport. I'm on melodic, ubuntu 18.04, x86_64 cpu. Here is what I found.

The base64 encoder called by XmlRpcValue.cpp may write two newlines to the end of the encoded string, but this is not accounted for in base64EncodedSize(). This results in a buffer overflow, which then causes ROS UDP Transport connection setup to fail.
    
One case which results in the double newline is when the total length of "node_name + topic_name + msg_type" is 33 bytes.

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcValue.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcValue.cpp
@@ -466,6 +466,8 @@ namespace XmlRpc {
       encoded += (encoded + 71) / 72;
       // for some input, the encoder puts two newlines at the end
       encoded += 1;
+      // add a buffer safety margin
+      encoded += 16;
       return encoded;
     }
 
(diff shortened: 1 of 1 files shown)


jacobperron:
Thanks for the patch! Could you provide a complete code example that I can run to verify the bug and fix?

Also, could you target the latest development branch (`noetic-devel`), and then I'll backport the fix to Melodic.

JavierUbago:
I've had this same issue in ROS noetic, my question with a code example that you can run is here: [https://robotics.stackexchange.com/questions/104346/issue-with-rostopic-name-length-affecting-node-communication](url))

sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #1571, opened 2019-01-07
Added a watcher thread handling logging of child's stderr

betab0t:
As described in issue #1566 - roslaunch wont log child's stderr to ~/.ros/ros/<node_name>-stderr.log.
This PR fixes the issue and enables both printing to screen and logging to file simultaneously.

Issue #1566, opened 2018-12-19
No project_name.node.log file generation when use C++


mxpfeng:
Dear all,
When I use C++ compile node, there is no project_name.node.log file generation in ./ros/log.
![log](https://user-images.githubusercontent.com/18302839/50193942-b851c600-0373-11e9-8007-be4f874eb1f6.png)
There is only stdout.log file, but I read the code find that 
https://github.com/ros/ros_comm/blob/c646e0f3a9a2d134c2550d2bf40b534611372662/tools/roslaunch/src/roslaunch/nodeprocess.py#L243-L245
I dont't understand what's the reason to judge “is_child_mode”. If  wipe off this judge the stderr.log will generation although node.log still doesn't.
Could you help me? Thanks a lot for you.

betab0t:
Hi @mxpfeng!

1. You are correct, there is a conflict here between the docs and the code you referenced...
- [docs say stderr should go both to screen and file if output="log"](http://wiki.ros.org/roslaunch/XML/node) :
> If 'log', the stdout/stderr output will be sent to a log file in $ROS_HOME/log, and stderr will continue to be sent to screen.
 - but in fact the code also checks if node executed on a remote machine('[is_child_mode](https://github.com/ros/ros_comm/blob/89ca9a7bf288b09d033a1ab6966ef8cfe39e1002/tools/roslaunch/src/roslaunch/core.py#L73)') in order for it to log stderr to file.

2. I didn't quite get what is the expected result for you... can you show an example maybe?

mxpfeng:
Hi @betab0t ,
  First of all, thanks for your reply. I just like to generate log files named after nodes in $ROS_HOME/log.
  The same as rospy node, like this:
  
![log](https://user-images.githubusercontent.com/18302839/50578868-739d5300-0e7a-11e9-8186-f018a295a52e.png)
  But I found many ways but could not solve it. So I read the source code, found the above problem. I think maybe stdout/stderr log file can solve the problem instead of node name log file? 
  Thank you again for your help.

betab0t:
@mxpfeng thank you, now i understand :+1: ... so there are two different issues here:
1. roscpp node wont log the ~/.log/log/<node_name>.log (seems like roscpp issue)
2. node(both roscpp & rospy) wont log error messages to ~/.log/log/<node_name>-stderr.log when using `output="log"` (roslaunch issue)

ill try to provide PR soon!

mxpfeng:
[...]

PR #2168, opened 2021-07-02
Added a thread to both print & log child process stderr

MirkoFerrati:
Noetic rebase of https://github.com/ros/ros_comm/pull/1571 as asked by @dirk-thomas the original PR. https://github.com/ros/ros_comm/pull/1571#issuecomment-702949261

Original issue was https://github.com/ros/ros_comm/issues/1566

### DIFF
--- a/tools/roslaunch/src/roslaunch/nodeprocess.py
+++ b/tools/roslaunch/src/roslaunch/nodeprocess.py
@@ -42,6 +42,8 @@ import signal
 import subprocess 
 import time
 import traceback
+import threading
+import sys
 
 import rospkg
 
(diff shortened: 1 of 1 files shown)


tools/roslaunch/src/roslaunch/nodeprocess.py
```diff
@@ -163,6 +165,19 @@ def create_node_process(run_id, node, master_uri, sigint_timeout=DEFAULT_TIMEOUT
...
+     """
+     Watcher for stderr of child process so we can print to screen while simultaneously log to file
+     """
+     while True:
+        buf = popen.stderr.readline()
+        sys.stderr.write(buf)
+        logfile.write(buf)
+
+        if not buf:
+            ret = popen.poll()
```
fujitatomoya, line 178: do we need to check if the child process is still alive for every time? what if child process exits after copying buffer from readline, it will call next loop then exception?

tools/roslaunch/src/roslaunch/nodeprocess.py
```diff
@@ -269,9 +282,9 @@ def _configure_logging(self):
                 mode = 'a'
             else:
                 mode = 'w'
+
```
fujitatomoya, line 285: probably it would be nice to avoid unnecessary difference.

sdo91:
any updates on this?

sloretz:
Thank you for the PR!

PR #2169, opened 2021-07-13
Fix warning related to Boost bind placeholders declared in global namespace.

edowson:
This pull request fixes the following warning in `clients/roscpp/include/ros/publisher.h` and `clients/roscpp/include/ros/node_handle.h`:

Warning: The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define  BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.

### DIFF
--- a/clients/roscpp/include/ros/node_handle.h
+++ b/clients/roscpp/include/ros/node_handle.h
@@ -47,7 +47,7 @@
 #include "ros/init.h"
 #include "common.h"
 
-#include <boost/bind.hpp>
+#include <boost/bind/bind.hpp>
 
 #include <xmlrpcpp/XmlRpcValue.h>
 
--- a/clients/roscpp/include/ros/publisher.h
+++ b/clients/roscpp/include/ros/publisher.h
@@ -32,7 +32,7 @@
 #include "ros/common.h"
 #include "ros/message.h"
 #include "ros/serialization.h"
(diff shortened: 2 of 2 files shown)


jacobperron (approved):
Thanks!

jacobperron:
Unfortunately, this caused several regressions downstream. It seems many packages were depending on the old header ([for example](https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__image_transport__ubuntu_focal_amd64__binary/20/)). While it probably makes sense for the downstream packages to be patched, in the meantime I'm going to revert this PR to unblock the buildfarm (https://github.com/ros/ros_comm/pull/2187).

edowson:
I would recommend fixing downstream packages and moving forward with the new headers, instead of taking the quick way out and reverting the original commit.

jacobperron:
+1 for fixing downstream packages, however I think it will take some time to get all of the fixes in and released. To unblock the buildfarm from producing new debs, I've temporarily reverted the change.

jacobperron:
To avoid out-right breaking downstream packages, I think also doing something as proposed here would be better: https://github.com/ros/ros_comm/pull/2187#issuecomment-925822949

PR #2170, opened 2021-07-14
fix delays from rospy long-running callbacks

### LINKED ISSUE #1901: Inconsistent message queue handling between roscpp and rospy
(Started at https://github.com/CopterExpress/clever/pull/218)

## The problem

While writing some image processing code in Python, we have noticed that if the code is being run inside the callback and is taking a while to complete, the latency would increase significantly over time. This happens even with `queue_size=1` set in publisher and subscriber, and is only partially affected by `buff_size`.

In order to reproduce the problem, I did the following:
* wrote a simple node that subscribed to a message, looked at the difference between `rospy.Time.now()` and the message timestamp, and blocked the callback thread for one second (using `rospy.sleep()`);
* wrote the same code in C++ (I've blocked the thread by using `std::this_thread::sleep_for`);
* looked at the resulting differences.

Typical message "staleness" (the difference between the moment when the callback was called and the message timestamp) was close to the time between two published messages for the C++ variant; in Python, however, it was closer to the time it took for the callback to finish.

## Test setup

I've tried to write a simple test to determine the staleness for different setups; the code is available here: https://github.com/sfalexrog/staleness_test. I also tried running the code in Travis: https://travis-ci.com/sfalexrog/staleness_test/.

There are three variants for publishing nodes (`nodelet` - the node is a C++ nodelet; `node_cpp` - the node is a C++ node; `node_py` - the node is a Python node) and three variants for subscribing nodes (which have the same names). Additionally, there is a variant of the `node_py` node, called a `node_py_largebuf`, which is basically the same as `node_py`, but with `buff_size` set to approximately the size of messages generated by the publisher during the callback.

Publishers simulate a camera that generates 640x480 RGB8 images at 100 Hz. Subsribers simulate image processing nodes with long-running code (the message callback blocks for 1 second). Publishing and subscribing nodelets use the same nodelet manager.

## Test results

Below are the average results from running on Kinetic (https://travis-ci.com/sfalexrog/staleness_test/jobs/291735129); column names describe the publisher type, row names describe the subscriber type:

|   |`nodelet`|`node_cpp`|`node_py`|
|---|---|---|---|
|`nodelet`|0.000253|0.005502|0.006024|
|`node_cpp`|0.005065|0.00517571|0.005521|
|`node_py`|**8.126688**|**8.159472**|**8.077343**|
|`node_py_largebuf`|**0.944017**|**0.944568**|**0.938216**|

Melodic results are similar: https://travis-ci.com/sfalexrog/staleness_test/jobs/291735130

## Questions

Is this behavior expected or documented? I could not find a warning about long-running subscribers in `rospy`, and the behavior is inconsistent with `roscpp`. We *can* work around this problem by creating a separate thread for long-running tasks, but this makes the code more complicated, and we'd prefer a simplier, more reliable Python API.

c-andy-martin:
Long-running callbacks in rospy can cause extreme amounts of buffering
resulting in unnecessary delay, essentially ignoring the queue_size
setting (#1901). This can already be somewhat mitigated by setting
buff_size to be larger than the amount of data that could be buffered
by a long running callback. However, setting buff_size to a correct
value is not possible for the user of the API if the amount of time in
the callback or the amount of data that would be transmitted is
unknown. Also, even with a correct buff_size and a queue_size of 1, the
received data may still be the oldest of all data transmitted while the
callback was running.

Fix the delays in such cases by running callbacks in a separate thread.
The receive_loop then calls recv() concurrently with the long running
callback, enforcing queue_size as new data is received. This fixes the
latency in the data when queue_size is set to be similar to roscpp.

This fixes #1901

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -782,6 +782,31 @@ class TCPROSTransport(Transport):
                 
             time.sleep(interval)
 
+    def callback_loop(self, msgs_callback):
+        while not self.done and not is_shutdown():
+            try:
+                with self.msg_queue_lock:
+                    # Data that was leftover from reading header may have made
+                    # messages immediately available (such as from a latched
+                    # topic). Go ahead and process anything we already have before
+                    # waiting.
+                    while self.msg_queue:
+                        msg = self.msg_queue.pop(0)
+                        # Be sure to not hold the message queue lock while calling
+                        # the callback, it may take a while.
+                        self.msg_queue_lock.release()
+                        msgs_callback([msg], self)
+                        self.msg_queue_lock.acquire()
+                    self.msg_queue_condition.wait()
+            except:
+                # in many cases this will be a normal hangup, but log internally
+                try:
+                    #1467 sometimes we get exceptions due to
+                    #interpreter shutdown, so blanket ignore those if
+                    #the reporting fails
+                    rospydebug("exception in callback loop for [%s], may be normal. Exception is %s",self.name, traceback.format_exc())
+                except: pass
+
     def receive_loop(self, msgs_callback):
         """
         Receive messages until shutdown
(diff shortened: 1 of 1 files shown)


sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #2171, opened 2021-07-15
fix non persistent rospy ServiceProxy is not thread-safe

### LINKED ISSUE #2078: non persistent rospy ServiceProxy is not thread-safe
Relevant code

https://github.com/ros/ros_comm/blob/090daa56f57d18c39946f107bcee5e3efefb2cac/clients/rospy/src/rospy/impl/tcpros_service.py#L481-L537

The self.transport gets set during a service proxy call, and then gets unset if the proxy is not persistent. 

This behavior is not thread safe if multiple threads are trying to use the same service proxy, as it could get unset as another thread is about to use the self.transport variable.

running the following should lead to errors
```python
     #!/usr/bin/env python

    import rospy
    from std_srvs.srv import Empty, EmptyResponse
    from threading import Timer

    class Test():
        def __init__(self):
            rospy.Service('/test', Empty, self.handle_request)
            self.proxy = rospy.ServiceProxy('/test', Empty, persistent=False)
        
        def handle_request(self, request):
            rospy.sleep(0.1)
            return EmptyResponse()
        
        def call_service(self):
            response = self.proxy.call()
            rospy.logwarn(response)
        
    if __name__ == "__main__":
        rospy.init_node("test")
        t = Test()
        a1 = Timer(2.01, t.call_service)
        a2 = Timer(2.02, t.call_service)
        a3 = Timer(2.03, t.call_service)
        a4 = Timer(2.04, t.call_service)
        a5 = Timer(2.05, t.call_service)
        a1.start()
        a2.start()
        a3.start()
        a4.start()
        a5.start()  
        rospy.spin()
```

kevinshchang:
fix #2078 

Add simple fix to threading issue. No new tests were added, but could be done eventually if wanted. 

Works to fix script referenced in issue.

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_service.py
+++ b/clients/rospy/src/rospy/impl/tcpros_service.py
@@ -510,7 +510,8 @@ class ServiceProxy(_Service):
             except TransportInitError as e:
                 # can be a connection or md5sum mismatch
                 raise ServiceException("unable to connect to service: %s"%e)
-            self.transport = transport
+            if self.persistent:
+                self.transport = transport
         else:
             transport = self.transport
 


jacobperron (approved):
LGTM, thanks for the patch!

PR #2173, opened 2021-07-17
Fix $(dirname) resolution in include tags

### LINKED ISSUE #1487: Unexpected $(dirname) behaviour when value assigned to arg.
The existing test is here, added as part of the overall feature in #1103: 

https://github.com/ros/ros_comm/blob/e96c407c64e1c17b0dd2bb85b67f388380527097/tools/roslaunch/test/xml/test-dirname.xml

If the test is augmented to be:

```
<launch>
  <param name="foo" value="$(dirname)/bar" />
  <include file="$(dirname)/test-dirname/included.xml">
    <arg name="baz" value="$(dirname)/baz" />
  </include>
</launch>
```

And then the `included.xml` file is changed to:

```
<launch>
  <arg name="baz" />

  <param name="bar" value="$(dirname)/baz" />
  <param name="baz" value="$(arg baz)" />
</launch>
```

You'd expect the `baz` parameter to have `tools/roslaunch/test/xml/baz` as its path, but instead it's `tools/roslaunch/test/xml/test-dirname/baz`. That is, dirname is being lazily resolved at the point where the value is assigned to the parameter, rather than when the arg value is assigned. Historically this hasn't mattered, as no other substitutions have this kind of context sensitivity.

I'm not sure how much effort this will be to fix, but I had an instance recently where it would have been nice to be able to pass a "back path" into an included file, and wasn't able to do so due to this problem.

rhaschke:
Fixes #1487. The problem was that arg values were evaluated (non-lazily) in the context of the newly included file already.
This PR first augments the unit test and then fixes it.

### DIFF
--- a/tools/roslaunch/src/roslaunch/xmlloader.py
+++ b/tools/roslaunch/src/roslaunch/xmlloader.py
@@ -613,6 +613,7 @@ class XmlLoader(loader.Loader):
             # error on) attempts to set the same arg twice.
             child_ns.pass_all_args = True
 
+        child_ns.filename = context.filename  # evaluate substitutions w.r.t. parent filename
         for t in [c for c in tag.childNodes if c.nodeType == DomNode.ELEMENT_NODE]:
             tag_name = t.tagName.lower()
             if tag_name == 'env':
@@ -621,6 +622,7 @@ class XmlLoader(loader.Loader):
                 self._arg_tag(t, child_ns, ros_config, verbose=verbose)
             else:
                 print("WARN: unrecognized '%s' tag in <%s> tag"%(t.tagName, tag.tagName), file=sys.stderr)
+        child_ns.filename = inc_filename  # restore filename
 
         # setup arg passing
         loader.process_include_args(child_ns)
--- a/tools/roslaunch/test/unit/test_xmlloader.py
+++ b/tools/roslaunch/test/unit/test_xmlloader.py
@@ -1080,5 +1080,7 @@ class TestXmlLoader(unittest.TestCase):
         for p in mock.params:
             param_d[p.key] = p.value
 
-        self.assertEquals(param_d['/foo'], self.xml_dir + '/bar')
-        self.assertEquals(param_d['/bar'], self.xml_dir + '/test-dirname/baz')
+        self.assertEqual(param_d['/base'], self.xml_dir)
+        self.assertEqual(param_d['/foo'], self.xml_dir + '/bar')
+        self.assertEqual(param_d['/baz'], self.xml_dir + '/baz')
+        self.assertEqual(param_d['/bar'], self.xml_dir + '/test-dirname/baz')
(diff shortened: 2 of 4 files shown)


rhaschke:
@mikepurvis, please review.

mikepurvis:
Awesome, thanks for tackling this, and I am 😬 at how simple of a fix it ended up being. LGTM.

rhaschke:
Can you comment on whether this might have undesired side effects? I guess the filename isn't used for any other substitution arg, is it?

mikepurvis:
It's definitely a change in end-user behaviour, but I can't image anyone relying on the current behaviour in this context since it's pretty counterintuitive.

As far as in-code side effects, I don't know enough about the roslaunch internals and frankly what I do know terrifies me. But fortunately the test coverage is decent, so I wouldn't expect breakages from this if the tests pass.

That said, it would be interesting to confirm if `roslaunch-check` works as expected here— there's some curious bits of code duplication between what roslaunch and roslaunch-check run.

rhaschke:
@jacobperron, I just came across this PR again and wondered that this is still open. Are there any concerns?

jacobperron (commented):
@rhaschke Thanks for the ping!

In the code block where we temporarily override the context filename, it looks like it may have some other side effects like changing the output in the case of an unknown attribute:

https://github.com/ros/ros_comm/blob/376ebdb74a4523fb1071f4d51689058bffb01ff9/tools/roslaunch/src/roslaunch/xmlloader.py#L224

I suppose we would rather point to the included launch file in this case, not the parent.

I haven't looked through the code thoroughly, but there may be other cases where the filename is used where we would rather not use the overridden name.

Maybe instead of overriding the `filename` member, we could instead introduce a new member `parent_filename` that we then use in the specific case of the `dirname` substitution? Then we avoid any side-effects that may happen if we modify `filename`. Thoughts?

rhaschke:
> I suppose we would rather point to the included launch file in this case, not the parent.

I don't agree. The `arg` and `env` tags in question still reside in the parent file name and thus should report the parent filename.
Consider the test example augmented with an unknown attribute `foo`:
```xml
<launch>
  <arg name="base" value="$(dirname)" />
  <param name="foo" value="$(dirname)/bar" />
  <include file="$(dirname)/test-dirname/included.xml" pass_all_args="True">
    <arg name="baz" value="$(dirname)/baz" foo=""/>
  </include>
</launch>
```
Currently, i.e. w/o this PR, this yields: 
`WARNING: [tools/roslaunch/test/xml/test-dirname/included.xml] unknown <arg> attribute 'foo'`

With this PR, we correctly get:
`WARNING: [tools/roslaunch/test/xml/test-dirname.xml] unknown <arg> attribute 'foo'`

rhaschke:
@jacobperron, ping again :wink:

jacobperron (approved):

jacobperron:
I took another look and the changes LGTM :+1:

PR #2180, opened 2021-08-26
rosbag: introduce external index file to speed up reading

### LINKED ISSUE #117: rosbag: Better support for large bag files in rosbag
I would like to suggest improving the support for large bag files. The problem that causes a slow start up time is that when Bag::open is called it scans for all the chunks in the bag file. The default chunk size is 768Kb. So for a 30GB file (30 min stereo video) around 40K chunks. I believe what is slowing things down is that each chunks needs a disk seek (~10ms) so it takes several minutes to start playing back the file.

I don't know the details of the Bag library, but here are some suggestions that might serve as a solution to the problem.
1. Allow for a configurable chunk size. e.g. if the chunk

size is 100MB in the example above then only 300 seeks are need and only ~3s spend on seeks instead of ~6min.
1. Run a re-index that makes a copy of all chunk indexes

and puts them to the end of the bag file so that they can be read in on pass.
1. Modify rosbag play so that it starts playing back the

file before finishing reading in all indexes. That is read the index and data in parallel or read everything sequentially and process each chunk header as soon as they are read from the file.

See a related question on answers.ros.org  http://answers.ros.org/question/1813/playing-back-large-files-with-rosbag

migration of trac ticket 3829:
  https://code.ros.org/trac/ros/ticket/3829

chunk_size.patch:

```
Index: include/rosbag/bag.h
===================================================================
--- include/rosbag/bag.h    (revision 16227)
+++ include/rosbag/bag.h    (working copy)
@@ -91,6 +91,8 @@
      */
     Bag(std::string const& filename, uint32_t mode = bagmode::Read);

+    Bag(std::string const& filename, uint32_t mode, uint32_t chunk_size);
+
     ~Bag();

     //! Open a bag file.
Index: include/rosbag/recorder.h
===================================================================
--- include/rosbag/recorder.h   (revision 16227)
+++ include/rosbag/recorder.h   (working copy)
@@ -102,6 +102,7 @@
     uint32_t        max_size;
     ros::Duration   max_duration;
     std::string     node;
+    uint32_t        chunk_size;

     std::vector<std::string> topics;
 };
Index: src/recorder.cpp
===================================================================
--- src/recorder.cpp    (revision 16227)
+++ src/recorder.cpp    (working copy)
@@ -102,7 +102,8 @@
     split(false),
     max_size(0),
     max_duration(-1.0),
-    node("")
+    node(""),
+    chunk_size(1048576 * 1)
 {
 }

@@ -350,6 +351,7 @@

     updateFilenames();
     try {
+        bag_.setChunkThreshold(options_.chunk_size);
         bag_.open(write_filename_, bagmode::Write);
     }
     catch (rosbag::BagException e) {
@@ -487,6 +489,7 @@
         string write_filename  = target_filename + string(".active");

         try {
+            bag_.setChunkThreshold(options_.chunk_size);
             bag_.open(write_filename, bagmode::Write);
         }
         catch (rosbag::BagException ex) {
Index: src/record.cpp
===================================================================
--- src/record.cpp  (revision 16227)
+++ src/record.cpp  (working copy)
@@ -50,6 +50,7 @@
     desc.add_options()
       ("help,h", "produce help message")
       ("all,a", "record all topics")
+      ("chunk-size,c", po::value<int>()->default_value(1), "Chunk size in MB (Default: 1)")
       ("regex,e", "match topics using regular expressions")
       ("exclude,x", po::value<std::string>(), "exclude topics matching regular expressions")
       ("quiet,q", "suppress console output")
@@ -120,6 +121,13 @@
         opts.max_size = 1048576 * S;
       }
     }
+    if (vm.count("chunk-size"))
+    {
+      int chunk_size = vm["chunk-size"].as<int>();
+      if (chunk_size <= 0) 
+        throw ros::Exception("Chunk size must be positive");
+      opts.chunk_size = 1048576 * chunk_size;
+    }
     if (vm.count("buffsize"))
     {
       int m = vm["buffsize"].as<int>();
Index: src/bag.cpp
===================================================================
--- src/bag.cpp (revision 16227)
+++ src/bag.cpp (working copy)
@@ -82,6 +82,22 @@
     open(filename, mode);
 }

+Bag::Bag(string const& filename, uint32_t mode, uint32_t chunk_size) :
+    compression_(compression::Uncompressed),
+    chunk_threshold_(chunk_size),
+    bag_revision_(0),
+    file_size_(0),
+    file_header_pos_(0),
+    index_data_pos_(0),
+    connection_count_(0),
+    chunk_count_(0),
+    chunk_open_(false),
+    curr_chunk_data_pos_(0),
+    decompressed_chunk_(0)
+{
+    open(filename, mode);
+}
+
 Bag::~Bag() {
     close();
 }
```

change_cunk_size.py

```
#!/usr/bin/env python

PKG = 'sensor_msgs'
import roslib; roslib.load_manifest(PKG)
import rospy
import rosbag

with rosbag.Bag('output.bag', 'w', chunk_threshold=100 * 1024 * 1024) as outbag:
    for topic, msg, t in rosbag.Bag('input.bag').read_messages():
        outbag.write(topic, msg, msg.header.stamp if msg._has_header else t)
```

change history:
Changed 11 months ago by hordurj 

I included a patch to rosbag record that adds a chunk-size option. The chunk size is specified in MB, I was trying to follow the convention of other parameters. I also included a simple script that can be used to convert the chunk size of a bag file.

I did a quick test on a ~30GB file that had ~40K chunks. It took around 7 min to load on my machine. When I changed the chunk size to 100MB it took 9 sec.

Is there a disadvantage to having large chunk size?

Changed 11 months ago by hordurj 

One thing that came up when using very large chunk size is that it can cause delays in the playback when a new chunk is loaded in. I fixed that by splitting the file reading and the playback into separate threads. I also added a simple cache manager because I noticed that sometimes a sequence of messages could come from different chunks causing the chunks to be reloaded multiple times. I'm currently just maintaining my own copy of rosbag. I would be happy to share these changes if there is any interest in incorporating them. But this might be just a niche use-case and the added complexity not worth it.

Changed 11 months ago by kwc 

@hordurj: thanks for the patch. We can't integrate this right now as things are frozen for Fuerte. Apologies we didn't get to it sooner, but we were already 4 weeks behind on schedule and had to focus on high priority items.

ghost:
Hello,

we propose to optionally use an external index file e.g. `*.idx` to speedup reading. This is particularly useful for bags which reside on a slow HDD or (even worse) on a mounted network drive. This can be quite slow. Our bags are about 100GB or even larger (they contain a lot of camera images, point clouds, etc).

The "traditional" way of reading a ROS bag requires to read in the whole file as the [index records](http://wiki.ros.org/Bags/Format/2.0#Index_data) (required to locate the individual serialized messages within the file) are spread across the whole file. This is very I/O intensive and requires some amount of time (~4min for 100GB on my HDD). In case of the mounted network drive, rosbag is almost unusable.

Compared to our large bag, the index file is just ~100MB. After reading this index file, the locations of the serialized messages inside the bag are known and the bag can be played normally. Since only 100MB have to be read, a bag on the disk starts to play almost instantly (2-3s). Depending on the bandwidth it is even reasonable to play bag from a network drive. It is just required to load the 100MB index file completely. Afterwards the the large bag file can be loaded sequentially step-by-step.

This index file is essentially a regular bag file, where the [large chunk records](http://wiki.ros.org/Bags/Format/2.0#Chunk), which contain the actual payload of the messages,  are pruned. To get the full meta data information even without chunk records it was additionally necessary to extend the data part of the [index data](http://wiki.ros.org/Bags/Format/2.0#Index_data) records by "chunk offset" (besides `time` and  `offset` <- offset of a message within a chunk). It describes the position of the chunk within the bag file. Since this index file has (almost) the format of a regular bag file, the code required to read them is almost the same. Only small changes were necessary.

These index files can be easily generated from an existing bag file. A corresponding script is also added in this pull request. Furthermore, this pull request is fully backwards compatible as reading with index file is only used when a file is found next to the bag (same name but with file extension ".idx"). Otherwise the bag is opened regularly.

Command to generate an index file:
`rosbag index-file 2021-08-09-08-55-29.bag`
Start player normally:
`rosbag play 2021-08-09-08-55-29.bag` (uses generated .`idx` file if it exists)

I think it would also address/fix #117.

### DIFF
--- a/tools/rosbag/src/player.cpp
+++ b/tools/rosbag/src/player.cpp
@@ -535,7 +535,16 @@ void Player::doPublish(MessageInstance const& m) {
         }
     }
 
-    while ((paused_ || delayed_ || !time_publisher_.horizonReached()) && node_handle_.ok())
+    // In rare cases it is possible to reach the time horizon immediately. This
+    // typically happens when a bag is played from a very slow disk or a mounted
+    // network drive. Then the 'publish' (at this point the message is actually
+    // fully read from the file) of the previous message took longer than the
+    // time horizon of the current message. In this case we want to read once
+    // from stdin to be able to pause, step, etc. Furthermore, we want to update
+    // the time and to print it.
+    bool check_stdin_once_ = time_publisher_.horizonReached();
+
+    while ((paused_ || delayed_ || !time_publisher_.horizonReached() || check_stdin_once_) && node_handle_.ok())
     {
         bool charsleftorpaused = true;
         while (charsleftorpaused && node_handle_.ok())
@@ -601,6 +610,12 @@ void Player::doPublish(MessageInstance const& m) {
                 else
                     charsleftorpaused = false;
             }
+        }
+
+        if (check_stdin_once_) {
+            time_publisher_.stepClock();
+            printTime();
+            break;
         }
 
         printTime();
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -1443,12 +1443,19 @@ class Bag(object):
             return isinstance(f, io.IOBase)  # Python 3...this will return false in Python 2 always
 
     def _open_read(self, f, allow_unindexed):
+        use_index_file = False
+
         if self._is_file(f):
             self._file     = f
             self._filename = None
         else:
-            self._file     = open(f, 'rb')
-            self._filename = f        
+            bag_index_path = f[:-len(".bag")] + ".idx"
+            if os.path.exists(bag_index_path):
+                use_index_file = True
+                self._file = open(bag_index_path, 'rb')
+            else:
+                self._file = open(f, 'rb')
+            self._filename = f
 
         self._mode = 'r'
 
@@ -1460,8 +1467,10 @@ class Bag(object):
             raise
 
         try:
-            self._create_reader()
+            self._create_reader(use_index_file)
             self._reader.start_reading()
+            if use_index_file:
+                self._file = open(f, 'rb')
         except ROSBagUnindexedException as ex:
             if not allow_unindexed:
                 self._close_file()
@@ -1533,13 +1542,14 @@ class Bag(object):
         self._file.close()
         self._file = None
 
-    def _create_reader(self):
+    def _create_reader(self, use_index_file=False):
         """
         @raise ROSBagException: if the bag version is unsupported
         """
         major_version, minor_version = int(self._version / 100), self._version % 100
         if major_version == 2:
             self._reader = _BagReader200(self)
+            self._reader.use_index_file = use_index_file
         elif major_version == 1:
             if minor_version == 1:
                 raise ROSBagException('unsupported bag version %d. Please convert bag to newer format.' % self._version)
@@ -2436,6 +2446,8 @@ class _BagReader200(_BagReader):
         self.decompressed_chunk     = None
         self.decompressed_chunk_io  = None
 
+        self.use_index_file = None
+
     def reindex(self):
         """
         Generates all bag index information by rereading the chunks.
@@ -2657,15 +2669,22 @@ class _BagReader200(_BagReader):
 
             # Read the chunk info records
             self.bag._chunks = [self.read_chunk_info_record() for i in range(self.bag._chunk_count)]
-    
-            # Read the chunk headers
-            self.bag._chunk_headers = {}
-            for chunk_info in self.bag._chunks:
-                self.bag._file.seek(chunk_info.pos)
-                self.bag._chunk_headers[chunk_info.pos] = self.read_chunk_header()
 
-            if not self.bag._skip_index:
-                self._read_connection_index_records()
+            if not self.use_index_file:
+                # Read the chunk headers
+                self.bag._chunk_headers = {}
+                for chunk_info in self.bag._chunks:
+                    self.bag._file.seek(chunk_info.pos)
+                    self.bag._chunk_headers[chunk_info.pos] = self.read_chunk_header()
+
+                if not self.bag._skip_index:
+                    self._read_connection_index_records()
+            else:
+                f = self.bag._file
+                index_file_size = os.stat(f.fileno()).st_size
+                while f.tell() < index_file_size:
+                    self.bag._curr_chunk_info = _ChunkInfo(0, 0, 0)  # set dummy
+                    self.read_connection_index_record(True)
 
         except ROSBagEncryptNotSupportedException:
             raise
@@ -2774,7 +2793,7 @@ class _BagReader200(_BagReader):
 
         return _ChunkHeader(compression, compressed_size, uncompressed_size, data_pos)
 
-    def read_connection_index_record(self):
+    def read_connection_index_record(self, extended=False):
         f = self.bag._file
 
         header = _read_header(f, _OP_INDEX_DATA)
@@ -2789,12 +2808,15 @@ class _BagReader200(_BagReader):
         record_size = _read_uint32(f) # skip the record data size
 
         index = []
-                
+        chunk_pos = self.bag._curr_chunk_info.pos
+
         for i in range(count):
             time   = _read_time  (f)
+            if extended:
+                chunk_pos = _read_uint64(f)
             offset = _read_uint32(f)
             
-            index.append(_IndexEntry200(time, self.bag._curr_chunk_info.pos, offset))
+            index.append(_IndexEntry200(time, chunk_pos, offset))
 
         return (connection_id, index)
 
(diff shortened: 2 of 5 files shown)


goldbattle:
I tested by manually building your edited rosbag and rosbag_storage packages and it works as expected for me. Additionally, it works through the C++ interface as expected:
```c++
rosbag::Bag bag;
bag.open(path_to_bag, rosbag::bagmode::Read);
```

In general a >100gb file now opens in around a second, compared to 1-2 minutes I was getting before off my external SSD. This is a great alternative to switching the chunks size that was done for the [TUM-VI dataset](https://vision.in.tum.de/data/datasets/visual-inertial-dataset). Using smaller chunks gives me smooth playback, and with this index file the opening is much faster.
https://vision.in.tum.de/data/datasets/visual-inertial-dataset#faq

goldbattle:
I actually get an error on a bag that I had to reindex on.
```
patrick@patrick-ThinkPad-P51:/media/patrick/RPNG_FLASH_5/d455$ rosrun rosbag rosbag index-file d455_room_07_stereo.bag 
Start reading bag indices of d455_room_07_stereo.bag ...
Generate index file d455_room_07_stereo.idx ...
  File "/home/patrick/workspace/catkin_ws_kalibr/src/rosbag_custom/rosbag/scripts/rosbag", line 35, in <module>
    rosbag.rosbagmain()
  File "/home/patrick/workspace/catkin_ws_kalibr/src/rosbag_custom/rosbag/src/rosbag/rosbag_main.py", line 1168, in rosbagmain
    cmds[cmd](argv[2:])
  File "/home/patrick/workspace/catkin_ws_kalibr/src/rosbag_custom/rosbag/src/rosbag/rosbag_main.py", line 678, in gen_idx_file_cmd
    index_file_op(args, options.force)
  File "/home/patrick/workspace/catkin_ws_kalibr/src/rosbag_custom/rosbag/src/rosbag/rosbag_main.py", line 986, in index_file_op
    op = _read_uint8_field(header, 'op')
  File "/home/patrick/workspace/catkin_ws_kalibr/src/rosbag_custom/rosbag/src/rosbag/bag.py", line 1994, in _read_uint8_field
    def _read_uint8_field (header, field): return _read_field(header, field, _unpack_uint8)
  File "/home/patrick/workspace/catkin_ws_kalibr/src/rosbag_custom/rosbag/src/rosbag/bag.py", line 1983, in _read_field
    raise ROSBagFormatException('expected "%s" field in record' % field)
rosbag.bag.ROSBagFormatException: expected "op" field in record
```

Seems to be due to this line here: https://github.com/ros/ros_comm/blob/981d29320ff6791865062e53fdd24ab760934671/tools/rosbag/src/rosbag/rosbag_main.py#L986

I added some small printouts and got this:

```python
        # copy all records to output bag (except for chunk records and index data records)
        print("Generate index file %s ..." % output_idx_raw_path)
        while input_bag_raw.tell() < input_bag_raw_size:
            header = _read_header(input_bag_raw)
            print(str(input_bag_raw.tell()) + " of " + str(input_bag_raw_size))
            print(header)
            op = _read_uint8_field(header, 'op')
```

Final print before crash:
```
148282766751 of 148306297758
{'conn': b'\x03\x00\x00\x00', 'count': b'\x01\x00\x00\x00', 'op': b'\x04', 'ver': b'\x01\x00\x00\x00'}
148282766812 of 148306297758
{'compression': b'none', 'op': b'\x05', 'size': b'pl\x0c\x00'}
148283581059 of 148306297758
{'conn': b'\x04\x00\x00\x00', 'count': b'\x01\x00\x00\x00', 'op': b'\x04', 'ver': b'\x01\x00\x00\x00'}
148283581079 of 148306297758
{}
```


I will try to upload the bag. Do you have an email I can send the link to?

AndreasR30:
@goldbattle Nice to see, that someone is using this feature. Sorry for the delay, Is this a private Rosbag? Im not sure whether it is a good idea to post my email adress here.

goldbattle:
Email me at ..... and I can send you the link.

Hommus:
Thanks for the PR!

AndreasR30:
@goldbattle Sorry I was quite busy in last weeks. This seems to be rather a bug in the reindex feature, since a record header shouldn't be empty as one can see in your debug output. Of course I could add a try-except around  `_read_uint8_field()` and skip this record. However, this wouldn't fix the original bug/issue. I will try to investigate the reindex feature, how a empty record header can happen.

valschmidt:
+1 for this feature!

lucasw:
@valschmidt have you looked at converting your bags to mcap with [`mcap convert`](https://mcap.dev/guides/cli#ros-bag-to-mcap-conversion)?  They're very similar to bags but are more flexible (they can store ros1 messages, ros2, protobuf, or anything else but you're more on your own), and I think outperform bags for reading back messages by a large amount.

(I'm working on command line tools https://github.com/lucasw/mcap_tools to record directly into mcap and play back messages as easily as rosbag record/play does, though I wouldn't recommend relying on those until they've had a lot more testing)

sloretz:
Thank you for the PR!

PR #2183, opened 2021-09-16
fix typo t_start -> self.t_start

k-okada:
fix typo local t_start is not defined here, need to use self.t_start

Cc: @furushchev


furushchev:
Thank you for the fix! You're totally right!

jmariscalharana:
Could someone merge this, please?

sloretz (approved):
Thank you for the PR!

PR #2187, opened 2021-09-22
Revert "Fix warning related to Boost bind placeholders declared in global namespace. (#2169)"

jacobperron:
This reverts commit 32942993e3f9afc8248be7b1f5944344fa2b4e30.

To unblock the buildfarm that's failing to build downstream packages transitively depending on this header being included.

### DIFF
--- a/clients/roscpp/include/ros/node_handle.h
+++ b/clients/roscpp/include/ros/node_handle.h
@@ -47,7 +47,7 @@
 #include "ros/init.h"
 #include "common.h"
 
-#include <boost/bind/bind.hpp>
+#include <boost/bind.hpp>
 
 #include <xmlrpcpp/XmlRpcValue.h>
 
--- a/clients/roscpp/include/ros/publisher.h
+++ b/clients/roscpp/include/ros/publisher.h
@@ -32,7 +32,7 @@
 #include "ros/common.h"
 #include "ros/message.h"
 #include "ros/serialization.h"
-#include <boost/bind/bind.hpp>
+#include <boost/bind.hpp>
 #include <boost/thread/mutex.hpp>
 
 namespace ros


sloretz (approved):

mjcarroll (approved):

jacobperron:
Here's are failing jobs that this PR should fix:

- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__bondcpp__ubuntu_focal_amd64__binary/24/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__control_toolbox__ubuntu_focal_amd64__binary/25/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__costmap_cspace__ubuntu_focal_amd64__binary/48/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__dwb_plugins__ubuntu_focal_amd64__binary/20/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__image_transport__ubuntu_focal_amd64__binary/20/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__interval_intersection__ubuntu_focal_amd64__binary/5/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__joystick_interrupt__ubuntu_focal_amd64__binary/42/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__laser_filters__ubuntu_focal_amd64__binary/28/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__map_server__ubuntu_focal_amd64__binary/25/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__moveit_core__ubuntu_focal_amd64__binary/39/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__nmea_comms__ubuntu_focal_amd64__binary/9/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__pid__ubuntu_focal_amd64__binary/16/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__power_monitor__ubuntu_focal_amd64__binary/10/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__prbt_hardware_support__ubuntu_focal_amd64__binary/5/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__rc_hand_eye_calibration_client__ubuntu_focal_amd64__binary/23/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__rc_silhouettematch_client__ubuntu_focal_amd64__binary/24/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__rm_base__ubuntu_focal_amd64__binary/3/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__ros_ign_bridge__ubuntu_focal_amd64__binary/19/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__roscpp_tutorials__ubuntu_focal_amd64__binary/32/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__rosrt__ubuntu_focal_amd64__binary/5/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__sick_tim__ubuntu_focal_amd64__binary/33/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__swri_roscpp__ubuntu_focal_amd64__binary/29/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__toposens_driver__ubuntu_focal_amd64__binary/12/
- https://build.ros.org/view/Nbin_uF64/job/Nbin_uF64__visp_camera_calibration__ubuntu_focal_amd64__binary/7/

There are probably several other jobs (that are blocked right now) that would have failed for the same reason.

mikepurvis:
Could there be some kind of compat-type option here where we fix roscpp, and then in the header create our own aliases for _1, _2, etc, but with the `[[deprecated]]` attribute? This would resolve the warnings for the vast majority of cases while still signaling to the above packages that a deprecated usage exists.

jacobperron:
@mikepurvis Thanks for the suggestion! This sounds okay to me. I'm happy to review a PR if you (or someone else) takes the time to implement it.

iwanders:
Ok, to get back to this. I tried implementing @mikepurvis 's solution with something like;

```cpp

#include <boost/bind/bind.hpp>

#define ROSCPP_BOOST_BIND_PROVIDER(name) \
[[deprecated("Roscpp used to provide the Bind placeholders (_1, _2, ...) in the global namespace is as boost provided them, this is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.")]]\
\
BOOST_STATIC_CONSTEXPR decltype(boost::placeholders::name) name;

ROSCPP_BOOST_BIND_PROVIDER(_1)
ROSCPP_BOOST_BIND_PROVIDER(_2)
ROSCPP_BOOST_BIND_PROVIDER(_3)
ROSCPP_BOOST_BIND_PROVIDER(_4)
ROSCPP_BOOST_BIND_PROVIDER(_5)
ROSCPP_BOOST_BIND_PROVIDER(_6)
ROSCPP_BOOST_BIND_PROVIDER(_7)
ROSCPP_BOOST_BIND_PROVIDER(_8)
ROSCPP_BOOST_BIND_PROVIDER(_9)
#undef ROSCPP_BOOST_BIND_PROVIDER
```

So effectively we don't have ros provide the global `_1` anymore but we provide it from `publisher.h` and we include the `boost/bind/bind.hpp` which doesn't produce the warning.

The `_1` is a static constexpr value, not a type. So if anyone now includes `<boost/bind.hpp>`, we have two instances of `_1` in the global scope and the resolution is ambiguous. Specifically, this happens in [signal9.h](https://github.com/ros/ros_comm/blob/07fb5469c71e10e4a05fa3a631897d9adece61c5/utilities/message_filters/include/message_filters/signal9.h#L45), which then uses the global placeholder `_7` in [this line](https://github.com/ros/ros_comm/blob/07fb5469c71e10e4a05fa3a631897d9adece61c5/utilities/message_filters/include/message_filters/signal9.h#L219), which is clearly a typo, but it shows the problem.

I thought of setting [BOOST_BIND_NO_PLACEHOLDERS](https://github.com/boostorg/bind/blob/develop/include/boost/bind.hpp#L32) here to ensure we don't ever have boost populate them. But then the behaviour is suddenly dependent on the include order, which is not something that's desirable.

Basically, I think the only workaround is to just set `BOOST_BIND_GLOBAL_PLACEHOLDERS` to `1` to silence the warnings. We can do this from `publisher.h`, that way roscpp still provides the placeholders for the downstream packages, without having so much of this deprecation spam in the compile logs.

mikepurvis:
Might it be possible to do it with a deprecated macro pragma instead, eg: https://stackoverflow.com/a/29297970/109517

But I guess fundamentally, we're probably not going to get a "real" fix for this given that it's only an issue on newer boosts than what is called out in REP-3 for Noetic.

iwanders:
No, unfortunately a macro doesn't work, that doesn't do namespacing and breaks any `using namespace`;

Example:
```cpp
namespace boost
{
namespace placeholders
{
  static constexpr int _1 = 1;
}
}

namespace std
{
namespace placeholders
{
  static constexpr int _1 = 5;
}
}

#define _1 _Pragma ("GCC warning \"'DEPRECATED_MACRO1' macro is deprecated\"") boost::placeholders::_1

int main()
{
  using namespace std::placeholders;
  return _1;
}
```

We should expect the `std::placeholders` to be returned here. However, we'll get the boost one; in the output of `gcc -E`;

```cpp
int main()
{
  using namespace std::placeholders;
  return
# 22 "foo.cpp"

# 22 "foo.cpp"
  boost::placeholders::_1;
}
```

Compiling shows the warning & return code is 1 instead of 5.

hariharan82:
when i try to build ROS workspace in Raspberry Pi 400 (debian buster), i get the following error
/home/pi/ros_catkin_ws/src/ros_comm/roscpp/include/ros/node_handle.h: In member function ‘ros::WallTimer ros::NodeHandle::createWallTimer(ros::WallDuration, void (T::*)(const ros::WallTimerEvent&), const boost::shared_ptr<X>&, bool, bool) const’:
/home/pi/ros_catkin_ws/src/ros_comm/roscpp/include/ros/node_handle.h:1436:74: error: ‘boost::placeholders’ has not been declared
     WallTimerOptions ops(period, boost::bind(callback, obj.get(), boost::placeholders::_1), 0);
i find this page most relevant to the issue. i tried every possible solution I found online but non worked.
#define BOOST_BIND_NO_PLACEHOLDERS and few others. Any help is much appreciated

jacobperron:
@hariharan82 Please open questions like this on https://answers.ros.org , which is our central question-and-answers site.

Issue #2189, opened 2021-10-01
message_filters.Cache sometimes misses messages from latched publishers


heuristicus:
Sometimes when using multiple caches in the same node on the same topic, which is latched, the expected behaviour of the cache does not happen.

What I expect to happen is that when the cache is initialised, it is immediately populated with the message on the latched topic. This should happen regardless of how many caches I create.

However, this is not the case. Due to the way that the cache is initialised, it is possible for the latched topic message to be missed entirely.

Here's a minimal working example:

```python
#!/usr/bin/env python

import rospy
from geometry_msgs.msg import PoseStamped
import message_filters


def first_cb(msg):
    rospy.loginfo("first callback called")


def second_cb(msg):
    rospy.loginfo("second callback called")


def third_cb(msg):
    rospy.loginfo("third callback called")


def fourth_cb(msg):
    rospy.loginfo("fourth callback called")


if __name__ == "__main__":
    rospy.init_node("message_filters_cache_test")
    test_topic = "/cache_test"
    # Create a latched publisher
    pub = rospy.Publisher(test_topic, PoseStamped, latch=True, queue_size=1)
    pub.publish(PoseStamped())
    rospy.sleep(1)

    rospy.loginfo("Starting first cache")
    cache_one = message_filters.Cache(
        message_filters.Subscriber(test_topic, PoseStamped)
    )
    cache_one.registerCallback(first_cb)

    rospy.loginfo("Starting second cache")
    cache_two = message_filters.Cache(
        message_filters.Subscriber(test_topic, PoseStamped)
    )
    cache_two.registerCallback(second_cb)

    rospy.loginfo("Starting third cache")
    cache_three = message_filters.Cache(
        message_filters.Subscriber(test_topic, PoseStamped)
    )
    cache_three.registerCallback(third_cb)

    rospy.loginfo("Starting fourth cache")
    cache_four = message_filters.Cache(
        message_filters.Subscriber(test_topic, PoseStamped)
    )
    cache_four.registerCallback(fourth_cb)

    rospy.spin()
```

I suggest running this script several times. The output that I would expect to see from this script is something like

```
[INFO] [1633092471.990110] [/message_filters_cache_test]: Starting first cache
[INFO] [1633092472.005591] [/message_filters_cache_test]: Starting second cache
[INFO] [1633092472.011472] [/message_filters_cache_test]: Starting third cache
[INFO] [1633092472.019149] [/message_filters_cache_test]: Starting fourth cache
[...]

PR #2190, opened 2021-10-04
Add option to not immediately start message_filters subscriber

heuristicus:
Solution for #2189.

This fixes an issue where sometimes caches do not receive the first message on the subscribed topic, if the message arrives immediately after initialising the subscriber. This can happen when the topic is latched, and is problematic if the latched topic never updates as the cache will never receive a message.

Example of starting a cache with delayed initialisation:

```python
sub = message_filters.Subscriber(test_topic, PoseStamped, auto_start=False)
cache = message_filters.Cache(sub)
cache.registerCallback(callback)
# Both the internal cache callback and our custom callback are now in place, so we can start
# the subscriber and guarantee that even if a message arrives instantly both callbacks are called
sub.start()  
```


sloretz:
Thank you for the PR!

PR #2191, opened 2021-10-08
Address DeprecationWarning of logger.warn

g-gemignani:


sloretz (approved):
Thank you for the PR!

PR #2195, opened 2021-10-20
Add optional timeout parameter to rosgraph.is_master_online()

### LINKED ISSUE #2194: `rosgraph.is_master_alive()` has unpredicable delay with certain URIs
### Issue

I've been having issues getting a remote node to detect when master is down (detailed [here](https://answers.ros.org/question/389077/remote-node-doesnt-shut-down-when-connection-with-master-is-lost/)) which I've pinpointed down to the `rosgraph.is_master_alive()` call I'm using to check the connection. In short, certain IPs cause the method to take an awful long time compared to others.

### Code

Below is the simplest example I've been able to create to reproduce the issue. Master is running at 127.0.1.1.

```py
if __name__ == "__main__":
    rospy.init_node("test_node")
    uri = "http://192.168.0.1:11311"
    status = "Online!" if rosgraph.is_master_online(uri) else "Offline!"
    print(status)
```

Which yields a result of

```
rosrun test_package test.py
(2m 10s later...)
Offline!
```

`uri = "http://127.0.0.1:11311"` yields an instantaneous result of `Online!`
`uri = "http://1.2.3.4:11311"` yields an instantaneous result of `Offline!`

After playing around with different IPs it seems that the issue occurs with IPs of the form `192.168.x.x`, though I'm unsure if that range of IPs is specific to my machine.

### Potential Solution

The workaround I'm currently using is to wrap the method call in a timeout.

``` py
@contextmanager
def timeout(time):
    signal.signal(signal.SIGALRM, raise_timeout)
    signal.alarm(time)
    try:
        yield
    except TimeoutError:
        pass
    finally:
        signal.signal(signal.SIGALRM, signal.SIG_IGN)


def raise_timeout(signum, frame):
    raise TimeoutError


def is_master_alive(uri, time):
    with timeout(time):
        return rosgraph.is_master_online(uri)


if __name__ == "__main__":
    rospy.init_node("test_node")
    uri = "http://192.168.0.1:11311"
    status = "Online!" if is_master_alive(uri, time=2) else "Offline!"
    print(status)
```

This feels like a lot of unnecessary code, so I think it'd be worthwhile to add an optional timeout parameter to `rosgraph.is_master_alive()`. The fact that the method returns at all suggests to me that there's some level of timeout already happening, but I think it'd be nice to allow it to be user-defined.

I'm happy to raise a PR if someone could point me in the right direction.

riv-tnoble:
Fixes #2194 by adding an optional `timeout` parameter to `is_online()`. Defaults to `None` to preserve existing behaviour.

The blocking code seems to come from deep within xmlrpclib, so the easiest way to add a timeout seems to be using `socket.setdefaulttimeout()`, which is reset after the call to `ServerProxy.getPid()` is made (within a `finally` block to ensure cleanup is performed even when an exception is raised).

Let me know if I should add any tests or apply formatting to the code!

### DIFF
--- a/tools/rosgraph/src/rosgraph/masterapi.py
+++ b/tools/rosgraph/src/rosgraph/masterapi.py
@@ -37,6 +37,8 @@ Master directly using XML-RPC, this API provides a safer abstraction in the even
 the Master API is changed.
 """
 
+import socket
+
 try:
     from xmlrpc.client import ServerProxy  # Python 3.x
 except ImportError:
@@ -71,13 +73,14 @@ ROSMasterException = MasterException
 Error = MasterError
 Failure = MasterFailure
 
-def is_online(master_uri=None):
+def is_online(master_uri=None, timeout=None):
     """
     @param master_uri: (optional) override environment's ROS_MASTER_URI
+    @param timeout: (optional) give up after n seconds
     @type  master_uri: str
     @return: True if Master is available
     """
-    return Master('rosgraph', master_uri=master_uri).is_online()
+    return Master('rosgraph', master_uri=master_uri).is_online(timeout)
 
 class Master(object):
     """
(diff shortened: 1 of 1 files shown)


sloretz:
Thank you for the PR!

PR #2200, opened 2021-11-03
Improve performance of ShapeShifter::read()

arneboe:
using `resize` causes an unnecessary call to `memset`. Replacing it with `insert` mitigates that. 
Especially when handling lots of large messages (e.g. images) the overhead caused by  `memset`  is substantial. 
This results in ~30% improved performance in my use-case.

### DIFF
--- a/tools/topic_tools/include/topic_tools/shape_shifter.h
+++ b/tools/topic_tools/include/topic_tools/shape_shifter.h
@@ -222,12 +222,8 @@ void ShapeShifter::write(Stream& stream) const {
 template<typename Stream>
 void ShapeShifter::read(Stream& stream)
 {
-  stream.getLength();
-  stream.getData();
-
   // stash this message in our buffer
-  msgBuf.resize(stream.getLength());
-  memcpy(msgBuf.data(), stream.getData(), stream.getLength());
+  msgBuf.insert(msgBuf.end(), stream.getData(), stream.getData() + stream.getLength());
 }
 
 } // namespace topic_tools


tools/topic_tools/include/topic_tools/shape_shifter.h
```diff
@@ -222,12 +222,8 @@ void ShapeShifter::write(Stream& stream) const {
 template<typename Stream>
 void ShapeShifter::read(Stream& stream)
 {
-  stream.getLength();
-  stream.getData();
-
   // stash this message in our buffer
-  msgBuf.resize(stream.getLength());
-  memcpy(msgBuf.data(), stream.getData(), stream.getLength());
+  msgBuf.insert(msgBuf.end(), stream.getData(), stream.getData() + stream.getLength());
```
sloretz, line 226: If I understand correctly, the original code is replacing the content of `msgBuf` with the content of the stream, while this new code appends the stream data to the end of msgBuf.

Maybe `assign` would be a better choice?

sloretz (changes requested):

tools/topic_tools/include/topic_tools/shape_shifter.h
arneboe: If the internal implementation of assign does not do a memset, then yes :)
When I originally implemented this, we were still using c++17 and assign was not available.

Maybe check the compiler explorer to be sure that it does not do memset

sloretz:
Thank you for the PR!

PR #2201, opened 2021-11-04
some fixes for the --repeat_latched option

1r0b1n0:
This should fix some problems with #1850 : 
**1**
The `repeat_latched` member variable bool was left in an uninitialized/random state.
On my machine, even without specifying "--repeat_latched", it was sometimes activated.

**2**
When repeating the latched messages in splitted bags, these should still be published as latched.

_Steps to reproduce :_ 
Start the following in several terminals :

-  `roscore`
- publish a latched topic : `rostopic pub -l /chatter_latched std_msgs/String "data: 'blabla'"`
- publish a periodic topic : `rostopic pub -r 2 /periodic std_msgs/Empty`
- launch a rosbag record : `rosrun rosbag record -a --repeat-latched --duration 5 --split`

after ~15 seconds stop all the nodes and the `rosbag record`

without the fix : when playing a splitted bag `rosrun rosbag play [...]_1.bag -s 2` : running `rostopic echo /chatter_latched` will show nothing

with this fix : when playing a splitted bag `rosrun rosbag play [...]_1.bag -s 2` : running `rostopic echo /chatter_latched` will show the message, because it is latched

### DIFF
--- a/tools/rosbag/src/recorder.cpp
+++ b/tools/rosbag/src/recorder.cpp
@@ -457,7 +457,7 @@ void Recorder::startWriting() {
         {
             // Overwrite the original receipt time, otherwise the new bag will
             // have a gap before the new messages start.
-            bag_.write(out.second.topic, now, *out.second.msg);
+            bag_.write(out.second.topic, now, *out.second.msg, out.second.connection_header);
         }
     }
 


tkazik:
Nice catch, thx a lot!
My 2c: I guess I would prefer the initialization to set the `repeat_latched` to `true`. As such, every bag is "contained" in itself by default.
Or are there points against that?

https://github.com/ros/ros_comm/blob/785826e1235182046d7b44026208f9633a9112a8/tools/rosbag/src/recorder.cpp#L100

1r0b1n0:
Hi,
I agree in my opinion the better choice would be to enable repeat_latched by default.
In this case the "--repeat-latched" option can be removed/ignored.

Martin-Oehler:
Can this be merged please? It contains essential fixes.

romainreignier:
Hi @mjcarroll what do you think about this?
Should we enable `repeat-latched` by default and remove the argument or keep the choice?
At least, the uninitialized variable issue should be merged, this is a bugfix. Do you want a separate PR?

mjcarroll:
> At least, the uninitialized variable issue should be merged, this is a bugfix. Do you want a separate PR?

Yes please, I have no issues with that part.  I'll need a moment to think about the other part.

1r0b1n0:
I created #2314  that will only fix the undefined variable.

romainreignier:
This PR has now been updated to only include the `connection_header` when writing a latched topic to a bag file. This allows to keep the latched info so `rosbag play` can publish it as latched.

AndreasNagel (approved):

sloretz (approved):
Thank you for the PR!

PR #2202, opened 2021-11-08
Fix determining supported kernel version for HTTP 1.1 in kernels with patch version 0

peci1:
I use ROS on a system with mainline kernels, and every time a new minor version is installed, ROS stops working.

That's because the uname of the kernel with patch version 0 looks like this: `5.15-051500-generic`. It does not correspond to the standard `x.y.z-whatever` pattern, which breaks rosgraph. The error I'm getting from rosgraph is:

    ValueError: invalid literal for int() with base 10: '15-051501-generic'

This patch removes all characters starting from the first non-numeric character in the minor version part. This should make the `int()` call pass.

I also added another workaround that returns True as early as kernel 5.x or newer is detected, not looking at the minor version at all. This should further limit the surface for more bugs like this.

I'm not sure how other kernel strings in other distros than Ubuntu can look like, but taking into account I'm the first one reporting such issue, I believe the kernel strings look more or less the same, thus this fix should be the "final" ™ one in this regard.

### DIFF
--- a/tools/rosgraph/src/rosgraph/xmlrpc.py
+++ b/tools/rosgraph/src/rosgraph/xmlrpc.py
@@ -45,6 +45,7 @@ The common entry point for most libraries is the L{XmlRpcNode} class.
 import errno
 import logging
 import platform
+import re
 import select
 import socket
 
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
this looks good to me, but

> That's because the uname of the kernel with patch version 0 looks like this: 5.15-051500-generic

isn't this supposed to be more like `5.15.0-051500-generic`?

peci1:
That's what I'd expect, but the PPA I'm using apparently doesn't follow the naming:
Snímek obrazovky pořízený 2021-11-08 17-27-43

PR #2205, opened 2021-11-23
fix char signedness issue for test_base64 on ARM and RISC-V

alexfanqi:
tested on riscv64 and amd64 machines, I am sure ARM should also work.

char on some machines like ARM and RISC-V is by default unsigned. Using signed literals will cause Wnarrow errors from compiler.

I am not sure if this is the most proper fix. And I guess it would be best if the bundled libb64 can be updated because it is more portable and clear to use signed char as decoder and encoder argument.

I am sorry about commit reference noises in the origin issue.


sloretz (approved):
Thank you for the PR!

PR #2152, opened 2021-04-06
Changes between 1.15.9 and 1.15.11 for backporting to Melodic

jacobperron:
The following list of changes has been integrated into ros_comm 1.15.11 (Noetic) since the last Melodic release (1.14.10)

**Backported** (in this PR):

* Replace message assertion with logging in order to have release modes to fail in compilation when msg type mismatches occur (#2096)
  * Bug fix

* Fix /use_sim_time parameter typo in roswtf error (#2102)
  * Fix typo

* rosbag: start player in paused state (#2086)
  * Bug fix

* Add missing Boost (#2108)
  * Missing dependency

* fix getNumPublishers() to only count fully connected (#2107)
  * Bug fix

* Fix AttributeError isAlive (#2092) 
  * Bug fix

* Portable fix to recent Windows build breaks (#2110)
  * Windows fix

* Fix "TypeError: not enough arguments for format string" (#2127) 
  * Bug fix

* Fix for deadlock issue 1980 (#2121)
  * Bug fix
  * **Tests not backported due to conflict**

* rosmsg: clean up test dependencies (#2103) 
  * Dependency fixes

* Fix HTTP for kernel < 4.16 (#2132)
  * Bug fix

* rosbag: recorder: Catch all exceptions in record thread. (#2151) 
  * Bug fix

* handle SIGINT in rosbag play (#2150) 
  * Bug fix

**Not backported:**

* Use Logger.warning() instead of the deprecated warn() (#2120)
  * Not necessary, Python 3

* fix issue 2141: raw_input does not exist in python 3 (#2143) 
  * Not necessary, Python 3

PR #2208, opened 2021-11-30
keep the persistent connection only if rosmaster supports http1.1

### LINKED ISSUE #2182: regression from #2132: ros::param::set limited to 32 kB on localhost on computers with kernel 4.x
Minimum non-working example (edited at 10 Nov 2021):

```C++
#include <ros/ros.h>

{
  ros::init(argc, argv, "node");
  ros::NodeHandle nh;
  nh.setParam("param", std::string(32654, 'a'));
  return 0;
}

```

Running this program with a ROS master on localhost results in the following error message on some computers:

    [ERROR 1630469544.713298952 /node] [setParam] Failed to contact master at [localhost:11311].  Retrying...

It seems that computers running a 4.x kernel have this 32 kB limit whereas 5.x kernels allow 100s of MBs.

Affected:
- Kontron KTQM87 with Core i7 4th gen (amd64 platform, Intel CPU, kernel 4.15)
- NVidia Xavier AGX (arm64 platform, Maxwell CPU, kernel 4.9)

Non-affected:
- Intel NUC 10th gen (amd64 platform, Intel CPU, kernel 5.4)
- Lenovo T14s notebook (amd64 platform, AMD Ryzen CPU, kernel 5.13)

`ulimit -a` doesn't show any related limitation:

```
core file size          (blocks, -c) unlimited
data seg size           (kbytes, -d) unlimited
scheduling priority             (-e) 40
file size               (blocks, -f) unlimited
pending signals                 (-i) 63543
max locked memory       (kbytes, -l) 65536
max memory size         (kbytes, -m) unlimited
open files                      (-n) 1024
pipe size            (512 bytes, -p) 8
POSIX message queues     (bytes, -q) 819200
real-time priority              (-r) 99
stack size              (kbytes, -s) 8192
cpu time               (seconds, -t) unlimited
max user processes              (-u) 63543
virtual memory          (kbytes, -v) unlimited
file locks                      (-x) unlimited
```

Decreasing the payload size to 32653 or less makes the program work on all computers.

I tried running the program with XmlRpc client debug logs and here's what I get around the `setParam` call:

```
XmlRpcClient::execute: method setParam (_connectionState IDLE).
XmlRpcClient::generateRequest: header is 115 bytes, content-length is 32654.
XmlRpcClient::writeRequest (attempt 1):
POST / HTTP/1.1
User-Agent: XMLRPC++ 0.7
Host: localhost:11311
Content-Type: text/xml
Content-length: 32654

<?xml version="1.0"?>
<methodCall><methodName>setParam</methodName>
<params><param><value>/node</value></param><param><value>/node/param</value></param><param><value>aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
XmlRpcSocket::nbWrite: send/write returned 32768.
XmlRpcSocket::nbWrite: send/write returned -1.
XmlRpcClient::close: fd 3.
XmlRpcSource::close: closing socket 3.
XmlRpcSocket::close: fd 3.
XmlRpcSource::close: done closing socket 3.
[ERROR 1630465298.375945115 /node] [setParam] Failed to contact master at [localhost:11311].  Retrying...
```

The two `nbWrite` are interesting - the second one ends with error, but the error message is not outputted. I tried manually reading errno in this case and the error is `EPIPE`. According to the docs that means that the other side (rosmaster) has closed the connection before it read all the data.

That would point towards a bug in rosmaster, but if I run a similar program using rospy, then I can set parameters as large as I want. So it is probably some incompatibility between the C++ XMLRPC client and Python XMLRPC server.

Also, if ROS master is running on a different computer in LAN, no combination of master/client computers leads to this error. It is solely affecting the case when master and the node are running on the same machine.

Last, if I edit the patch from https://github.com/ros/ros_comm/pull/2132 and force HTTP/1.1 even on the 4.x kernels, the 32 kB limit is away. @jikawa-az @fujitatomoya @emersonknapp any ideas how to get both the fix from #2132 and allow larger params again?

iuhilnehc-ynos:
to fix https://github.com/ros/ros_comm/issues/2182

Signed-off-by: Chen Lihui <lihui.chen@sony.com>

### DIFF
--- a/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
+++ b/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
@@ -118,6 +118,9 @@ namespace XmlRpc {
     // each thread should have its own client.
     bool _executing;
 
+    // True if the server supports HTTP1.1
+    bool _server_http_1_1;
+
     // True if the server closed the connection
     bool _eof;
 
--- a/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
+++ b/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
@@ -118,9 +118,6 @@ namespace XmlRpc {
     // each thread should have its own client.
     bool _executing;
 
-    // True if the server supports HTTP1.1
-    bool _server_http_1_1;
-
     // True if the server closed the connection
     bool _eof;
 
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -56,7 +56,6 @@ XmlRpcClient::XmlRpcClient(const char* host, int port, const char* uri/*=0*/)
   _sendAttempts(0),
   _bytesWritten(0),
   _executing(false),
-  _server_http_1_1(false),
   _eof(false),
   _isFault(false),
   _contentLength(0)
@@ -84,7 +83,6 @@ XmlRpcClient::close()
 {
   XmlRpcUtil::log(4, "XmlRpcClient::close: fd %d.", getfd());
   _connectionState = NO_CONNECTION;
-  _server_http_1_1 = false;
   _disp.exit();
   _disp.removeSource(this);
   XmlRpcSource::close();
@@ -131,13 +129,6 @@ XmlRpcClient::execute(const char* method, XmlRpcValue const& params, XmlRpcValue
 
   if (_connectionState != IDLE || ! parseResponse(result))
     return false;
-
-  // close() if server not supports HTTP1.1
-  // otherwise, reusing the socket to write leads to a SIGPIPE because
-  // the remote server could shut down the corresponding socket.
-  if (!_server_http_1_1) {
-    close();
-  }
 
   XmlRpcUtil::log(1, "XmlRpcClient::execute: method %s completed.", method);
   _response = "";
@@ -420,9 +411,7 @@ XmlRpcClient::readHeader()
   char *lp = 0;                       // Start of content-length value
 
   for (char *cp = hp; (bp == 0) && (cp < ep); ++cp) {
-    if ((ep - cp > 15) && (strncmp(cp, "HTTP/1.1 200 OK", 15) == 0))
-      _server_http_1_1 = true;
-    else if ((ep - cp > 16) && (strncasecmp(cp, "Content-length: ", 16) == 0))
+    if ((ep - cp > 16) && (strncasecmp(cp, "Content-length: ", 16) == 0))
       lp = cp + 16;
     else if ((ep - cp > 4) && (strncmp(cp, "\r\n\r\n", 4) == 0))
       bp = cp + 4;
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -132,7 +132,7 @@ XmlRpcClient::execute(const char* method, XmlRpcValue const& params, XmlRpcValue
     return false;
   }
 
-  // close() if server not supports HTTP1.1
+  // close() if server does not supports HTTP1.1
   // otherwise, reusing the socket to write leads to a SIGPIPE because
   // the remote server could shut down the corresponding socket.
   if (_header.find("HTTP/1.1 200 OK", 0, 15) != 0) {
--- a/utilities/xmlrpcpp/src/XmlRpcClient.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcClient.cpp
@@ -127,10 +127,20 @@ XmlRpcClient::execute(const char* method, XmlRpcValue const& params, XmlRpcValue
   double msTime = -1.0;   // Process until exit is called
   _disp.work(msTime);
 
-  if (_connectionState != IDLE || ! parseResponse(result))
+  if (_connectionState != IDLE || ! parseResponse(result)) {
+    _header = "";
     return false;
+  }
+
+  // close() if server not supports HTTP1.1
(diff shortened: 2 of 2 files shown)


fujitatomoya (approved):

fujitatomoya:
@sloretz @jacobperron @mjcarroll friendly ping, requesting maintainer review.

utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
```diff
@@ -118,6 +118,9 @@ namespace XmlRpc {
     // each thread should have its own client.
     bool _executing;
 
+    // True if the server supports HTTP1.1
+    bool _server_http_1_1;
```
jacobperron: This change breaks ABI.

jacobperron (commented):
The changes look okay to me, but this breaks ABI. I'd rather not release an ABI-breaking change into an already released distribution. Maybe there's a way to make this fix ABI-compatible?

utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcClient.h
fujitatomoya: > but this breaks ABI.

our bad, that should be kept...sorry.

iuhilnehc-ynos: I knew it breaks ABI, but it's clear.
I tried to use `_connectionState`, but some source code might be as 

```cpp
enum ClientConnectionState { NO_CONNECTION, CONNECTING, WRITE_REQUEST, READ_HEADER, READ_RESPONSE, IDLE,
        CLIENT_CONNECT_STATE_MASK = 0xFFFF, SERVER_HTTP_DISCONNECT = 0x10000};

check header to set _connectionState:
  _connectionState |= SERVER_HTTP_DISCONNECT;

set:
_connectionState = READ_HEADER;
  -->
_connectionState |= READ_HEADER;


if (_connectionState == WRITE_REQUEST)
   -->
if (_connectionState & CLIENT_CONNECT_STATE_MASK == WRITE_REQUEST)

check if SERVER_HTTP_DISCONNECT:
if (_connectionState & SERVER_HTTP_DISCONNECT)
```

<del>do you think it's acceptable?</del>
It seems to not work fine.

iuhilnehc-ynos: Do you have any other suggestions? I'd like to hear that.

iuhilnehc-ynos: I'll make use of the `_header` member variable, make it [not clean](https://github.com/ros/ros_comm/blob/63c9bbb5c2480a232530cfa13308a22d727f20b8/utilities/xmlrpcpp/src/XmlRpcClient.cpp#L457) after `XmlRpcClient::readHeader`, and then check it [here](https://github.com/ros/ros_comm/blob/63c9bbb5c2480a232530cfa13308a22d727f20b8/utilities/xmlrpcpp/src/XmlRpcClient.cpp#L132) and clean it.

peci1:
I confirm this patch fixed #2182 for me on Jetson Xavier AGX (kernel 4.9.253) running Melodic. Also, the performance from #2132 is not affected.

Thanks a lot for investigating!

utilities/xmlrpcpp/src/XmlRpcClient.cpp
```diff
@@ -127,10 +127,20 @@ XmlRpcClient::execute(const char* method, XmlRpcValue const& params, XmlRpcValue
   double msTime = -1.0;   // Process until exit is called
   _disp.work(msTime);
 
-  if (_connectionState != IDLE || ! parseResponse(result))
+  if (_connectionState != IDLE || ! parseResponse(result)) {
+    _header = "";
     return false;
+  }
+
+  // close() if server not supports HTTP1.1
```
jacobperron:
```suggestion
  // close() if server does not supports HTTP1.1
```

jacobperron (approved):
LGTM, thanks for the fix!

peci1:
Should I take care of the Melodic backport? Or is it planned?

jacobperron:
> Should I take care of the Melodic backport? Or is it planned?

I think a maintainer can include it in a backport PR, e.g. https://github.com/ros/ros_comm/pull/2152

PR #2228, opened 2022-01-27
rosbag::Recorder: allow pausing of message recording

rhaschke:
This is a rework of #1294 addressing [comments by @dirk-thomas](https://github.com/ros/ros_comm/pull/1294#issuecomment-363184836).
The primary goal is to provide an API to pause and continue message recording. As suggested by Dirk, this PR
- keeps pausing independent from snapshot triggering: while pausing stops message recording (into the memory queue), the latter triggers writing to a new file (but always continues recording)
- uses `std_msgs::Bool` to pause/continue recording

The new API is not exposed to the cmdline, because the snapshot API isn't either (and I don't need it to be exposed).

### DIFF
--- a/tools/rosbag/include/rosbag/recorder.h
+++ b/tools/rosbag/include/rosbag/recorder.h
@@ -54,6 +54,7 @@
 #include <ros/ros.h>
 #include <ros/time.h>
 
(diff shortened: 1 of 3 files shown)


sloretz:
Thank you for the PR!

Issue #723, opened 2015-12-30
roslaunch: add support for "param" tag in substitution args


mishrasubhransu:
Say the contents of a file tf_stp_area1.yaml were 

```
world2map: "127 108 0 0 0 0.73454 0.67855 world map"
```

I feel it would be very convenient to be able to do the following in a launch file

```
<launch>
  <arg name="location" value="area1"/>
  <rosparam command="load" file="$(find my_pkg)/params/tf_stp_$(arg location).yaml" />
  <node pkg="tf2_ros" type="static_transform_publisher" name="world_to_map" args="$(param world2map)" />
</launch>
```

hkorre:
Was just looking for a way to do this exact thing today.

doisyg:
same here

doisyg:
same here

tahsinkose:
Would lift the redundant work of duplicating numbers and strings into launch file. This should also work for `<arg> ` substitution.

eva-dierkes:
I'm searching for the same solution...

tahsinkose:
I have implemented it. If you don't want to wait for official merge, you can directly clone my own fork and try it. I couldn't do an extensive testing, so I'm waiting feedbacks from you for any type of problems.

rickstaa:
@tahsinkose  Thanks for implementing this! I think this would be an amazing feature to have.

tahsinkose:
I'm glad that it helps for you @rickstaa . Unfortunately, I don't have the time to re-inspect the PR and have it merged into the master. I know that @dirk-thomas and other maintainers have shifted their focus on ROS 2 development, which has a Python-based launch system, therefore they may not be as helpful as they can for the ROS 1. 

If you like you can take on this simple feature and have it pass the tests.

rickstaa:
@tahsinkose Ah thanks I understand. Thanks again for the nice pull request!

doisyg:
I would love to see it merged too

mechatheo:
Hello,

this issue is already quite old, i saw a recent PR: https://github.com/ros/ros_comm/pull/2097 which solves it. What is the state of this, is there a good chance that this will be accepted?

tahsinkose:
I'm expecting some maintainer to have a look at it. As most of the efforts are channeled into ROS2 dev, there is not much people left to accept/reject/review stuff. Who should I ping for this?

FabianSchurig:
I would love to see it merged, too :)

[...]

PR #2229, opened 2022-02-01
[roslaunch] add 'command' attribute to 'arg' tag

### LINKED ISSUE #1767: Add command attribute to <arg> tag same as <param>.
This would help a lot, it should be easy as the functionality is already implemented for <param>, and partially it would provide similar functionality as #723 was seeking.

standmit:
There is very useful feature of setting parameter value from shell command output (attribute 'command' of 'param' tag). But the functionality is too limited: it is impossible to pass a command result to included launch or to node arguments. Extending this feature to arguments provides more ability for customization. It resolves #1767. Also, it can be used as a workaround for problem, described in #723

### DIFF
--- a/tools/roslaunch/src/roslaunch/loader.py
+++ b/tools/roslaunch/src/roslaunch/loader.py
@@ -107,6 +107,71 @@ def convert_value(value, type_):
     else:
         raise ValueError("Unknown type '%s'"%type_)        
 
+def exec_command(recepient, command, verbose=False):
+    try:
+        if type(command) == unicode:
+            command = command.encode('utf-8') #attempt to force to string for shlex/subprocess
+    except NameError:
+        pass
+    if verbose:
+        print("... executing command [%s]" % command)
+    import subprocess, shlex #shlex rocks
+    try:
+        if os.name != 'nt':
+            command = shlex.split(command)
+        else:
+            cl = shlex.split(command, posix=False)  # use non-posix method on Windows
+
+            # On Linux, single quotes are commonly used to enclose a path to escape spaces.
+            # However, on Windows, the single quotes are treated as part of the arguments.
+            # Special handling is required to remove the extra single quotes.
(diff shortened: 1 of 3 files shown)


tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -107,6 +107,71 @@ def convert_value(value, type_):
...
+        if p.returncode != 0:
+            raise ValueError("Cannot load command for [%s]: command [%s] returned with code [%s]"%(recepient, command, p.returncode))
+    except OSError as e:
+        if e.errno == errno.ENOENT:
+            raise ValueError("Cannot load command for [%s]: no such command [%s]"%(recepient, command))
+        raise
+    if c_value is None:
+        raise ValueError("command: unable to get output of command [%s]"%command)
+    return c_value
+
```
standmit, line 174: Just moved from Loader.param_value(...)

tools/roslaunch/src/roslaunch/loader.py
```diff
@@ -485,69 +550,7 @@ def param_value(self, verbose, name, ptype, value, textfile, binfile, command):
...
-                    c_value = c_value.decode('utf-8')
-                if p.returncode != 0:
-                    raise ValueError("Cannot load command parameter [%s]: command [%s] returned with code [%s]"%(name, command, p.returncode))
-            except OSError as e:
-                if e.errno == errno.ENOENT:
-                    raise ValueError("Cannot load command parameter [%s]: no such command [%s]"%(name, command))
-                raise
-            if c_value is None:
-                raise ValueError("parameter: unable to get output of command [%s]"%command)
-            return convert_value(c_value, ptype)
```
standmit, line 550: Is moved outside the class

standmit:
@jacobperron , could you review this?

sloretz:
Thank you for the PR!

Issue #2212, opened 2021-12-21
AttributeError: 'NoneType' object has no attribute 'close'


MatthijsBurgh:
I have no clue of the following exception is caused by incorrect implemetnation on my side or caused by a bug in rospy.
```
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_pubsub.py", line 185, in robust_connect_subscriber
    conn.receive_loop(receive_cb)	    
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_base.py", line 846, in receive_loop
    self.close()
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_base.py", line 858, in close
    self.socket.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

MatthijsBurgh:
@jacobperron friendly ping

jacobperron:
The error message makes it sound like it's a bug in rospy.

@MatthijsBurgh Can you provide simple steps to reproduce the error?

MatthijsBurgh:
@jacobperron I will get back to you when I have been able to isolate the issue into a reproducible error.

The situation in which I experienced this error the first time, was in a node launching other nodes.

Now I also experience this in nodes with an action_server simple_client. In both situations it happens during stopping of a node.
```
Exception in thread /hero/local_planner/action_server/result:
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_pubsub.py", line 185, in robust_connect_subscriber
    conn.receive_loop(receive_cb)	    
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_base.py", line 846, in receive_loop
    self.close()
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/impl/tcpros_base.py", line 858, in close
    self.socket.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

[...]

PR #2233, opened 2022-03-15
Add a workaround for a race condition while closing the socket (#2212)

madmage:
This is not a fix but rather a workaround to avoid the exception when the socket is closed from two different threads (race condition). A proper fix would require to redesign the whole class to be thread safe. See https://github.com/ros/ros_comm/issues/2212.


jacobperron (approved):

jacobperron:
Given the complexity involved in the code, I think this workaround is okay in lieu of a more involved fix.

PR #2242, opened 2022-05-14
A _7 was missed in previous commit to convert everything to boost::placeholders::_N

lucasw:


mikepurvis:
Haha, whoops. Thanks for the fix.

FYI @seanyen

PR #2248, opened 2022-05-30
Add missing include

jspricke:


peci1 (approved):
LGTM. In what use-case did you need this?

jspricke:
>LGTM. In what use-case did you need this?

I'm planning to update Debian to the latest class_loader and pluginlib 
which don't expose the header anymore.

mjcarroll (approved):

PR #2250, opened 2022-06-06
Fix #2249: Topic statistics do not handle well ROS time jumping back

### LINKED ISSUE #2249: Topic statistics do not handle well ROS time jumping back - denial of service
# How I found it

I have a bag file which I replay from the beginning a few times in succession, and one of my nodes processing IMU messages stopped being able to process messages for no obvious reason. In the end, it was processing e.g. at 0.1 Hz while the messages are coming at 100 Hz. Restarting the node fixes the issue temporarily.

# Analysis

I accidentaly enabled topic statistics by setting `/enable_statistics` ROS param to true. With topic statistics enabled, this bug manifested.

https://github.com/ros/ros_comm/blob/dd78ac8af128bb8eb992d6431bb9f994658ea6ab/clients/roscpp/src/libros/statistics.cpp#L118

`stats.last_publish` is updated with the last time when a message was published to `/statistics`. If then a bag file starts playing again, the check referenced above will for long not be true (actually never again if you're playing the same bag).

There is a list in `stats.arrival_time_list` that is cleared only when a stats message is published:

https://github.com/ros/ros_comm/blob/dd78ac8af128bb8eb992d6431bb9f994658ea6ab/clients/roscpp/src/libros/statistics.cpp#L252

So this list is growing without bounds when the bag is replayed. Each received message adds a record to this list. With IMU messages @ 100 Hz, this quickly comes to the order of 1e5 and even higher orders.

Finally, this growing list is being copied twice in each call of the `callback()` function:

https://github.com/ros/ros_comm/blob/dd78ac8af128bb8eb992d6431bb9f994658ea6ab/clients/roscpp/src/libros/statistics.cpp#L84

and

https://github.com/ros/ros_comm/blob/dd78ac8af128bb8eb992d6431bb9f994658ea6ab/clients/roscpp/src/libros/statistics.cpp#L258

These copies are what is slowing the subscriber down to unacceptable slowness after a few bag replays. Whenever I pause the node in GDB, `PollManager` is caught copying this list:

```
std::list<ros::Time>::_M_assign_dispatch<std::_List_const_iterator<ros::Time> > list.tcc:313
std::list<ros::Time>::operator= list.tcc:285
ros::StatisticsLogger::StatData::operator= statistics.h:91
ros::StatisticsLogger::callback 0x00007f98cbce233a
ros::Subscription::handleMessage 0x00007f98cbd38e19
ros::TransportPublisherLink::handleMessage 0x00007f98cbd1881d
ros::TransportPublisherLink::onMessage 0x00007f98cbd1894c
boost::function4<void, const boost::shared_ptr<ros::Connection> &, const boost::shared_array<unsigned char> &, unsigned int, bool>::operator() function_template.hpp:759
ros::Connection::readTransport 0x00007f98cbc9a24c
boost::function1<void, const boost::shared_ptr<ros::Transport> &>::operator() function_template.hpp:759
ros::TransportTCP::socketUpdate 0x00007f98cbd1343d
boost::function1<void, int>::operator() function_template.hpp:759
ros::PollSet::update 0x00007f98cbd49d43
ros::PollManager::threadFunc 0x00007f98cbcd5365
<unknown> 0x00007f98c4411bcd
start_thread 0x00007f98c50076db
clone 0x00007f98c8d2161f
```

I'll sketch a PR fixing this by detecting ROS time jumping back and resetting the timestamp check.

peci1:
First, I'm submitting a [failing test](https://build.ros.org/job/Npr__ros_comm__ubuntu_focal_amd64/520/testReport/).

It makes 4 repetitions of receiving 10000 messages, each time resetting ROS time from 1.0 onwards. First repetition is okay. If #2249 is not fixed, the following iterations will take considerably longer because the internal lists are never cleared.

### DIFF
--- a/clients/roscpp/src/libros/statistics.cpp
+++ b/clients/roscpp/src/libros/statistics.cpp
@@ -33,6 +33,11 @@
 #include "std_msgs/Header.h"
 #include "ros/param.h"
 
+//! \brief Hysteresis setting for recognition when ROS time has jumped back (in seconds).
+#ifndef ROS_STATISTICS_JUMP_BACK_DELTA
+#define ROS_STATISTICS_JUMP_BACK_DELTA 3
+#endif
+
 namespace ros
 {
 
@@ -85,6 +90,19 @@ void StatisticsLogger::callback(const boost::shared_ptr<M_string>& connection_he
   }
   previous_connection_id = connection_id;
 
+  // Handle ROS time jumping back
+  if (received_time + ros::Duration(ROS_STATISTICS_JUMP_BACK_DELTA) < stats.last_publish)
+  {
+    ROS_DEBUG("ROS time has jumped back, resetting topic statistics.");
+    stats.last_publish = received_time;
+    // clear the window
+    stats.age_list.clear();
+    stats.arrival_time_list.clear();
+    stats.dropped_msgs = 0;
+    stats.stat_bytes_last = 0;
+    map_[callerid] = stats;
+  }
+  
   stats.arrival_time_list.push_back(received_time);
 
   if (dropped)
--- a/test/test_roscpp/test/CMakeLists.txt
+++ b/test/test_roscpp/test/CMakeLists.txt
@@ -150,6 +150,7 @@ add_rostest(launch/search_param.xml)
 
 add_rostest(launch/stamped_topic_statistics_with_empty_timestamp.xml)
 add_rostest(launch/topic_statistic_frequency.xml DEPENDENCIES ${PROJECT_NAME}-publisher_rate ${PROJECT_NAME}-subscriber ${PROJECT_NAME}-topic_statistic_frequency)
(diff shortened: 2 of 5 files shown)


peci1:
And the fix comes.

peci1:
Is there actually reason to copy the whole stats structure twice in each callback? Wouldn't it be better to modify it directly via an `auto&`? Could the reason be keeping the lists consistent even in usage by different threads? But that would mean some data would be lost if more threads were overwriting the stats structure...

peci1:
@mjcarroll @sloretz friendly ping

sloretz:
Thank you for the PR!

peci1:
This PR comes with a unit test that fails without this PR and succeeds with it. The test does something that should be completely valid - having topic statistics enabled, receiving lots of messages and then resetting ROS time back.

sloretz:

* it only supports ROS 1

Issue #139, opened 2013-01-31
roslaunch: option to disable all logging to file system


tfoote:
I have an embedded computer on the robot which runs off of an SD card.  It is highly desirable to be able to disable output to the file system, notably logging, but ROS does not support this.

Please see discussion at http://answers.ros.org/question/594/how-can-i-completely-disable-writing-logs-to

It seems that rosconsole, rospy, roscpp, and rosout all need to be modified, but I don't really understand.

trac data:
- Owner: **straszheim**
- Reporter: **veltrop**
- Reported at: **Sun Apr  3 22:57:55 2011**
- URL: https://code.ros.org/trac/ros/ticket/3438

tfoote:
[kwc] It appears there are some issues fully disabling logging, but you can greatly reduce with:

rosconsole.config: replace INFO/WARN with FATAL

python_logging.config: replace DEBUG with CRITICAL

If you want to disable roslaunch logging, use the attached patch.

tfoote:
[veltrop] Excellent, the roslaunch logging patch is very helpful.

tfoote:
[veltrop] Followed your steps, plus recompiled everything with this in my $ROS_ROOT/rosconfig.cmake:
set(ROS_BUILD_TYPE MinSizeRel)
set(ROS_COMPILE_FLAGS "-DROSCONSOLE_MIN_SEVERITY=5 ${ROS_COMPILE_FLAGS}")

I still get logging that begins with [roscpp_internal] and [DEBUG]

tfoote:
[kwc] Right, that's one the issues I'm referring to.  It appears that the roscpp_internal logger may not work with the config file, but I need to investigate more before filing a separate ticket.

tfoote:
[kwc] transferring ticket to troy to look at.  The basic issue is: "how do I disable logging of the roscpp_internal logger".  After some basic experiments with the logging config file, I'm not sure this logger is respecting the external configuration.

tfoote:
[straszheim] compile with ROSCPP_LOG_DISABLE defined to disable the roscpp_internal logger completely.  In r13673.  Leaving ticket open pending confirmation that this does the trick...

tfoote:
[straszheim] ROSCPP_LOG_DISABLE is gone.   In r13691 ROSCPP_LOG_DEBUG has been converted:

#define ROSCPP_LOG_DEBUG(...) \ ROS_DEBUG_NAMED("roscpp_internal", **VA_ARGS**)

[...]

PR #2252, opened 2022-06-27
Add an environment variable to disable rosout

gstorer:
There are a lot of questions around asking about how to disable logging in ROS. One of the more difficult things to do is to turn off rosout. Our solution was to patch all of our ROS installs and set an environment variable to disable logging.

ROSOUT_DISABLE_FILE_LOGGING exists but rosout still runs which is unnecessary for our deployed systems.

https://github.com/ros/ros_comm/issues/139
https://answers.ros.org/question/9627/how-can-i-completely-disable-writing-logs-to-filesystem/

### DIFF
--- a/tools/roslaunch/resources/roscore.xml
+++ b/tools/roslaunch/resources/roscore.xml
@@ -8,6 +8,6 @@
   <group ns="/">
     <param name="rosversion" command="rosversion roslaunch" />
     <param name="rosdistro" command="rosversion -d" />
-    <node pkg="rosout" type="rosout" name="rosout" respawn="true"/>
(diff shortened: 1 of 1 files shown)


peci1:
I think this can be quite easily solved by crafting your own `roscore.xml` file. It is first searched in `ROS_ETC_DIR`:

https://github.com/ros/ros_comm/blob/842f0f026924323f605495da0b80493f15f0bdce/tools/roslaunch/src/roslaunch/config.py#L80

So if you want this to be configurable during runtime, just create an alternative `ROS_ETC_DIR` with the altered `roslaunch.xml`. You can then select whether you want rosout or not by just changing `ROS_ETC_DIR`.

gstorer:
Yes, that would indeed solve our problems. I wasn't aware of the ROS_ETC_DIR variable. I will close this pull request.

Issue #2167, opened 2021-06-30
relay_field broken in python3 / ROS noetic


marcelino-pensa:
I am trying to run the example for `relay_field` as described in [this link](https://github.com/ros/ros_comm/pull/639).

Terminal 1: `rosrun rospy_tutorials talker.py`

Terminal 2:
```
rosrun topic_tools relay_field /chatter /image sensor_msgs/Image "header:
  seq: 0
  stamp: {secs: 0, nsecs: 0}
  frame_id: m.data
height: 0
width: 0
encoding: ''
is_bigendian: 0
step: 0
data: ''"
```

Terminal 3: 
`rostopic echo /image`

The problem arises at first in Terminal 2, with the following error:
```
[ERROR] [1625092139.961614]: bad callback: <bound method RelayField.callback of <__main__.RelayField object at 0x7fb62ebc1550>>
Traceback (most recent call last):
  File "/opt/ros/noetic/lib/python3/dist-packages/rospy/topics.py", line 750, in _invoke_callback
    cb(msg)
  File "relay_field", line 106, in callback
    pub_args = _eval_in_dict_impl(msg_generation, None, {'m': m})
  File "relay_field", line 40, in _eval_in_dict_impl
    res[k] = _eval_in_dict_impl(v, globals_, locals_)
  File "relay_field", line 41, in _eval_in_dict_impl
    elif (type_ is str) or (type_ is unicode):
NameError: name 'unicode' is not defined
```

According with [this forum](https://stackoverflow.com/questions/19877306/nameerror-global-name-unicode-is-not-defined-in-python-3), this is related to an update in python3. 
I went ahead and replaced the check `(type_ is unicode)` with `(type_ is bytes)` in the [code for relay_field](https://github.com/ros/ros_comm/blob/c7c5bcb22780e7f6dbbff9bc923afca34fd612b6/tools/topic_tools/scripts/relay_field#L41), and things seem to work. However, whenever I go to terminal 3 and run `rostopic echo /image`, Terminal 2 starts printing the following error:

```
[ERROR] [1625092400.464262]: bad callback: <bound method RelayField.callback of <__main__.RelayField object at 0x7f697c585580>>
Traceback (most recent call last):
  File "/opt/ros/noetic/lib/python3/dist-packages/sensor_msgs/msg/_Image.py", line 138, in serialize
    buff.write(struct.Struct('<I%ss'%length).pack(length, _x))
struct.error: argument for 's' must be a bytes object

[...]

PR #639, opened 2015-07-15
[topic_tools] relay_field which allows relay topic fields to another fields

wkentaro:

PR #2253, opened 2022-06-27
Update topic_tools relay_field to Python 3

RyanDMott:
### Solves
[Issue 2167](https://github.com/ros/ros_comm/issues/2167)

### Content
Implementing the "unicode" to "bytes" migration the author of [Issue 2167](https://github.com/ros/ros_comm/issues/2167) describes.

After this change, Python's 2to3 only suggests 
 - using list() to encase the result of a dictionary's items() before iterating over it (which is not useful) and 
 - deleting the import of print_function from future (which we might as well do).

### Explanation
To address the second half of [issue #2167](https://github.com/ros/ros_comm/pull/639#issuecomment-121757055), it is the OP's example which is at fault, not relay_field. The example attempts to write a string to a field ("data") whose type has changed from "unicode" to "bytes" in Noetic. Replacing the empty string with empty brackets [] fixes that example. The example in the file's help looks intact.

### Testing
On my copy of Noetic the attached example launch files run when the edit in this PR is made. Both show the frame_id being transferred but the timestamp changed to one second.

[test_relay_field.tar.gz](https://github.com/ros/ros_comm/files/8995956/test_relay_field.tar.gz)

I have not tested beyond these two examples.

### DIFF
--- a/tools/topic_tools/scripts/relay_field
+++ b/tools/topic_tools/scripts/relay_field
@@ -16,7 +16,6 @@ Example:
          frame_id: m.data"
 """
 
-from __future__ import print_function
 import argparse
 import sys
 import copy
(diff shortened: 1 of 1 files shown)


peci1 (changes requested):
LGTM, but a test would help here. I think starting by copying `relay.test` and changing it to relay a field would be enough.

sloretz (approved):
Thank you for the PR!

PR #2256, opened 2022-07-01
Added feature to enable or disable logs through service call

Lakshay011:
**Issue aimed**
Basically right now there is no way to stop logging during the runtime of rospy node ,event though using service call one cannot disable logs, they can only change the log level of the node.
So to overcome this issue I have basically added a feature to enable/disable logs for any rospy node using base service `set_logger_level`.

**Approach**
To enable/disable logs, one has to pass enable/disable parameter in logger, part of service request.
_Sample Service call to disable logging_
`rosservice call /node_name/set_logger_level "logger: 'disable' level: 'info'" `
This will make sure no logs will get dump.

Moreover the logs will automatically get enable if any other logger is set in service request.
Example:
`rosservice call /node_name/set_logger_level "logger: 'rospy' level: 'info'" `
or using enable parameter
`rosservice call /node_name/set_logger_level "logger: 'enable' level: 'info'" `

**Future**
Once this PR is approved I will raise a PR for roscpp part (already tested & created), so that any node can be disable using service call.

### DIFF
--- a/clients/rospy/src/rospy/client.py
+++ b/clients/rospy/src/rospy/client.py
@@ -157,14 +157,23 @@ _names_to_logging_levels = {
       'ERROR':    logging.ERROR,
       'FATAL':    logging.CRITICAL,
       }
+_level_to_change_logs = {
+      "ENABLE":    logging.NOTSET,
+      "DISABLE":   logging.CRITICAL
+}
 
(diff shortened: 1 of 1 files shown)


sloretz:
Thank you for the PR!

PR #2257, opened 2022-07-06
support `rosbag play --loop` in `rospy.Timer`

v4hn:
I fail to understand why this issue still exists after all these years, though I really should not be surprised...

Without the patch the Timer thread just dies on rosbag loop *with no way of detecting and restarting it* easily.
The patch changes the behavior to assume it slept enough and calls the callback again.
If the callback logic requires resets, this can be detected from the TimerEvent passed to the callback.

This addresses a use case that is currently fully broken, so there is no need to worry about incompatible behavior.


sloretz (approved):
Thank you for the PR!

PR #2259, opened 2022-07-20
Fix get_type_and_topic_info() exception in Python3

KindDragon:
Calling this function in Python 3 can produce the exception "TypeError: unhashable type: 'dict_keys'", we should use proper conversion to list to fix this exception

### DIFF
--- a/tools/rosbag/src/rosbag/bag.py
+++ b/tools/rosbag/src/rosbag/bag.py
@@ -857,7 +857,7 @@ class Bag(object):
         # load our list of topics and optionally filter
         if topic_filters is not None:
             if not isinstance(topic_filters, list):
-                topic_filters = [topic_filters]
+                topic_filters = list(topic_filters)
                 
             topics = topic_filters
         else:


tools/rosbag/src/rosbag/bag.py
```diff
@@ -857,7 +857,7 @@ def get_type_and_topic_info(self, topic_filters=None):
         # load our list of topics and optionally filter
         if topic_filters is not None:
             if not isinstance(topic_filters, list):
-                topic_filters = [topic_filters]
+                topic_filters = list(topic_filters)
```
sloretz, line 860: The doc string says `topic_filters` may either be a string, or a list of strings.

```
@type topic_filters: either a single str or a list of str.
```

If given `topic_filters = "foo"`, the line `topic_filters = [topic_filters]` makes `topic_filters` =  `["foo"]`, but IIUC this change would make it equal to `["f", "o", "o"]`.

sloretz (changes requested):
Thank you for the PR! I don't think it can be accepted as is. Some equivalent to an `isinstance()` check to handle the special case of `{}.keys()` could work though.

tools/rosbag/src/rosbag/bag.py
KindDragon: Ok, but it already 2 years for this PR. Please make fixes anyway you like 👍

sloretz: Thank you for your patience these two years. Unfortunately I don't have the bandwidth to make and test the changes, so I'll close this for now.

PR #2261, opened 2022-07-29
Ensure latched messages are updated on every split

ggorjup:
## Summary
This fixes an issue with #1850 - latched messages from the same publisher are not updated in the `Recorder`.

## Steps to Reproduce
1. Create and run a node that periodically publishes latched messages, for example:
```python
#!/usr/bin/env python3
import rospy
from std_msgs.msg import String

pub = rospy.Publisher('chatter', String, latch=True)
rospy.init_node('talker')
for i in range(20):
    pub.publish(str(i))
    rospy.sleep(5)
```
2. Record the latched topic in splits:
`rosrun rosbag record -a --repeat-latched --duration 10 --split`
3. Check messages on the target topic in the last split:
`rostopic echo -b <last_split>.bag chatter`

### Expected behavior
The **first message** in the **last split** is the value of the latched topic when the split was started (e.g., `data: "18"`).

### Actual behavior
The **first message** in the **last split** is the value of the latched topic when the `rosbag record` was run (e.g., `data: "0"`).

## Implications
This behavior causes issues when recording latched topics that are not often updated. If the `rosbag record` is a long-running process, the latched topic values in some splits may be several hours old (if the publisher does not send an update within the split time frame).

## Cause
Latched messages are stored as a `std::map<std::pair<std::string, std::string>, OutgoingMessage>`, where the key is a pair of strings (topic and publisher name) and the value is the latched message.

When updating the `latched_msgs_` map, the `std::map::insert()` function is used. This behaves as expected when the key does not yet exist in the map. However, if the key exists, the old value is not updated:
> Because element keys in a map are unique, the insertion operation checks whether each inserted element has a key equivalent to the one of an element already in the container, and if so, the element is not inserted, returning an iterator to this existing element
>
> -- <cite>[std::map::insert](https://cplusplus.com/reference/map/map/insert/)</cite>

This effectively means that the first value (latched message) that was inserted into the map will persist across the whole `rosbag record`.

## Solution
Check for insertion success and assign value manually if required (see commit).

### DIFF
--- a/tools/rosbag/src/recorder.cpp
+++ b/tools/rosbag/src/recorder.cpp
@@ -347,7 +347,11 @@ void Recorder::doQueue(const ros::MessageEvent<topic_tools::ShapeShifter const>&
                 ros::M_string::const_iterator it2 = out.connection_header->find("callerid");
                 if (it2 != out.connection_header->end())
                 {
-                    latched_msgs_.insert({{subscriber->getTopic(), it2->second}, out});
+                    auto const result = latched_msgs_.insert({{subscriber->getTopic(), it2->second}, out});
+                    if (not result.second)
+                    {
+                        result.first->second = out;
+                    }
                 }
             }
         }
--- a/tools/rosbag/src/recorder.cpp
+++ b/tools/rosbag/src/recorder.cpp
@@ -348,7 +348,7 @@ void Recorder::doQueue(const ros::MessageEvent<topic_tools::ShapeShifter const>&
                 if (it2 != out.connection_header->end())
                 {
                     auto const result = latched_msgs_.insert({{subscriber->getTopic(), it2->second}, out});
-                    if (not result.second)
+                    if (not result.second)  // The map::insert function does not update values of existing keys
                     {
                         result.first->second = out;
                     }


ghost:
I am running into issues caused by this, which strongly limits the viability of `--repeat-latched`.

@mjcarroll 
Would it be possible to have a look at merging this bugfix?

tools/rosbag/src/recorder.cpp
```diff
@@ -347,7 +347,11 @@ void Recorder::doQueue(const ros::MessageEvent<topic_tools::ShapeShifter const>&
                 ros::M_string::const_iterator it2 = out.connection_header->find("callerid");
                 if (it2 != out.connection_header->end())
                 {
-                    latched_msgs_.insert({{subscriber->getTopic(), it2->second}, out});
+                    auto const result = latched_msgs_.insert({{subscriber->getTopic(), it2->second}, out});
```
AndreasNagel, line 350:
```suggestion
                    latched_msgs_[{subscriber->getTopic(), it2->second}] = out;
```

AndreasNagel: The `[]` operator does exactly what's expected here, in case the key doesn't exist in the map, it will be created and when it does exist, it will be updated.

AndreasNagel (approved):
Suggested a change, but even without it, merging what is here is better than keeping a broken implementation around.

tools/rosbag/src/recorder.cpp
ggorjup: Thanks for the suggestion - the `[]` operator is indeed cleaner since it removes the need for the extra manual check (in lines 351-354).

I'll try to update it shortly (I don't have the test environment handy anymore).

ggorjup: @AndreasNagel I tried your suggestion and unfortunately it fails to build:
`error: no matching function for call to ‘rosbag::OutgoingMessage::OutgoingMessage()’`

I believe this is due to the behavior of the `[]` operator when a key is not found in the map. From the [docs](https://cplusplus.com/reference/map/map/operator[]/):
> If k does not match the key of any element in the container, the function inserts a new element with that key and returns a reference to its mapped value.

When a key is missing, the `[]` operator tries to create an `OutgoingMessage` object with an empty constructor that does not exist: `OutgoingMessage()`. (The object seems to only have one constructor that requires 4 arguments.)

I didn't find an easy way to go around this without creating an empty constructor for `OutgoingMessage` - I'd prefer to avoid introducing a change like that.

AndreasNagel: Good point, I missed that `OutgoingMessage` doesn't have a default constructor. I agree, there's no point in creating one just to use the `[]` operator. There's insert_or_assign in c++17, but nothing with equivalent functionality in c++14, so what you have there is probably as good as it gets. Still hoping, that this change will get merged one day.

kaarelvark (approved):

kaarelvark:
Behavior has changed after #2351
Previously latched topics were repeated in the beginning of the bag without being latched.
Now those topics are latched, but they behave just as you described: first message is always the same in all split bags..

nicolas-cadart:
What's the status of this fix? We are still experiencing this issue on ROS noetic, which is a major problem for checking the real state of long term topics in splitted bags...

ggorjup:
@nicolas-cadart the fix is ready, just needs to be merged.

fabianhirmann:
The issue that the first published message on a topic is repeated instead of the last published message on a topic is still persistent in the current version.

@sloretz I saw that you have reviewed and merged the last few merge requests in this repo so you seem to be an active maintainer. Could you have a look on it and merge it if accepted? I believe the merge request is already ready to be merged.

fabianhirmann (approved):

sloretz (approved):
Thanks for the PR, and thanks for the reviews everyone! I'm going to assume the reviewers tested this change on real systems.

PR #2267, opened 2022-08-24
Expose is_shutdown_requested in rospy namespace.

MichaelGrupp:
Just like the C++ `ros::isShuttingDown()`, this can be useful in certain
cases where shutdown needs to be checked before all shutdown handlers
finished.

The difference to `is_shutdown()` is already clearly described in the
docstring of `is_shutdown_requested()` (see [here](https://github.com/ros/ros_comm/blob/dd78ac8af128bb8eb992d6431bb9f994658ea6ab/clients/rospy/src/rospy/core.py#L465)).


peci1 (approved):
LGTM.

sloretz (approved):
Thank you for the PR!

PR #2269, opened 2022-08-24
Fix/Improve $(eval ...)

rhaschke:
Evaluation of global symbols in list comprehension context (e.g. `$(eval [arg(var) for var in ['fuga', 'arg']])`) was failing:
`AttributeError: '_DictWrapper' object has no attribute 'get'`

Essentially, symbols need to be passed as the `globals` argument to `eval()` and this argument needs to inherit from `dict`.

Additionally, this PR improves the discarding of private symbols (starting with double underscore). Previously, any double underscores were considered offending, even those in the middle of a symbol or in literal text.
These are the same changes, I have implemented for `xacro`.

### DIFF
--- a/tools/roslaunch/src/roslaunch/substitution_args.py
+++ b/tools/roslaunch/src/roslaunch/substitution_args.py
@@ -38,6 +38,7 @@ by roslaunch and xacro, but it is not yet a top-level ROS feature.
 """
 
 import os
+import math
 
 try:
     from cStringIO import StringIO # Python 2.x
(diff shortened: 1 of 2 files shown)


rhaschke:
Ping @jacobperron

peci1:
This would make writing list comprehension expressions in launch files much easier! Thanks for the PR.

peci1 (approved):

peci1:
@mjcarroll @sloretz friendly request for review. This PR is good to go according to me.

sloretz:
Thank you for the PR!

PR #2277, opened 2022-09-13
Log errors in XmlRpcpp

### LINKED ISSUE #2274: xmlrpcpp: Errors are not logged
The default XmlRpc error handler [discard error messages](https://github.com/ros/ros_comm/blob/dd78ac8af128bb8eb992d6431bb9f994658ea6ab/utilities/xmlrpcpp/src/XmlRpcUtil.cpp#L66). It warns about the danger of throwing an exception, but it could at least log the error message.

Hugal31:
Errors were not logged at all. Fix this by using the default log handler with level 0.

Closes #2274


peci1:
I guess the error logging code could then be simplified by removing the `#ifdef` for Windows and just relay everything to the normal log handler.

sloretz (approved):
Thank you for the PR!

PR #2278, opened 2022-09-13
Fix EINTR handling in XmlRpcDispatch::work

### LINKED ISSUE #2275: xmlrpcpp: EINTR is not handled correctly
The syscalls (mainly the blocking one, [poll](https://github.com/ros/ros_comm/blob/dd78ac8af128bb8eb992d6431bb9f994658ea6ab/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp#L173)) are not handling EINTR correctly: they should retry instead of returning (see https://stackoverflow.com/questions/2957759/using-gprof-with-sockets).

This make profiling ROS programs impossible, as it will cause random errors in the XmlRpcClient.

Hugal31:
Only leave XmlRpcDispatchwork if the `poll` error is **not** EINTR.

The EINTR check should probably be at every I/O syscalls, but I think this `poll` call is the only one supposed to block.

Closes #2275

### DIFF
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -176,12 +176,16 @@ XmlRpcDispatch::work(double timeout)
     {
 #if defined(_WINDOWS)
       XmlRpcUtil::error("Error in XmlRpcDispatch::work: error in poll (%d).", WSAGetLastError());
-#else
-      if(errno != EINTR)
-        XmlRpcUtil::error("Error in XmlRpcDispatch::work: error in poll (%d).", nEvents);
-#endif
       _inWork = false;
       return;
+#else
+      if(errno != EINTR)
+      {
+        XmlRpcUtil::error("Error in XmlRpcDispatch::work: error in poll (%d).", nEvents);
+        _inWork = false;
+        return;
+      }
+#endif
     }
 
     // Process events
--- a/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
+++ b/utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
@@ -185,6 +185,8 @@ XmlRpcDispatch::work(double timeout)
         _inWork = false;
         return;
       }
+      // If we receive EINTR, the revents will be empty and no handleEvents will be called.
+      // It will loop back to the poll unless the timeout is reached.
 #endif
     }
 


peci1:
So what would the function do when EINTR is returned? It would continue to the next part and...? Isn't returning from `work()` the correct thing to do - the next iteration would call `poll()` again, wouldn't it?

Hugal31:
> So what would the function do when EINTR is returned? It would continue to the next part and...? Isn't returning from `work()` the correct thing to do - the next iteration would call `poll()` again, wouldn't it?

If EINTR is returned, it would continue, ignore the "process events" block as there is no event, and re-do a poll if there the deadline is not passed.

If we return (the current behavior), the caller would have to handle this case. Right now, `XmlRpcClient::execute` assumes the dispatcher returned because of an error or because the work was done. This caused me issue when debugging or profiling a ROS program (I can't remember exactly, it was quite some time ago).

utilities/xmlrpcpp/src/XmlRpcDispatch.cpp
```diff
@@ -176,12 +176,16 @@ XmlRpcDispatch::work(double timeout)
     {
 #if defined(_WINDOWS)
       XmlRpcUtil::error("Error in XmlRpcDispatch::work: error in poll (%d).", WSAGetLastError());
+      _inWork = false;
+      return;
 #else
       if(errno != EINTR)
+      {
         XmlRpcUtil::error("Error in XmlRpcDispatch::work: error in poll (%d).", nEvents);
+        _inWork = false;
```
peci1, line 185: Also, `timeout_ms` has to be recomputed from `_end_time` and `getTime()`. There is probably a larger problem here with `timeout_ms`, as the current implementation calls all `poll()`s with `timeout_ms` regardless of how much time has elapsed from the beginning of the call. Maybe the recomputation of `timeout_ms` should be moved right before the `poll()` call.

peci1 (changes requested):
Right, according to Linux `poll()` documentation, EINTR should only be returned when no events happened, so `revents` should be empty. In that case, the whole "process events" block will probably be a no-op. However, it was actually pretty difficult to figure this out.

I wonder if it weren't easier for code maintainability if there was an explicit `continue` in case of EINTR, with a comment that the `poll()` has to be tried again. The only possible issue I see is that in such case the timeout and clear checks on the end of the loop would be skipped. Another option would be to skip just the "process events" for loop in case poll was interrupted. Or just add a comment explaining what is going to happen.

I also checked the Windows part, but there doesn't seem to be anything similar on Windows. `WSAPoll()` can only be interrupted by a call to a deprecated API function. So it is probably correct that this PR leaves the Windows part as is.

Hugal31:
I agree that a `continue` would be more readable, but we can't skip the deadline check. Maybe we can add a comment about how the following block will be ignored, or change the loop to check for the deadline at the beginning (maybe using a do while).

peci1:
> Maybe we can add a comment

That should be sufficient.

peci1 (approved):
Thanks for the update!

LGTM.

sloretz (approved):
Thank you for the PR!

PR #2280, opened 2022-09-23
use itertools count for atomic increments

graingert:


peci1:
The provided code looks really suspicious. How is it guaranteed that the instance created by `count()` call is not garbage-collected? It would look more safe and descriptive to first store this instance and then use a lambda calling the iterator.

Apart from that, does this PR solve a particular performance issue, or does it just modernize the code?

For other reviewers: `itertools.count()` is implemented in C++, so it locks GIL instead of a mutex. This might be actually faster. Or might not...

graingert:
This is the style used in the stdlib see https://github.com/python/cpython/blob/3.11/Lib/asyncio/tasks.py#L32

The count object can't be garbage collected because the bound method keeps a reference to it

peci1:
Thanks for the explanation. And regarding the performance part? Is there some benefit?

sloretz:
Thank you for the PR!

Since [ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160) and this seems like a style change, I'm going to close this one.

Issue #2238, opened 2022-04-26
xmlrpc.XmlRpcNode consumes 100%+ cpu after being shutdown


drjsmith:
**TLDR**: calling `shutdown()` on a `XmlRpcNode` does not fully shut it down and leaves a thread running. 
I am currently on Noetic/Ubuntu 20.04/python 3.8. 

**What is going on:**
- Calling `start()` results in a new thread running [`self.server.serve_forever()`](https://github.com/ros/ros_comm/blob/3f8eca4a1f13996b33c03a5c4ee3a69fb0db3668/tools/rosgraph/src/rosgraph/xmlrpc.py#L321), where `server` is of type `ThreadingXMLRPCServer` which has a base class of `socketserver.BaseServer`.
- That function, defined for `BaseServer` in `socketserver.py:215`, will not exit until the loop's condition is satisfied: `while not self.__shutdown_request:`
- `self.__shutdown_request` can be set by calling `shutdown()` on `BaseServer`
- However, `XmlRpcNode`'s [`shutdown()`](https://github.com/ros/ros_comm/blob/3f8eca4a1f13996b33c03a5c4ee3a69fb0db3668/tools/rosgraph/src/rosgraph/xmlrpc.py#L215
) method never calls `self.server.shutdown()`, only ` server.socket.close()` and `server.server_close()`. I'm pretty sure those last 2 functions do the same thing, since:
```
class TCPServer(BaseServer):
    ...
    def server_close(self):
        self.socket.close()
```
The thread is then stuck in an infinite loop. Before shutting down `XmlRpcNode`, the thread not use a significant amount of cpu; afterwards, however, the thread consumes 100% of a single core (as shown by `top`).

It seems to me that the solution may be as simple as adding `self.server.shutdown()` to `XmlRpcNode`'s `shutdown()` function, though I'm not certain whether it should be before or after closing the socket or if there are other considerations. 

**Why this matters to me:**
[...]

PR #2281, opened 2022-09-27
Thread is stucking on shutdown

l1va:
Fixing the issue with 100% CPU usage after the shutdown 
https://github.com/ros/ros_comm/issues/2238

### DIFF
--- a/tools/roslaunch/src/roslaunch/parent.py
+++ b/tools/roslaunch/src/roslaunch/parent.py
@@ -279,7 +279,12 @@ class ROSLaunchParent(object):
         
         if self.server:
             try:
-                self.server.shutdown("roslaunch parent complete")
+                import sys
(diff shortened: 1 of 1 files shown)


peci1:
I can't follow the reasoning why this proposed solution would work. It is difficult to dig through all the indirections involved.

What is the difference in calling shutdown with args or without? Does it take a different code path?

This would also deserve a test to show the actual problem and that this PR solves it.

sloretz:
Thank you for the PR!

I'm not sure how this would fix the issue either. Since [ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160), I don't think this one will get merged before EOL, so I'm going to close it.

Issue #2283, opened 2022-09-28
ros::Timer & ros::ServiceServer finalization failure


aurzenligl:
Delivering #2121 broke existing `ros::Timer` and `ros::ServiceServer` usages, introducing possibility of using destroyed objects by last callback call. Affected usages in public projects look like this:
https://github.com/ros-planning/moveit/blob/1.1.10/moveit_ros/moveit_servo/include/moveit_servo/collision_check.h
https://github.com/ros-planning/moveit/blob/1.1.10/moveit_ros/moveit_servo/src/collision_check.cpp
One production system I'm aware of was affected as well.

Before #2121 stop/shutdown post-condition ensured callback is not being executed, now it doesn't, allowing destruction of referred-to data and undefined behavior when last callback call continues. Following example illustrates such crash scenario:
https://github.com/aurzenligl/study/blob/master/ros-timer/src/race.cpp

When `ros::Timer` or `ros::ServiceServer` is created with a callback capturing data by reference, there is no easy way to guarantee that lifetime of callback execution doesn't exceed the lifetime of data it uses. Problem occurs in code bases where objects composing timers or srv-servers (capturing `this`) are created/destroyed repeatedly while spinners are running.

Phrasing that documentation uses to describe stop/shutdown/dtor behavior is as such:
- [ros::Timer](http://docs.ros.org/en/noetic/api/roscpp/html/classros_1_1Timer.html) "(...) no more callbacks will be called"
- [ros::ServiceServer](http://docs.ros.org/en/noetic/api/roscpp/html/classros_1_1ServiceServer.html) "(...) service callback will stop being called"
- [ros::Subscriber](http://docs.ros.org/en/noetic/api/roscpp/html/classros_1_1Subscriber.html) "(...) will stop being called"
- [ros::NodeHandle::createTimer](http://docs.ros.org/en/noetic/api/roscpp/html/classros_1_1NodeHandle.html) "(...) callback will no longer be called"

It's imprecise wrt waiting for currently executing callback, but since all above classes did block in stop/shutdown before #2121, and `ros::Subscriber` still blocks, I assume proper/expected/consistent behavior is to block.

iwanders:
> Following example illustrates such crash scenario: https://github.com/aurzenligl/study/blob/master/ros-timer/src/race.cpp

Yikes, that's definitely not good. Thanks for a minimal example to reproduce this failure.

> I assume proper/expected/consistent behavior is to block.
&
[...]

PR #2284, opened 2022-09-28
Fix for timer/srv-server finalization issue 2283

aurzenligl:
As explained in #2283, delivering #2121 opened a possibility for a race condition during ros::Timer and ros::ServiceServer finalization: https://github.com/aurzenligl/study/blob/master/ros-timer/src/race.cpp

In order to remedy the situation, but still allow the [scenario](https://gist.github.com/iwanders/ede48fb649fd47f9b1f9a52c527b463c) mentioned in #2121 to succeed, I propose to define post-condition of removeById this way: callback is not and will not be executed, unless removal happened in the context of callback (self-removal). It's compatible with user code prior to #2121, as then removeByID always blocked (even on self-removal, leading to deadlock).

This allows to satisfy both use cases:
- https://gist.github.com/iwanders/ede48fb649fd47f9b1f9a52c527b463c:
doesn't deadlock as long as timer stop is called w/o external mtx locked
(see: selfRemoveCallbackWhileExecuting testcase),
- https://github.com/aurzenligl/study/blob/master/ros-timer/src/race.cpp:
doesn't crash as ros::Timer::stop guarantees no spinner uses any cb-captured data anymore
(see: removeCallbackWhileExecuting testcase).

FYI: @iwanders, @jacobperron, @fujitatomoya

### DIFF
--- a/clients/roscpp/src/libros/callback_queue.cpp
+++ b/clients/roscpp/src/libros/callback_queue.cpp
@@ -166,9 +166,10 @@ void CallbackQueue::removeByID(uint64_t removal_id)
     }
 
     {
-      boost::unique_lock<boost::shared_mutex> rw_lock(id_info->calling_rw_mutex, boost::defer_lock);
-      if (rw_lock.try_lock())
+      // Unless we're removing from callback, we lock the calling mutex to ensure callback is not being executed.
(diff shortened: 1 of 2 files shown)


sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #2286, opened 2022-09-30
rosbag reindex bugfix - seek to truncated position after broken chunk

### LINKED ISSUE #2282: rosbag reindex outputs invalid bags when run on truncated files
The rosbag reindex command is documented as "repairing broken bag files", but when run on a broken bag file it will output a bag that is indexed, but illegal per the ROS bag spec.

This script can be used to demonstrate the issue:
https://gist.github.com/wkalt/836cd80f1c6a9b0e87e5e24771c7b8e0

Steps to reproduce:
```
wget https://assets.foxglove.dev/demo.bag
truncate -s 30000000 demo.bag
rosbag reindex demo.bag

// using go program above
$ go run main.go demo.bag
2022/09/27 09:18:10 got zero length header at offset 29578569
exit status 1
```

The resulting bag violates the spec in that it is not a sequence of back to back `<header length><header><data length><data>` records, with each header containing an "op" key, as the bag spec mandates. However, ROS tooling accepts the bag due to reliance on the message index, which is well-formed. The index simply does not point at the junk data present in the file, which occurs after the final chunk and its message indexes.

This causes problems for tooling that does not use the index and instead performs a linear scan. It is possible this issue does not present if you are lucky enough to truncate the file on a message boundary.

emersonknapp:
Fixes #2282

The truncate operation left the current `f.tell()` read head at the pre-truncated position, so the chunk infos that are written on closing the file started writing at this pre-truncated position, leaving dangling broken chunk data in between the last good chunk and the first file-end chunkinfo.

### DIFF
--- a/test/test_rosbag/test/test_bag.py
+++ b/test/test_rosbag/test/test_bag.py
@@ -261,6 +261,57 @@ class TestRosbag(unittest.TestCase):
 
             msgs = list(rosbag.Bag(reindex_filename, 'r'))
 
+    def _read_all_records(self, bag_path):
+        """
+        Throw exception if malformed records found, otherwise return number of records in file.
+        """
+        with rosbag.Bag(bag_path, 'r') as b:
+            file_header_pos = b._file_header_pos
+
+        total_records_read = 0
+        with open(bag_path, 'r+b') as f:
+            f.seek(0, os.SEEK_END)
+            total_bytes = f.tell()
+            f.seek(file_header_pos)
+            while f.tell() < total_bytes:
+                bag._read_sized(f)
+                total_records_read += 1
+
+        return total_records_read
+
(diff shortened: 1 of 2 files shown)


emersonknapp:
Added a regression test - verified that test fails without `f.seek` change, passes with it.

emersonknapp:
@jacobperron @mjcarroll @sloretz a review would be much appreciated :) hopefully I have made it easy with the regression test.
 
This bug is a far corner case, but it's real: someone using a format-compliant parser that is reading records without using the index will find invalid records in the file, even after using the `reindex` command which purports to fix the file.

peci1:
Looks good to me, tests pass.

mjcarroll (approved):

Issue #2221, opened 2022-01-13
rosgraph fails to start on Kernels with no minor version number


c-rizz:
Hello! I recently updated my machine to kernel 5.16 and on this specific kernel rqt fails to start because of rosgraph crashing. This because of line https://github.com/ros/ros_comm/blob/52b0556dadf3ec0c0bc72df4fc202153a53b539e/tools/rosgraph/src/rosgraph/xmlrpc.py#L92 
This happens because platform.release() returns '5.16-051600-generic' and not something like e.g. '5.16.0-051600-generic'. I believe this is a relatively common occurrence. The problem can be solved by changing the line to`release = platform.release().replace("-",".").split('.')`
Do I submit a pull request?

Thank you for the great tools!

MatthijsBurgh:
You mean missing patch number 😉

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #2290, opened 2022-10-11
More robust Linux release parsing

### LINKED ISSUE #2289: Kernel release parsing
I am running Ubuntu 20.04 with the 6.0 kernel, installed via `ppa:tuxinvader/lts-mainline`.

I am not sure there are any rules that apply to the release name of a kernel. But it is `6.0-060000-generic`. While for 5.15 is `5.15.0-48-generic`. So the big difference here is that the patch version is not provided.

In rosgraph.xmlrpc you split the release on dots. So normally this would caused the last element to contain the patch version and all other release info. Now this is included in the minor version, which causes the current error.

```py
  File "/opt/ros/noetic/bin/roslaunch", line 34, in <module>
    import roslaunch
  File "/opt/ros/noetic/lib/python3/dist-packages/roslaunch/__init__.py", line 57, in <module>
    from .launch import ROSLaunchRunner
  File "/opt/ros/noetic/lib/python3/dist-packages/roslaunch/launch.py", line 55, in <module>
    from roslaunch.nodeprocess import create_master_process, create_node_process, DEFAULT_TIMEOUT_SIGINT, DEFAULT_TIMEOUT_SIGTERM
  File "/opt/ros/noetic/lib/python3/dist-packages/roslaunch/nodeprocess.py", line 53, in <module>
    from rosmaster.master_api import NUM_WORKERS
  File "/opt/ros/noetic/lib/python3/dist-packages/rosmaster/__init__.py", line 35, in <module>
    from .main import rosmaster_main
  File "/opt/ros/noetic/lib/python3/dist-packages/rosmaster/main.py", line 43, in <module>
    import rosmaster.master
  File "/opt/ros/noetic/lib/python3/dist-packages/rosmaster/master.py", line 45, in <module>
    import rosgraph.xmlrpc
  File "/opt/ros/noetic/lib/python3/dist-packages/rosgraph/xmlrpc.py", line 103, in <module>
    class SilenceableXMLRPCRequestHandler(SimpleXMLRPCRequestHandler):
  File "/opt/ros/noetic/lib/python3/dist-packages/rosgraph/xmlrpc.py", line 104, in SilenceableXMLRPCRequestHandler
    if _support_http_1_1():
  File "/opt/ros/noetic/lib/python3/dist-packages/rosgraph/xmlrpc.py", line 94, in _support_http_1_1
    platform_minor = int(release[1])
ValueError: invalid literal for int() with base 10: '0-060000-generic'
```
https://github.com/ros/ros_comm/blob/f5fa3a168760d62e9693f10dcb9adfffc6132d22/tools/rosgraph/src/rosgraph/xmlrpc.py#L94

I suggest to change https://github.com/ros/ros_comm/blob/f5fa3a168760d62e9693f10dcb9adfffc6132d22/tools/rosgraph/src/rosgraph/xmlrpc.py#L92
to the following:
```py
 release = platform.release().split('-')[0].split('.')
```

MatthijsBurgh:
This more robust than the current code.

The current code leaves all the remaining release info into the patch version number of the version. When the version doesn't include a patch version number. The minor version number is polluted. Which causes the error of #2289.

By splitting on the dash. Only the major, minor(, patch) is used. And the remaining release info is ignored.

Fixes #2289

### DIFF
--- a/tools/rosgraph/src/rosgraph/xmlrpc.py
+++ b/tools/rosgraph/src/rosgraph/xmlrpc.py
@@ -89,7 +89,7 @@ def _support_http_1_1():
     if platform.system() != 'Linux':
         return True
     minimum_supported_major, minimum_supported_minor = (4, 16)
-    release = platform.release().split('.')
+    release = platform.release().split('-')[0].split('.')
     platform_major = int(release[0])
     platform_minor = int(release[1])
     if platform_major < minimum_supported_major:


MatthijsBurgh:
Friendly ping @jacobperron @mjcarroll @sloretz

MatthijsBurgh:
Friendly ping @jacobperron @mjcarroll @sloretz

peci1:
Related issue: https://github.com/ros/ros_comm/issues/2221

peci1:
There already is a PR addressing this exact issue: https://github.com/ros/ros_comm/pull/2202 . Can you compare the two approaches?

MatthijsBurgh:
@peci1 I think the if condition you introduced at https://github.com/ros/ros_comm/pull/2202/files#diff-8418703149a6073387df09d2d41702b620fd1d0b84a6f526341204a461c8a965R98 is a shortcut to return faster. But not really related to the bug.

I think the changes regarding the `release` and `platform_X` variables in this PR are more elegant than in #2202. As it doesn't introduce a dependency on the `re` module. Also I think a lot of people are not really familiar with regex and therefore the code is less readable. Also the `release` variable in this PR is not polluted with build information, which is still the case in #2022.

Therefore I think this PR is a more cleaner solution than #2022. Though the added condition mentioned at the start of this comment could still be merged, but I recommend to do it in a separate PR as it isn't needed for the bugfix.

peci1:
> @peci1 I think the if condition you introduced at https://github.com/ros/ros_comm/pull/2202/files#diff-8418703149a6073387df09d2d41702b620fd1d0b84a6f526341204a461c8a965R98 is a shortcut to return faster. But not really related to the bug.

Correct. It might just be over-engineering. It's not needed. However, it could be moved a few lines higher to avoid calling `re.sub` at all if major version differs.

> it doesn't introduce a dependency on the `re` module

That's a non-issue. `re` is a part of all standard Python distributions.

> Also the `release` variable in this PR is not polluted with build information, which is still the case in #2022.

I don't see a problem there. `release` is just a local variable. It only matters how you use it.

> Therefore I think this PR is a more cleaner solution than #2022.

Cleaner, in terms of code readability, yes. However, the regex approach does not care what is the actual separator (your approach assumes `-`). The regex will just find first non-numeric character and remove everything starting with it. If we're seeking for making the code more bulletproof, I'd still prefer the regex.

MatthijsBurgh:
> I don't see a problem there. `release` is just a local variable. It only matters how you use it.

True, but keeping it clean is still very useful for future changes and understanding of the code.

> Cleaner, in terms of code readability, yes. However, the regex approach does not care what is the actual separator (your approach assumes `-`). The regex will just find first non-numeric character and remove everything starting with it. If we're seeking for making the code more bulletproof, I'd still prefer the regex.

I first split on `-` to drop all extra info except from the major.minor.patch information. I am not sure if there is any standard on the release names, but I have never seen any other character used there. In case there is no `-`, the first split will just return a list of size one. The second split on `.`, will split the major, minor(, patch) version into a list. I think this is sufficient robust enough.

peci1:
To gather facts, I've looked up how Linux builds the kernel release string (`uname -r`, which should be what `platform.release()` returns).

It starts with:

`KERNELVERSION = $(VERSION)$(if $(PATCHLEVEL),.$(PATCHLEVEL)$(if $(SUBLEVEL),.$(SUBLEVEL)))$(EXTRAVERSION)` ([here](https://github.com/torvalds/linux/blob/b7b275e60bcd5f89771e865a8239325f86d9927d/Makefile#L373))

Then it continues with:

`"$(KERNELVERSION)$$($(CONFIG_SHELL) $(srctree)/scripts/setlocalversion $(srctree))"` ([here](https://github.com/torvalds/linux/blob/b7b275e60bcd5f89771e865a8239325f86d9927d/Makefile#L2097)).

And finishes with appending whatever [setlocalversion](https://github.com/torvalds/linux/blob/b7b275e60bcd5f89771e865a8239325f86d9927d/scripts/setlocalversion) script outputs.

[KConfig docs](https://github.com/torvalds/linux/blob/d4013bc4d49f6da8178a340348369bb9920225c9/Documentation/kbuild/makefiles.rst#9-kbuild-variables) say:

> $(EXTRAVERSION) defines an even tinier sublevel for pre-patches or additional patches. It is usually some non-numeric string such as "-pre4", and is often blank.

So, technically, there is nothing that would define that a dash has to be the first character of `EXTRAVERSION` or the localversion string.  The only definition is that it "is usually a non-numeric string". Well, if it'd start with a number, parsing would be difficult, so I'd translate that as "is a string that does not start with a number". And that's exactly what the regex checks.

On the other hand, I've checked a lot of various kernels on various machines (even release candidate kernels), and in all cases, there is either a sublevel version (the second dot), or the extraversion is delimited by `-`. So practically, your PR could work 99.9% of cases. The regex approach should work in 100% of cases. I think now it's up to the maintainers (@sloretz , @mjcarroll) to decide which way to take. But please, finally take an action in this problem - it's apparent it is bugging several developers and will only get worse in the future.

MatthijsBurgh:
Friendly ping @sloretz @mjcarroll

MatthijsBurgh:
Less friendly ping @sloretz @mjcarroll

mjcarroll:
I prefer the approach in #2202, I will merge and release that today.

PR #2294, opened 2022-10-22
Allow forwarding of working directory and launch file args to add_rostest from gtest wrapper functions

daniel-s-ingram:
I'm not great with cmake, so there very well may be a better way to do what I'm trying to do here. If this is acceptable, I'll add some unit tests and what not, but I wanted to go ahead and open the PR to get some feedback on whether or not there's a better way to do this.

This would have been more straightforward using a named argument for the sources, but I didn't want to break backwards compatibility.


peci1:
Generally, this PR could work. A test-case (utilizing the added features) is almost a must to be sure it actually works.

Could you also describe the scenario where this PR helps?

sloretz:
Thank you for the PR!

Issue #1951, opened 2020-05-11
Rosbag wrapper does not recognize option "--min-space"


romayiii:
Got this error while executing the command:

```
$ rosbag record --min-space=100M /rosout
Usage: rosbag record TOPIC1 [TOPIC2 TOPIC3 ...]

rosbag: error: no such option: --min-space
```
This seems to be only an issue with the wrapper since the command:

```
$ rosrun rosbag record --min-space=100M /rosout
[ INFO] [1589211746.361097235]: Subscribing to /rosout
[ INFO] [1589211746.367071524]: Recording to '2020-05-11-17-42-26.bag'.
```
works successfully.

dirk-thomas:
Please consider to contribute a pull request to add the option to the Python CLI.

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

PR #2298, opened 2022-11-14
Add --min-space option to rosbag record Python cli (#1951)

Jeffxzj:
The "--min-space" option is missed in the Python rosbag record cli. This fix issue #1951.


sloretz (approved):
Thank you for the PR!

PR #2299, opened 2022-11-18
Fix error reporting TCPROSServer

g-gemignani:
Stumbled across this little bug while hunting for another one

### DIFF
--- a/clients/rospy/src/rospy/impl/tcpros_base.py
+++ b/clients/rospy/src/rospy/impl/tcpros_base.py
@@ -322,17 +322,21 @@ class TCPROSServer(object):
                 header = read_ros_handshake_header(sock, StringIO(), buff_size)
             else:
                 header = read_ros_handshake_header(sock, BytesIO(), buff_size)
-            
+
             if 'topic' in header:
                 err_msg = self.topic_connection_handler(sock, client_addr, header)
             elif 'service' in header:
                 err_msg = self.service_connection_handler(sock, client_addr, header)
             else:
                 err_msg = 'no topic or service name detected'
+
+            if isinstance(err_msg, dict):
+                err_msg = err_msg.get('error', 'unhandled connection')
+
             if err_msg:
                 # shutdown race condition: nodes that come up and down
                 # quickly can receive connections during teardown.
-                
+
                 # We use is_shutdown_requested() because we can get
                 # into bad connection states during client shutdown
                 # hooks.


mjcarroll:
Can you describe the bug that this is fixing please?

peci1:
@g-gemignani The fix looks reasonable, but I agree with Michael a little context would help. In what situation did you face an error? How did it manifest?

g-gemignani:
Hi and thank you for the answer. Yes, you are right, I was in a rush when I opened the fix and did not attach more info. Here it is:
I was debugging a python interactive CLI that was spawning ros nodes in child processes upon issuing commands. The library was using multiprocessing (https://docs.python.org/3/library/multiprocessing.html).
The issue that I was getting was a cascade of:

> Inbound TCP/IP connection failed: can't concat dict to bytes
Inbound TCP/IP connection failed: can't concat dict to bytes
Inbound TCP/IP connection failed: can't concat dict to bytes
Inbound TCP/IP connection failed: can't concat dict to bytes
Inbound TCP/IP connection failed: can't concat dict to bytes
Inbound TCP/IP connection failed: can't concat dict to bytes
Inbound TCP/IP connection failed: can't concat dict to bytes
Inbound TCP/IP connection failed: can't concat dict to bytes

After digging a bit, I found out that the issue was in ros_comm:
> Traceback (most recent call last):
  File "/home/toru/catkin_ws/install/lib/python3/dist-packages/rospy/impl/tcpros_base.py", line 350, in _tcp_server_callback
    write_ros_handshake_header(sock, {'error' : err_msg})
  File "/home/toru/catkin_ws/install/lib/python3/dist-packages/rosgraph/network.py", line 416, in write_ros_handshake_header
    s = encode_ros_handshake_header(header)
  File "/home/toru/catkin_ws/install/lib/python3/dist-packages/rosgraph/network.py", line 403, in encode_ros_handshake_header
    fields = [k + b"=" + v for k, v in sorted(encoded_header.items())]
  File "/home/toru/catkin_ws/install/lib/python3/dist-packages/rosgraph/network.py", line 403, in <listcomp>
    fields = [k + b"=" + v for k, v in sorted(encoded_header.items())]
TypeError: can't concat dict to bytes
Inbound TCP/IP connection failed: can't concat dict to bytes

where encoder_header was:
`{b'error': {'error': 'unhandled connection'}}`
and [this line of code](https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosgraph/src/rosgraph/network.py#L403) was expecting encoder_header to be a dict[str: str].
Hence, the fix.
Please let me know if you need more info!

peci1:
Is there a way to reproduce the problem?

g-gemignani:
I tried to reproduce the issue but failed to do so. If you do not think that this fix is worth merging, feel free to decline it. Up to you :)

peci1:
I'm not a maintainer of rospy, so I'll leave that decision on @mjcarroll .

sloretz:
Thank you for the PR!

Since [ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160) and you weren't able to reproduce the issue, I'm going to close this one.

PR #2305, opened 2022-12-07
[topic_tools/mux] add wait_publisher_initialization option in mux

knorth55:
`mux` node does not have `sleep` between `advertise` and `publish`.
this causes dropping first message publication.
this does not matter when the topic's frequency is high.
but when the topic's frequency is low, dropping the first message causes huge time loss.

This PR add option to add `0.1ms` sleep between `advertise` and `publish`.
This change avoids dropping the first message.

### Sample launch

Publish topic in 10 second.
without this PR, the first message of `/node_mux` will drop,
so we need to wait 20 seconds to get `/node_mux` topic.

```xml
<launch>
  <arg name="wait_publisher_initialization" default="false" />

  <node name="node1" pkg="rostopic" type="rostopic"
        args="pub -r 0.1 /node1 std_msgs/String node1" 
        output="screen" />
  <node name="node2" pkg="rostopic" type="rostopic"
        args="pub -r 0.1 /node2 std_msgs/String node2"
        output="screen" />
  <node name="mux" pkg="topic_tools" type="mux"
        args="/node_mux /node1 /node2"
        output="screen">
    <rosparam subst_value="true">
      wait_publisher_initialization: $(arg wait_publisher_initialization) 
    </rosparam>
  </node>
</launch>
```

## `wait_publisher_initialization`: true (this PR)

After 10 seconds, `/node1`, `/node2` and `/node_mux` arrived!

```
$ rostopic echo /node1
data: "node1"
---
```

```
$ rostopic echo /node2
data: "node2"
---
```

```
$ rostopic echo /node_mux
data: "node1"
---
```

## `wait_publisher_initialization`: false (current behavior)

After 10 seconds, `/node1` and `/node2` arrive, but `/node_mux` did not arrived.

```
$ rostopic echo /node1
data: "node1"
---
```

```
$ rostopic echo /node2
data: "node2"
---
```

```
$ rostopic echo /node_mux
```

### DIFF
--- a/tools/topic_tools/src/mux.cpp
+++ b/tools/topic_tools/src/mux.cpp
@@ -52,6 +52,7 @@ static ros::NodeHandle *g_node = NULL;
 static bool g_lazy = false;
 static bool g_latch = false;
 static bool g_advertised = false;
+static bool g_wait_pub_init = false;
 static string g_output_topic;
 static ros::Publisher g_pub;
 static ros::Publisher g_pub_selected;
@@ -152,6 +153,11 @@ void in_cb(const boost::shared_ptr<ShapeShifter const>& msg,
   {
     ROS_INFO("advertising");
     g_pub = msg->advertise(*g_node, g_output_topic, 10, g_latch, conn_cb);
+    // we need sleep for publisher initialization
+    // otherwise the first topic will drop.
+    if (g_wait_pub_init) {
+      usleep(100000);
+    }
     g_advertised = true;
     
     // If lazy, unregister from all but the selected topic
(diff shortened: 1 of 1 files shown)


knorth55:
@mjcarroll can you review this PR?
this first topic drop causes unintended robot behavior.

Our case:
In our lab, we use `mux` for `speech-to-text` result, because there are several running `speech-to-text` engines to select.
When a person talk to the robot, the robot always ignore the person at first, which is caused by this first topic drop.
Therefore, we need to speak the same thing twice when `mux` start publishing the topic.

In order to avoid this behavior, we make this `wait_publisher_initialization` option for `mux`.
This PR does not change the default behavior.

peci1 (approved):
These changes look good to me. You might consider making this interval configurable, as some systems might need longer time (e.g. if there are nodes running across several computers). In that case, I'd prefer adding a second parameter with the interval length.

knorth55:
@peci1 Thank you for your advice! I add `wait_publisher_second` param

mjcarroll (approved):

PR #2315, opened 2023-02-21
Maintain constness of accessed XmlRpcValue struct

rhaschke:
Return value should not change constness of `this` pointer!
This augments/completes #1978.


peci1 (approved):
Logically, this change makes sense. I'm a bit afraid there could be some downstream code depending on the old twisted logic, though. Not sure this would pass the API stability requirements.

I ran [extensive test suite](https://github.com/ctu-vras/ros-utils/blob/master/cras_cpp_common/test/test_xmlrpc_value_utils.cpp) on my [xmlrpcvalue wrapper](https://github.com/ctu-vras/ros-utils/blob/master/cras_cpp_common/include/cras_cpp_common/xmlrpc_value_utils.hpp) and found no issues. But that's just one library.

peci1:
Just a note: The getter does not change constness of `this->`. It just returns one non-const value from the const `struct` (map-like) object. Yes, it seems silly, but the current API allows that. The constness is applied just to the internal pointer, not to the pointed-to value.

sloretz (approved):
Thank you for the PR!

PR #2320, opened 2023-03-23
Fix error "s is not defined" and whitespace cleanup

oss-vineet:


peci1 (changes requested):
The fix you've provided seems good. However, the whitespace changes are generally not acceptable when mixed together with functional changes.

Please, split this to two PRs: one with the functional fix, and one with the whitespace changes only. However, I'm not even sure somebody would approve the whitespace PR, as that brings no value and complicates things like git blame.

PR #2325, opened 2023-04-09
roscpp: Fixed remapping of private parameters in anonymous nodes

### LINKED ISSUE #2324: Remapping private topics of anonymous nodes works differently in rospy and roscpp
When I launch a node from CLI via rosrun and do not set any name, the nodes with anonymous flag generate a random name.

If I pass a local topic remap (`~input:=test`), rospy correctly remaps this topic, while roscpp does not.

The problem seems to be in this part of this_node code:

https://github.com/ros/ros_comm/blob/842f0f026924323f605495da0b80493f15f0bdce/clients/roscpp/src/libros/this_node.cpp#L120-L170

It first sets `name_` to the `init_node()` argument, possibly replaces it with `__name` remap, and then it calls `names::init()`.

However, `names::init()` calls `names::resolve()`, which in turn calls `this_node::getName()`:

https://github.com/ros/ros_comm/blob/842f0f026924323f605495da0b80493f15f0bdce/clients/roscpp/src/libros/names.cpp#L167-L170

But the name in `this_node` is only the non-anonymous one - the anonymizing part is appended only after `names::init()` finishes.

Minimal code showing the problem is this:

```c++
#include <ros/ros.h>

{
  ros::init(argc, argv, "node", ros::init_options::AnonymousName);
  
  ros::NodeHandle nh ("~");
  ROS_ERROR("%s", nh.resolveName("test").c_str());
}
```

run as `node ~test:=a` to see the problem. Running with `node ~test:=a __name:=t` works as expected, because explicit name setting disables the anonymous node behavior.

Please note that running the node via roslaunch and `name="$(anon node)"` would not trigger this issue as it does not actually trigger the anonymous init option of the node (it directly generates a random name which is passed to the node in `__name:=` remap).

For comparison, here is the Python code that works correctly:

```py
import rospy

rospy.init_node("node", anonymous=True)
rospy.logerr(rospy.names.resolve_name("~test"))
```

The solution would be calling `names::init()` once more after the name mangling is finished. Would that be okay?

peci1:
Fixes #2324.

This fix runs all private name remaps through another round of remapping after the anonymous node name is finished. The first remapping round is done with incomplete node name, which resulted in inability to remap private names for anonymous nodes without `__name:=` arg.

The new behavior corresponds to what rospy is doing.

The impact on existing code should be close to zero. Anything launched from a launch file will not be affected. The only affected usages are manual `rosrun` commands (and if somebody does some crazy things around `ros::init()`).

I did not find a good way to remove the wrong remappings, so they remain there. I.e. `g_remappings` will contain both `/node/topic->newtopic` and `/node_12321321/topic->newtopic`. This isn't correct, but it could actually serve as a backwards compatibility feature for the case if someone was working around #2324 by explicitly remapping the incomplete name.

Real-world tools affected by #2324 can be found e.g. here: http://docs.ros.org/en/latest/api/cras_topic_tools/html/classcras_1_1ThrottleMessagesNodelet.html#details . It is an improved version of `topic_tools throttle` which, if no topic args are given through CLI, subscribes to `~input` topic. When running as `rosrun cras_topic_tools throttle ~input:=/remapped`, the remap does not happen without this PR. Appending `__name:=foo` makes the remap happy, which is confusing.

### DIFF
--- a/clients/roscpp/src/libros/this_node.cpp
+++ b/clients/roscpp/src/libros/this_node.cpp
@@ -167,6 +167,22 @@ void ThisNode::init(const std::string& name, const M_string& remappings, uint32_
     char buf[200];
     std::snprintf(buf, sizeof(buf), "_%llu", (unsigned long long)WallTime::now().toNSec());
     name_ += buf;
+    
+    // names::init() above did not yet know the full name of the node, so the
+    // remappings for private names it set are wrong. This second call fixes
+    // that. However, the old remappings (without the anonymizing part) still
+    // remain set. There is no easy way to remove them, so we keep them there.
+    M_string privateRemappings;
+    for (const auto& leftRight : remappings)
+    {
+      const auto& left = leftRight.first;
+      if (!left.empty() && left[0] == '~')
+      {
+        privateRemappings[left] = leftRight.second;
+      }
+    }
+    
+    names::init(privateRemappings);
   }
 
   ros::console::setFixedFilterToken("node", name_);
--- a/test/test_roscpp/test/CMakeLists.txt
+++ b/test/test_roscpp/test/CMakeLists.txt
@@ -156,3 +156,5 @@ add_rostest(launch/multiple_latched_publishers.xml)
 
 # Test spinners
 add_rostest(launch/spinners.xml)
+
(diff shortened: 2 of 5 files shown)


sloretz:
Thank you for the PR!

peci1:
This PR comes with a rostest that tests the described behavior.

sloretz:

* it only supports ROS 1

PR #2328, opened 2023-04-11
Fix error "s is not defined" (reopening #2320 again)

oss-vineet:
This time, without the whitespace changes.

However seeing that in #2320 @peci1 mentioned that whitespace changes complicate `git blame`, I will not submit the whitespace changes again.

### DIFF
--- a/clients/rospy/src/rospy/topics.py
+++ b/clients/rospy/src/rospy/topics.py
@@ -1236,7 +1236,7 @@ class _TopicManager(object):
         elif reg_type == Registration.SUB:
             rmap = self.subs
         else:
-            raise TypeError("invalid reg_type: %s"%s)
+            raise TypeError("invalid reg_type: %s"%reg_type)
         return rmap.get(resolved_name, None)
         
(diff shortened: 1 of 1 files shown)


clients/rospy/src/rospy/topics.py
```diff
@@ -1236,7 +1236,7 @@ def get_impl(self, reg_type, resolved_name):
         elif reg_type == Registration.SUB:
             rmap = self.subs
         else:
-            raise TypeError("invalid reg_type: %s"%s)
+            raise TypeError("invalid reg_type: %s"%reg_type)
```
peci1:
```suggestion
            raise TypeError("invalid reg_type: %s"%(reg_type,))
```
I'd suggest adding the brackets to make the argument a 1-element tuple. It it most probably not needed here, but it is a good Python habit.

peci1 (approved):
Thanks for taking the time to improve this PR. LGTM!

mjcarroll (approved):

PR #2329, opened 2023-05-05
Rostopic fix

prarobo:
Currently loading a message from file does not apply argument substitution for each published message. This fixes that.

### DIFF
--- a/tools/rostopic/src/rostopic/__init__.py
+++ b/tools/rostopic/src/rostopic/__init__.py
@@ -35,34 +35,34 @@
 
 # make sure we aren't using floor division
 from __future__ import division, print_function
-
-NAME='rostopic'
-
-import argparse
-import os
-import sys
-import math
-import socket
-import time
-import traceback
+import rospy
+import rosgraph
+import roslib.message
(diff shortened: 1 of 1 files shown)


tools/rostopic/src/rostopic/__init__.py
```diff
@@ -1939,7 +1939,7 @@ def param_publish(pub, msg_class, param_name, rate, verbose):
         if rospy.is_shutdown():
             break
 
-def stdin_publish(pub, msg_class, rate, once, filename, verbose):
+def stdin_publish(pub, msg_class, rate, once, filename, verbose, substitute_keywords):
```
peci1, line 1942:
```suggestion
def stdin_publish(pub, msg_class, rate, once, filename, verbose, substitute_keywords=False):
```

As this is a library downstream code could be using, the old API must be retained.

peci1 (changes requested):

sloretz:
Thank you for the PR!

Unfortunately I don't think we should merge this one because it would change an API in a released ROS distro.

PR #2335, opened 2023-07-03
Add bool return to _base_logger, log* functions

owenclaxton-QUT:
Add a bool return to expose whether log succeeded for:
- _base_logger
- log* functions (log*, log*_throttle, log*_once, etc...) 

**Explanation**
Provides the user a trigger for actions which may be sensitive to logging (i.e. creating a 'static' diagnostic display). The return will be true, if:
- is_initialized() succeeded
- it was 'time' for a throttled logger to write a new log event
In both these cases, true represents a change to console, false represents no change.

**Specific example use case:**
When working with the log*_throttle functions, I needed a way to detect whether a loginfo_throttle was actually printing, as I needed to know how many lines to escape and clear before writing new content on-top for a 'static' diagnostic display. The bool already existed, but just needed to be exposed to the user.

### DIFF
--- a/clients/rospy/src/rospy/core.py
+++ b/clients/rospy/src/rospy/core.py
@@ -157,6 +157,28 @@ def _frame_to_caller_id(frame):
 
 def _base_logger(msg, args, kwargs, throttle=None,
                  throttle_identical=False, level=None, once=False):
+    """
+    Base logger function that handles logging of normal, throttled, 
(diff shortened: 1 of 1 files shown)


peci1 (approved):
This change looks okay to me, but it needs a second eye.

sloretz (approved):
Thank you for the PR!

PR #2348, opened 2023-09-12
Update: #include <boost/bind.hpp> -> <boost/bind/bind.hpp> for boost 1.73

rhaschke:
Since boost 1.73, i.e. on Ubuntu 22.04, the old header issues a deprecation warning:
```
/usr/include/boost/bind.hpp:36:1: note: ‘#pragma message:
The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated.
Please use <boost/bind/bind.hpp> + using namespace boost::placeholders,
or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.’
```

While the new header <boost/bind/bind.hpp> is available for a long time (on 18.04 already) and the source has been updated to use boost::placeholders in #2023, the header was not yet updated.

### DIFF
--- a/clients/roscpp/include/ros/node_handle.h
+++ b/clients/roscpp/include/ros/node_handle.h
@@ -47,7 +47,7 @@
 #include "ros/init.h"
 #include "common.h"
 
-#include <boost/bind.hpp>
+#include <boost/bind/bind.hpp>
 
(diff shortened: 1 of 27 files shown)


sloretz (approved):
Thank you for the PR!

PR #2351, opened 2023-10-02
Fix latched topic not latched in splitted rosbags

### LINKED ISSUE #2350: rosbag "repeat latched" messages are not latched
# Description

When using rosbag record with splits and `--repeat-latched`, the latched messages in the subsequent bags are not marked as latched (and also lose the callerId).

This caused issue when I tried to replay a rosbag from the middle using `-s`.

# Steps to reproduce

```
# Latched topic publisher, in the background
rostopic pub -l /foo std_msgs/Bool true & 
# Non latched topic at 10Hz, so the bags are not empty, in the background
rostopic pub /bar std_msgs/Bool false -r 10 & 

# Record at least 20 seconds
rosbag record /foo /bar -O foo.bag --split --duration 10 --repeat-latched
```

Kill the rosbag, the two publishers, and try to get `/foo` from the middle of foo_1.bag:

```
rostopic echo /foo # In another terminal
rosbag play foo_1.bag -s 5
```

Expected: the rostopic node displays `data: True`
Actual: nothing happen.

Hugal31:
When writing latched topics in new rosbags splits, the header was not stored, which lost the "latching" flag. Store the header in new splits to avoid this.

Closes #2350

### DIFF
--- a/tools/rosbag/src/recorder.cpp
+++ b/tools/rosbag/src/recorder.cpp
@@ -457,7 +457,7 @@ void Recorder::startWriting() {
         {
             // Overwrite the original receipt time, otherwise the new bag will
             // have a gap before the new messages start.
-            bag_.write(out.second.topic, now, *out.second.msg);
+            bag_.write(out.second.topic, now, *out.second.msg, out.second.connection_header);
         }
     }
 


peci1 (approved):
LGTM!

pbelanger-avid:
Hey, just wondering if there's anything in particular that's holding up this PR? My team recently ran into this same issue and arrived at the same fix. It would be great if this could be included, Thanks!

AndreasNagel (approved):

sloretz (approved):
Thank you for the PR!

PR #2353, opened 2023-10-04
Fixed indefinite loop in roslogging.py

### LINKED ISSUE #2352: RospyLogger stuck in indefinte loop
By importing rosbag the standard logger is overwritte.
When sending a warning or error the logger is stuck in an indefinte loop.
This can cause other python libraries to malfunction (for example Flask).
To reproduce
```
import rosbag
import logging

logger = logging.getLogger('myLogger') # gets a RospyLogger object
logger.warn('message') # gets stuck in infinite loop
```
Infinite Loop is inside `def findCaller(self, *args, **kwargs):` in  [roslogging.py](https://github.com/ros/ros_comm/blob/noetic-devel/tools/rosgraph/src/rosgraph/roslogging.py).

Cause of the issue is:
```
file_name, lineno, func_name = super(RospyLogger, self).findCaller(*args, **kwargs)[:3] # returns file_name = ".../rosgraph/roslogging.py" lineno = 57  call func_name = 'findCaller'
```
returns itself as the caller:

Afterwards it tries to find a matching stackframe (skipping the first), but there is none. 
The loop to do so is a while loop, but the condition does not change, as `f` is not reassigned if the last frame is reached.
```
while hasattr(f, "f_code"):
            # Search for the right frame using the data already found by parent class.
            co = f.f_code
            filename = os.path.normcase(co.co_filename)
            if filename == file_name and f.f_lineno == lineno and co.co_name == func_name:
                break
            if f.f_back:
                f = f.f_back
```

There are two things to fix:
- Change the `super().findCaller()` so it does not return itself rather than the intended function. (Maybe stacklevel)
- Prevent infinite loop by adding a else statement that does something (Throw Error?)

Alternativly do not set the global Logging class to RaspyLogger.

StephanOLC:
Raising an ValueError if the caller function can not be found inside RospyLogger::findCaller() instead of looping indefinitly.
Fixes #2352 partially.
It prevents the function from looping indefinitely, but the correct frame is still not found.

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -69,6 +69,8 @@ class RospyLogger(logging.getLoggerClass()):
                 break
             if f.f_back:
                 f = f.f_back
+            else: # Reached the last stack frame and found no matching one.
+                raise ValueError("Could not find function [%s] on the framestack"%func_name)
 
         # Jump up two more frames, as the logger methods have been double wrapped.
         if f is not None and f.f_back and f.f_code and f.f_code.co_name == '_base_logger':


peci1:
What if the calling function is the top-level one? Wouldn't that also satisfy the condition that there is no previous frame?

Could you try adding a unit test that fails without the fix and succeeds with it?

StephanOLC:
If the top level function is the caller it should be caught by the if clause before.
 `if filename == file_name and f.f_lineno == lineno and co.co_name == func_name: 
                break`
The logic i changed affects going up the stack if the current frame is not the correct one.
`            if f.f_back:
                f = f.f_back
            else: # Reached the last stack frame and found no matching one.
                raise ValueError("Could not find function [%s] on the framestack"%func_name)`

A unit test could certainly be written to test it.

yobol1024:
@peci1 Hi! I recently take over a program importing rospy (depend on rosgraph), however it is blocked when I try to run it. But I am new to the rosgraph. I follow the suggestion from @StephanOLC, then it works. But I think it isn't a good choice to modify directly raw code. Can you do something to fix it, or give me some helpful advice to skip it? Thanks!

peci1:
I definitely cannot fix it in this repo because I'm not the maintainer. It would be up to the maintainers whether they are willing to accept this PR without the tests I asked, or if they will require them.

yobol1024:
> I definitely cannot fix it in this repo because I'm not the maintainer. It would be up to the maintainers whether they are willing to accept this PR without the tests I asked, or if they will require them.

Thanks for the reply! The problem was accidentally caused by a teammate's improper implementation. I skip invoking RospyLogger.findCaller by excluding rosgraph imports, as rosgraph should never have been introduced.

sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #2297, opened 2022-11-13
[ROS-O] required patches on current systems

jspricke:

PR #2354, opened 2023-10-06
apply patch to fix build with boost 1.83.0

daizhirui:


jspricke:
I've pushed a proper fix as part of #2297

sloretz (approved):
Thank you for the PR!

PR #871, opened 2016-08-19
fix static destruction order

rhaschke:
This is an attempt to fix #870.

According to [C++ standard](http://en.cppreference.com/w/cpp/utility/program/exit) `static` variables and registered `atexit` handlers are destroyed/called in _reverse_ order of their occurence.
The ROS source base seems to assume, that `atexit` handlers are called _before_ destruction of any `static` variables.

For example, the singletons `TopicManager` ... `XMLRPCManager` are created in `ros::start()` after having `ros::init()` registering the `atexit` handler. Hence, they will be destroyed before the call to the `atexit` handler, which attempted to access them and call their shutdown methods after destruction.

Use of global variables should be avoided in general, because their destruction order is determined by library linking. Using singletons the order of destruction can be determined by calling order.

This bug fix should be applied to Indigo and Jade as well and the source base should be checked for similar issues. I tracked the issue down with [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer) using`-DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -O1"`.

PR #2355, opened 2023-10-06
Fix occasional crash during shutdown when explicitly calling ros::start but not ros::shutdown

dgossow:
## Summary

In the case that user code calls `ros::start` explicitly, the `ros::NodeHandle` class will not call `ros::shutdown` when the last instance is destroyed.

In order to make sure that the program can exit error-free nonetheless, currently `ros::init` registers the ``ros::atexitCallback`` callback to be executed after the user's `main` has finished, which performs clean-up of resources that were set up in `init`.

However, `atexitCallback` is registered _before_ the various ROS singletons are instantiated, which means that it is called _after_ the singletons have been destroyed.

[This PR](https://github.com/ros/ros_comm/pull/871) addressed most of the problem, however with one remaining issue: `ros::start` creates a thread, which will continue to execute after singletons have been destroyed and until `ros::atexitCallback` is called. So, in rare circumstances, this thread can access singleton objects after they have been destroyed.

This PR fixes the issue by ensuring that `ros::shutdown` is called after `main` has returned, but _before_ singletons are destroyed.

## Detailed Description

We came across a quite rare crash of ROS nodes during shutdown at my company with the following backtrace:

main thread:
```
#0  0x00007f0076baead3 in futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x29708d8) at ../sysdeps/unix/sysv/linux/futex-internal.h:88
#1  0x00007f0076baead3 in __pthread_cond_wait_common (abstime=0x0, mutex=0x2970888, cond=0x29708b0) at pthread_cond_wait.c:502
#2  0x00007f0076baead3 in __pthread_cond_wait (cond=0x29708b0, mutex=0x2970888) at pthread_cond_wait.c:655
#3  0x00007f00773f1b5d in boost::condition_variable::wait(boost::unique_lock<boost::mutex>&) () at /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1
#4  0x00007f00773e9d14 in boost::thread::join_noexcept() () at /usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1
#5  0x00007f00ae60784d in ros::shutdown() () at /opt/ros/melodic/lib/libroscpp.so
#6  0x00007f00ae607a49 in ros::atexitCallback() () at /opt/ros/melodic/lib/libroscpp.so
#7  0x00007f007362f031 in __run_exit_handlers (status=0, listp=0x7f00739d7718 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true, run_dtors=run_dtors@entry=true) at exit.c:108
#8  0x00007f007362f12a in __GI_exit (status=<optimized out>) at exit.c:139
#9  0x00007f007360dc8e in __libc_start_main (main=0x419a90 <main(int, char**)>, argc=27, argv=0x7ffdf94ccdc8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7ffdf94ccdb8) at ../csu/libc-start.c:344
#10 0x00000000004176ca in _start () at /usr/include/boost/asio/ssl/impl/error.ipp:89
```

second thread:
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >'
  what():  boost: mutex lock failed in pthread_mutex_lock: Invalid argument
Stack trace (most recent call last) in thread 245:
#18   Object "", at 0xffffffffffffffff, in 
#17   Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7fae78dc561e, in clone
#16   Object "/lib/x86_64-linux-gnu/libpthread.so.0", at 0x7fae7c25f6da, in start_thread
#15   Object "/usr/lib/x86_64-linux-gnu/libboost_thread.so.1.65.1", at 0x7fae7caa0bcc, in boost::this_thread::interruption_requested()
#14   Object "/opt/ros/melodic/lib/libroscpp.so", at 0x7faeb3cbdef0, in ros::internalCallbackQueueThreadFunc()
#13   Object "/opt/ros/melodic/lib/libroscpp.so", at 0x7faeb3c7e2fa, in ros::CallbackQueue::callAvailable(ros::WallDuration)
#12   Object "/opt/ros/melodic/lib/libroscpp.so", at 0x7faeb3c7c558, in ros::CallbackQueue::callOneCB(ros::CallbackQueue::TLS*)
#11   Object "/opt/ros/melodic/lib/libroscpp.so", at 0x7faeb3cb267a, in ros::TimerManager<ros::SteadyTime, ros::WallDuration, ros::SteadyTimerEvent>::TimerQueueCallback::call()
#10   Object "/opt/ros/melodic/lib/libroscpp.so", at 0x7faeb3cad644, in ros::TransportPublisherLink::onRetryTimer(ros::SteadyTimerEvent const&)
#9    Object "/opt/ros/melodic/lib/libroscpp.so", at 0x7faeb3ca32c2, in ros::TransportTCP::connect(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int)
#8    Object "/opt/ros/melodic/lib/libroscpp.so", at 0x7faeb3ca1e34, in ros::TransportTCP::initializeSocket()
#7    Object "/opt/ros/melodic/lib/libroscpp.so", at 0x7faeb3cdbb31, in ros::PollSet::addSocket(int, boost::function<void (int)> const&, boost::shared_ptr<ros::Transport> const&)
#6    Object "/opt/ros/melodic/lib/libroscpp.so", at 0x7faeb3c2a368, in void boost::throw_exception<boost::lock_error>(boost::lock_error const&)
#5    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7fae796ddd53, in __cxa_throw
#4    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7fae796ddb20, in std::terminate()
#3    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7fae796ddae5, in std::rethrow_exception(std::__exception_ptr::exception_ptr)
#2    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7fae796d7956, in 
#1    Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7fae78ce47f0, in abort
#0    Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7fae78ce2e87, in gsignal
Aborted (Signal sent by tkill() 231 0)
```

After digging through the source code, I found that this is indeed a bug in ROS itself, which happens in the following circumstances:
* The user code calls `ros::start`, but not `ros::shutdown`.
* There is some TCP transport publisher link which is dropped and happens to try to reconnect after the program's `main()` has ended

In this case, the sequence that leads to the crash is the following:
* [In `ros::init()`, the `ros::atexitCallback` is registered to be called after `main()` has ended.](https://github.com/boostorg/thread/blob/boost-1.71.0/include/boost/thread/detail/thread.hpp#L248-L259)
* In `ros::start()`, [Singletons are created](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/init.cpp#L321-L330) and [a thread which calls `internalCallbackQueueThreadFunc` is started](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/init.cpp#L410), [which calls callbacks from `ros::g_internal_callback_queue`](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/init.cpp#L276-L286).
* [TransportPublisherLink creates a timer which calls `TransportPublisherLink::onRetryTimer`](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/transport_publisher_link.cpp#L285-L287) and pushes this onto `ros::g_internal_callback_queue`, [accessed via ros::getInternalCallbackQueue()](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/transport_publisher_link.cpp#L255).

Once the user's `main()` function exits:
* g_internal_queue_thread is destructed, [which results in the underlying thread being detached](https://www.boost.org/doc/libs/1_79_0/doc/html/thread/thread_management.html#thread.thread_management.thread.destructor), i.e. `internalCallbackQueueThreadFunc` continues to run
* all singletons are destructed, including PollManager.
* `atexitCallback` is called and calls `ros::shutdown`, which sets `g_shutting_down` to true, signaling the `internalCallbackQueueThreadFunc` to exit, and waits for the thread to join.
* The `internalCallbackQueueThreadFunc` thread however [has already entered the call to `queue->callAvailable`](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/init.cpp#L284), which in turn calls `TransportPublisherLink::onRetryTimer`.
* [`TransportPublisherLink::onRetryTimer` calls `PollManager::instance()`](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/transport_publisher_link.cpp#L233), which returns the address of the destructed singleton object
* [The following call to `TransportTCP::connect`](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/transport_publisher_link.cpp#L234) results in an exception, since it tries to lock an already-destructed mutex.

The fix in this PR works around this problem by making sure that the `internalCallbackQueueThreadFunc` thread is ended before singletons are destroyed.

It does this by stopping the thread in the destructor of a `InternalQueueJoiningThread` instance which is declared as a static variable inside of the `initInternalQueueJoiningThread` function.

This function is called after the singleton instances (which are also static function-scope variables) are created, which means that the compiler must generate code that destroys the `InternalQueueJoiningThread` object, and thus joins the thread, before the other Singletons are destroyed.

Note that, as mentioned above, init.cpp registers [`atexitCallback`](https://github.com/ros/ros_comm/blob/noetic-devel/clients/roscpp/src/libros/init.cpp#L146-L154) specifically for the case that `ros::shutdown` was not called by the user code. However, it fails to address the issue described above since it is called after singletons have been destroyed but before the `internalCallbackQueueThreadFunc` thread has ended.

This was meant to be addressed by [this PR,](https://github.com/ros/ros_comm/pull/871), but the fix was incomplete.

### DIFF
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -304,9 +304,9 @@ struct InternalQueueJoiningThread
 
 void initInternalQueueJoiningThread()
 {
-  // Constructs the thread on first use, and joins it on shutdown.
-  // This is used to avoid the situation where the thread continues to run
-  // after singletons have been destroyed.
+  // This function must be called after the ROS singletons have been created
+  // via their respective getInstance() functions in order to ensure that
+  // the thread stops before the singletons are destroyed.
   // For more details, see: https://github.com/ros/ros_comm/pull/2355
   static InternalQueueJoiningThread internal_queue_joining_thread;
 }
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -98,7 +98,6 @@ static CallbackQueuePtr g_internal_callback_queue;
 
 static bool g_initialized = false;
 static bool g_started = false;
-static bool g_atexit_registered = false;
 static boost::mutex g_start_mutex;
 static bool g_ok = false;
 static uint32_t g_init_options = 0;
@@ -143,14 +142,17 @@ void requestShutdown()
   g_shutdown_requested = true;
 }
 
-void atexitCallback()
+struct ShutdownCaller
 {
-  if (ok() && !isShuttingDown())
+  ~ShutdownCaller()
   {
-    ROSCPP_LOG_DEBUG("shutting down due to exit() or end of main() without cleanup of all NodeHandles");
-    g_started = false; // don't shutdown singletons, because they are already destroyed
     shutdown();
   }
+};
+
+void initShutdownCaller()
+{
+  static ShutdownCaller shutdown_caller;
 }
 
 void shutdownCallback(XmlRpc::XmlRpcValue& params, XmlRpc::XmlRpcValue& result)
@@ -285,33 +287,6 @@ void internalCallbackQueueThreadFunc()
   }
 }
 
-struct InternalQueueJoiningThread
-{
-  InternalQueueJoiningThread()
-  {
-      g_internal_queue_thread = boost::thread(internalCallbackQueueThreadFunc);
-  }
-
-  ~InternalQueueJoiningThread()
-  {
-    if (g_internal_queue_thread.joinable())
-    {
-      g_shutting_down = true;
-      g_internal_queue_thread.join();
-    }
-  }
-};
-
-void initInternalQueueJoiningThread()
-{
-  // This function must be called after the ROS singletons have been created
-  // via their respective getInstance() functions in order to ensure that
-  // the thread stops before the singletons are destroyed.
-  // For more details, see: https://github.com/ros/ros_comm/pull/2355
-  static InternalQueueJoiningThread internal_queue_joining_thread;
-}
-
-
 bool isStarted()
 {
   return g_started;
@@ -434,7 +409,9 @@ void start()
 
   if (g_shutting_down) goto end;
 
-  initInternalQueueJoiningThread();
+  g_internal_queue_thread = boost::thread(internalCallbackQueueThreadFunc);
+
+  initShutdownCaller();
 
   getGlobalCallbackQueue()->enable();
 
@@ -475,12 +452,6 @@ void check_ipv6_environment() {
 
 void init(const M_string& remappings, const std::string& name, uint32_t options)
 {
-  if (!g_atexit_registered)
-  {
-    g_atexit_registered = true;
-    atexit(atexitCallback);
-  }
-
   if (!g_global_queue)
   {
     g_global_queue.reset(new CallbackQueue);
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -622,6 +622,10 @@ void deInit()
   //ros::console::deregister_appender(g_rosout_appender);
   delete g_rosout_appender;
   g_rosout_appender = 0;
+
+  // preserve legacy behavior
+  g_shutting_down = true;
+  g_ok = false;
 }
 
 void shutdown()
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -608,6 +608,10 @@ bool ok()
 
 void deInit()
 {
+  boost::recursive_mutex::scoped_lock lock(g_shutting_down_mutex);
+  if (g_shutting_down)
+    return;
+
   ros::console::shutdown();
 
   g_global_queue->disable();
@@ -626,12 +630,13 @@ void shutdown()
   else
     g_shutting_down = true;
 
+  if (g_internal_queue_thread.get_id() != boost::this_thread::get_id())
+  {
+    g_internal_queue_thread.join();
+  }
+
   if (g_started)
   {
-    if (g_internal_queue_thread.get_id() != boost::this_thread::get_id())
-    {
-      g_internal_queue_thread.join();
-    }
 
     TopicManager::instance()->shutdown();
     ServiceManager::instance()->shutdown();
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -160,7 +160,13 @@ struct DeinitCaller
 {
   ~DeinitCaller()
   {
+  boost::recursive_mutex::scoped_lock lock(g_shutting_down_mutex);
+  // Only call deInit if ros::shutdown has not been called, otherwise
+  // we'd call it twice.
+  if (!g_shutting_down)
+  {
     deInit();
+  }
   }
 };
 
@@ -608,11 +614,6 @@ bool ok()
 
 void deInit()
 {
-  static bool deinitialized = false;
-  if (deinitialized)
-    return;
-  deinitialized = true;
-  
   ros::console::shutdown();
 
   g_global_queue->disable();
@@ -646,14 +647,11 @@ void shutdown()
     XMLRPCManager::instance()->shutdown();
   }
 
-  if (g_initialized)
-  {
-    deInit();
-  }
-
   g_started = false;
   g_ok = false;
   Time::shutdown();
+
+  deInit();
 }
 
 const std::string& getDefaultMasterURI() {
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -98,6 +98,8 @@ static CallbackQueuePtr g_internal_callback_queue;
 
 static bool g_initialized = false;
 static bool g_started = false;
+static bool g_deinit_atexit_registered = false;
+static bool g_shutdown_atexit_registered = false;
 static boost::mutex g_start_mutex;
 static bool g_ok = false;
 static uint32_t g_init_options = 0;
@@ -142,37 +144,16 @@ void requestShutdown()
   g_shutdown_requested = true;
 }
 
-struct ShutdownCaller
-{
-  ~ShutdownCaller()
-  {
-    shutdown();
-  }
-};
-
-void installShutdownCaller()
-{
-  static ShutdownCaller shutdown_caller;
-}
-
 void deInit();
-struct DeinitCaller
+
+void deinitAtexitCallback()
 {
-  ~DeinitCaller()
-  {
-  boost::recursive_mutex::scoped_lock lock(g_shutting_down_mutex);
   // Only call deInit if ros::shutdown has not been called, otherwise
   // we'd call it twice.
-  if (!g_shutting_down)
+  if (!isShuttingDown())
   {
     deInit();
   }
-  }
-};
-
-void installDeinitCaller()
-{
-  static DeinitCaller deinit_caller;
 }
 
 void shutdownCallback(XmlRpc::XmlRpcValue& params, XmlRpc::XmlRpcValue& result)
@@ -431,7 +412,11 @@ void start()
 
   g_internal_queue_thread = boost::thread(internalCallbackQueueThreadFunc);
 
-  installShutdownCaller();
+  if (!g_shutdown_atexit_registered)
+  {
+    g_shutdown_atexit_registered = true;
+    atexit(shutdown);
+  }
 
   getGlobalCallbackQueue()->enable();
 
@@ -472,6 +457,12 @@ void check_ipv6_environment() {
 
 void init(const M_string& remappings, const std::string& name, uint32_t options)
 {
+  if (!g_deinit_atexit_registered)
+  {
+    g_deinit_atexit_registered = true;
+    atexit(deinitAtexitCallback);
+  }
+  
   if (!g_global_queue)
   {
     g_global_queue.reset(new CallbackQueue);
@@ -500,8 +491,6 @@ void init(const M_string& remappings, const std::string& name, uint32_t options)
 
     g_initialized = true;
   }
-
-  installDeinitCaller();
 }
 
 void init(int& argc, char** argv, const std::string& name, uint32_t options)
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -608,10 +608,11 @@ bool ok()
 
 void deInit()
 {
-  boost::recursive_mutex::scoped_lock lock(g_shutting_down_mutex);
-  if (g_shutting_down)
+  static bool deinitialized = false;
+  if (deinitialized)
     return;
-
+  deinitialized = true;
+  
   ros::console::shutdown();
 
   g_global_queue->disable();
--- a/clients/roscpp/src/libros/init.cpp
+++ b/clients/roscpp/src/libros/init.cpp
@@ -150,9 +150,23 @@ struct ShutdownCaller
   }
 };
 
-void initShutdownCaller()
+void installShutdownCaller()
 {
   static ShutdownCaller shutdown_caller;
+}
+
+void deInit();
+struct DeinitCaller
+{
+  ~DeinitCaller()
+  {
+    deInit();
+  }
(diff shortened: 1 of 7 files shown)


dgossow:
@mjcarroll could you PTAL?

clients/roscpp/src/libros/init.cpp
```diff
@@ -285,6 +285,33 @@ void internalCallbackQueueThreadFunc()
...
+  InternalQueueJoiningThread()
+  {
+      g_internal_queue_thread = boost::thread(internalCallbackQueueThreadFunc);
+  }
+
+  ~InternalQueueJoiningThread()
+  {
+    if (g_internal_queue_thread.joinable())
+    {
+      g_shutting_down = true;
```
rhaschke: Why do you need to set this flag? This essentially prohibits `shutdown()` from being executed.

rhaschke (commented):
Generally, this change looks reasonable. But I don't understand why you need to set the flag. The destruction order ensured by the compiler should already guarantee that the thread finishes before singleton destruction.

clients/roscpp/src/libros/init.cpp
dgossow: > Why do you need to set this flag?

My motivation was to signal the internalCallbackQueueThreadFunc thread to exit its main loop.

```
void internalCallbackQueueThreadFunc()
{
  disableAllSignalsInThisThread();

  CallbackQueuePtr queue = getInternalCallbackQueue();

  while (!g_shutting_down)
  {
    queue->callAvailable(WallDuration(0.1));
  }
}
```

> This essentially prohibits shutdown() from being executed.

Right, thanks for catching this.

Let me look again at how to properly do this.

dgossow:
@rhaschke could you take another look?

test/test_roscpp/test/src/missing_call_to_shutdown.cpp
```diff
@@ -0,0 +1,84 @@
...
+  if (ros::isStarted())
+  {
+    std::cerr << "ERROR: ros::isStarted() returned true after ROS has been de-initialized!" << std::endl;
+  }
+  if (!ros::isShuttingDown())
+  {
+    std::cerr << "ERROR: ros::isShuttingDown() returned false after ROS has been de-initialized!" << std::endl;
+  }
+
+  if (hasError)
```
rhaschke: How can `hasError` become true?

clients/roscpp/src/libros/init.cpp
```diff
@@ -447,12 +452,6 @@ void check_ipv6_environment() {
 
 void init(const M_string& remappings, const std::string& name, uint32_t options)
 {
-  if (!g_atexit_registered)
-  {
-    g_atexit_registered = true;
-    atexit(atexitCallback);
```
rhaschke: I would expect you to call `initShutdownCaller()` here too. 
`shutdownCallback()` is only called by an explicit remote shutdown request, isn't it?

rhaschke (commented):
Switching from boolean variables to static objects is probably the only way to correctly fix that issue. But I have one open question (see below).
If you take the effort to write a test, why don't you make a true unit test from it, i.e. adding some assertions?

dgossow:
> The destruction order ensured by the compiler should already guarantee that the thread finishes before singleton destruction.

I guess because of [this](https://en.cppreference.com/w/cpp/utility/program/exit)?

> d) For each local object obj with static storage duration, obj is destroyed as if a function calling the destructor of obj were registered with [std::atexit](https://en.cppreference.com/w/cpp/utility/program/atexit) at the completion of the constructor of obj.

> c) If the completion of the initialization of a static object A was sequenced-before the call to [std::atexit](https://en.cppreference.com/w/cpp/utility/program/atexit) for some function F, the call to F during termination is sequenced-before the start of the destruction of A

Order of de-initialization used to be:
* singleton destructors
* atexitCallback
* g_internal_queue_thread destructor

Now it is
* ShutdownCaller destructor
* singleton destructors
* g_internal_queue_thread destructor

rhaschke:
I'm afraid we're talking past each other. My point is that the shutdown handler isn't registered in `ros::init()` at all anymore.
I would expect that you _replace_ the `atexit` registration with `ShutdownCaller` instantiation, i.e. calling `initShutdownCaller()`, but you just removed the `atexit` registration w/o replacing it with something else.

The `ShutdownCaller` is only initialized in `shutdownCallback()`, which is only called on an external `shutdown` request:
https://github.com/ros/ros_comm/blob/842f0f026924323f605495da0b80493f15f0bdce/clients/roscpp/src/libros/init.cpp#L322

Actually, I would expect the call to `initShutdownCaller()` _only_ in `ros::init()` and not in the latter callback.

test/test_roscpp/test/src/missing_call_to_shutdown.cpp
dgossow: fixed

dgossow:
> The ShutdownCaller is only initialized in shutdownCallback(), which is only called on an external shutdown request:

The GitHub diff view is a little misleading here. `initShutdownCaller` is now called in `void start()` and not shutdownCallback`

> I'm afraid we're talking past each other. My point is that the shutdown handler isn't registered in ros::init() at all anymore.

Sorry, I didn't respond to that comment yet.

Yes, that is true. With this change, `ros::init` does not register any type of exit handler anymore. Only `ros::start` does.

`atexitHandler` used to call `ros::shutdown` (with `g_started` set to `false`).

My understanding was that `shutdown` only has to be called if `start` has been called. But you are right, some of the resources it frees are allocated in `init()`.

Looking at the shutdown code in `ros::shutdown`:

```cpp
void shutdown()
{
  boost::recursive_mutex::scoped_lock lock(g_shutting_down_mutex);
  if (g_shutting_down)
    return;
  else
    g_shutting_down = true;

  ros::console::shutdown(); // ROSCONSOLE_AUTOINIT is called in ros::init

  g_global_queue->disable(); // created in init()
  g_global_queue->clear();

  if (g_internal_queue_thread.get_id() != boost::this_thread::get_id())
  {
    g_internal_queue_thread.join();
  } // created in start()

  //ros::console::deregister_appender(g_rosout_appender);
  delete g_rosout_appender; // created in start()
  g_rosout_appender = 0;

  if (g_started)
  {
    TopicManager::instance()->shutdown();
    ServiceManager::instance()->shutdown();
    PollManager::instance()->shutdown();
    ConnectionManager::instance()->shutdown();
    XMLRPCManager::instance()->shutdown();
  }

  g_started = false;
  g_ok = false;
  Time::shutdown(); // ros::Time::init(); is called in start()
}
```

dgossow:
>  If you take the effort to write a test, why don't you make a true unit test from it, i.e. adding some assertions?

the assertions are made in the `atexit` callback, this is why gtest is not used, if this was your question.

the missing assignment of `hasError` was a bug, it's fixed now

dgossow:
I'll see if I can fix the mismatch between `init`, `start` and `shutdown`

dgossow:
> Switching from boolean variables to static objects is probably the only way to correctly fix that issue.

Yes, it's a mess and much more of a rabbit hole than I signed up for.

I pushed another fix.

clients/roscpp/src/libros/init.cpp
dgossow: replaced with the `installDeinitCaller` call at the end of the function

dgossow:
oh what have I gotten myself into. This code is such a f-ing mess

test/test_roscpp/test/src/missing_call_to_shutdown.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+  if (!ros::isShuttingDown())
+  {
+    std::cerr << "ERROR: ros::isShuttingDown() returned false after ROS has been de-initialized!" << std::endl;
+    hasError = true;
+  }
+
+  if (hasError)
+  {
+    std::_Exit(1);
```
rhaschke: So, the unit test fails via a non-zero exit code? :+1:

dgossow: yes, though I will have to write a wrapper executable as well, rostest wants an XML file as output

dgossow:
Alright.

It seems like the current implementation preserves the previous behavior, except for joining the thread before singletons are destructed.

I also tested that introducing various bugs does indeed fail the tests.

dgossow:
@rhaschke Thanks a lot for the quick review cycle. would you mind taking another look?

clients/roscpp/src/libros/init.cpp
```diff
@@ -143,14 +142,37 @@ void requestShutdown()
...
+  if (!g_shutting_down)
+  {
+    deInit();
+  }
+  }
+};
+
+void installDeinitCaller()
+{
+  static DeinitCaller deinit_caller;
```
dgossow: I could also revert to the previous pattern using `std::atexit`. This would be functionally the same and minimize cosmetic code changes

dgossow: done.

rhaschke:
> would you mind taking another look?

I'm in a meeting now and then driving home. I will have a look into that later in the evening again.

dgossow:
@rhaschke happy to also jump on a call if you want to talk through the changes here.

I opened this PR on a whim, but now that I've spent a full day on it, I'm kind of invested in it.

dgossow:
@rhaschke I also added a more readable explanation to the PR description.

rhaschke (approved):
Looks good. Great to have unit tests for this now!

dgossow:
@mjcarroll could you merge this?

dgossow:
@ros/ros_comm-maintainers could someone give this a final review and merge?

mjcarroll (approved):

mjcarroll:
This looks good to me, @sloretz do you want to take a look before we merge?

sloretz (approved):

sloretz:
Thank you for the PR and clear description!

test/test_roscpp/test/src/missing_call_to_shutdown.cpp
meyerj, line 1: This test case introduced a dependency on `rosbash` which provides the `rosrun` executable. So add
`<test_depend>rosbash</test_depend>` to `package.xml`? Or would it be possible to just use a relative path here (`./`) and make sure that the working directory is set accordingly in `test/test_roscpp/test/launch/missing_call_to_shutdown.xml`?

sloretz: Adding a `<test_depend>` could work as long as nothing in `https://github.com/ros/ros` depends on `ros/ros_comm`, which I think is the case. I'm cautious about hard coding a relative path and setting the test working directory, but that could work too.

Another option to remove the `rosbash` dependency:

1. Remove `${PROJECT_NAME}-missing_call_to_shutdown`
2. Rename `${PROJECT_NAME}-missing_call_to_shutdown_impl` to `${PROJECT_NAME}-missing_call_to_shutdown`
3. Make `test/test_roscpp/test/launch/missing_call_to_shutdown.xml` call `test_roscpp-missing_call_to_shutdown` twice: once with `args="0"` and once with `args="1"`.

PR #2357, opened 2023-10-27
Use atomic bool instead of volatile uint32_t

4c3y:
- Use `boost::atomic<bool>` instead of `volatile uint32_t`
- Replaced int literals with bool literals
- Fixed `#include` order


peci1:
This PR would break the existing API and ABI. Such changes are usually not accepted for released ROS distros.

sloretz:
Thank you for the PR!

Unfortunately I don't think we should merge this one. As said above, we can't break ABI or API in released ROS distros

PR #2361, opened 2023-11-22
[ROS-O] Fix rostest target names when build dir is inside source dir

v4hn:
It's a common approach to build cmake projects in a `build` folder in the project root. It's just that `rosbuild`/`catkin` does not do it, so the changed name never mattered.

When you try, the target name includes another `build_` string and there is *exactly one* package in the whole ecosystem that relies on this not happening: https://github.com/ros/ros_comm/blob/845f74602c7464e08ef5ac6fd9e26c97d0fe42c9/test/test_rosbag/bag_migration_tests/CMakeLists.txt#L87-L103


sloretz (approved):
Thank you for the PR!

PR #2363, opened 2023-11-29
Fix segfault with default-constructed rosbag::ChunkedFile::swap

### LINKED ISSUE #2362: Default constructed rosbag::ChunkedFile segfault on swap
`rosbag::ChunkedFile::swap` segfault if the ChunkedFile is default-constructed, because it tries to write to `read_stream_->file_` while `read_stream_` is null.

This causes `rosbag::Bag` to segfault as well when using `std::swap`, for instance when adding rosbags to a growing vector.

This code triggers the segfault:

```c++
#include <rosbag/bag.h>

int main()
{
  rosbag::Bag bag1;
  rosbag::Bag bag2;
  std::swap(bag1, bag2);
}
```

Hugal31:
Only update the read and write stream if their pointers are not null.

Closes #2362.


sloretz (approved):
Thank you for the PR!

PR #1872, opened 2020-01-28
[windows] Workaround for Python 2 xmlrpc performance issues.

seanyen:
There is a ROS performance slowness issue that we root caused to an existing issue for Python 2. One discussion can be found here: https://stackoverflow.com/questions/2617615/slow-python-http-server-on-localhost

Instead of asking Windows developers to manually override the `ROS_MASTER_URI` and `ROS_IP` to workaround it, I am proposing that, by default, make them to be IP address directly.

This fixes https://github.com/ms-iot/ROSOnWindows/issues/6.

PR #2364, opened 2023-11-30
[windows] Remove python2 specific hack

gftabor:
This PR https://github.com/ros/ros_comm/pull/1872 made windows act differently than linux because of a specific python 2 delay doing unnecessary DNS checks for localhost.

The underlying code in python is different for python3, so the hack should no longer necessary. Given that, it's better to reunify the codebase.

Fixes 
https://github.com/RoboStack/ros-noetic/issues/430

Standard workflow is source workspace/catkin install and then export environment variables. The current hack makes that workflow not work on windows.


sloretz (approved):
Thank you for the PR!

PR #2366, opened 2024-01-14
Make rosnode info show topic types of unconnected subscribers

Sanoronas:
When running a node which has unconnected Subscribers rosnode info is unable to infer topic type.
In contrast rostopic info is able to resolve this.

- change rosnode info to use `master.getTopicTypes()` instead of `master.getPublishedTopics('/')`
- rename `pub_topics` variable to `topic_types` to match new content

### DIFF
--- a/tools/rosnode/src/rosnode/__init__.py
+++ b/tools/rosnode/src/rosnode/__init__.py
@@ -498,7 +498,7 @@ def get_node_info_description(node_name):
     # go through the master system state first
     try:
         state = master.getSystemState()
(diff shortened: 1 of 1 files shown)


peci1 (approved):
LGTM, I've tested this locally.

Theoretically, this might introduce much higher traffic on systems with many topics, but I guess it's not a big problem because `rosnode info` is more a diagnostic tool than something that should be used as a part of a final system.

sloretz (approved):
Thank you for the PR!

PR #2368, opened 2024-01-26
properly shutdown xml rpc server

### LINKED ISSUE #2238: xmlrpc.XmlRpcNode consumes 100%+ cpu after being shutdown

jamesdarrenmuir:
stop CPU usage spike when a rospy node receives a signals.SIGINT interrupt but before it terminates the process (e.g. pressing CTRL+C while the node is sleeping due to a rospy.sleep() call no longer sends CPU usage of a core to 100%)

should also fix https://github.com/ros/ros_comm/issues/2238

### DIFF
--- a/tools/rosgraph/src/rosgraph/xmlrpc.py
+++ b/tools/rosgraph/src/rosgraph/xmlrpc.py
@@ -229,8 +229,8 @@ class XmlRpcNode(object):
             if handler:
                 handler._shutdown(reason)
             if server:
-                server.socket.close()
                 server.server_close()
+                server.shutdown()
                 
     def start(self):
         """


peci1:
This apparently causes problems elsewhere - see the timing out unit tests...

jamesdarrenmuir:
I looked at the build log but I'm still not sure which test failed.

I saw these lines in the build log:
```
14:05:56 Scanning dependencies of target _run_tests_rosgraph_nosetests_test
14:05:56 -- run_tests.py: execute commands
14:05:56   "/usr/bin/cmake" -E make_directory /tmp/ws/test_results/rosgraph
14:05:56   /usr/bin/nosetests3 -P --process-timeout=60 --where=/tmp/ws/src/ros_comm/tools/rosgraph/test --with-xunit --xunit-file=/tmp/ws/test_results/rosgraph/nosetests-test.xml
14:05:56 /tmp/ws/src/ros_comm/tools/rosgraph/test/test_names.py:97: SyntaxWarning: "is" with a literal. Did you mean "=="?
14:05:56   if 'ROS_NAMESPACE' is os.environ:
14:05:56 ........ERROR: Invalid remapping argument 'foo:=bar:=baz'
14:05:56 ..........invalid ROS_IP (must be a valid IPv4 or IPv6 address)
14:05:56 invalid ROS_IP (must be a valid IPv4 or IPv6 address)
14:05:56 .invalid ROS_IP (must be a valid IPv4 or IPv6 address)
14:05:56 .invalid ROS_IP (must be a valid IPv4 or IPv6 address)
14:05:56 ................................Build timed out (after 120 minutes). Marking the build as failed.
14:05:56 Build was aborted
```

But I'm not sure which of these, if any, caused the test failure.

jamesdarrenmuir:
I am running the tests on my machine now (`catkin test`), but I am getting failures when testing the latest `noetic-devel` branch as well.

jamesdarrenmuir:
Ok, it seems like running the tests in the `rosgraph` package gets stuck. I will look more into this later.

sloretz:
Thank you for the PR!

Given that [ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160), and this one maybe isn't the right solution, I think it's unlikely this will get merged before EOL. I'm going to close it instead.

PR #2370, opened 2024-02-20
Fixed ROSCONSOLE_FORMAT with microseconds

peci1:
Currently, using the `%f` modifier for microseconds in `ROSCONSOLE_FORMAT` only works for C++ nodes. Python nodes do not interpret it:

```
$ export ROSCONSOLE_FORMAT='[${severity}] [${time:%d/%m/%Y %T.%f}] [${node}]: ${message}'

$ rosrun joint_state_publisher_gui joint_state_publisher_gui  # Python node
[INFO] [20/02/2024 15:19:57.%f, 0.000000] [/joint_state_publisher_gui]: Centering

$ rosrun tf2_ros static_transform_publisher 0 0 0 0 0 0 a b  # C++ node
[ INFO] [20/02/2024 14:41:05.796267] [/static_transform_publisher_1708440065774087215]: Spinning until killed publishing a to b
```

That is because the Python part calls `time.strftime()`, which uses the integer seconds. If, on the other hand, `datetime` is used, it uses a fractional seconds representation and is this [capable of printing the microseconds](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes).

This PR fixes the issue by switching to `datetime` objects. It also updates the unit test to test the feature.

### DIFF
--- a/tools/rosgraph/src/rosgraph/roslogging.py
+++ b/tools/rosgraph/src/rosgraph/roslogging.py
@@ -255,7 +255,7 @@ class RosStreamHandler(logging.Handler):
         while '${walltime:' in msg:
             tag_end_index = msg.index('${walltime:') + len('${walltime:')
             time_format = msg[tag_end_index: msg.index('}', tag_end_index)]
-            time_str = time.strftime(time_format)
+            time_str = datetime.datetime.now().strftime(time_format)
             msg = msg.replace('${walltime:' + time_format + '}', time_str)
 
         msg = msg.replace('${thread}', str(record.thread))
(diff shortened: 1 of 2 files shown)


peci1:
This PR is similar to https://github.com/ros/ros_comm/pull/2367 but chose the other mentioned approach by using `datetime`. This PR also fixes both places where time strings can appear. Moreover, this PR adds a unit test.

rafa-martin:
I think they will merge this feature...

peci1:
@rafa-martin Thanks for stopping by. Would you be able to review this PR? In the low maintenance mode Noetic has, non-maintainer reviews are almost a requirement to get things moving.

rafa-martin (approved):
It seems a good solution, also it includes enough tests to validate it.

peci1:
Thanks!

sloretz (approved):
Thanks for the PR!

PR #2388, opened 2025-05-09
ROS-O just the required commits

sloretz:
#2297 got updated with commits that change the behavior of some packages. I'm opening a new PR set to a few commits back to avoid changing the `ros-o:obese-devel` branch

@jspricke FYI


sloretz:
LGTM. Verified desktop_full builds with this change.

jspricke:
Awesome, thanks for your work!

peci1:
I see a new problem that might be connected to this PR: https://github.com/ros/ros_comm/issues/2389 .

jspricke:
@peci1 I can't reproduce this on Debian (which includes this patches). Can you bisect where it is coming from instead of blaming some PR?

peci1:
Ah, right. I blamed this PR because it was the last one that touched the relevant files, but the error actually started appearing since #2277 (which makes sense, as it only started showing the errors).

I've debugged this locally and the culprit was in a few abruptly killed nodes that did not unregister from rosmaster. The `/requestTopics` XMLRPC request then tried to contact these dead sockets, so the error was actually correct.

Sorry for the noise.

PR #2391, opened 2025-05-16
Revert: Maintain constness of accessed XmlRpcValue struct

### LINKED ISSUE #2390: XmlRpcValue - accessing a non-existing key throws std::out_of_range
One side-effect of [this change](https://github.com/ros/ros_comm/pull/2315) is that retrieving a value of a non-existing key now throws an out of range instead of returning a default value.

```
#include "xmlrpcpp/XmlRpcValue.h"
#include <string>
#include <iostream>

void bla(const XmlRpc::XmlRpcValue& value) {
	const std::string& does_not_exist = value["this_does_not_exist"]; // this line now throws an exception
	std::cout << "does_not_exist:" << does_not_exist << std::endl;
}

int main() {
	XmlRpc::XmlRpcValue value;
	value["this_exists"] = "this exists";
	std::cout << "this_exists:" << value["this_exists"] << std::endl;
	bla(value);
	return 0;
}

```

sloretz:
Edit: This now reverts #2315  @rhaschke FYI

Reverting because it changed behavior in a way that broke user's builds. See #2390 for more information.

Fixes #2390

### DIFF
--- a/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcValue.h
+++ b/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcValue.h
@@ -105,9 +105,9 @@ namespace XmlRpc {
     XmlRpcValue const& operator[](int i) const { assertArray(i+1); return _value.asArray->at(i); }
     XmlRpcValue& operator[](int i)             { assertArray(i+1); return _value.asArray->at(i); }
 
-    XmlRpcValue const& operator[](std::string const& k) const { assertStruct(); return _value.asStruct->at(k); }
+    XmlRpcValue& operator[](std::string const& k) const { assertStruct(); return (*_value.asStruct)[k]; }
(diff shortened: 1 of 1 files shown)


sloretz:
`noetic-devel` without this PR:

```
this_exists:this exists
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
Aborted (core dumped)
```

which is raised here:
https://github.com/ros/ros_comm/blob/b6c57e76a764252cf50d8d24053f32e2ad54a264/utilities/xmlrpcpp/include/xmlrpcpp/XmlRpcValue.h#L110

This PR:

```
this_exists:this exists
terminate called after throwing an instance of 'XmlRpc::XmlRpcException'
Aborted (core dumped)
```

which is raised here: https://github.com/ros/ros_comm/blob/41a72bd22a81deb8a729a08c6912de5dadaed083/utilities/xmlrpcpp/src/XmlRpcValue.cpp#L78

So this PR doesn't restore the original behavior either

